PR #14, opened 2020-10-29
HEEDLS-130 Fix build

FridaTveit:
* Add a GitHub Actions workflow which will build and test the code. It won't run the data tests since they need to use a database.

* Fix issue with latest migration. Looks like there was a field missed when creating the new FilteredLearningActivity. And the ApplyLPDefaultsToSelfEnrol field was already present in the Customisations table so couldn't be added again.

* Fix web and data unit tests. There were a few different errors, some to do with the difference between using an obfuscated database and a non-obfuscated one. I think it would be best to stick with the obfuscated version if possible.

* Fix linting issues.

### DIFF
--- a/DigitalLearningSolutions.Data.Migrations/202010161445_ApplyLPDefaultSPs.cs
+++ b/DigitalLearningSolutions.Data.Migrations/202010161445_ApplyLPDefaultSPs.cs
@@ -2,12 +2,10 @@
 {
     using FluentMigrator;
     [Migration(202010161445)]
-    public class ApplyLPDefaultSPs :Migration
+    public class ApplyLPDefaultSPs : Migration
     {
         public override void Up()
         {
-            Alter.Table("Customisations")
-               .AddColumn("ApplyLPDefaultsToSelfEnrol").AsBoolean().NotNullable().WithDefaultValue(0);
             Create.Table("FilteredAssets")
                .WithColumn("ID").AsInt32().NotNullable().PrimaryKey()
                .WithColumn("Title").AsString(255).NotNullable()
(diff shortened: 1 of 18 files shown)


DigitalLearningSolutions.Data.Migrations/202010161445_ApplyLPDefaultSPs.cs
```diff
@@ -20,12 +18,13 @@ public override void Up()
...
-                .WithColumn("CompletedDate").AsDateTime().Nullable()
-                .WithColumn("Duration").AsInt32().Nullable()
-                .WithColumn("Outcome").AsInt32().Nullable();
+                 .WithColumn("FilteredAssetID").AsInt32().NotNullable().ForeignKey("FilteredAssets", "ID")
+                 .WithColumn("LaunchedDate").AsDateTime().NotNullable().WithDefault(SystemMethods.CurrentDateTime)
+                 .WithColumn("LaunchCount").AsInt32().NotNullable().WithDefaultValue(1)
+                 .WithColumn("CompletedDate").AsDateTime().Nullable()
+                 .WithColumn("Duration").AsInt32().Nullable()
+                 .WithColumn("Outcome").AsInt32().Nullable()
+                 .WithColumn("Bookmarked").AsBoolean().Nullable();
```
FridaTveit, line 27: I guessed that this was missed accidentally? It was referenced in the sql script below. Let me know if this should be fixed in another way 🙂

DigitalLearningSolutions.Data.Migrations/202010161445_ApplyLPDefaultSPs.cs
```diff
@@ -2,12 +2,10 @@
 {
     using FluentMigrator;
     [Migration(202010161445)]
-    public class ApplyLPDefaultSPs :Migration
+    public class ApplyLPDefaultSPs : Migration
     {
         public override void Up()
         {
-            Alter.Table("Customisations")
-               .AddColumn("ApplyLPDefaultsToSelfEnrol").AsBoolean().NotNullable().WithDefaultValue(0);
```
FridaTveit, line 10: This was already present in the db as far as I can tell? 🙂

kevwhitt-hee:
Happy with those changes. I was aware of some problems with migration resulting from differences between the live and development databases (which migrations should help to avoid in future) but hadn't got round to fixing. There were some tests that I was struggling to interpret the error messages on too. I'm happy to merge your changes if you are ready?

FridaTveit:
Excellent! Sure, I'm happy for it to be merged as long as you're happy with my changes 🙂

PR #29, opened 2020-11-20
HEEDLS-NONE Fix down migrations

FridaTveit:
Just checking that all the down migrations work before applying the migrations to UAT, just in case there's a need to migrate down.

The change to the sql script isn't shown because it was originally saved with an encoding git didn't like. I've now saved it with UTF-8 encoding so changes should show up from now on. The only change in the script is that I removed:
```
/****** Object:  StoredProcedure [dbo].[InsertCustomisation_V3]    Script Date: 16/10/2020 14:40:40 ******/
DROP PROCEDURE [dbo].[InsertCustomisation_V3]
GO
```
That stored procedure hadn't been added by the apply script so I assume it was in the drop script by a mistake 🙂


kevwhitt-hee:
The mistake was omitting the CREATE script from the apply procedure called in the Up() migration. Sorry! This DB change relates to a change in the legacy application rather than the new one making it harder to spot mistakes. The DROP PROCEDURE deletion needs reverting and the CREATE ROCEDURE needs adding to the Up() SQL. Or would it be better handled in a new migration?

FridaTveit:
@kevwhitt-hee I see! I think if nothing in the current code relies on it being part of that migration, it would be better for you to add it in a new migration. Does that sound okay? 🙂

PR #30, opened 2020-11-20
HEEDLS-NONE Add script to fix inconsistencies in obfuscated database

FridaTveit:
There were some changes made to the database after the obfuscated database backup was created and before the project was setup to use migrations for database changes. Therefore there are some inconsistencies between the database backup and the live
& UAT databases. I've added a script to fix this which should be run on any locally created databases and on test.


kevwhitt-hee (approved):
Looks good - thank you

PR #36, opened 2020-11-24
HEEDLS-201 Change test old system url

FridaTveit:
Plan for SCORM player testing documented here: https://swiki.softwire.com/display/HEE/Testing.

The only code change necessary was to change the test application to point to the UAT version of the old system. This isn't strictly necessary, as we shouldn't be doing SCORM testing on test (we should be testing locally and on UAT). But just in case some content works and is launched during testing it would be better for the tracking for that content to be updated in the UAT database. As we're just using a test account it shouldn't matter much either way but might as well be safe 🙂


kwunkiuma (approved):
Looks good.

SChandarana (approved):

PR #37, opened 2020-11-24
HEEDLS-203 Menu Header Course information back end

benjimarshall:
Subtask of HEEDLS-173

### Changes
1. Add the following pieces of information to the `CourseContent` model:
    - Course title (`"{Application Name} - {Customisation Name}"`)
    - Average course duration
    - Centre name
    - Centre banner text (nullable)
2. Rewrite the `GetCourseContent` method in `CourseService`, to run a query similar to the one on the HEEDLS-173 ticket to get this information, trimmed to only get the required information
3. Modify the `InitialMenuViewModel` to collect these new fields and assign them to its own fields, including renaming one of the fields for consistency (`Name` to `Title`) 
4. Amend tests to use the new fields of the `CourseContent` and `InitialMenuViewModel`

### Testing
Refactored and amended the tests that use this code to cover the new fields. After this, ran all unit tests and they pass.

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
@@ -24,17 +24,17 @@
         public void Get_course_content_should_return_course()
         {
             // When
-            const int customisationId = 1;
+            const int customisationId = 1604;
             var result = courseContentService.GetCourseContent(customisationId);
 
             // Then
             var expectedCourse = new CourseContent(
-                1,
-                "Entry Level - Win XP, Office 2003/07 OLD",
-                "Standard",
-                "N/A",
-                "North West Boroughs Healthcare NHS Foundation Trust",
-                "xxxxxxxxxxxxxxxxxxxx"
+                1604,
+                "Level 2 - Microsoft Word 2007",
+                "Word Core 07 Testing",
+                "3h 56m",
+                "Northumbria Healthcare NHS Foundation Trust",
+                "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
             );
             result.Should().BeEquivalentTo(expectedCourse);
         }
--- a/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
@@ -28,12 +28,14 @@
             var result = courseContentService.GetCourseContent(customisationId);
 
             // Then
-            var expectedCourse = new CourseContent
-            {
-                Id = 1,
-                CustomisationName = "Standard",
-                ApplicationName = "Entry Level - Win XP, Office 2003/07 OLD"
-            };
+            var expectedCourse = new CourseContent(
+                1,
+                "Entry Level - Win XP, Office 2003/07 OLD",
+                "Standard",
+                "N/A",
+                "North West Boroughs Healthcare NHS Foundation Trust",
+                "xxxxxxxxxxxxxxxxxxxx"
+            );
             result.Should().BeEquivalentTo(expectedCourse);
         }
     }
--- a/DigitalLearningSolutions.Data/Models/CourseContent/CourseContent.cs
+++ b/DigitalLearningSolutions.Data/Models/CourseContent/CourseContent.cs
@@ -2,11 +2,11 @@
 {
     public class CourseContent
     {
-        public int Id { get; }
-        public string Title { get; }
-        public string AverageDuration { get; }
-        public string CentreName { get; }
-        public string? BannerText { get; }
+        public int Id { get; private set; }
+        public string Title { get; private set; }
+        public string AverageDuration { get; private set; }
+        public string CentreName { get; private set; }
+        public string? BannerText { get; private set; }
 
         public CourseContent(
             int id,
--- a/DigitalLearningSolutions.Data/Models/CourseContent/CourseContent.cs
+++ b/DigitalLearningSolutions.Data/Models/CourseContent/CourseContent.cs
@@ -2,8 +2,26 @@
 {
     public class CourseContent
     {
-        public string CustomisationName { get; set; }
-        public string ApplicationName { get; set; }
-        public int Id { get; set; }
+        public int Id { get; }
+        public string Title { get; }
+        public string AverageDuration { get; }
+        public string CentreName { get; }
+        public string? BannerText { get; }
+
+        public CourseContent(
+            int id,
+            string applicationName,
+            string customisationName,
+            string averageDuration,
+            string centreName,
+            string? bannerText
+        )
+        {
+            Id = id;
+            Title = $"{applicationName} - {customisationName}";
+            AverageDuration = averageDuration;
+            CentreName = centreName;
+            BannerText = bannerText;
+        }
     }
 }
--- a/DigitalLearningSolutions.Data/Services/CourseContentService.cs
+++ b/DigitalLearningSolutions.Data/Services/CourseContentService.cs
@@ -24,10 +24,16 @@
         public CourseContent? GetCourseContent(int customisationId)
         {
             return connection.QueryFirstOrDefault<CourseContent>(
-                @"SELECT C.CustomisationID AS Id, C.CustomisationName, A.ApplicationName 
-                      FROM Customisations as C
-                      JOIN Applications AS A ON C.ApplicationID = A.ApplicationID
-                      WHERE C.CustomisationID = @customisationId",
+                @"SELECT Customisations.CustomisationID AS id,
+                         Applications.ApplicationName,
+	                     Customisations.CustomisationName,
+	                     dbo.GetMinsForCustomisation(Customisations.CustomisationID) AS AverageDuration,
+	                     Centres.CentreName,
+	                     Centres.BannerText
+                  FROM Applications
+                  INNER JOIN Customisations ON Applications.ApplicationID = Customisations.ApplicationID
+                  INNER JOIN Centres ON Customisations.CentreID = Centres.CentreID
+                  WHERE Customisations.CustomisationID = @customisationId;",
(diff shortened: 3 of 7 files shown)


DigitalLearningSolutions.Data/Models/CourseContent/CourseContent.cs
```diff
@@ -2,8 +2,26 @@
...
     public class CourseContent
     {
-        public string CustomisationName { get; set; }
-        public string ApplicationName { get; set; }
-        public int Id { get; set; }
+        public int Id { get; }
+        public string Title { get; }
+        public string AverageDuration { get; }
+        public string CentreName { get; }
+        public string? BannerText { get; }
```
FridaTveit, line 9: Can these have private getters now that you're using a constructor? 🙂

DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
```diff
@@ -28,12 +28,14 @@ public void Get_course_content_should_return_course()
...
-            {
-                Id = 1,
-                CustomisationName = "Standard",
-                ApplicationName = "Entry Level - Win XP, Office 2003/07 OLD"
-            };
+            var expectedCourse = new CourseContent(
+                1,
+                "Entry Level - Win XP, Office 2003/07 OLD",
+                "Standard",
+                "N/A",
```
FridaTveit, line 35: Have you found a course which doesn't have this set to N/A? Might be useful to use one which doesn't have that for the tests, and would be good to see what the format usually is? 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/InitialMenuViewModel.cs
```diff
@@ -4,11 +4,17 @@
...
+        public string CentreName { get; }
+        public string? BannerText { get; }
 
         public InitialMenuViewModel(CourseContent courseContent)
         {
-            Name = $"{courseContent.ApplicationName} - {courseContent.CustomisationName}";
+            Title = courseContent.Title;
+            AverageDuration = courseContent.AverageDuration;
+            CentreName = courseContent.CentreName;
+            BannerText = courseContent.BannerText;
```
FridaTveit, line 17: You might as well just store the course content object now? 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -24,10 +24,16 @@ public CourseContentService(IDbConnection connection, ILogger<CourseContentServi
...
+                @"SELECT Customisations.CustomisationID AS id,
+                         Applications.ApplicationName,
+	                     Customisations.CustomisationName,
+	                     dbo.GetMinsForCustomisation(Customisations.CustomisationID) AS AverageDuration,
+	                     Centres.CentreName,
+	                     Centres.BannerText
+                  FROM Applications
+                  INNER JOIN Customisations ON Applications.ApplicationID = Customisations.ApplicationID
+                  INNER JOIN Centres ON Customisations.CentreID = Centres.CentreID
+                  WHERE Customisations.CustomisationID = @customisationId;",
```
FridaTveit, line 36: Could you also make this return the progress id? Found in the progress table by searching with the customisation id and candidate id (will need to be passed in). This will be needed later to fetch section info, and will be needed by Frank I've just realised 🙂

FridaTveit (commented):
A few comments, otherwise looks great. Nice quick work! 🙂

DigitalLearningSolutions.Data/Models/CourseContent/CourseContent.cs
benjimarshall: I don't think so, because the view model will need to be able to access these fields 🙂 I might be missing something here though

FridaTveit: Ah sorry, I meant `setter` 🤦

DigitalLearningSolutions.Data/Services/CourseContentService.cs
FridaTveit: Actually looking at some of the upcoming tickets I think this might make more sense as a separate method (because we might need to create the progress record if it doesn't exist). Sorry about the confusion @benjimarshall!

FridaTveit: @kwunkiuma if you create a service method to get the progress id (which will find it from progress table based on customisation id and candidate id), in addition to your update service method. Then you can call that in the controller to get the progress id, before calling your update method. Does that make sense? And do you think it fits with what you're doing @benjimarshall? 🙂

kwunkiuma: Yes, I should be able to do that.

DigitalLearningSolutions.Data/Models/CourseContent/CourseContent.cs
benjimarshall: Sure! Done in 4ec78e9 🙂

DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
benjimarshall: The general format is along the lines of "5h 32m", stored as a string. I've changed the test case in 4ec78e9 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/InitialMenuViewModel.cs
benjimarshall: Yeah that's much neater. Refactored in 4ec78e9 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/IntitialMenuViewModelTests.cs
```diff
@@ -30,10 +30,10 @@ public void Initial_menu_should_have_name()
...
 
             // Then
-            initialMenuViewModel.Title.Should().Be($"{applicationName} - {customisationName}");
-            initialMenuViewModel.AverageDuration.Should().Be(averageDuration);
-            initialMenuViewModel.CentreName.Should().Be(centreName);
-            initialMenuViewModel.BannerText.Should().Be(bannerText);
+            initialMenuViewModel.CourseContent.Title.Should().Be($"{applicationName} - {customisationName}");
+            initialMenuViewModel.CourseContent.AverageDuration.Should().Be(averageDuration);
+            initialMenuViewModel.CourseContent.CentreName.Should().Be(centreName);
+            initialMenuViewModel.CourseContent.BannerText.Should().Be(bannerText);
```
FridaTveit, line 36: You can just check that the course content is equal to the expected course content 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
benjimarshall: I think I follow what's going on here 🙂 Just to confirm, there's nothing with progressIDs that you want added to this branch?

DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
```diff
@@ -24,17 +24,17 @@ public void Setup()
...
-                1,
-                "Entry Level - Win XP, Office 2003/07 OLD",
-                "Standard",
-                "N/A",
-                "North West Boroughs Healthcare NHS Foundation Trust",
-                "xxxxxxxxxxxxxxxxxxxx"
+                1604,
+                "Level 2 - Microsoft Word 2007",
+                "Word Core 07 Testing",
+                "3h 56m",
```
FridaTveit, line 35: Nice, good to know this is the format 🙂 Also @SChandarana worth noting that it can be N/A. Worth checking what the old system displays in that case (whether it displays N/A or nothing) and doing the same 🙂

FridaTveit (commented):
One small suggestion, otherwise looks great 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
FridaTveit: No, nothing needed on this branch 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/IntitialMenuViewModelTests.cs
benjimarshall: Ah yes! Done in 8fbaf7c 🙂

FridaTveit (approved):
Looks great 🙂

DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
benjimarshall: The old system displays the duration in the same format. I've not managed to find an example where the duration is "N/A", but given that the underlying SQL function is just returning a string of either that format, or "N/A", I suspect the duration displayed by the old website will be "N/A" for courses without a duration. 

A major part of the SQL function is formatting, as well producing an average duration in minutes to format. I wonder if it would be nicer to bring that closer to the front end and instead have this query just return a nullable number of minutes. What do you think? 🙂

FridaTveit: We could do but I'd be tempted to leave it like this, at least if we don't have any issues with it. If we can reuse something like this from the old system without needing to modify it all then I think it's worth doing 🙂

FridaTveit: (I mainly meant to as a comment for Shiv to bear in mind when implementing the front end, sorry if that wasn't clear 🙂 )

benjimarshall: Ah I see. That makes sense 🙂

PR #38, opened 2020-11-24
HEEDLS-198 Update login count and duration

kwunkiuma:
Added an method for updating LoginCount and Duration in Progress; this required a new method for obtaining ProgressID from Progress using CandidateID and CustomisationID which was also added. This update will be ran whenever a learning menu is opened.

Ran and passed all unit tests then manually tested by changing the stored value then opening the page on Firefox.

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
@@ -12,6 +12,7 @@
     internal class CourseContentServiceTests
     {
         private CourseContentService courseContentService;
+        private CourseContentTestHelper courseContentTestHelper;
 
         [SetUp]
         public void Setup()
@@ -19,6 +20,7 @@
             var connection = ServiceTestHelper.GetDatabaseConnection();
             var logger = A.Fake<ILogger<CourseContentService>>();
             courseContentService = new CourseContentService(connection, logger);
+            courseContentTestHelper = new CourseContentTestHelper(connection);
         }
 
         [Test]
@@ -41,41 +43,43 @@
         [Test]
         public void Get_progress_id_should_return_progress_id()
         {
-            // When
+            // Given
             const int candidateId = 9;
             const int customisationId = 259;
+
+            // When
             var result = courseContentService.GetProgressId(candidateId, customisationId);
 
             // Then
-            var expectedProgressId = 10;
+            const int expectedProgressId = 10;
             result.Should().Be(expectedProgressId);
         }
 
         [Test]
         public void Update_login_count_should_not_increment_login_if_no_new_session()
         {
+            // Given
+            const int progressId = 10;
+            var expectedLoginCount = courseContentTestHelper.GetLoginCount(progressId);
+
             // When
-            const int candidateId = 9;
-            const int customisationId = 259;
-            var progressId = courseContentService.GetProgressId(candidateId, customisationId);
-            var expectedLoginCount = CourseContentTestHelper.GetLoginCount(progressId);
             courseContentService.UpdateLoginCountAndDuration(progressId);
-            var result = CourseContentTestHelper.GetLoginCount(progressId);
+            var result = courseContentTestHelper.GetLoginCount(progressId);
 
             // Then
             result.Should().Be(expectedLoginCount);
         }
 
         [Test]
-        public void Update_duration_should_not_increment_login_if_no_new_session()
+        public void Update_duration_should_not_increment_duration_if_no_new_session()
         {
+            // Given
+            const int progressId = 10;
+            var expectedDuration = courseContentTestHelper.GetDuration(progressId);
+
             // When
-            const int candidateId = 9;
-            const int customisationId = 259;
-            var progressId = courseContentService.GetProgressId(candidateId, customisationId);
-            var expectedDuration = CourseContentTestHelper.GetDuration(progressId);
             courseContentService.UpdateLoginCountAndDuration(progressId);
-            var result = CourseContentTestHelper.GetDuration(progressId);
+            var result = courseContentTestHelper.GetDuration(progressId);
 
             // Then
             result.Should().Be(expectedDuration);
@@ -84,36 +88,38 @@
         [Test]
         public void Update_login_count_should_increment_login_if_new_session()
         {
-            // When
+            // Given
             const int candidateId = 9;
             const int customisationId = 259;
+            const int progressId = 10;
             const int duration = 5;
-            const int active = 0;
             var loginTime = new DateTime(2010, 9, 23);
-            var progressId = courseContentService.GetProgressId(candidateId, customisationId);
-            var initialLoginCount = CourseContentTestHelper.GetLoginCount(progressId);
-            CourseContentTestHelper.InsertSession(candidateId, customisationId, loginTime, duration, active);
+            var initialLoginCount = courseContentTestHelper.GetLoginCount(progressId);
+
+            // When
+            courseContentTestHelper.InsertSession(candidateId, customisationId, loginTime, duration);
             courseContentService.UpdateLoginCountAndDuration(progressId);
-            var result = CourseContentTestHelper.GetLoginCount(progressId);
+            var result = courseContentTestHelper.GetLoginCount(progressId);
 
             // Then
             result.Should().Be(initialLoginCount + 1);
         }
 
         [Test]
-        public void Update_duration_should_increment_login_if_new_session()
+        public void Update_duration_should_increment_duration_if_new_session()
         {
-            // When
+            // Given
             const int candidateId = 9;
             const int customisationId = 259;
+            const int progressId = 10;
             const int duration = 5;
-            const int active = 0;
             var loginTime = new DateTime(2010, 9, 23);
-            var progressId = courseContentService.GetProgressId(candidateId, customisationId);
-            var initialDuration = CourseContentTestHelper.GetDuration(progressId);
-            CourseContentTestHelper.InsertSession(candidateId, customisationId, loginTime, duration, active);
+            var initialDuration = courseContentTestHelper.GetDuration(progressId);
+
+            // When
+            courseContentTestHelper.InsertSession(candidateId, customisationId, loginTime, duration);
             courseContentService.UpdateLoginCountAndDuration(progressId);
-            var result = CourseContentTestHelper.GetDuration(progressId);
+            var result = courseContentTestHelper.GetDuration(progressId);
 
             // Then
             result.Should().Be(initialDuration + duration);
--- a/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
@@ -1,5 +1,6 @@
 ﻿namespace DigitalLearningSolutions.Data.Tests.Services
 {
+    using System;
     using DigitalLearningSolutions.Data.Models.CourseContent;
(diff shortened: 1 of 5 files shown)


DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -31,5 +33,43 @@ public CourseContentService(IDbConnection connection, ILogger<CourseContentServi
...
             );
         }
+
+        public int GetProgressId(int candidateId, int customisationId)
+        {
+            return connection.QueryFirstOrDefault<int>(
+                @"SELECT ProgressId
+                        FROM Progress
+                        WHERE CandidateId = @candidateId
+                          AND customisationId = @customisationId",
```
FridaTveit, line 43: I'm guessing this works since you've tested it, but I'd change it to match the database casing (uppercase ID, uppercase c in CustomisationID) just for consistency 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -31,5 +33,43 @@ public CourseContentService(IDbConnection connection, ILogger<CourseContentServi
...
+                new { candidateId, customisationId }
+            );
+        }
+
+        public void UpdateLoginCountAndDuration(int progressId)
+        {
+            var numberOfAffectedRows = connection.Execute(
+                @"UPDATE Progress
+	                    SET LoginCount = (SELECT COALESCE(COUNT(*), 0)
+		                    FROM Sessions as s
```
FridaTveit, line 53: Another couple of consistency things:
* We generally write any keywords (such as "AS") in uppercase 
* We generally capitalise the table aliases, so "S" instead of "S"

🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -31,5 +33,43 @@ public CourseContentService(IDbConnection connection, ILogger<CourseContentServi
                 new { customisationId }
             );
         }
+
+        public int GetProgressId(int candidateId, int customisationId)
+        {
+            return connection.QueryFirstOrDefault<int>(
```
FridaTveit, line 39: So currently this will return 0 as the progress id if there's no progress record which is not what we want. However HEEDLS-202 is going to handle this. Worth just leaving a TODO comment here saying something like:
```
// TODO HEEDLS-202: change QueryFirstOrDefault to creating progress record if not found
```
Just so we don't forget 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -8,6 +8,8 @@
     public interface ICourseContentService
     {
         CourseContent? GetCourseContent(int customisationId);
+        int GetProgressId(int candidateId, int customisationId);
+        void UpdateLoginCountAndDuration(int progressId);
     }
 
     public class CourseContentService : ICourseContentService
```
FridaTveit, line 15: You need to update the CourseContentServiceTests to add tests for your new methods 🙂

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
```diff
@@ -36,6 +36,9 @@ public IActionResult Index(int customisationId)
                 return RedirectToAction("StatusCode", "LearningSolutions", new { code = 404 });
             }
 
+            var progressId = courseContentService.GetProgressId(User.GetCandidateId(), customisationId);
+            courseContentService.UpdateLoginCountAndDuration(progressId);
```
FridaTveit, line 40: Please add a test to the LearningMenuController checking that this method is called. We should actually have a test for the GetCourseContent method being called as well, would you mind adding that too? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
```diff
@@ -36,5 +37,86 @@ public void Get_course_content_should_return_course()
...
+        [Test]
+        public void Get_progress_id_should_return_progress_id()
+        {
+            // When
+            const int candidateId = 9;
+            const int customisationId = 259;
+            var result = courseContentService.GetProgressId(candidateId, customisationId);
+
+            // Then
+            var expectedProgressId = 10;
```
FridaTveit, line 50: Should be `const int` (the IDE should suggest this to you) 🙂

DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
```diff
@@ -36,5 +37,86 @@ public void Get_course_content_should_return_course()
...
+        }
+
+        [Test]
+        public void Update_login_count_should_not_increment_login_if_no_new_session()
+        {
+            // When
+            const int candidateId = 9;
+            const int customisationId = 259;
+            var progressId = courseContentService.GetProgressId(candidateId, customisationId);
+            var expectedLoginCount = CourseContentTestHelper.GetLoginCount(progressId);
```
FridaTveit, line 61: The lines above this comment should be grouped under `// Given`. The tests so be split into `// Given` (setup), `// When` (execute the method you want to test) and `// Then` (check the result) 🙂

DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
```diff
@@ -36,5 +37,86 @@ public void Get_course_content_should_return_course()
...
+            courseContentService.UpdateLoginCountAndDuration(progressId);
+            var result = CourseContentTestHelper.GetLoginCount(progressId);
+
+            // Then
+            result.Should().Be(expectedLoginCount);
+        }
+
+        [Test]
+        public void Update_duration_should_not_increment_login_if_no_new_session()
```
FridaTveit, line 70: Should be `should_not_increment_duration` 🙂

DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
```diff
@@ -36,5 +37,86 @@ public void Get_course_content_should_return_course()
...
+        }
+
+        [Test]
+        public void Update_login_count_should_increment_login_if_new_session()
+        {
+            // When
+            const int candidateId = 9;
+            const int customisationId = 259;
+            const int duration = 5;
+            const int active = 0;
```
FridaTveit, line 91: `active` isn't actually relevant to the test so can probably just always be set to 0 in the helper method 🙂

DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
```diff
@@ -36,5 +37,86 @@ public void Get_course_content_should_return_course()
...
+            result.Should().Be(expectedProgressId);
+        }
+
+        [Test]
+        public void Update_login_count_should_not_increment_login_if_no_new_session()
+        {
+            // When
+            const int candidateId = 9;
+            const int customisationId = 259;
+            var progressId = courseContentService.GetProgressId(candidateId, customisationId);
```
FridaTveit, line 60: Your tests ideally shouldn't rely on any method in the tested class other than the one they're testing. So I'd suggest hardcoding the progress id here 🙂

DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
```diff
@@ -85,5 +85,38 @@ public void ContentViewer_should_render_view()
...
+            const int progressId = 13;
+            var defaultCourseContent = CourseContentHelper.CreateDefaultCourseContent(CustomisationId);
+            A.CallTo(() => courseContentService.GetCourseContent(CustomisationId)).Returns(defaultCourseContent);
+            A.CallTo(() => courseContentService.GetProgressId(CandidateId, CustomisationId)).Returns(progressId);
+
+            // When
+            controller.Index(CustomisationId);
+
+            // Then
+            A.CallTo(() => courseContentService.GetCourseContent(CustomisationId)).MustHaveHappened();
```
FridaTveit, line 102: This assertion isn't necessary as it isn't relevant to what's being tested. I'd suggest testing that this is called in a separate test 🙂

FridaTveit (commented):
A few comments, let me know if you have any questions about any of them 🙂

DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
```diff
@@ -36,5 +39,90 @@ public void Get_course_content_should_return_course()
...
+            var loginTime = new DateTime(2010, 9, 23);
+            var initialDuration = courseContentTestHelper.GetDuration(progressId);
+
+            // When
+            courseContentTestHelper.InsertSession(candidateId, customisationId, loginTime, duration);
+            courseContentService.UpdateLoginCountAndDuration(progressId);
+            var result = courseContentTestHelper.GetDuration(progressId);
+
+            // Then
+            result.Should().Be(initialDuration + duration);
```
FridaTveit, line 125: Actually there's probably one more case we need to test; that the duration and login count don't get updated if the login time is not between the first submitted time and the submitted time. So I'd suggest adding two tests for that (one for duration and one for login) 🙂

DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
```diff
@@ -85,5 +85,50 @@ public void ContentViewer_should_render_view()
...
+
+            // When
+            controller.Index(1);
+
+            // Then
+            A.CallTo(() => courseContentService.GetCourseContent(customisationId)).MustHaveHappened();
+        }
+
+        [Test]
+        public void Index_valid_customisation_id_should_call_the_course_content_service()
```
FridaTveit, line 103: This test name isn't quite accurate, it doesn't describe what it should call on the course content service. Something like `should_update_login_and_duration` would be more descriptive I think 🙂

DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
```diff
@@ -85,5 +85,50 @@ public void ContentViewer_should_render_view()
             result.Should().BeViewResult()
                 .Model.Should().BeEquivalentTo(expectedModel);
         }
+
+        [Test]
+        public void Index_always_calls_checks_course_content()
```
FridaTveit, line 90: `calls_get_course_content` is more descriptive 🙂

DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
```diff
@@ -85,5 +85,50 @@ public void ContentViewer_should_render_view()
...
+        public void Index_invalid_customisation_id_should_not_call_the_course_content_service()
+        {
+            // Given
+            A.CallTo(() => courseContentService.GetCourseContent(CustomisationId)).Returns(null);
+
+            // When
+            controller.Index(CustomisationId);
+
+            // Then
+            A.CallTo(() => courseContentService.GetCourseContent(CustomisationId)).MustHaveHappened();
```
FridaTveit, line 129: This check isn't necessary

FridaTveit (commented):
The tests are looking good! Just a few small tweaks needed 🙂

FridaTveit (approved):
Looks good 🙂

PR #40, opened 2020-11-25
HEEDLS-204 menu header course information front end

SChandarana:
image

Few things I wasn't sure about. In the old website the test centre was the title for the box that contained all of the different tasks etc, I just put it on a new line as I wasn't sure where exactly it should go (since I'm not sure where the task etc will be viewed). 
I used secondary text colour for the banner and use a `<br/>` so that the banner text would be grouped with the centre name as using a new `<p>` would leave a large gap between them. I wasn't actually able to find an example of the banner text in the old website when I was looking through it so I wasn't exactly sure how it used to look.

### DIFF
--- a/DigitalLearningSolutions.Web/Views/LearningMenu/Index.cshtml
+++ b/DigitalLearningSolutions.Web/Views/LearningMenu/Index.cshtml
@@ -2,12 +2,19 @@
 @model InitialMenuViewModel
 
 @{
-  ViewData["Application"] = "Learning Menu";
+  ViewData["HeaderPrefix"] = "";
+  ViewData["Application"] = Model.CourseContent.Title;
   ViewData["Title"] = "Learning Menu";
 }
 
 <div class="nhsuk-grid-row">
   <div class="nhsuk-grid-column-full">
     <h1 id="page-heading">Course: @Model.CourseContent.Title</h1>
+    <h2 class="nhsuk-u-secondary-text-color">Average Course Time: @Model.CourseContent.AverageDuration</h2>
+    <p>
+      <span>Centre: @Model.CourseContent.CentreName</span>
+      <br/>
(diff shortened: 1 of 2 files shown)


DigitalLearningSolutions.Web/Views/LearningMenu/Index.cshtml
```diff
@@ -2,12 +2,19 @@
...
+  ViewData["Application"] = Model.CourseContent.Title;
   ViewData["Title"] = "Learning Menu";
 }
 
 <div class="nhsuk-grid-row">
   <div class="nhsuk-grid-column-full">
     <h1 id="page-heading">Course: @Model.CourseContent.Title</h1>
+    <h2 class="nhsuk-u-secondary-text-color">Average Course Time: @Model.CourseContent.AverageDuration</h2>
+    <p>
+      <span>Centre: @Model.CourseContent.CentreName</span>
```
FridaTveit, line 15: I'd leave out `Centre: ` 🙂

SChandarana:
I've noticed in the design docs that I could use a bottom padding override in the paragraph do avoid using a `<br/>`

FridaTveit (commented):

SChandarana:
image

FridaTveit (approved):
Looks good 🙂

SChandarana:
@FridaTveit Before I merge this PR I was wondering what your thoughts on the header link are. Currently the entire link is clickable (it redirects you to the old website learning portal). This is the same as the main learning portal page however it doesn't seem like the whole header should be a link as you wouldn't expect clicking "Excel 2013 for the Workplace" to redirect you to the portal page. Does that make sense?

Here are what the headers look like when you hover over them. I believe the self assessment page has the header text unclickable and the NHS logo redirects to the learning portal page.
image

image

FridaTveit:
@SChandarana good point. I'll raise a bug and discuss with Kevin (as this affects more than just the new pages we're adding) 🙂

PR #41, opened 2020-11-25
Heedls 174 menu header close button

SChandarana:
image

Button will redirect you to the current courses page. The code is a copy of what is used in the _layout.cshtml for the SelfAssessment page. May want to consider using it in a partial in future if it is used enough.

Wondering if it would be better to use a Chevron rather than the close icon as it is a "return" to rather than a "close"

### DIFF
--- a/DigitalLearningSolutions.Web/Views/LearningMenu/Index.cshtml
+++ b/DigitalLearningSolutions.Web/Views/LearningMenu/Index.cshtml
@@ -18,9 +18,9 @@
 
 <div class="nhsuk-back-link">
   <a class="nhsuk-back-link__link" asp-action="Current" asp-controller="LearningPortal">
-    <svg class="nhsuk-icon nhsuk-icon__chevron-left" focusable='false' xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" aria-hidden="true" focusable="false">
-      <path d="M13.41 12l5.3-5.29a1 1 0 1 0-1.42-1.42L12 10.59l-5.29-5.3a1 1 0 0 0-1.42 1.42l5.3 5.29-5.3 5.29a1 1 0 0 0 0 1.42 1 1 0 0 0 1.42 0l5.29-5.3 5.29 5.3a1 1 0 0 0 1.42 0 1 1 0 0 0 0-1.42z"></path>
+    <svg class="nhsuk-icon nhsuk-icon__chevron-left" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" aria-hidden="true">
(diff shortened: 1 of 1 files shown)


DigitalLearningSolutions.Web/Views/LearningMenu/Index.cshtml
```diff
@@ -15,3 +15,12 @@
     <p class="nhsuk-u-secondary-text-color">@Model.CourseContent.BannerText</p>
   </div>
 </div>
+
+<div class="nhsuk-back-link">
+  <a class="nhsuk-back-link__link" asp-action="Current" asp-controller="LearningPortal">
+    <svg class="nhsuk-icon nhsuk-icon__chevron-left" focusable='false' xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" aria-hidden="true" focusable="false">
+      <path d="M13.41 12l5.3-5.29a1 1 0 1 0-1.42-1.42L12 10.59l-5.29-5.3a1 1 0 0 0-1.42 1.42l5.3 5.29-5.3 5.29a1 1 0 0 0 0 1.42 1 1 0 0 0 1.42 0l5.29-5.3 5.29 5.3a1 1 0 0 0 1.42 0 1 1 0 0 0 0-1.42z"></path>
+    </svg>
+    Return to Courses
```
FridaTveit, line 24: "Return to your current activities" might be better 🙂 ("activities" rather than "courses" because they've changed to use that terminology on the front end I've noticed, probably because it includes self assessments in addition to courses 🙂 )

DigitalLearningSolutions.Web/Views/LearningMenu/Index.cshtml
```diff
@@ -15,3 +15,12 @@
     <p class="nhsuk-u-secondary-text-color">@Model.CourseContent.BannerText</p>
   </div>
 </div>
+
+<div class="nhsuk-back-link">
+  <a class="nhsuk-back-link__link" asp-action="Current" asp-controller="LearningPortal">
```
FridaTveit, line 20: I think you're right and a back chevron might be better than the x. We might need to change it later (at least for some of the other pages) because we might want some different back buttons (e.g. to bring the user back to the previous section). But we can deal with that if and when we get to it 🙂

FridaTveit (commented):

SChandarana:
image

DigitalLearningSolutions.Web/Views/LearningMenu/Index.cshtml
SChandarana: If we are referring to them as activities now, would we want to change the main page to have the phrase "Launch Activity" instead of "Launch Course" in a different ticket/PR

FridaTveit (approved):
Looks good 🙂

DigitalLearningSolutions.Web/Views/LearningMenu/Index.cshtml
FridaTveit: No because that button only appears on courses 🙂

FridaTveit: Self assessments say "Launch self assessment"

PR #43, opened 2020-11-26
HEEDLS-208 Initial Menu Page Front End

SChandarana:
I tried to reuse code from the course cards (but this time utilising the nhsuk course card css class). I had to hard code in the padding as it by default it was far too large.

With this method, the return to activity button is stuck all the way at the bottom. Would we prefer to put it just above the section cards?

What still needs to be added is:

- A `SectionCardViewModel` which I can give to the section card partial
- A list of `Sections` (custom class) added to the `courseContent` @benjimarshall is doing this on his end
- Change the current `for` loop to a `foreach section in courseContent.Sections`
- Change the hard coded values to use values from the View Models

image

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/TestHelpers/CourseSectionHelper.cs
+++ b/DigitalLearningSolutions.Web.Tests/TestHelpers/CourseSectionHelper.cs
@@ -14,7 +14,7 @@
                 sectionName,
                 hasLearning,
                 percentComplete
-                );
+            );
         }
     }
 }
--- a/DigitalLearningSolutions.Web.Tests/TestHelpers/CourseSectionHelper.cs
+++ b/DigitalLearningSolutions.Web.Tests/TestHelpers/CourseSectionHelper.cs
@@ -1,0 +1,20 @@
+﻿namespace DigitalLearningSolutions.Web.Tests.TestHelpers
+{
+    using DigitalLearningSolutions.Data.Models.CourseContent;
+
+    internal class CourseSectionHelper
+    {
+        public static CourseSection CreateDefaultCourseSection(
+            string sectionName = "SectionName",
+            bool hasLearning = true,
+            double percentComplete = 15.0
+        )
+        {
+            return new CourseSection(
+                sectionName,
+                hasLearning,
+                percentComplete
+                );
+        }
+    }
+}
--- a/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/IntitialMenuViewModelTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/IntitialMenuViewModelTests.cs
@@ -1,5 +1,6 @@
 ﻿namespace DigitalLearningSolutions.Web.Tests.ViewModels.LearningMenu
 {
+    using DigitalLearningSolutions.Data.Models.CourseContent;
     using DigitalLearningSolutions.Web.Tests.TestHelpers;
     using DigitalLearningSolutions.Web.ViewModels.LearningMenu;
     using FluentAssertions;
@@ -8,7 +9,7 @@
     public class InitialMenuViewModelTests
     {
         [Test]
-        public void Initial_menu_should_have_name()
+        public void Initial_menu_should_have_values_for_all_courseContent_fields()
         {
             // Given
             const int customisationId = 12;
@@ -25,13 +26,21 @@
                 centreName,
                 bannerText
             );
+            const string sectionName = "TestName";
+            const bool hasLearning = true;
+            const double percentComplete = 12.00;
+            var section = new CourseSection(sectionName, hasLearning, percentComplete);
+            expectedCourseContent.Sections.Add(section);
 
             // When
             var initialMenuViewModel = new InitialMenuViewModel(expectedCourseContent);
 
             // Then
-            initialMenuViewModel.CourseContent.Title.Should().Be($"{applicationName} - {customisationName}");
-            initialMenuViewModel.CourseContent.Should().BeEquivalentTo(expectedCourseContent);
+            initialMenuViewModel.Title.Should().Be($"{applicationName} - {customisationName}");
+            initialMenuViewModel.Id.Should().Be(customisationId);
+            initialMenuViewModel.AverageDuration.Should().Be(averageDuration);
+            initialMenuViewModel.CentreName.Should().Be(centreName);
+            initialMenuViewModel.BannerText.Should().Be(bannerText);
         }
     }
 }
--- a/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/IntitialMenuViewModelTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/IntitialMenuViewModelTests.cs
@@ -1,5 +1,6 @@
 ﻿namespace DigitalLearningSolutions.Web.Tests.ViewModels.LearningMenu
 {
+    using System.Linq;
     using DigitalLearningSolutions.Data.Models.CourseContent;
     using DigitalLearningSolutions.Web.Tests.TestHelpers;
     using DigitalLearningSolutions.Web.ViewModels.LearningMenu;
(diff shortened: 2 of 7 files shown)


FridaTveit:
@SChandarana Can you add a screenshot showing the full page, including the exit link and the footers? 🙂

FridaTveit:
The backend work for this has been merged so this can now be updated to:
* Use the view model values and add any necessary helper methods to the view model
* Update the view model tests if any helper methods are added

🙂

SChandarana:
Interesting, I'll have a quick play around to get this commit history to look correct

SChandarana:
I decided to put the back link just before the all the course sections were displayed but let me know if it would be preferred to have it at the bottom
image

image

SChandarana:
Just noticed I'll need to add tests for the view models. I'll start working on those now :)

DigitalLearningSolutions.Web/ViewModels/LearningMenu/InitialMenuViewModel.cs
```diff
@@ -1,14 +1,18 @@
 ﻿namespace DigitalLearningSolutions.Web.ViewModels.LearningMenu
 {
+    using System.Collections.Generic;
+    using System.Linq;
     using DigitalLearningSolutions.Data.Models.CourseContent;
 
     public class InitialMenuViewModel
     {
         public CourseContent CourseContent { get; }
+        public IEnumerable<SectionCardViewModel> Sections { get; }
```
FridaTveit, line 10: The sections are already part of the CourseContent so I don't think we should also store them in an additional variable. We can instead add a helper method to the model for getting the percentage complete 🙂

FridaTveit (commented):
One comment and as you've pointed out you need to add tests 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/InitialMenuViewModel.cs
SChandarana: I mainly did this to follow what had been done for the course cards. If I were to remove this, how would I pass the specific section to the partial cshtml I have made? As if I just pass `InitialMenuViewModel` it won't know which section to use. The other alternative is to have the foreach loop within the partial class and then it would work with `InitialMenuViewModel`. What do you think is the best way?

FridaTveit: Ah I see. In that case I would keep the section card view model but remove the CourseContent variable and instead store the parts of the course content that you need. Does that make sense? 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/IntitialMenuViewModelTests.cs
```diff
@@ -25,13 +26,21 @@ public void Initial_menu_should_have_name()
...
             var initialMenuViewModel = new InitialMenuViewModel(expectedCourseContent);
 
             // Then
-            initialMenuViewModel.CourseContent.Title.Should().Be($"{applicationName} - {customisationName}");
-            initialMenuViewModel.CourseContent.Should().BeEquivalentTo(expectedCourseContent);
+            initialMenuViewModel.Title.Should().Be($"{applicationName} - {customisationName}");
+            initialMenuViewModel.Id.Should().Be(customisationId);
+            initialMenuViewModel.AverageDuration.Should().Be(averageDuration);
+            initialMenuViewModel.CentreName.Should().Be(centreName);
+            initialMenuViewModel.BannerText.Should().Be(bannerText);
```
FridaTveit, line 43: One test for each assertion please 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/IntitialMenuViewModelTests.cs
```diff
@@ -25,13 +26,21 @@ public void Initial_menu_should_have_name()
                 centreName,
                 bannerText
             );
+            const string sectionName = "TestName";
+            const bool hasLearning = true;
+            const double percentComplete = 12.00;
+            var section = new CourseSection(sectionName, hasLearning, percentComplete);
+            expectedCourseContent.Sections.Add(section);
```
FridaTveit, line 33: You're not testing the sections here so no need to add a section. Worth testing that the sections list gets converted to sections view model list in a separate test 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionCardViewModelTest.cs
```diff
@@ -0,0 +1,45 @@
...
+    class SectionCardViewModelTest
+    {
+        [Test]
+        public void Section_should_return_empty_string_if_has_learning_is_false()
+        {
+            // Given
+            const string sectionName = "TestName";
+            const bool hasLearning = false;
+            const double percentComplete = 12.00;
+            var section = new CourseSection(sectionName, hasLearning, percentComplete);
```
FridaTveit, line 18: Probably worth creating a helper method like we have for creating course content, so that you can only override the fields you care about when constructing a section 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionCardViewModelTest.cs
```diff
@@ -0,0 +1,45 @@
...
+
+            // When
+            var sectionCardViewModel = new SectionCardViewModel(section);
+
+            // Then
+            sectionCardViewModel.HasLearning.Should().BeFalse();
```
FridaTveit, line 24: This check isn't really necessary. The percentage complete is what we really care about 🙂

SChandarana:
I added some tests but I wasn't too sure what exactly I needed to add. Something that is missing is the test for the section list in `InitialMenuViewModel` which should tests if there is a list of sections within the view model as I wasn't sure how to actually test this (as it has a list of `SectionCardViewModel` and not a list of `CourseSection`)

DigitalLearningSolutions.Web/ViewModels/LearningMenu/SectionCardViewModel.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace DigitalLearningSolutions.Web.ViewModels.LearningMenu
+{
+    using DigitalLearningSolutions.Data.Models.CourseContent;
+
+    public class SectionCardViewModel
+    {
+        public readonly string Title;
+        public readonly bool HasLearning;
+        public readonly double PercentComplete;
```
FridaTveit, line 9: Do these need to be public, I think only the title needs to be public? In particular the percent complete should not be public because we want the method to be used to get the percentage complete.

And for the field(s) which need to be public I'd suggest using properties with a public getter and no setter, just for consistency 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/SectionCardViewModel.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        public SectionCardViewModel(CourseSection section)
+        {
+            Title = section.Title;
+            HasLearning = section.HasLearning;
+            PercentComplete = section.PercentComplete;
+        }
+
+        public string GetPercentComplete()
```
FridaTveit, line 18: Instead of having this as method you could store the percent complete string which you can create in the constructor. Not very important, but it means you don't need to store more fields than you need 🙂

FridaTveit (commented):

FridaTveit:
To test the list of sections you create the initial page view model with a list of sections. Then you create a list of section view models and check that they match the list of the initial page view model 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/IntitialMenuViewModelTests.cs
```diff
@@ -11,27 +13,116 @@ public class InitialMenuViewModelTests
...
+            const int customisationId = 12;
+            var expectedCourseContent = CourseContentHelper.CreateDefaultCourseContent(
+                customisationId: customisationId
+            );
+
+            // When
+            var initialMenuViewModel = new InitialMenuViewModel(expectedCourseContent);
+
+            // Then
+            initialMenuViewModel.Id.Should().Be(12);
```
FridaTveit, line 43: Should reuse the customisation id variable here, for clarity 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/IntitialMenuViewModelTests.cs
```diff
@@ -11,27 +13,116 @@ public class InitialMenuViewModelTests
...
```
FridaTveit, line 59: Wrong check?

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/IntitialMenuViewModelTests.cs
```diff
@@ -11,27 +13,116 @@ public class InitialMenuViewModelTests
...
+
+        [Test]
+        public void Initial_menu_should_have_list_of_section_card_view_model()
+        {
+            // Given
+            const string sectionName = "TestSectionName";
+            const bool hasLearning = true;
+            const double percentComplete = 12.00;
+            var expectedCourseContent = CourseContentHelper.CreateDefaultCourseContent();
+            var section = CourseSectionHelper.CreateDefaultCourseSection(sectionName, hasLearning, percentComplete);
```
FridaTveit, line 117: Looks like you've forgotten to add your new helper to the PR? 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/IntitialMenuViewModelTests.cs
```diff
@@ -11,27 +13,116 @@ public class InitialMenuViewModelTests
...
+            var expectedSection = new SectionCardViewModel(section);
+            expectedCourseContent.Sections.Add(section);
+
+            // When
+            var initialMenuViewModel = new InitialMenuViewModel(expectedCourseContent);
+
+            // Then
+            initialMenuViewModel.Sections.First().Should().BeEquivalentTo(expectedSection);
```
FridaTveit, line 125: Probably worth checking the length as well. Unless you can just assert that the lists are equivalent? 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionCardViewModelTest.cs
```diff
@@ -0,0 +1,43 @@
...
+        [Test]
+        public void Section_should_return_correct_string_if_has_learning_is_true()
+        {
+            // Given
+            const bool hasLearning = true;
+            const double percentComplete = 12.00;
+            var section = CourseSectionHelper.CreateDefaultCourseSection(
+                hasLearning: hasLearning,
+                percentComplete: percentComplete
+                );
```
FridaTveit, line 34: Should be indented less 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/SectionCardViewModel.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace DigitalLearningSolutions.Web.ViewModels.LearningMenu
+{
+    using DigitalLearningSolutions.Data.Models.CourseContent;
+
+    public class SectionCardViewModel
+    {
+        public string Title { get; }
+        public string PercentComplete { get; private set; }
+        private bool HasLearning;
```
FridaTveit, line 9: Do you need to store this? 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/SectionCardViewModel.cs
```diff
@@ -0,0 +1,19 @@
...
+        
+
```
FridaTveit, line 11: Unneeded empty line 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web.Tests/TestHelpers/CourseSectionHelper.cs
```diff
@@ -0,0 +1,20 @@
...
+            string sectionName = "SectionName",
+            bool hasLearning = true,
+            double percentComplete = 15.0
+        )
+        {
+            return new CourseSection(
+                sectionName,
+                hasLearning,
+                percentComplete
+                );
```
FridaTveit, line 17: Indentation a bit off 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/IntitialMenuViewModelTests.cs
```diff
@@ -40,23 +41,23 @@ public void Initial_menu_should_have_id()
...
-            initialMenuViewModel.Id.Should().Be(12);
+            initialMenuViewModel.Id.Should().Be(customisationId);
         }
 
         [Test]
         public void Initial_menu_should_have_averageDuration()
         {
             // Given
-            const int customisationId = 12;
+            const string averageDuration = "TestDuration";
```
FridaTveit, line 51: Very minor but I think it's best to set this to the format we expect it to be in. Otherwise it can be a bit confusing 🙂

FridaTveit (commented):
Two minor things, otherwise looks good 🙂

FridaTveit (approved):
Looks good 🙂

PR #44, opened 2020-11-26
HEEDLS-209 Back end for initial course page

benjimarshall:
### Changes
Add Section class, containing basic data about a candidate's progress
through a specific section.

Amend GetCourseContent class to contain a list of sections, and the get
method in the CourseContentService to collect all of this.

Modify LearningMenuController to use the candidate ID to find a course
overview, so details about progress through each section can be given.

Amend tests to reflect new structure of models, and add some to cover
the edge cases of getting a course.

### Testing
Ran all tests after updating them, all tests passed.

### DIFF
--- a/DigitalLearningSolutions.Data/Models/CourseContent/CourseContent.cs
+++ b/DigitalLearningSolutions.Data/Models/CourseContent/CourseContent.cs
@@ -5,12 +5,12 @@
 
     public class CourseContent
     {
-        public int Id { get; private set; }
-        public string Title { get; private set; }
-        public string AverageDuration { get; private set; }
-        public string CentreName { get; private set; }
-        public string? BannerText { get; private set; }
-        public List<CourseSection> Sections { get; private set; } = new List<CourseSection>();
+        public int Id { get; }
+        public string Title { get; }
+        public string AverageDuration { get; }
+        public string CentreName { get; }
+        public string? BannerText { get; }
+        public List<CourseSection> Sections { get; } = new List<CourseSection>();
 
         public CourseContent(
             int id,
--- a/DigitalLearningSolutions.Data/Models/CourseContent/CourseContent.cs
+++ b/DigitalLearningSolutions.Data/Models/CourseContent/CourseContent.cs
@@ -1,5 +1,8 @@
 ﻿namespace DigitalLearningSolutions.Data.Models.CourseContent
 {
+    using System.Collections.Generic;
+    using DigitalLearningSolutions.Data.Models.Courses;
+
     public class CourseContent
     {
         public int Id { get; private set; }
@@ -7,6 +10,7 @@
         public string AverageDuration { get; private set; }
         public string CentreName { get; private set; }
         public string? BannerText { get; private set; }
+        public List<CourseSection> Sections { get; private set; } = new List<CourseSection>();
 
         public CourseContent(
             int id,
--- a/DigitalLearningSolutions.Data/Services/CourseContentService.cs
+++ b/DigitalLearningSolutions.Data/Services/CourseContentService.cs
@@ -50,6 +50,7 @@
                   LEFT JOIN Progress ON Customisations.CustomisationID = Progress.CustomisationID AND Progress.CandidateID = @candidateId
                   LEFT JOIN aspProgress ON aspProgress.ProgressID = Progress.ProgressID AND aspProgress.TutorialID = Tutorials.TutorialID
                   WHERE Customisations.CustomisationID = @customisationId
+                    AND ((Progress.CandidateID IS NULL) OR (Progress.SystemRefreshed = 0 AND Progress.RemovedDate IS NULL))
                   GROUP BY
                          Sections.SectionID,
                          Customisations.CustomisationID,
@@ -102,7 +103,7 @@
 	                    WHERE Progress.ProgressID = @progressId",
                 new { progressId }
             );
-            
+
             if (numberOfAffectedRows < 1)
             {
                 logger.LogWarning(
--- a/DigitalLearningSolutions.Data/Services/CourseContentService.cs
+++ b/DigitalLearningSolutions.Data/Services/CourseContentService.cs
@@ -1,13 +1,14 @@
 ﻿namespace DigitalLearningSolutions.Data.Services
 {
     using System.Data;
+    using System.Linq;
     using Dapper;
     using DigitalLearningSolutions.Data.Models.CourseContent;
     using Microsoft.Extensions.Logging;
 
     public interface ICourseContentService
     {
-        CourseContent? GetCourseContent(int customisationId);
+        CourseContent? GetCourseContent(int candidateId, int customisationId);
         int GetProgressId(int candidateId, int customisationId);
         void UpdateLoginCountAndDuration(int progressId);
     }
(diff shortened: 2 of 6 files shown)


DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -23,21 +24,52 @@ public CourseContentService(IDbConnection connection, ILogger<CourseContentServi
...
         }
 
-        public CourseContent? GetCourseContent(int customisationId)
+        public CourseContent? GetCourseContent(int candidateId, int customisationId)
         {
-            return connection.QueryFirstOrDefault<CourseContent>(
-                @"SELECT Customisations.CustomisationID AS id,
+            CourseContent? courseContent = null;
+            return connection.Query<CourseContent, CourseSection, CourseContent>(
+                @"SELECT TOP 10000 Customisations.CustomisationID AS id,
```
FridaTveit, line 31: Should `TOP 1000` be here? 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -23,21 +24,52 @@ public CourseContentService(IDbConnection connection, ILogger<CourseContentServi
...
+                  INNER JOIN Sections ON Sections.ApplicationID = Applications.ApplicationID
                   INNER JOIN Centres ON Customisations.CentreID = Centres.CentreID
-                  WHERE Customisations.CustomisationID = @customisationId;",
-                new { customisationId }
-            );
+                  LEFT JOIN Tutorials ON Sections.SectionID = Tutorials.SectionID
+                  LEFT JOIN Progress ON Customisations.CustomisationID = Progress.CustomisationID AND Progress.CandidateID = @candidateId
+                  LEFT JOIN aspProgress ON aspProgress.ProgressID = Progress.ProgressID AND aspProgress.TutorialID = Tutorials.TutorialID
+                  WHERE Customisations.CustomisationID = @customisationId
+                  GROUP BY
```
FridaTveit, line 53: Is the group by necessary? 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -23,21 +24,52 @@ public CourseContentService(IDbConnection connection, ILogger<CourseContentServi
...
+                         Sections.SectionID, Customisations.CustomisationID,
```
FridaTveit, line 54: Customisation id should be on a new line for consistency 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data/Services/CourseContentService.cs
benjimarshall: I think it might be, because of the aggregate functions `SUM` and `COUNT` on line 43. There might be another way of doing this though, maybe using subqueries? 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
benjimarshall: Nope, sorry about that! Fixed in 0635e51 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
benjimarshall: Whoops, fixed in 0635e51 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -23,21 +24,53 @@ public CourseContentService(IDbConnection connection, ILogger<CourseContentServi
...
+                         END) AS PercentComplete
                   FROM Applications
                   INNER JOIN Customisations ON Applications.ApplicationID = Customisations.ApplicationID
+                  INNER JOIN Sections ON Sections.ApplicationID = Applications.ApplicationID
                   INNER JOIN Centres ON Customisations.CentreID = Centres.CentreID
-                  WHERE Customisations.CustomisationID = @customisationId;",
-                new { customisationId }
-            );
+                  LEFT JOIN Tutorials ON Sections.SectionID = Tutorials.SectionID
+                  LEFT JOIN Progress ON Customisations.CustomisationID = Progress.CustomisationID AND Progress.CandidateID = @candidateId
```
FridaTveit, line 50: You also need to check here that SystemRefreshed is 0 and RemovedDate is NULL 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
FridaTveit: OKay, fair enough 🙂

FridaTveit (commented):
One thing, otherwise looks good 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
benjimarshall: Sorry I hadn't realised that those fields needed to be checked! This query was closely based on the `uspReturnSectionsForCandCust_V2` stored procedure and I don't think this was being checked there. Then again I don't understand everything going on in that query, so it might have been being checked in a different way.

I've added a check in 48d2aed 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -50,6 +50,7 @@ FROM Applications
                   LEFT JOIN Progress ON Customisations.CustomisationID = Progress.CustomisationID AND Progress.CandidateID = @candidateId
                   LEFT JOIN aspProgress ON aspProgress.ProgressID = Progress.ProgressID AND aspProgress.TutorialID = Tutorials.TutorialID
                   WHERE Customisations.CustomisationID = @customisationId
+                    AND ((Progress.CandidateID IS NULL) OR (Progress.SystemRefreshed = 0 AND Progress.RemovedDate IS NULL))
```
FridaTveit, line 53: Why the check for candidate id being null? In what case would it be null? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data/Services/CourseContentService.cs
FridaTveit: I assume the stored procedure doesn't check that because it's checking for a specific progress id 🙂 And yes I should have mentioned those fields needed checking earlier, I only recently realised when helping Frank with his ticket. When the old code fetches a progress id it will check that the customisation id and candidate id match + checking the removed date and system refreshed. That's because there can be more than one progress entry per user & customisation. For example if they unenrol from a course, a new progress entry should be created if they enrol again 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
benjimarshall: When the candidate is not doing this course, or more strictly when there isn't an entry in the progress table for the given customisation and candidate IDs. We still want to get the general the course information, as per our chat on Slack on Wednesday evening, just with the percentage completion set to 0.

The way I implemented this was to `LEFT JOIN` on the Progress table, to allow the progress-related fields to be `null`. The only potentially `null` field that is selected is the `PercentComplete` field, which is overwritten with 0 if `Progress.CandidateId` is `null` Is this a sensible way of solving that problem? 🙂

FridaTveit: Ah okay, that makes sense. Thanks for explaining 🙂

DigitalLearningSolutions.Data/Models/CourseContent/CourseContent.cs
```diff
@@ -1,12 +1,16 @@
...
+    using DigitalLearningSolutions.Data.Models.Courses;
+
     public class CourseContent
     {
         public int Id { get; private set; }
         public string Title { get; private set; }
         public string AverageDuration { get; private set; }
         public string CentreName { get; private set; }
         public string? BannerText { get; private set; }
+        public List<CourseSection> Sections { get; private set; } = new List<CourseSection>();
```
FridaTveit, line 13: Something I noticed when looking through the code yesterday, you can just remove the `private set` as that's equivalent to not having a set at all. The IDE should warn you of this by showing it as greyed out 🙂

FridaTveit (commented):
One small thing I didn't notice ealier, otherwise looks good 🙂

DigitalLearningSolutions.Data/Models/CourseContent/CourseContent.cs
benjimarshall: Yeah I think misunderstood one of your reviews for #37 / HEEDLS-203 here. Fixed in 7c72149 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
benjimarshall: The fact it needed explaining suggests there probably should be a comment about this in the code, which I've added in e54a187 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
benjimarshall: Ah I see! Thanks for explaining that 🙂

FridaTveit (approved):
Looks great 🙂

PR #45, opened 2020-11-26
Put tests with changes in transaction scope

kwunkiuma:
Added transaction scopes around tests that change the database so that the changes are not made.

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
@@ -1,6 +1,7 @@
 ﻿namespace DigitalLearningSolutions.Data.Tests.Services
 {
     using System;
+    using System.Transactions;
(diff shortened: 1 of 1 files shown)


DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
```diff
@@ -98,13 +99,16 @@ public void Update_login_count_should_not_increment_login_count_if_session_time_
             var loginTime = new DateTime(2011, 9, 23);
             var expectedLoginCount = courseContentTestHelper.GetLoginCount(progressId);
```
FridaTveit, line 101: All the update tests should be in transaction scope, also the ones were we don't expect them to change anything. Because if those tests fail then they will have changed the db, and we still don't want that change to persist 🙂

FridaTveit (commented):

FridaTveit (approved):
Looks good 🙂

PR #47, opened 2020-11-27
HEEDLS-172 Add brand logo to couse page

benjimarshall:
## Changes
Add brand logos to course page if centre logo is not available.

Add a LogoService to manage finding logos for a page, which replaces
GetCentreLogo in the CentresService. This will find the best logo based
on the given criteria, for a centreId and customisationId. If a
customisationId is not given the centre logo is still returned if it
exists.

Replace the CentreLogo model with the more general Logo model for both
centre and brand logos. Move the resizing of the logos from the service
to CSS (because that requires height and width fields for the logos,
which are not recorded in the Brand table unlike the Centre logo).

Pass the customisationId from the LearningMenu view up to the _Layout
and into the LogoViewComponent, who uses it to call the LogoService to
get the logo.

Refactor and simplify the logo ViewModel and View, and control its size
using a CSS class in layout.css instead of in these classes.

Remove GetCentreLogo tests, and add tests for the LogoService.
## Testing
Added LogoService unit tests, and all tests. Tested the alt text with
a screen reader, and tested in Chrome, Edge, IE11, and Firefox for each 
of the three types of logo. 

### Current activities when centre does have a logo:
current_activities_centre_logo
### Current activities when centre does not have a logo:
current_activities_no_centre_logo

### Course page when there is no logo available, or a mismatch between course's and user's `centreId`
course_page_no_logo
### Course page with a centre logo
course_page_centre_logo
### Course page when there is no centre logo, but there is a matching brand logo
course_page_brand_logo


DigitalLearningSolutions.Data.Tests/Services/LogoServiceTests.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        [Test]
+        public void GetLogo_Should_Return_Null_LogoUrl_When_No_Logo_Is_Available()
+        {
+            // When
+            var result = logoService.GetLogo(3, 100);
+
+            // Then
+            result.LogoUrl.Should().BeNull();
```
FridaTveit, line 62: It probably makes more sense for the whole logo object to be null in this case? 🙂

DigitalLearningSolutions.Data/Services/LogoService.cs
```diff
@@ -0,0 +1,55 @@
...
+                   WHERE Centres.CentreID = @centreId;
+
+                  SELECT Brands.BrandName AS LogoName, Brands.BrandLogo AS LogoData
+                    FROM Customisations
+                         INNER JOIN Applications
+                         ON Applications.ApplicationID = Customisations.ApplicationID
+
+                         INNER JOIN Brands
+                         ON Applications.BrandID = Brands.BrandID
+                   WHERE Customisations.CustomisationID = @customisationId AND CentreID = @centreId;",
```
FridaTveit, line 38: Why use two queries instead of just one? 🙂

DigitalLearningSolutions.Data/Models/Logo.cs
```diff
@@ -0,0 +1,29 @@
+﻿namespace DigitalLearningSolutions.Data.Models
+{
+    using System;
+
+    public class Logo
```
FridaTveit, line 5: What about the logic to scale the image, is that no longer needed? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data/Models/Logo.cs
benjimarshall: The logic to scale the image needs the current native width and height of the image, which are not fields in the Brands table so to use the old logic we would either need to parse the JPEG to find these values. The approach I took was to let the image coming out of the LogoService be whatever size it likes, and control its size using CSS `max-height` and `max-width`, using the same values as what was in the previous code (width of `280px` and height of `75px`). Using this approach the image scaling logic isn't needed for it is done by the browser instead. Which approach do you prefer? 🙂

DigitalLearningSolutions.Data/Services/LogoService.cs
benjimarshall: It felt like I was querying two discrete pieces of data (the Centre Logo, and the Brand Logo) so two queries felt natural. 

I could join the queries on the `centreId` and have a logo constructor that takes two sets of logo information and stores the first non-null one, if you think that would be better? 🙂

DigitalLearningSolutions.Data/Models/Logo.cs
FridaTveit: Okay. As long as the centre logos look the same as they did before then I'm happy 🙂

DigitalLearningSolutions.Data/Services/LogoService.cs
FridaTveit: Yes, I think I would prefer that 🙂

DigitalLearningSolutions.Data/Models/Logo.cs
benjimarshall: I've check 5 examples and they look the same, and I'm fairly sure its functionally equivalent 🙂

DigitalLearningSolutions.Data.Tests/Services/LogoServiceTests.cs
```diff
@@ -0,0 +1,97 @@
...
+            var result = logoService.GetLogo(25, 1176);
+
+            // Then
+            result.Should().NotBeNull();
+            result.LogoUrl.Should().NotBeNullOrEmpty();
+            result.Name.Should().Be("Royal Cornwall Hospitals NHS Trust");
+        }
+
+        [Test]
+        public void Get_Logo_Should_Return_Centre_logo_When_CustomisationId_Is_Null()
```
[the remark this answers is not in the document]
benjimarshall: Yep, sorry about that! Fixed in 60805a3 🙂

DigitalLearningSolutions.Data.Tests/Services/LogoServiceTests.cs
benjimarshall: Yeah I agree. Fixed when I refactored the query in 60805a3 🙂

DigitalLearningSolutions.Data/Services/LogoService.cs
benjimarshall: Refactored to this in 60805a3 🙂

DigitalLearningSolutions.Web/ViewComponents/LogoViewComponent.cs
```diff
@@ -1,35 +1,26 @@
...
+        private readonly ILogoService logoService;
+
+        public LogoViewComponent(ILogoService logoService)
+        {
+            this.logoService = logoService;
+        }
+
+        public IViewComponentResult Invoke(int? customisationId)
+        {
+            var centreId = UserClaimsPrincipal.GetCustomClaimAsInt(CustomClaimTypes.UserCentreId);
```
FridaTveit, line 19: There's a helper method for this, `User.GetCentreId` 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web/ViewComponents/LogoViewComponent.cs
benjimarshall: Ah I didn't realise that, thanks! 🙂 I had to cast types to make this work like this: `((ClaimsPrincipal) User).GetCentreId()`, I hope that's ok! Done in 1f0fb8c 🙂

DigitalLearningSolutions.Data/Models/Logo.cs
```diff
@@ -7,23 +7,45 @@ public class Logo
...
-                string base64Logo = Convert.ToBase64String(logoData);
-                LogoUrl = $"data:{logoMimeType};base64,{base64Logo}";
+                LogoUrl = brandLogoData;
+                Name = brandName!;
+                return;
             }
 
-            Name = logoName;
+            LogoUrl = null;
+            Name = "";
```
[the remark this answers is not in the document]
benjimarshall: Yeah that makes sense! Changed over in 1f0fb8c 🙂

DigitalLearningSolutions.Data/Services/LogoService.cs
```diff
@@ -19,37 +19,28 @@ public LogoService(IDbConnection connection)
...
+                         LEFT JOIN Applications
                          ON Applications.ApplicationID = Customisations.ApplicationID
 
-                         INNER JOIN Brands
+                         LEFT JOIN Brands
                          ON Applications.BrandID = Brands.BrandID
-                   WHERE Customisations.CustomisationID = @customisationId AND CentreID = @centreId;",
-                new { centreId, customisationId });
+                   WHERE Centres.CentreID = @centreId;",
+                new { centreId, customisationId }).FirstOrDefault();
```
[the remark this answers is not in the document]
benjimarshall: Yep! Done in 1f0fb8c 🙂

DigitalLearningSolutions.Data/Models/Logo.cs
```diff
@@ -7,23 +7,45 @@ public class Logo
...
+        private string? ParseLogoData(byte[]? logoData, string? logoMimeType = null)
         {
+            logoMimeType ??= "jpeg";
+
+            if (logoData == null || logoData.Length < 10)
+            {
+                return null;
+            }
+
+            string base64Logo = Convert.ToBase64String(logoData);
```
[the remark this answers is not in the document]
benjimarshall: Yep! Done in 1f0fb8c 🙂

DigitalLearningSolutions.Data/Services/LogoService.cs
```diff
@@ -0,0 +1,55 @@
...
+        public LogoService(IDbConnection connection)
+        {
+            this.connection = connection;
+        }
+
+        public Logo? GetLogo(int? centreId, int? customisationId)
+        {
+            try
+            {
+                var logo = connection.QueryFirstOrDefault<Logo>(
```
FridaTveit, line 25: Could just return here 🙂

FridaTveit (commented):
One tiny thing, otherwise looks good 🙂

DigitalLearningSolutions.Data/Services/LogoService.cs
benjimarshall: Yep! Fixed in 7aa6523 🙂

FridaTveit (approved):
Looks good 🙂

PR #48, opened 2020-11-27
HEEDLS-213 Add session state to app

FridaTveit:
Did a quick proof of concept to check using cookies to store session id would work. Most of it has just been added as details to HEEDLS-212 but this seemed worth just adding now 🙂


benjimarshall (approved):
Looks good to me! 🙂

PR #49, opened 2020-11-30
Create progress entry on first launch

kwunkiuma:
Add a new method to check if a given entry exists in the progress table and executes the uspCreateProgressRecord_V3 stored procedure if the entry doesn't exist.

The stored procedure has a few checks and doesn't insert the entry if any fail. This will be logged but the program will still try to run GetProgress method so this has been kept as QueryFirstOrDefault to prevent crashing if the entry wasn't added.

Ran and passed all unit tests and manually tested on Firefox, "Current Activities" will show newly enrolled courses.

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
@@ -173,17 +173,20 @@
             const int centreId = 53;
             var initialDoesProgressExist = courseContentTestHelper.DoesProgressExist(candidateId, customisationId);
 
-            // When
-            var result = courseContentService.GetOrCreateProgressId(candidateId, customisationId, centreId);
+            using (new TransactionScope())
+            {
+                // When
+                var result = courseContentService.GetOrCreateProgressId(candidateId, customisationId, centreId);
 
-            // Then
-            initialDoesProgressExist.Should().BeTrue();
-            const int expectedProgressId = 10;
-            result.Should().Be(expectedProgressId);
+                // Then
+                initialDoesProgressExist.Should().BeTrue();
+                const int expectedProgressId = 10;
+                result.Should().Be(expectedProgressId);
+            }
         }
 
         [Test]
-        public void Insert_new_progress_should_insert_progress()
+        public void Get_or_create_progress_id_should_insert_new_progress_if_does_not_exist()
         {
             // Given
             const int candidateId = 187251;
@@ -195,12 +198,12 @@
             {
                 // When
                 var progressId = courseContentService.GetOrCreateProgressId(candidateId, customisationId, centreId);
-                var result = courseContentTestHelper.DoesProgressExist(candidateId, customisationId);
+                var isProgressAdded = courseContentTestHelper.DoesProgressExist(candidateId, customisationId);
 
                 //Then
                 progressId.Should().NotBeNull();
                 initialDoesProgressExist.Should().BeFalse();
-                result.Should().BeTrue();
+                isProgressAdded.Should().BeTrue();
             }
         }
 
--- a/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
@@ -165,46 +165,21 @@
         }
 
         [Test]
-        public void Get_progress_id_should_return_progress_id()
+        public void Get_existing_progress_id_should_return_progress_id()
         {
             // Given
             const int candidateId = 9;
             const int customisationId = 259;
+            const int centreId = 53;
+            var initialDoesProgressExist = courseContentTestHelper.DoesProgressExist(candidateId, customisationId);
 
             // When
-            var result = courseContentService.GetProgressId(candidateId, customisationId);
+            var result = courseContentService.GetOrCreateProgressId(candidateId, customisationId, centreId);
 
             // Then
+            initialDoesProgressExist.Should().BeTrue();
             const int expectedProgressId = 10;
             result.Should().Be(expectedProgressId);
-        }
-
-        [Test]
-        public void Does_progress_exist_returns_true_for_existing_progress()
-        {
-            // Given
-            const int candidateId = 9;
-            const int customisationId = 259;
-
-            // When
-            var result = courseContentService.DoesProgressExist(candidateId, customisationId);
-
-            // Then
-            result.Should().BeTrue();
-        }
-
-        [Test]
-        public void Does_progress_exist_returns_false_for_non_existing_progress()
-        {
-            // Given
-            const int candidateId = 123;
-            const int customisationId = 123;
-
-            // When
-            var result = courseContentService.DoesProgressExist(candidateId, customisationId);
-
-            // Then
-            result.Should().BeFalse();
         }
 
         [Test]
@@ -214,15 +189,16 @@
             const int candidateId = 187251;
             const int customisationId = 17468;
             const int centreId = 549;
-            var initialDoesProgressExist = courseContentService.DoesProgressExist(candidateId, customisationId);
+            var initialDoesProgressExist = courseContentTestHelper.DoesProgressExist(candidateId, customisationId);
 
             using (new TransactionScope())
             {
                 // When
-                courseContentService.InsertNewProgress(candidateId, customisationId, centreId);
-                var result = courseContentService.DoesProgressExist(candidateId, customisationId);
+                var progressId = courseContentService.GetOrCreateProgressId(candidateId, customisationId, centreId);
+                var result = courseContentTestHelper.DoesProgressExist(candidateId, customisationId);
 
                 //Then
+                progressId.Should().NotBeNull();
                 initialDoesProgressExist.Should().BeFalse();
                 result.Should().BeTrue();
             }
@@ -361,7 +337,6 @@
         {
             // Given
             const int progressId = 10;
-            var initialSubmittedTime = courseContentTestHelper.GetSubmittedTime(progressId);
 
             using (new TransactionScope())
             {
--- a/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
@@ -180,6 +180,55 @@
         }
 
         [Test]
+        public void Does_progress_exist_returns_true_for_existing_progress()
+        {
+            // Given
+            const int candidateId = 9;
+            const int customisationId = 259;
+
+            // When
+            var result = courseContentService.DoesProgressExist(candidateId, customisationId);
+
+            // Then
+            result.Should().BeTrue();
+        }
+
+        [Test]
+        public void Does_progress_exist_returns_false_for_non_existing_progress()
+        {
+            // Given
+            const int candidateId = 123;
+            const int customisationId = 123;
+
+            // When
+            var result = courseContentService.DoesProgressExist(candidateId, customisationId);
+
+            // Then
+            result.Should().BeFalse();
+        }
+
+        [Test]
+        public void Insert_new_progress_should_insert_progress()
+        {
+            // Given
+            const int candidateId = 187251;
+            const int customisationId = 17468;
+            const int centreId = 549;
+            var initialDoesProgressExist = courseContentService.DoesProgressExist(candidateId, customisationId);
+
+            using (new TransactionScope())
+            {
+                // When
+                courseContentService.InsertNewProgress(candidateId, customisationId, centreId);
+                var result = courseContentService.DoesProgressExist(candidateId, customisationId);
+
+                //Then
+                initialDoesProgressExist.Should().BeFalse();
+                result.Should().BeTrue();
+            }
+        }
+
+        [Test]
         public void Update_progress_should_not_increment_login_count_if_no_new_session()
         {
             // Given
--- a/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
@@ -165,7 +165,7 @@
         }
 
         [Test]
-        public void Get_existing_progress_id_should_return_progress_id()
+        public void Get_or_create_progress_id_should_return_progress_id_if_exists()
         {
             // Given
             const int candidateId = 9;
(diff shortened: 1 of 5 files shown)


DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -80,18 +82,67 @@ GROUP BY
...
                         WHERE CandidateID = @candidateId
-                          AND CustomisationID = @customisationId",
+                          AND CustomisationID = @customisationId
+                          AND SystemRefreshed = 0
+                          AND RemovedDate IS NULL",
                 new { candidateId, customisationId }
             );
         }
 
+        public void InsertNewProgress(int candidateId, int customisationId, int centreId)
```
FridaTveit, line 111: I wouldn't expect this to be three different public methods. I think there should be one public method to GetProcessId (could be renamed to GetOrCreateProgressId to be clear) which will return an existing one or create a new one 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -80,18 +82,67 @@ GROUP BY
...
+                          AND SystemRefreshed = 0
+                          AND RemovedDate IS NULL",
+                new { candidateId, customisationId }
+            ).Any();
+        }
+
         public int GetProgressId(int candidateId, int customisationId)
         {
-            // TODO HEEDLS-202: change QueryFirstOrDefault to creating progress record if not found
             return connection.QueryFirstOrDefault<int>(
```
FridaTveit, line 100: As commented before we should not use QueryFirstOrDefault. This will return progress id 0 if a progress id can't be found which is not what we want.

DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -80,18 +82,67 @@ GROUP BY
...
+                    EnrollmentMethodID = 1,
+                    EnrolledByAdminID = 0
+                },
+                commandType: CommandType.StoredProcedure
+            );
+
+            switch (errorCode)
+            {
+                case 1:
+                    logger.LogWarning(
```
FridaTveit, line 129: These should all be errors 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -80,18 +82,67 @@ GROUP BY
...
+                case 100:
+                    logger.LogWarning(
+                        "Not enrolled candidate on course as customisation id doesn't match centre id. " +
+                        $"Candidate id: {candidateId}, customisation id: {customisationId}, centreId{centreId}");
+                    break;
+                case 101:
+                    logger.LogWarning(
+                        "Not enrolled candidate on course as candidate id doesn't match centre id. " +
+                        $"Candidate id: {candidateId}, customisation id: {customisationId}, centreId{centreId}");
+                    break;
```
FridaTveit, line 142: We probably want a default case as well, I'd expect the IDE to warn you about that?

FridaTveit (commented):
A few comments, let me know if  you have any questions 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
kwunkiuma: The IDE doesn't give any warnings, we don't expect to obtain any error codes that aren't listed so I didn't think we needed to define it. Should I include it just in case?

FridaTveit: I'm pretty sure there'll be a small green squigly line under the switch keyword, at least there is when I try it. Yes you should have a default case because the stored procedure may be modified 🙂

DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
```diff
@@ -165,20 +165,45 @@ public void Get_course_content_of_unstarted_course_without_learning_should_retur
...
         {
             // Given
             const int candidateId = 9;
             const int customisationId = 259;
+            const int centreId = 53;
+            var initialDoesProgressExist = courseContentTestHelper.DoesProgressExist(candidateId, customisationId);
 
             // When
-            var result = courseContentService.GetProgressId(candidateId, customisationId);
+            var result = courseContentService.GetOrCreateProgressId(candidateId, customisationId, centreId);
```
FridaTveit, line 177: This can now make changes to the db so should be inside transaction scope 🙂

DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
```diff
@@ -165,20 +165,45 @@ public void Get_course_content_of_unstarted_course_without_learning_should_retur
...
 
             // Then
+            initialDoesProgressExist.Should().BeTrue();
             const int expectedProgressId = 10;
             result.Should().Be(expectedProgressId);
         }
 
+        [Test]
+        public void Insert_new_progress_should_insert_progress()
```
FridaTveit, line 186: This doesn't explain which method is being tested 🙂

DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
```diff
@@ -165,20 +165,45 @@ public void Get_course_content_of_unstarted_course_without_learning_should_retur
...
+            const int candidateId = 187251;
+            const int customisationId = 17468;
+            const int centreId = 549;
+            var initialDoesProgressExist = courseContentTestHelper.DoesProgressExist(candidateId, customisationId);
+
+            using (new TransactionScope())
+            {
+                // When
+                var progressId = courseContentService.GetOrCreateProgressId(candidateId, customisationId, centreId);
+                var result = courseContentTestHelper.DoesProgressExist(candidateId, customisationId);
```
FridaTveit, line 198: This isn't a very descriptive name, as it isn't really the "result" 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -80,18 +80,55 @@ GROUP BY
...
+        public int? GetOrCreateProgressId(int candidateId, int customisationId, int centreId)
         {
-            // TODO HEEDLS-202: change QueryFirstOrDefault to creating progress record if not found
-            return connection.QueryFirstOrDefault<int>(
-                @"SELECT ProgressId
-                        FROM Progress
-                        WHERE CandidateID = @candidateId
-                          AND CustomisationID = @customisationId",
-                new { candidateId, customisationId }
+            if (DoesProgressExist(candidateId, customisationId))
```
FridaTveit, line 85: You don't really need this as a separate check. You can just get the progress id and check if it's null 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -80,18 +80,55 @@ GROUP BY
...
+            {
+                return GetProgressId(candidateId, customisationId);
+            }
+
+            var errorCode = connection.QueryFirst<int>(
```
FridaTveit, line 90: Do you need `QueryFirst`? Would just `Query` work? 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -80,18 +80,55 @@ GROUP BY
...
-        }
 
+            switch (errorCode)
+            {
+                case 0:
+                    return GetProgressId(candidateId, customisationId);
+                case 1:
+                    logger.LogError(
+                        "Not enrolled candidate on course as progress already exists. " +
+                        $"Candidate id: {candidateId}, customisation id: {customisationId}, centreId{centreId}");
```
FridaTveit, line 110: Should have colon and space after centre id

DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -80,18 +80,55 @@ GROUP BY
...
+                    break;
+                case 101:
+                    logger.LogError(
+                        "Not enrolled candidate on course as candidate id doesn't match centre id. " +
+                        $"Candidate id: {candidateId}, customisation id: {customisationId}, centreId{centreId}");
+                    break;
+                default:
+                    logger.LogError(
+                        "Not enrolled candidate on course as stored procedure failed. " +
+                        $"Candidate id: {candidateId}, customisation id: {customisationId}, centreId{centreId}");
```
FridaTveit, line 125: Probably worth also logging the error code and noting in the message that it's an unknown error code

DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -119,5 +156,31 @@ public void UpdateProgress(int progressId)
...
+                          AND CustomisationID = @customisationId
+                          AND SystemRefreshed = 0
+                          AND RemovedDate IS NULL",
+                new { candidateId, customisationId }
+            ).Any();
+        }
+
+        private int GetProgressId(int candidateId, int customisationId)
+        {
+            return connection.QueryFirst<int>(
```
FridaTveit, line 175: Won't this throw an exception if there's no progress id?

DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
```diff
@@ -128,7 +141,20 @@ public void Index_invalid_customisation_id_should_not_update_login_and_duration(
             controller.Index(CustomisationId);
 
             // Then
-            A.CallTo(() => courseContentService.GetProgressId(A<int>._, A<int>._)).MustNotHaveHappened();
+            A.CallTo(() => courseContentService.GetOrCreateProgressId(A<int>._, A<int>._, A<int>._)).MustNotHaveHappened();
```
FridaTveit, line 144: No need to test this here as it's already being covered by your other test 🙂

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
```diff
@@ -36,8 +36,14 @@ public IActionResult Index(int customisationId)
                 return RedirectToAction("StatusCode", "LearningSolutions", new { code = 404 });
             }
 
-            var progressId = courseContentService.GetProgressId(User.GetCandidateId(), customisationId);
-            courseContentService.UpdateProgress(progressId);
+            var progressId = courseContentService.GetOrCreateProgressId(User.GetCandidateId(), customisationId, User.GetCentreId().Value);
```
FridaTveit, line 39: You should handle cetre id being null. Can the stored procedure handle the centre id being null?

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
```diff
@@ -36,8 +36,14 @@ public IActionResult Index(int customisationId)
                 return RedirectToAction("StatusCode", "LearningSolutions", new { code = 404 });
             }
 
-            var progressId = courseContentService.GetProgressId(User.GetCandidateId(), customisationId);
-            courseContentService.UpdateProgress(progressId);
+            var progressId = courseContentService.GetOrCreateProgressId(User.GetCandidateId(), customisationId, User.GetCentreId().Value);
+
+            if (progressId == null)
+            {
```
FridaTveit, line 42: Worth logging an error here I think 🙂

FridaTveit (commented):
A few comments, let me know if you have any questions 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
kwunkiuma: Query returns IEnumerable<int> even though it's guaranteed to only return one value.

DigitalLearningSolutions.Data/Services/CourseContentService.cs
kwunkiuma: The method is currently only called after DoesProgressExist but if we change it to return int? like suggested above then I can put it in a try catch.

DigitalLearningSolutions.Data/Services/CourseContentService.cs
FridaTveit: Ah fair enough, this is fine then 🙂

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
kwunkiuma: The service should have already logged the error, should I log it again?

FridaTveit: Yes I think it's worth it so there's a nice trace of errors in the logs. Something like "Redirecting to 500 as no progress id returned for user ..." 🙂 We should probably add a similar log when we're redirecting to 404 actually

kwunkiuma:
Forgot to refactor DoesProgressExist, will do that soon.

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
```diff
@@ -30,16 +30,23 @@ ICourseContentService courseContentService
...
         {
             var courseContent = courseContentService.GetCourseContent(User.GetCandidateId(), customisationId);
+            var centreId = User.GetCentreId();
 
-            if (courseContent == null)
+            if (courseContent == null || centreId == null)
             {
+                logger.LogError(
+                    "Redirecting to 404 as course/centre id was not found. " +
+                    $"Candidate id: {User.GetCandidateId()}, customisation id: {customisationId}");
```
FridaTveit, line 39: Should include the centre id in this log

FridaTveit (commented):
A couple of comments and as you spotted the DoesProgressExist method still needs to be removed 🙂

DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
```diff
@@ -165,18 +165,46 @@ public void Get_course_content_of_unstarted_course_without_learning_should_retur
         }
 
         [Test]
-        public void Get_progress_id_should_return_progress_id()
+        public void Get_existing_progress_id_should_return_progress_id()
```
FridaTveit, line 168: This name is slightly inconsistent with the one below 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -119,5 +159,25 @@ public void UpdateProgress(int progressId)
                 );
             }
         }
+
+        public int? GetProgressId(int candidateId, int customisationId)
```
FridaTveit, line 163: This should be private 🙂

FridaTveit (commented):
A couple of minor things, otherwise looks good 🙂

FridaTveit (approved):
Looks good 🙂

PR #50, opened 2020-11-30
HEEDLS-207 Fix layout for long titles on narrow screens

benjimarshall:
## Changes
Normally service names are links in a header, where the link covers the
NHS logo and the text. It doesn't make sense for the text to be a link
here, but keep the same layout by using the CSS classes in a `<div>`
instead of on the `<a>` tag.

## Testing
Tested using Chrome, Firefox, Edge and IE11, all using narrow viewports and large zooms. Tested on Chrome and IE11 using a screenreader. The only cases when the header layout breaks down (menu button overlapping text mainly) are for extremely narrow viewports combined with large zooms. In these cases it occurs in exactly the same way for other NHS pages (tested with [this one](https://service-manual.nhs.uk/design-system/components) mainly), so none of the changes we have made have changed this behaviour. 

### Using a 1920p wide viewport, 100% zoom
large_desktop_header

### Using a 1920p wide viewport, 300% zoom
desktop_significant_zoom

### Using a 500p wide viewport, 100% zoom
narrow_viewport

### Using a 500p wide viewport, 250% zoom. Note these parameters will cause a similar failure on other NHS websites but this is included for completeness.
narrow_viewport_significant_zoom

### DIFF
--- a/DigitalLearningSolutions.Web/Views/Shared/_Layout.cshtml
+++ b/DigitalLearningSolutions.Web/Views/Shared/_Layout.cshtml
@@ -42,17 +42,18 @@
     <a class="nhsuk-skip-link" href="#maincontent">Skip to main content</a>
     <header class="nhsuk-header nhsuk-header--organisation nhsuk-header--white" role="banner">
       <div class="nhsuk-width-container nhsuk-header__container">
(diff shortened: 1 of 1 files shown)


FridaTveit (approved):
Looks good 🙂 FYI @kevwhitt-hee

PR #52, opened 2020-12-01
HEEDLS-212 Add SessionService (and tests) to track sessions

benjimarshall:
## Changes

- Add `SessionService` to manage interactions with sessions, with methods to:
    - Start or restart a session
    - Update the duration of a session
    - Make a session as inactive
This is the DB side of HEEDLS-212. This will need to be connected up to the Controllers in a future PR.

## Testing
Added unit tests to check the three methods of the service to ensure they behave as expected.

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Helpers/SessionTestHelper.cs
+++ b/DigitalLearningSolutions.Data.Tests/Helpers/SessionTestHelper.cs
@@ -1,0 +1,45 @@
+﻿namespace DigitalLearningSolutions.Data.Tests.Helpers
+{
+    using System.Collections.Generic;
+    using Dapper;
+    using DigitalLearningSolutions.Data.Models;
+    using FluentAssertions;
+    using Microsoft.Data.SqlClient;
+
+    public class SessionTestHelper
+    {
+        private SqlConnection connection;
+
+        public SessionTestHelper(SqlConnection connection)
+        {
+            this.connection = connection;
+        }
+
+        public Session? GetSession(int sessionId)
+        {
+            return connection.QueryFirstOrDefault<Session>(
+                @"SELECT SessionID, CandidateID, CustomisationID, LoginTime, Duration, Active
+                    FROM Sessions
+                   WHERE SessionID = @sessionId",
+                new { sessionId });
+        }
+
+        public IEnumerable<Session> GetCandidateSessions(int candidateId)
+        {
+            return connection.Query<Session>(
+                @"SELECT SessionID, CandidateID, CustomisationID, LoginTime, Duration, Active
+                    FROM Sessions
+                   WHERE CandidateID = @candidateId",
+                new { candidateId });
+        }
+
+        public static void SessionsShouldBeApproximatelyEquivalent(Session session1, Session session2)
+        {
+            session1.CandidateId.Should().Be(session2.CandidateId);
+            session1.CustomisationId.Should().Be(session2.CustomisationId);
+            session1.Duration.Should().Be(session2.Duration);
+            session1.Active.Should().Be(session2.Active);
+            session1.LoginTime.Should().BeCloseTo(session2.LoginTime, 10000);
+        }
+    }
+}
--- a/DigitalLearningSolutions.Data.Tests/Helpers/SessionTestHelper.cs
+++ b/DigitalLearningSolutions.Data.Tests/Helpers/SessionTestHelper.cs
@@ -1,5 +1,6 @@
 ﻿namespace DigitalLearningSolutions.Data.Tests.Helpers
 {
+    using System;
     using System.Collections.Generic;
     using Dapper;
     using DigitalLearningSolutions.Data.Models;
@@ -33,13 +34,28 @@
                 new { candidateId });
         }
 
+        public static Session CreateDefaultSession(
+            int sessionId,
+            int candidateId = 101,
+            int customisationId = 1240,
+            DateTime? loginTime = null,
+            int duration = 0,
+            bool active = true
+        )
+        {
+            loginTime ??= DateTime.Now;
+            return new Session(sessionId, candidateId, customisationId, loginTime.Value, duration, active);
+        }
+
         public static void SessionsShouldBeApproximatelyEquivalent(Session session1, Session session2)
         {
+            const int tenSecondsInMilliseconds = 10000;
+
             session1.CandidateId.Should().Be(session2.CandidateId);
             session1.CustomisationId.Should().Be(session2.CustomisationId);
             session1.Duration.Should().Be(session2.Duration);
             session1.Active.Should().Be(session2.Active);
-            session1.LoginTime.Should().BeCloseTo(session2.LoginTime, 10000);
+            session1.LoginTime.Should().BeCloseTo(session2.LoginTime, tenSecondsInMilliseconds);
         }
     }
 }
--- a/DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
@@ -38,14 +38,7 @@
                 newSession.Should().NotBeNull();
                 SessionTestHelper.SessionsShouldBeApproximatelyEquivalent(
                     newSession!,
-                    new Session(
-                        sessionId,
-                        candidateId,
-                        customisationId,
-                        DateTime.Now,
-                        0,
-                        true
-                    )
+                    SessionTestHelper.CreateDefaultSession(sessionId, candidateId, customisationId)
                 );
             }
         }
@@ -66,14 +59,7 @@
                 newSession.Should().NotBeNull();
                 SessionTestHelper.SessionsShouldBeApproximatelyEquivalent(
                     newSession!,
-                    new Session(
-                        sessionId,
-                        candidateId,
-                        customisationId,
-                        DateTime.Now,
-                        0,
-                        true
-                    )
+                    SessionTestHelper.CreateDefaultSession(sessionId)
                 );
             }
         }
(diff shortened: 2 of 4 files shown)


DigitalLearningSolutions.Data.Tests/Helpers/SessionTestHelper.cs
```diff
@@ -0,0 +1,45 @@
...
+                new { candidateId });
+        }
+
+        public static void SessionsShouldBeApproximatelyEquivalent(Session session1, Session session2)
+        {
+            session1.CandidateId.Should().Be(session2.CandidateId);
+            session1.CustomisationId.Should().Be(session2.CustomisationId);
+            session1.Duration.Should().Be(session2.Duration);
+            session1.Active.Should().Be(session2.Active);
+            session1.LoginTime.Should().BeCloseTo(session2.LoginTime, 10000);
```
FridaTveit, line 42: Is the 10000 milliseconds? 🙂

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
```diff
@@ -0,0 +1,166 @@
...
+                const int customisationId = 1240;
+                var sessionId = sessionService.StartOrRestartSession(candidateId, customisationId);
+
+                // Then
+                var newSession = sessionTestHelper.GetSession(sessionId);
+
+                newSession.Should().NotBeNull();
+                SessionTestHelper.SessionsShouldBeApproximatelyEquivalent(
+                    newSession!,
+                    new Session(
```
FridaTveit, line 69: Could have a helper method to create a session, like we have for e.g. course content, so that you can only override the fields you care about 🙂

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
```diff
@@ -0,0 +1,166 @@
...
+        public void UpdateSessionDuration_Should_Only_Update_Given_Active_Sessions()
+        {
+            const int twoMinutesInMilliseconds = 120 * 1000;
+
+            using (new TransactionScope())
+            {
+                // When
+                const int candidateId = 9;
+                const int sessionId = 473;
+                var startingSessions = sessionTestHelper.GetCandidateSessions(candidateId);
```
FridaTveit, line 127: I think some of these tests, especially this would, would benefit from having a `// Given` section as well, to make it a bit clearer what's actually being tested 🙂

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
```diff
@@ -0,0 +1,166 @@
...
+                var newSessions = sessionTestHelper.GetCandidateSessions(candidateId).ToList();
+
+                newSessions
+                    .Where(session => session.SessionId != sessionId)
+                    .Should()
+                    .BeEquivalentTo(startingSessions.Where(session => session.SessionId != sessionId));
+
+                var newSession = newSessions.First(session => session.SessionId == sessionId);
+                DateTime.Now.Should().BeCloseTo(
+                    newSession.LoginTime.AddMinutes(newSession.Duration),
```
FridaTveit, line 141: Is duration in minutes? 🙂

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
```diff
@@ -0,0 +1,166 @@
...
+                // Then
+                var newSessions = sessionTestHelper.GetCandidateSessions(candidateId).ToList();
+
+                newSessions
+                    .Where(session => session.SessionId != sessionId)
+                    .Should()
+                    .BeEquivalentTo(startingSessions.Where(session => session.SessionId != sessionId));
+
+                var newSession = newSessions.First(session => session.SessionId == sessionId);
+                DateTime.Now.Should().BeCloseTo(
```
FridaTveit, line 140: I think this assertion is the wrong way around. It should be `actual.Should().SomeAssertion(expected)` 🙂

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
```diff
@@ -0,0 +1,166 @@
...
+            {
+                // When
+                const int candidateId = 9;
+                const int sessionId = 473;
+                var startingSessions = sessionTestHelper.GetCandidateSessions(candidateId);
+
+                sessionService.UpdateSessionDuration(sessionId);
+
+                // Then
+                var newSessions = sessionTestHelper.GetCandidateSessions(candidateId).ToList();
```
FridaTveit, line 132: I think `updatedSessions` might be more descriptive? As otherwise it's easy to mistake these for sessions which have just been added 🙂

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
```diff
@@ -0,0 +1,166 @@
...
+
+                newSessions
+                    .Where(session => session.SessionId != sessionId)
+                    .Should()
+                    .BeEquivalentTo(startingSessions.Where(session => session.SessionId != sessionId));
+
+                var newSession = newSessions.First(session => session.SessionId == sessionId);
```
FridaTveit, line 139: `activeSession` might be more descriptive? 🙂

DigitalLearningSolutions.Data/Services/SessionService.cs
```diff
@@ -0,0 +1,52 @@
...
+        private readonly IDbConnection connection;
+
+        public SessionService(IDbConnection connection)
+        {
+            this.connection = connection;
+        }
+
+        private const string stopSessionsSql =
+            @"UPDATE [Sessions] SET [Active] = 0
+              WHERE [CandidateId] = @candidateId;";
```
FridaTveit, line 24: Minor but I don't think the square brackets are necessary? Shouldn't cause an issue but might be worth removing for consistency 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
benjimarshall: Yep, the duration is calculated using `DATEDIFF(minute, [LoginTime], GetUTCDate())` in `SessionService` which calculates to the nearest minute 🙂

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
benjimarshall: I'm not quite sure what you mean here. These lines are to check the values in the session that has just been inserted are correct using equality for most of the fields, but using approximate equality for the `LoginTime` (because I won't know precisely which millisecond the event will be created at). How could I use a create-session method to help with this? 🙂

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
FridaTveit: Okay, great. I assume it was also in minutes in the old code/query? 🙂

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
FridaTveit: Sorry, I just meant to create the Session object. We already have a few methods like that e.g. to create a CourseContent object. So instead of having to call the constructor with all the fields that requires, you call your helper method which has defaults for every constructor parameter so you only need to specify the parameters you really care about in your test. Does that make sense? 🙂

DigitalLearningSolutions.Data/Services/SessionService.cs
benjimarshall: Yep, removed nearly all of them in a4825ca 🙂 The only one I left was `[SessionID]`, which doesn't strictly need the square brackets but it is very similar to the keyword `Session_ID` so I think it is worth keeping the square brackets to stop any accidents from happening. I can remove that pair as well if you would prefer 🙂

DigitalLearningSolutions.Data.Tests/Helpers/SessionTestHelper.cs
benjimarshall: Yep! I moved it into a `const` in a4825ca 🙂

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
benjimarshall: I hadn't come across that before. I've added given sections for the tests that collect `startingSessions` before interacting with the `SessionService` in a4825ca 🙂

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
benjimarshall: Yep, sorry! Fixed in a4825ca 🙂

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
benjimarshall: Yeah I agree. Changed in a4825ca 🙂

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
benjimarshall: Yeah I agree. Changed in a4825ca 🙂

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
benjimarshall: Ah I see! I've done that in a4825ca 🙂

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
benjimarshall: That was part of the query on the HEEDLS-212 ticket description. I assume that was part of the old code but I can check if you like? 🙂

FridaTveit: I see, great. Yes that's from the old code so no need to check. Thanks for confirming 🙂

DigitalLearningSolutions.Data/Services/SessionService.cs
```diff
@@ -0,0 +1,52 @@
...
+    public class SessionService : ISessionService
+    {
+        private readonly IDbConnection connection;
+
+        public SessionService(IDbConnection connection)
+        {
+            this.connection = connection;
+        }
+
+        private const string stopSessionsSql =
```
FridaTveit, line 22: Sorry missed this earlier, this should be declared above the constructor 🙂

FridaTveit (commented):
One tiny thing, otherwise looks great 🙂

DigitalLearningSolutions.Data/Services/SessionService.cs
benjimarshall: Fixed in 91b2354 🙂

FridaTveit (approved):
Looks good 🙂

PR #55, opened 2020-12-02
HEEDLS-202 Create progress entry on reenrollment

kwunkiuma:
Modify the GetCourseContent query such that it returns a null progress entry for removed courses.

Ran and passed all unit tests and manually tested on Firefox.

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
@@ -1,6 +1,7 @@
 ﻿namespace DigitalLearningSolutions.Data.Tests.Services
 {
     using System;
+    using System.Linq;
     using System.Transactions;
     using DigitalLearningSolutions.Data.Models.CourseContent;
     using DigitalLearningSolutions.Data.Services;
@@ -162,6 +163,21 @@
                 }
             );
             result.Should().BeEquivalentTo(expectedCourse);
+        }
+
+        [Test]
+        public void Get_course_content_for_removed_course_should_return_course()
+        {
+            // Given
+            const int candidateId = 210962;
+            const int customisationId = 24635;
+
+            // When
+            var result = courseContentService.GetCourseContent(candidateId, customisationId);
+
+            // Then
+            result.Should().NotBeNull();
+            result!.Sections.Where(section => section.PercentComplete != 0).Should().BeEmpty();
         }
 
         [Test]
--- a/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
@@ -1,7 +1,6 @@
 ﻿namespace DigitalLearningSolutions.Data.Tests.Services
 {
     using System;
-    using System.Linq;
     using System.Transactions;
     using DigitalLearningSolutions.Data.Models.CourseContent;
     using DigitalLearningSolutions.Data.Services;
@@ -166,7 +165,7 @@
         }
 
         [Test]
-        public void Get_course_content_for_removed_course_should_return_course()
+        public void Get_course_content_for_removed_and_refreshed_course_should_return_new_course()
         {
             // Given
             const int candidateId = 210962;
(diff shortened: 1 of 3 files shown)


DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -54,10 +54,9 @@ FROM Applications
                   INNER JOIN Sections ON Sections.ApplicationID = Applications.ApplicationID
                   INNER JOIN Centres ON Customisations.CentreID = Centres.CentreID
                   LEFT JOIN Tutorials ON Sections.SectionID = Tutorials.SectionID
-                  LEFT JOIN Progress ON Customisations.CustomisationID = Progress.CustomisationID AND Progress.CandidateID = @candidateId
+                  LEFT JOIN Progress ON Customisations.CustomisationID = Progress.CustomisationID AND Progress.CandidateID = @candidateId AND Progress.RemovedDate IS NULL AND Progress.SystemRefreshed = 0
```
FridaTveit, line 57: Probably worth adding a test for this specific scenario, e.g. getting the course content when there are progress entries with removed date or system refreshed set 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -54,10 +54,9 @@ FROM Applications
                   INNER JOIN Sections ON Sections.ApplicationID = Applications.ApplicationID
```
FridaTveit, line 54: I assume you've tested the specific AC which was failing earlier, i.e. enrolling on a course which you've removed? 🙂

FridaTveit (commented):

FridaTveit:
Oh and probably worth @benjimarshall reviewing this as well 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
kwunkiuma: Yes, seems to be working for that case now.

benjimarshall (approved):
This looks good! I have been talking about this with Frank throughout. It makes sense to be only joining courses that have not been removed. If removing a course would make `Progress.RemovedDate IS NULL` or `Progress.SystemRefreshed = 0` (which I think so but @FridaTveit please could you confirm?) then this should work nicely!

FridaTveit:
@benjimarshall removing a course makes Progress.RemovedDate not null. SystemRefreshed gets set to 1 in another way (I'm not entirely sure how, I can check with Kevin) but I believe it should work the same based on the old code 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
FridaTveit: Great 🙂

benjimarshall:
> @benjimarshall removing a course makes Progress.RemovedDate not null. SystemRefreshed gets set to 1 in another way (I'm not entirely sure how, I can check with Kevin) but I believe it should work the same based on the old code 🙂

Yep that's what I meant `Progress.RemovedDate `, sorry! It sounds like it should work as correctly 🙂

kwunkiuma:
I've added the test for removed courses but I suppose we should wait for clarification on `Progress.SystemRefreshed`.

DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
```diff
@@ -164,6 +165,21 @@ public void Get_course_content_of_unstarted_course_without_learning_should_retur
...
+            // Given
+            const int candidateId = 210962;
+            const int customisationId = 24635;
+
+            // When
+            var result = courseContentService.GetCourseContent(candidateId, customisationId);
+
+            // Then
+            result.Should().NotBeNull();
+            result!.Sections.Where(section => section.PercentComplete != 0).Should().BeEmpty();
```
FridaTveit, line 180: I think comparing to an expected course content, like in the test above, will be a more thorough and easier to understand test 🙂

FridaTveit (commented):
One comment and test for system refreshed missing otherwise looks good 🙂 

I don't think there's any need to wait to add the system refreshed test, what I'm asking Kevin about is how it's being set not how we should use it. So I think it's unlikely to change this behaviour (just possibly how I manually test it). If it does need to change then we could always raise a bug/separate story for that, depending on what the change would be 🙂

FridaTveit:
Did that make sense about the SystemRefreshed test @kwunkiuma? 🙂

kwunkiuma:
It seems that all entries with `SystemRefreshed = 1` also have a non-null value for `RemovedDate`, should I add a test entry with `SystemRefreshed = 1` and a null `RemovedDate`?

FridaTveit:
Ah I see! Yes I think in that case it would make sense to have three tests:
* One which uses a real entry which has RemovedDate set and SystemRefreshed set
* One which uses a test entry which has just RemovedDate set
* One which uses a test entry which has just SystemRefreshed set

Does that make sense? 🙂

kwunkiuma:
Yup, there are real entries with just RemovedDate set so I should be able to just use that directly.

FridaTveit:
Great 🙂 It sounds like it's unlikely that SystemRefreshed will be set without the RemovedDate being set, but I think we'd better make sure we can handle it just in case 🙂

FridaTveit (approved):
Looks good 🙂

PR #57, opened 2020-12-02
HEEDLS-207 Fix header link on UAT

FridaTveit:
I was hoping the relative urls would be relative to the application, but they're not so small fix needed 🙂


benjimarshall (approved):
Looks good! Sorry that I didn't manage to work out how to do this in the first place! 🙂

FridaTveit:
No worries I should have probably suggested testing it. But deploying to UAT might have been the easiest way to test it properly anyway 🙂

PR #58, opened 2020-12-02
HEEDLS-207 Add app root path to header link

FridaTveit:


benjimarshall (approved):
Looks great! Thanks for sorting this 🙂

PR #59, opened 2020-12-02
HEEDLS-212 Add session tracking to LearningMenu controller

benjimarshall:
## Changes
- Inject SessionService into Web project.
- Maintain Session state in the LearningMenu using the SessionService.
- Add a mock HttpContext.Session class and use this to test the new
LearningMenu controller session tracking.

## Testing
Add new unit tests and check they run. Tested links in Chrome, Firefox, Edge and IE11

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
@@ -3,16 +3,17 @@
     using System;
     using System.Linq;
     using System.Transactions;
-    using DigitalLearningSolutions.Data.Models;
     using DigitalLearningSolutions.Data.Services;
     using DigitalLearningSolutions.Data.Tests.Helpers;
     using FluentAssertions;
+    using Microsoft.AspNetCore.Http;
     using NUnit.Framework;
 
     internal class SessionServiceTests
     {
         private SessionService sessionService;
         private SessionTestHelper sessionTestHelper;
+        private ISession httpContextSession;
 
         [SetUp]
         public void Setup()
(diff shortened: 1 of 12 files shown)


DigitalLearningSolutions.Web.Tests/ControllerHelpers/MockHttpContextSession.cs
```diff
@@ -0,0 +1,47 @@
+﻿namespace DigitalLearningSolutions.Web.Tests.ControllerHelpers
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.AspNetCore.Http;
+
+    internal class MockHttpContextSession : ISession
```
FridaTveit, line 9: Would it not be sufficient/possible to mock the class using FakeItEasy (i.e. `A.Fake`)? 🙂

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
```diff
@@ -56,10 +64,37 @@ public IActionResult Index(int customisationId)
...
+
         public IActionResult ContentViewer()
         {
             var model = new ContentViewerViewModel(config);
             return View(model);
         }
+
+        private void StartOrUpdateSession(int candidateId, int customisationId)
+        {
+            var currentSessionId = HttpContext.Session.GetInt32($"SessionID-{customisationId}");
```
FridaTveit, line 84: Hm, it would be nice if the complication of interacting with `HttpContext.Session` could also be extracted to the SessionService. What do you think? 🙂

FridaTveit (commented):
Generally looks good! I've left a couple of questions which might affect the structure a bit so I'll hold off on reviewing the rest in detail until those have been resolved 🙂 (just to be clear, resolution might very well be to keep things exactly as they are, I just think it's worth discussing 🙂 )

DigitalLearningSolutions.Web.Tests/ControllerHelpers/MockHttpContextSession.cs
benjimarshall: I spent a while trying to get this to work. Unfortunately I depend on static extension methods in the `LearningMenuController` (in particular the `GetInt32` and `SetInt32` methods), which I don't think FakeItEasy supports 🙂 I might have been missing something though

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
benjimarshall: Yeah that might be nice. How would I inject the `HttpContext` into the `SessionService`? 🙂

DigitalLearningSolutions.Web.Tests/ControllerHelpers/MockHttpContextSession.cs
FridaTveit: Ah I see! I did wonder if that was why. In that case creating our own mock sounds like a good solution 🙂

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
FridaTveit: I think you'd probably need to pass it (or just the session) to each method. Which is a bit annoying but it's just one extra parameter so shouldn't be too bad. What do you think? 🙂

benjimarshall: Changed over to this in d29430b 🙂

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
```diff
@@ -3,80 +3,106 @@
...
+
                 // When
                 const int candidateId = 29;
                 const int customisationId = 100;
-                var sessionId = sessionService.StartOrRestartSession(candidateId, customisationId);
+                sessionService.StartOrUpdateSession(candidateId, customisationId, httpContextSession);
 
                 // Then
-                var newSession = sessionTestHelper.GetSession(sessionId);
+                httpContextSession.Keys.Should().BeEquivalentTo($"SessionID-{customisationId}");
```
FridaTveit, line 41: It's preferable to have as few assertions as possible per test. This makes the tests easier to understand and means if they're broken then it should be easier to identify why. So I'd suggest testing that the http context session has been set correctly separately from testing that the session in the db has been updated correctly 🙂 If it would make it easier you could even split the session service up so you'd have:
* SessionDataService (or similar) which is equivalent to the session service you had before introducing the http context session
* SessionService which calls the SessionDataService methods and also handles the http context session

What do you think? I think keeping it all in one service is also fine if you prefer, as long as you split the tests up so each test is only testing one thing 🙂

FridaTveit (commented):
Nice, the controller logic is very simple and easy to understand now 😄 I have one comment on the tests, otherwise looks good! 🙂

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
benjimarshall: Yeah I agree that's a much better way of doing it! I've refactored to `SessionDataService`/`SessionService` approach in e4f7ed2 🙂

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
```diff
@@ -1,152 +1,99 @@
...
-                    newSession!,
-                    SessionTestHelper.CreateDefaultSession(sessionId)
-                );
-            }
+            sessionService = new SessionService(sessionDataService);
         }
 
         [Test]
-        public void StartOrRestartSession_Should_Stop_Other_Sessions()
+        public void Index_should_StartOrRestartSession_for_course_not_in_session()
```
FridaTveit, line 32: I don't think this test name matches what's being tested? 🙂

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
```diff
@@ -1,152 +1,99 @@
...
+            A.CallTo(() => sessionDataService.StartOrRestartSession(CandidateId, newCourseId)).MustHaveHappenedOnceExactly();
+            A.CallTo(() => sessionDataService.StartOrRestartSession(A<int>._, A<int>._))
+                .WhenArgumentsMatch((int candidateId, int customisationId) =>
+                    candidateId != CandidateId || customisationId != newCourseId)
+                .MustNotHaveHappened();
+
+            A.CallTo(() => sessionDataService.UpdateSessionDuration(A<int>._)).MustNotHaveHappened();
+
+            httpContextSession.Keys.Should().BeEquivalentTo($"SessionID-{newCourseId}");
+            httpContextSession.GetInt32($"SessionID-{newCourseId}").Should().Be(DefaultSessionId);
```
FridaTveit, line 53: I'd still split this into two tests; one for checking the correct methods are being called on the session data service, one for checking the http context session has been updated 🙂

FridaTveit (commented):
A couple of minor comments on the tests, otherwise looks great! 🙂

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
benjimarshall: Whoops! Fixed in 90a4212 🙂

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
benjimarshall: Yeah that looks like a better approach. Refactored in 90a4212 🙂

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
```diff
@@ -48,13 +48,27 @@ public void Index_should_StartOrRestartSession_for_course_not_in_session()
...
+            sessionService.StartOrUpdateSession(CandidateId, newCourseId, httpContextSession);
 
+            // Then
             httpContextSession.Keys.Should().BeEquivalentTo($"SessionID-{newCourseId}");
             httpContextSession.GetInt32($"SessionID-{newCourseId}").Should().Be(DefaultSessionId);
         }
 
         [Test]
-        public void Index_should_UpdateSession_for_course_in_session()
+        public void StartOrUpdateSession_should_UpdateSession_for_course_in_session()
```
FridaTveit, line 71: This test also needs to be split up 🙂

FridaTveit (commented):
One test still needs splitting up, otherwise looks good! 🙂

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
benjimarshall: This was actually split in two, I just forgot to delete the second part of the assertions! Fixed in 1cd9ace 🙂

FridaTveit (approved):
Looks good! 🙂

PR #61, opened 2020-12-02
Round percentage completion to nearest integer

benjimarshall:
### Changes
Add a rounding to the percentage completion using string formatting.

### Testing
Tested on Google Chrome only because this is just changing a string before it gets passed to the View.

### Screenshot
percent_rounded

### DIFF
--- a/DigitalLearningSolutions.Web/ViewModels/LearningMenu/SectionCardViewModel.cs
+++ b/DigitalLearningSolutions.Web/ViewModels/LearningMenu/SectionCardViewModel.cs
@@ -10,7 +10,7 @@
         public SectionCardViewModel(CourseSection section)
         {
             Title = section.Title;
-            PercentComplete = section.HasLearning ? $"{section.PercentComplete}% Complete" : "";
+            PercentComplete = section.HasLearning ? $"{section.PercentComplete:f0}% Complete" : "";
         }
     }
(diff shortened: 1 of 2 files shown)


DigitalLearningSolutions.Web/ViewModels/LearningMenu/SectionCardViewModel.cs
```diff
@@ -10,7 +10,7 @@ public class SectionCardViewModel
         public SectionCardViewModel(CourseSection section)
         {
             Title = section.Title;
-            PercentComplete = section.HasLearning ? $"{section.PercentComplete}% Complete" : "";
+            PercentComplete = section.HasLearning ? $"{section.PercentComplete:f0}% Complete" : "";
```
FridaTveit, line 13: A test for this would be useful? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web/ViewModels/LearningMenu/SectionCardViewModel.cs
benjimarshall: Test added in cf9a73b 🙂

FridaTveit (approved):
Looks good 🙂

PR #64, opened 2020-12-03
HEEDLS-215 Section Page Basic Info Back End

SChandarana:
I've stripped a lot of the stuff from the stored procedure and resulted in this query. I've tested that the query works but there are some weird things with it (such as the nested inner join). Do you have any suggestions on how to make this query better? And also whether it would be better to use Benji's query for courses and extend/change that. (Especially since in future I believe I will also need to do the multi mapping etc)

Once we've decided on the query I can begin writing the tests :D

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
@@ -35,6 +35,8 @@ namespace DigitalLearningSolutions.Data.Tests.Services
 
             // Then
             var expectedSectionContent = new SectionContent(
+                "Erin Test 01",
+                "Office 2013 Essentials for the Workplace",
                 "Working with Microsoft Office applications",
                 0,
                 28,
@@ -94,6 +96,8 @@ namespace DigitalLearningSolutions.Data.Tests.Services
 
             // Then
             var expectedSectionContent = new SectionContent(
+                "Testing",
+                "Excel 2013 for the Workplace",
                 "Entering data",
                 4,
                 28,
--- a/DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
@@ -115,5 +115,97 @@ namespace DigitalLearningSolutions.Data.Tests.Services
             );
             result.Should().BeEquivalentTo(expectedSectionContent);
         }
+
+        [Test]
+        public void Get_section_content_should_return_null_if_archived_date_is_null()
+        {
+            // When
+            const int customisationId = 14212;
+            const int candidateId = 23031;
+            const int sectionId = 261;
+            var result = sectionContentService.GetSectionContent(customisationId, candidateId, sectionId);
+
+            // Then
+            result.Should().BeNull();
+        }
+
+        [Test]
+        public void Get_section_content_should_return_null_if_diag_status_and_is_assessed_and_tutorial_status_are_not_equal_to_one()
+        {
+            // When
+            const int customisationId = 1530;
+            const int candidateId = 23573;
+            const int sectionId = 74;
+            var result = sectionContentService.GetSectionContent(customisationId, candidateId, sectionId);
+
+            // Then
+            result.Should().BeNull();
+        }
+
+        [Test]
+        public void Get_section_content_should_return_content_if_only_diag_status_is_one()
+        {
+            // When
+            const int customisationId = 5982;
+            const int candidateId = 59561;
+            const int sectionId = 74;
+            var result = sectionContentService.GetSectionContent(customisationId, candidateId, sectionId);
+
+            // Then
+            var expectedSectionContent = new SectionContent(
+                "Level 2 - MS Word 2007 DIAGNOSTIC TESTING",
+                "Level 2 - Microsoft Word 2007",
+                "Working with documents",
+                0,
+                28,
+                false,
+                0
+            );
+            result.Should().BeEquivalentTo(expectedSectionContent);
+        }
+
+        [Test]
+        public void Get_section_content_should_return_content_if_only_is_assessed_is_one()
+        {
+            // When
+            const int customisationId = 2684;
+            const int candidateId = 196;
+            const int sectionId = 74;
+            var result = sectionContentService.GetSectionContent(customisationId, candidateId, sectionId);
+
+            // Then
+            var expectedSectionContent = new SectionContent(
+                "Styles and Working with References",
+                "Level 2 - Microsoft Word 2007",
+                "Working with documents",
+                0,
+                28,
+                false,
+                0
+            );
+            result.Should().BeEquivalentTo(expectedSectionContent);
+        }
+
+        [Test]
+        public void Get_section_content_should_return_content_if_only_tutorial_status_is_one()
+        {
+            // When
+            const int customisationId = 1499;
+            const int candidateId = 6;
+            const int sectionId = 74;
+            var result = sectionContentService.GetSectionContent(customisationId, candidateId, sectionId);
+
+            // Then
+            var expectedSectionContent = new SectionContent(
+                "Test",
+                "Level 2 - Microsoft Word 2007",
+                "Working with documents",
+                3,
+                28,
+                true,
+                50
+            );
+            result.Should().BeEquivalentTo(expectedSectionContent);
+        }
     }
 }
--- a/DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
@@ -45,7 +45,33 @@ namespace DigitalLearningSolutions.Data.Tests.Services
         }
 
         [Test]
-        public void Get_section_content_should_return_null_on_invalid_input()
+        public void Get_section_content_should_return_null_if_customisation_id_is_invalid()
+        {
+            //When
+            const int customisationId = 0;
+            const int progressId = 173218;
+            const int sectionId = 382;
+            var result = sectionContentService.GetSectionContent(customisationId, progressId, sectionId);
+
+            // Then
+            result.Should().BeNull();
+        }
+
+        [Test]
+        public void Get_section_content_should_return_null_if_progress_id_is_invalid()
+        {
+            //When
+            const int customisationId = 15853;
+            const int progressId = 0;
+            const int sectionId = 382;
+            var result = sectionContentService.GetSectionContent(customisationId, progressId, sectionId);
+
+            // Then
+            result.Should().BeNull();
+        }
+
+        [Test]
+        public void Get_section_content_should_return_null_if_section_id_is_invalid()
         {
             //When
             const int customisationId = 15853;
@@ -55,6 +81,26 @@ namespace DigitalLearningSolutions.Data.Tests.Services
 
             // Then
             result.Should().BeNull();
+        }
+
+        [Test]
+        public void Get_section_content_should_return_correct_section_when_percent_complete_is_not_zero()
+        {
+            // When
+            const int customisationId = 19262;
+            const int progressId = 201058;
+            const int sectionId = 1011;
+            var result = sectionContentService.GetSectionContent(customisationId, progressId, sectionId);
+
+            // Then
+            var expectedSectionContent = new SectionContent(
+                "Entering data",
+                4,
+                28,
+                true,
+                25
+            );
+            result.Should().BeEquivalentTo(expectedSectionContent);
         }
     }
 }
--- a/DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
@@ -1,0 +1,60 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace DigitalLearningSolutions.Data.Tests.Services
+{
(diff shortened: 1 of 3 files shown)


DigitalLearningSolutions.Data/Services/SectionContentService.cs
```diff
@@ -0,0 +1,61 @@
...
+        public SectionContentService(IDbConnection connection, ILogger<SectionContentService> logger)
+        {
+            this.connection = connection;
+            this.logger = logger;
+        }
+
+        public SectionContent? GetSectionContent(int customisationId, int progressId, int sectionId)
+        {
+            return connection.Query<SectionContent>(
+                @"SELECT Sections.SectionName,
```
FridaTveit, line 28: The indentation is a bit inconsistent in this query. Fixing that will make it a bit easier to read 🙂

DigitalLearningSolutions.Data/Services/SectionContentService.cs
```diff
@@ -0,0 +1,61 @@
...
+	                    AND (Progress.ProgressID = @progressId) 
+	                    AND (Sections.SectionID = @sectionId)
+                    GROUP BY
+                        Sections.SectionID, 
+	                    Sections.SectionName, 
+	                    Sections.AverageSectionMins, 
+	                    Progress.CandidateID, 
+	                    Progress.CustomisationID;",
+                new { customisationId, progressId, sectionId }
+            ).FirstOrDefault();
```
FridaTveit, line 58: Can use `QueryFirstOrDefault` instead 🙂

DigitalLearningSolutions.Data/Services/SectionContentService.cs
```diff
@@ -0,0 +1,61 @@
...
+    using Dapper;
+    using DigitalLearningSolutions.Data.Models.SectionContent;
+    using Microsoft.Extensions.Logging;
+
+    public interface ISectionContentService
+    {
+        SectionContent? GetSectionContent(int customisationId, int progressId, int sectionId);
+    }
+
+    public class SectionContentService : ISectionContentService
```
FridaTveit, line 14: You need to add tests for this 🙂

FridaTveit (commented):
A few comments and tests are missing 🙂 

I'd also expect the front end work to include using this service in the controller and passing the values to the front end. But fine to do that in a separate PR 🙂 

The query seems okay. Formatting it more consistently should make it easier to read though, so I'll review again once that's fixed 🙂 I'm not quite sure what you mean by extending Benji's query? Do you mean the query to get the course content? That fetches different information to what we want here so we couldn't reuse that 🙂

DigitalLearningSolutions.Data/Services/SectionContentService.cs
SChandarana: I wasn't really sure of the best way to indent the "FROM" part of the query as there is a sort of nested JOIN. I chose to indent the nested part as follows
```
FROM aspProgress
INNER JOIN Progress ON aspProgress.ProgressID = Progress.ProgressID 
INNER JOIN Sections 
    INNER JOIN Tutorials ON Sections.SectionID = Tutorials.SectionID 
    INNER JOIN CustomisationTutorials ON Tutorials.TutorialID = CustomisationTutorials.TutorialID 
ON aspProgress.TutorialID = Tutorials.TutorialID                    
```
Is there a better way to do this?

FridaTveit: Good point, it is a bit of a tricky one. I think just a few tweaks are necessary to make sure all statements at the same level are indented the same. So everything in the SELECT section should be indented the same, the inner joins at the same level should be indented the same etc. Does that make sense? For the from section I'd probably do something like 
```
FROM aspProgress
  INNER JOIN Progress ON aspProgress.ProgressID = Progress.ProgressID 
  INNER JOIN Sections 
    INNER JOIN Tutorials ON Sections.SectionID = Tutorials.SectionID 
    INNER JOIN CustomisationTutorials ON Tutorials.TutorialID = CustomisationTutorials.TutorialID 
    ON aspProgress.TutorialID = Tutorials.TutorialID 
  INNER JOIN Customisations ON Progress.CustomisationID = Customisations.CustomisationID
  LEFT OUTER JOIN AssessAttempts ON Progress.ProgressID = AssessAttempts.ProgressID AND Sections.SectionNumber = AssessAttempts.SectionNumber
WHERE
```
🙂

SChandarana:
Ah whoops, I left my WIP commit in rather than amending, just ignore that. I'm working on the tests now but thought it would be good to put the query up for review.
I solved the nested INNER LOOP issues with a restructuring of the SQL joins so they were done in a different order but still resulted in the same thing

FridaTveit:
Looks good, other than missing tests 🙂

DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
```diff
@@ -0,0 +1,60 @@
...
+                0,
+                28,
+                true,
+                0
+            );
+            result.Should().BeEquivalentTo(expectedSectionContent);
+        }
+
+        [Test]
+        public void Get_section_content_should_return_null_on_invalid_input()
```
FridaTveit, line 48: What's invalid about this input? I made a similar comment on Benji's most recent PR, might help to have a look at the solution we ended up with there 🙂

DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
```diff
@@ -0,0 +1,60 @@
...
+        [SetUp]
+        public void Setup()
+        {
+            var connection = ServiceTestHelper.GetDatabaseConnection();
+            var logger = A.Fake<ILogger<SectionContentService>>();
+            sectionContentService = new SectionContentService(connection, logger);
+        }
+
+        [Test]
+        public void Get_section_content_should_return_section_content()
```
FridaTveit, line 28: I think it would be useful to test a few more cases. For example testing that a section which has percentage complete is fetched correctly 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data/Models/SectionContent/SectionContent.cs
```diff
@@ -2,20 +2,26 @@
 {
     public class SectionContent
     {
+        public string CustomisationName { get; }
+        public string ApplicationName { get; }
```
FridaTveit, line 6: I think this should be joined together as Title, like it is for course content 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data/Services/SectionContentService.cs
```diff
@@ -0,0 +1,65 @@
...
+        private readonly IDbConnection connection;
+        private readonly ILogger<SectionContentService> logger;
+
+        public SectionContentService(IDbConnection connection, ILogger<SectionContentService> logger)
+        {
+            this.connection = connection;
+            this.logger = logger;
+        }
+
+        public SectionContent? GetSectionContent(int customisationId, int progressId, int sectionId)
```
FridaTveit, line 25: One thing I've just noticed, this needs to take the candidate id rather than the progress id. As we should only return content if the candidate is enrolled on the course. If there is no progress record for the course then anything which relies on the progress record should be initialised to a sensible default value, e.g. 0 percent complete 🙂

DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
```diff
@@ -0,0 +1,110 @@
...
+                "Excel 2013 for the Workplace",
+                "Entering data",
+                4,
+                28,
+                true,
+                25
+            );
+            result.Should().BeEquivalentTo(expectedSectionContent);
+        }
+    }
```
FridaTveit, line 109: We should also have a test for if there's no progress record, or the progress record is invalid (due to having removed date or system reset) 🙂

FridaTveit (commented):
Spotted an issue with the query, sorry for not noticing this sooner! Otherwise looks good 🙂

DigitalLearningSolutions.Data/Services/SectionContentService.cs
SChandarana: Can this be done using the query `GetOrCreateProgressId` frank made for course content? And if so would it be useful to extract it to a different helper method rather than have it within the course content service

FridaTveit: Yes, it should use that. It's not really a problem to have it in the course content service though, as it will be called from the controller. Does that make sense? So I won't expect that to be done in this PR 🙂

DigitalLearningSolutions.Data/Services/SectionContentService.cs
```diff
@@ -0,0 +1,65 @@
...
+	                    INNER JOIN Tutorials ON Tutorials.TutorialID = aspProgress.TutorialID
+	                    INNER JOIN CustomisationTutorials ON CustomisationTutorials.TutorialID = Tutorials.TutorialID
+	                    INNER JOIN Sections ON Sections.SectionID = Tutorials.SectionID 
+	                    INNER JOIN Customisations ON Progress.CustomisationID = Customisations.CustomisationID
+	                    INNER JOIN Applications ON Customisations.ApplicationID = Applications.ApplicationID
+	                    LEFT OUTER JOIN AssessAttempts ON Progress.ProgressID = AssessAttempts.ProgressID AND Sections.SectionNumber = AssessAttempts.SectionNumber
+                    WHERE (CustomisationTutorials.CustomisationID = Progress.CustomisationID)
+	                    AND (Progress.CustomisationID = @customisationId)
+	                    AND (Progress.ProgressID = @progressId) 
+	                    AND (Sections.SectionID = @sectionId)
```
FridaTveit, line 53: One more thing; there are a few checks which are in the WHERE of the stored procedure to fetch sections, but are missing here. In particular checking CustomisationTutorials.Status, CustomisationTutorials.DiagStatus, Sections.ArchivedDate and Customisations.IsAssessed. We should probably also check that the customisation is active, although that's not in the stored procedure. Just confirming that with Kevin now 🙂

FridaTveit (commented):
One more thing I noticed about the query, sorry for not spotting earlier! 🙂

SChandarana:
I believe this push should be everything we need. Just let me know if there are anymore tests that i need to add. @kwunkiuma is working on the combining of FE and BE so I can work on the next section page backend stuff, so he'll add the controller behaviour we oultlined earlier

DigitalLearningSolutions.Data/Services/SectionContentService.cs
```diff
@@ -0,0 +1,78 @@
...
+        private readonly IDbConnection connection;
+        private readonly ILogger<SectionContentService> logger;
+
+        public SectionContentService(IDbConnection connection, ILogger<SectionContentService> logger)
+        {
+            this.connection = connection;
+            this.logger = logger;
+        }
+
+
```
kwunkiuma, line 25: Unecessary newline.

kwunkiuma (commented):
Spotted one minor thing.

DigitalLearningSolutions.Data/Services/SectionContentService.cs
```diff
@@ -0,0 +1,77 @@
...
+                            ON CustomisationTutorials.TutorialID = Tutorials.TutorialID
+                        INNER JOIN Customisations
+                            ON Customisations.CustomisationID = CustomisationTutorials.CustomisationID
+                        INNER JOIN Applications
+                            ON Applications.ApplicationID = Customisations.ApplicationID
+                        INNER JOIN Sections
+                            ON Sections.SectionID = Tutorials.SectionID
+                        LEFT JOIN Progress
+                            ON Progress.CustomisationID = Customisations.CustomisationID
+                            AND Progress.CandidateID = @candidateId
```
FridaTveit, line 54: Should also check here that removed date is null and system refreshed is 0, as in the query for the initial menu 🙂

DigitalLearningSolutions.Data/Services/SectionContentService.cs
```diff
@@ -0,0 +1,77 @@
...
+                        LEFT JOIN aspProgress
+                            ON aspProgress.TutorialID = CustomisationTutorials.TutorialID
+                            and aspProgress.ProgressID = Progress.ProgressID
+                    WHERE
+                        CustomisationTutorials.CustomisationID = @customisationId
+	                    AND Sections.SectionID = @sectionId
+	                    AND (Sections.ArchivedDate IS NULL)
+	                    AND (CustomisationTutorials.DiagStatus = 1 OR Customisations.IsAssessed = 1 OR CustomisationTutorials.Status = 1)
```
FridaTveit, line 62: I think we should have some more tests checking that this `WHERE` is working correctly, i.e. having a test checking that:
* an archived section is not returned
* a section where DiagStatus != 1 && IsAssessed != 1 && Status != 0 is not returned
* a section when only one of `CustomisationTutorials.DiagStatus = 1 OR Customisations.IsAssessed = 1 OR CustomisationTutorials.Status = 1` is true is returned

FridaTveit (commented):

DigitalLearningSolutions.Data/Services/SectionContentService.cs
SChandarana: Yep I'll add those in :D 
Should it be Status != 1?

FridaTveit: Yes, sorry 🤦

FridaTveit (approved):
Looks good 🙂

PR #65, opened 2020-12-03
HEEDLS-216 Add customisation ID to section page

kwunkiuma:
Add customisation ID to the section page URL and section card view model. The customisation ID is needed for the backend methods and session tracking.

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/IntitialMenuViewModelTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/IntitialMenuViewModelTests.cs
@@ -145,10 +145,11 @@
             const string sectionName = "TestSectionName";
             const bool hasLearning = true;
             const double percentComplete = 12.00;
+            const int customisationId = 1;
             var expectedCourseContent = CourseContentHelper.CreateDefaultCourseContent();
             var section = CourseSectionHelper.CreateDefaultCourseSection(sectionName: sectionName, hasLearning: hasLearning, percentComplete: percentComplete);
             expectedCourseContent.Sections.Add(section);
-            var expectedSection = new SectionCardViewModel(section, 1);
+            var expectedSection = new SectionCardViewModel(section, customisationId);
             var expectedSectionList = new List<SectionCardViewModel>
             {
                 expectedSection
--- a/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/IntitialMenuViewModelTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/IntitialMenuViewModelTests.cs
@@ -146,9 +146,9 @@
             const bool hasLearning = true;
             const double percentComplete = 12.00;
             const int customisationId = 1;
-            var expectedCourseContent = CourseContentHelper.CreateDefaultCourseContent();
+            var courseContent = CourseContentHelper.CreateDefaultCourseContent(customisationId: customisationId);
             var section = CourseSectionHelper.CreateDefaultCourseSection(sectionName: sectionName, hasLearning: hasLearning, percentComplete: percentComplete);
-            expectedCourseContent.Sections.Add(section);
+            courseContent.Sections.Add(section);
             var expectedSection = new SectionCardViewModel(section, customisationId);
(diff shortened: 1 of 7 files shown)


DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/IntitialMenuViewModelTests.cs
```diff
@@ -148,7 +148,7 @@ public void Initial_menu_should_have_list_of_section_card_view_model()
             var expectedCourseContent = CourseContentHelper.CreateDefaultCourseContent();
             var section = CourseSectionHelper.CreateDefaultCourseSection(sectionName: sectionName, hasLearning: hasLearning, percentComplete: percentComplete);
             expectedCourseContent.Sections.Add(section);
-            var expectedSection = new SectionCardViewModel(section);
+            var expectedSection = new SectionCardViewModel(section, 1);
```
FridaTveit, line 151: The customisation id should be set when creating the course content and then that variable referenced again here. Otherwise it's not clear what this 1 represents or where it's coming from 🙂 Same for all the other tests

FridaTveit (commented):

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/IntitialMenuViewModelTests.cs
```diff
@@ -145,10 +145,11 @@ public void Initial_menu_should_have_list_of_section_card_view_model()
             const string sectionName = "TestSectionName";
             const bool hasLearning = true;
             const double percentComplete = 12.00;
+            const int customisationId = 1;
             var expectedCourseContent = CourseContentHelper.CreateDefaultCourseContent();
             var section = CourseSectionHelper.CreateDefaultCourseSection(sectionName: sectionName, hasLearning: hasLearning, percentComplete: percentComplete);
             expectedCourseContent.Sections.Add(section);
-            var expectedSection = new SectionCardViewModel(section, 1);
+            var expectedSection = new SectionCardViewModel(section, customisationId);
```
FridaTveit, line 152: So the reason we expect it to have this customisation id is because that's the customisation id from the course content. At the moment we're relying on the default value for the customisation id when creating a course content being 1. This makes the test harder to understand and will be harder to debug if it's failing. So instead we should be passing in that customisation id when creating the course content. Does that make sense? 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/IntitialMenuViewModelTests.cs
```diff
@@ -145,10 +145,11 @@ public void Initial_menu_should_have_list_of_section_card_view_model()
             const string sectionName = "TestSectionName";
             const bool hasLearning = true;
             const double percentComplete = 12.00;
+            const int customisationId = 1;
             var expectedCourseContent = CourseContentHelper.CreateDefaultCourseContent();
```
FridaTveit, line 149: Not due to this change but this should not have a name prefixed by `expected`. As we're not using this as an expected value, so just `courseContent` would be a better name 🙂

FridaTveit (commented):
Slight misunderstanding about what I meant for the tests, let me know if it's not clear now or you have any questions about it 🙂

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
```diff
@@ -72,8 +72,8 @@ public IActionResult Close()
             return RedirectToAction("Current", "LearningPortal");
         }
 
-        [Route("/LearningMenu/Section/{sectionId:int}")]
-        public IActionResult Section(int sectionId)
+        [Route("/LearningMenu/{customisationId:int}/{sectionId:int}")]
+        public IActionResult Section(int customisationId, int sectionId)
```
benjimarshall, line 76: Now that we have a `customisationId` in this controller method, we can add the session tracking call to `sessionService.StartOrUpdateSession(...)`. Let me know if you want a hand getting that set up 🙂

kwunkiuma:
Hopefully the changes to the tests were what you meant, also added session tracking to the sections page.

FridaTveit (approved):
Looks good 🙂

PR #66, opened 2020-12-04
Fix border for Initial Menu Page

SChandarana:
image


DigitalLearningSolutions.Web/Styles/learningMenu/index.scss
```diff
@@ -1,7 +1,13 @@
-﻿.incomplete {
-  border-left: 6px solid #d5281b;
+﻿@import "~nhsuk-frontend/packages/core/all";
+
+.section-card {
```
FridaTveit, line 3: I don't think this is necessarily a very good name, as we're likely to use it for all cards not just sections. Maybe `learning-menu-card`? 🙂

FridaTveit (commented):
One suggestion about naming, otherwise looks good 🙂

FridaTveit (approved):
Looks good 🙂

PR #73, opened 2020-12-09
HEEDLS-225 do not add session for invalid course

benjimarshall:
## Changes
- Fix session adding bug for invalid course
- Add tests to cover this

## Testing
Add and modify the unit tests, and run them all.

### DIFF
--- a/DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
@@ -37,8 +37,6 @@
             var centreId = User.GetCentreId();
             var courseContent = courseContentService.GetCourseContent(candidateId, customisationId);
 
-            sessionService.StartOrUpdateSession(candidateId, customisationId, HttpContext.Session);
-
             if (courseContent == null || centreId == null)
             {
                 logger.LogError(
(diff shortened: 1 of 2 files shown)


FridaTveit (commented):
A few of the comments on your other PR also apply here, in particular: 
* Not using "missing progress" in the test names
* Redirecting to 404 instead of 500 if unable to create progress (feel free to do in the other PR if you prefer but it's probably less likely to create conflicts if you do it here)

Otherwise looks good 🙂

benjimarshall:
Done both of those bits in 89bdba6 🙂

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
```diff
@@ -51,9 +51,9 @@ public IActionResult Index(int customisationId)
             if (progressId == null)
             {
                 logger.LogError(
-                    "Redirecting to 500 as no progress id was returned. " +
+                    "Redirecting to 404 as no progress id was returned. " +
                     $"Candidate id: {candidateId}, customisation id: {customisationId}, centre id: {centreId}");
-                return RedirectToAction("StatusCode", "LearningSolutions", new { code = 500 });
+                return RedirectToAction("StatusCode", "LearningSolutions", new { code = 404 });
```
FridaTveit, line 56: Do we not have a test for this which needs updating? If we don't, could you add one? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
benjimarshall: Yep, added in 149fac0 🙂

FridaTveit (approved):
Looks good 🙂

PR #74, opened 2020-12-09
Fix too many sections on initial menu

benjimarshall:
## Changes

- Inner join against the CustomisationsTutorials to make sure we only
consider tutorials selected for this course customisation.
- Fix a percentage complete counting bug, which when counting the number
of tutorials that need to be completed it considered all potential
tutorials instead of just the ones selected for this course.

I spotted the percentage counting bug when discussing this ticket with Frank 
and fixing this bug solved the main issue of the ticket (pretty much by accident)
so I said I would push it up.

## Testing
Added a test for the bug in this ticket.

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
@@ -126,13 +126,8 @@
             expectedCourse.Sections.AddRange(
                 new[]
                 {
-                    new CourseSection("Introducing Outlook", 96, false, 0),
-                    new CourseSection("Writing and sending emails", 97, false, 0),
-                    new CourseSection("Managing emails", 98, false, 0),
                     new CourseSection("Using the Calendar", 99, true, 300/ 11.0),
-                    new CourseSection("Working with Contacts", 100, false, 0),
-                    new CourseSection("Using Tasks", 101, false, 0),
-                    new CourseSection("Using Notes and  the  Journal", 102, true, 0),
+                    new CourseSection("Using Notes and  the  Journal", 102, true, 0)
                 }
             );
             result.Should().BeEquivalentTo(expectedCourse);
@@ -160,13 +155,8 @@
             expectedCourse.Sections.AddRange(
                 new[]
                 {
-                    new CourseSection("Introducing Outlook", 96, false, 0),
-                    new CourseSection("Writing and sending emails", 97, false, 0),
-                    new CourseSection("Managing emails", 98, false, 0),
                     new CourseSection("Using the Calendar", 99, true, 0),
-                    new CourseSection("Working with Contacts", 100, false, 0),
-                    new CourseSection("Using Tasks", 101, false, 0),
-                    new CourseSection("Using Notes and  the  Journal", 102, true, 0),
+                    new CourseSection("Using Notes and  the  Journal", 102, true, 0)
                 }
             );
             result.Should().BeEquivalentTo(expectedCourse);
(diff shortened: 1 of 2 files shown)


FridaTveit:
This is HEEDLS-226 I believe? @kwunkiuma did you hand this over to @benjimarshall, or have you been pairing on it? Or have you both been working on it separately?

DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
```diff
@@ -380,6 +380,39 @@ public void Get_course_content_for_refreshed_course_should_return_new_course()
...
+        public void Get_course_content_should_only_return_sections_with_tutorials_in_customisation()
+        {
+            // Given
+            const int candidateId = 254480;
+            const int customisationId = 24224;
+
+            using (new TransactionScope())
+            {
+                // When
+                courseContentTestHelper.UpdateSystemRefreshed(candidateId, customisationId, true);
```
FridaTveit, line 393: Is this needed? 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -56,7 +56,9 @@ FROM Applications
                   INNER JOIN Customisations ON Applications.ApplicationID = Customisations.ApplicationID
                   INNER JOIN Sections ON Sections.ApplicationID = Applications.ApplicationID
                   INNER JOIN Centres ON Customisations.CentreID = Centres.CentreID
-                  LEFT JOIN Tutorials ON Sections.SectionID = Tutorials.SectionID
+                  INNER JOIN Tutorials ON Sections.SectionID = Tutorials.SectionID
+                  INNER JOIN CustomisationTutorials ON Customisations.CustomisationID = CustomisationTutorials.CustomisationID
+                                                   AND Tutorials.TutorialID = CustomisationTutorials.TutorialID
```
FridaTveit, line 61: We also need to ensure that 
```
Sections.ArchivedDate != NULL && 
(CustomisationTutorials.Status == 1 || CustomisationTutorial.DiagStatus == 1 || Customisations.IsAssessed)
```
(just pseudocode to show the logic)

FridaTveit (commented):
A couple of comments, otherwise looks good. I'm also a bit confused about who's owning this now. I'm very happy for you to pair on tickets or swap tickets, but do let me know when you do so that I know who's working on what and so that we can avoid conflicts or duplicated work 🙂

FridaTveit:
@benjimarshall pushed to the wrong PR? 🙂

benjimarshall:
Yep reverted 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
benjimarshall: I've added those in fd5e084 🙂

I've been using `Sections.ArchivedDate IS NULL` instead of `IS NOT NULL`, which I think was what you meant 🙂

DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
benjimarshall: Nope, removed in fd5e084 🙂

kwunkiuma:
Sorry for not mention handing this over to Benji, we were working on the issue together since he's more familiar with the query then he managed to find the solution. We were planning to let you know by updating the JIRA board but it seems to be having issues today.

benjimarshall:
Yeah when I spotted and fixed that other bug Frank and I thought I had covered the bug as well, though I didn't realise those extra bits needed to be added in, so I just pushed it up.

I would have put this on JIRA if that wasn't down at the moment, though we probably should have sent you a message instead so sorry about that! 🙂

FridaTveit:
Ah okay. Yes, in future it would be great if you could send me a message if you're changing what you're working on. Both because it's useful for me to know who's working on what and avoid conflicts, but also because it means I can ensure our highest priority tickets are being worked on 🙂

FridaTveit (approved):
Looks good 🙂 How much manual testing have you done? I think this is the kind of thing where it would be useful to compare the behaviour with the old system, so have both running and check that the same number of sections are being shown in both for at least a few courses 🙂

benjimarshall:
We've done a little bit of manual testing. Visual Studio keeps crashing for me when I try to run both websites at once, but Frank managed to get it to work and we've checked a few of different courses and it seems consistent for all the ones we've tried 🙂

PR #75, opened 2020-12-09
HEEDLS-180 Add default values to TutorialContent query

benjimarshall:
## Changes
- When a candidate is not enrolled on a course but queries a tutorial in
it, return the tutorial with progress values set to a default (eg.
TimeSpent = 0).
- Adjust the TutorialContentService tests to reflect this.
- Automatically enrol a user after returning a tutorial if they are
not already enrolled is part of the LearningMenuController already.

## Testing
Add unit tests for this default behaviour.

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Services/TutorialContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/TutorialContentServiceTests.cs
@@ -55,7 +55,7 @@
         {
             // Given
             const int candidateId = 1;
-            const int customisationId = -1;
+            const int customisationId = 1378;
             const int sectionId = 74;
             const int tutorialId = 50;
 
@@ -72,7 +72,7 @@
             // Given
             const int candidateId = 1;
             const int customisationId = 1379;
-            const int sectionId = 100;
+            const int sectionId = 75;
             const int tutorialId = 50;
 
             // When
--- a/DigitalLearningSolutions.Data.Tests/Services/TutorialContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/TutorialContentServiceTests.cs
@@ -55,7 +55,7 @@
         {
             // Given
             const int candidateId = 1;
-            const int customisationId = 1378;
+            const int customisationId = -1;
             const int sectionId = 74;
             const int tutorialId = 50;
 
@@ -72,7 +72,7 @@
             // Given
             const int candidateId = 1;
             const int customisationId = 1379;
-            const int sectionId = 75;
+            const int sectionId = 100;
             const int tutorialId = 50;
 
             // When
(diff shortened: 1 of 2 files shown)


DigitalLearningSolutions.Data.Tests/Services/TutorialContentServiceTests.cs
```diff
@@ -55,7 +55,7 @@ public void Get_tutorial_content_should_return_null_if_customisation_id_invalid(
         {
             // Given
             const int candidateId = 1;
-            const int customisationId = 1378;
+            const int customisationId = -1;
```
FridaTveit, line 58: Why this change? 🙂

DigitalLearningSolutions.Data.Tests/Services/TutorialContentServiceTests.cs
```diff
@@ -72,7 +72,7 @@ public void Get_tutorial_content_should_return_null_if_section_id_invalid()
             // Given
             const int candidateId = 1;
             const int customisationId = 1379;
-            const int sectionId = 75;
+            const int sectionId = 100;
```
FridaTveit, line 75: Why this change? 🙂

DigitalLearningSolutions.Data/Services/TutorialContentService.cs
```diff
@@ -52,17 +46,22 @@ FROM Tutorials
...
+                         LEFT JOIN TutStatus
+                         ON aspProgress.TutStat = TutStatus.TutStatusID
+                   WHERE CustomisationTutorials.CustomisationID = @customisationId
                      AND Tutorials.SectionId = @sectionId
                      AND Tutorials.TutorialID = @tutorialId
                      AND Customisations.Active = 1
                      AND CustomisationTutorials.Status = 1
-                     AND Progress.RemovedDate IS NULL
-                     AND Progress.SystemRefreshed = 0;",
+                     AND (Progress.CandidateID IS NULL OR (Progress.RemovedDate IS NULL AND Progress.SystemRefreshed = 0));",
```
FridaTveit, line 64: You did these checks in the join with progress I believe, in the query for the initial menu. Any reason for doing it differently here? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data/Services/TutorialContentService.cs
benjimarshall: I had forgotten I had did that, and doing this in the join would be nicer here I think, so I've changed that in b8e8702 🙂

DigitalLearningSolutions.Data.Tests/Services/TutorialContentServiceTests.cs
benjimarshall: Similar to the previous test, I got a bit confused here and changed it when I didn't need to. I don't think this should have been changed, and have reverted this in b8e8702 🙂

DigitalLearningSolutions.Data.Tests/Services/TutorialContentServiceTests.cs
benjimarshall: I got a bit confused here. I don't think this should have been changed, and have reverted this in b8e8702 🙂

FridaTveit (approved):
Looks good 🙂 FYI @SChandarana this is similar to the change I asked you to make to the section query 🙂

PR #76, opened 2020-12-09
HEEDLS-184 - Launch tutorial button

benjimarshall:
## Changes
- Add link to Play Tutorial button
- Add route to ContentViewer for this button to link to.

## Testing
Unit tests still run. I can't think of any ways to test this outside of clicking the button and checking it works (which I have done). The ContentViewer action and view model will get more tests in the next ticket, all that's changed is a button's link and the route of the action.


FridaTveit (approved):
Looks good 🙂

PR #77, opened 2020-12-09
HEEDLS-NONE Add more tests for initial menu SQL query

FridaTveit:
Since we've had issues with our queries not matching the old stored procedures I thought tests like this might be useful. If you come across a course worth adding to the test list (e.g. because it wasn't working at any point, or it has some interesting properties) then feel free to add it. I'll try to find time to hunt for some more useful courses to add to the test list. I'll also try to add something similar for the section and tutorial pages 🙂


benjimarshall (approved):
Looks good to me!

PR #79, opened 2020-12-10
HEEDLS-227 SCORM Content viewer back end

benjimarshall:
## Changes
- Rename `TutorialContent` to `TutorialInformation` - this is overall information for a tutorial
- Add a new `TutorialContent` class, to be used by the `ContentViewer` action and view
- Add a new `GetTutorialContent` method to the `TutorialContentService` to get one of these objects
- Add tests for this new `TutorialContentService` method

## Testing
Add and run unit tests for this new method.

### DIFF
--- a/DigitalLearningSolutions.Data/Services/TutorialContentService.cs
+++ b/DigitalLearningSolutions.Data/Services/TutorialContentService.cs
@@ -12,6 +12,7 @@
             int sectionId,
             int tutorialId
         );
+        TutorialContent? GetTutorialContent(int customisationId, int sectionId, int tutorialId);
     }
 
     public class TutorialContentService : ITutorialContentService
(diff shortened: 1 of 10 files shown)


FridaTveit:
@benjimarshall I think this includes one of my commits as well which it probably isn't meant to? 🙂

benjimarshall:
Yep working on it 🙂

benjimarshall:
That should be consistent now. Sorry about that!

DigitalLearningSolutions.Data/Services/TutorialContentService.cs
```diff
@@ -65,5 +76,33 @@ FROM Tutorials
                      AND CustomisationTutorials.Status = 1;",
             new { candidateId, customisationId, sectionId, tutorialId });
         }
+
+        public TutorialContent? GetTutorialContent(int customisationId, int sectionId, int tutorialId)
+        {
+            return connection.QueryFirstOrDefault<TutorialContent?>(
```
FridaTveit, line 82: Do you need the question mark here? 🙂

FridaTveit (commented):
One tiny question, otherwise looks good 🙂

DigitalLearningSolutions.Data/Services/TutorialContentService.cs
benjimarshall: I think I need one in the return type (because this method returns null if the `customisationId`/`sectionId`/`tutorialId` values are inconsistent with one another, but I don't think I need it in the `QueryFirstOrDefault<...>` generic. I had this in the other method, `GetTutorialInformation`, so I've removed them both in ed148f7 🙂

FridaTveit: Yes, that's what I meant sorry. I agree it's needed in the return type 🙂

FridaTveit (approved):
Looks good 🙂

PR #80, opened 2020-12-10
HEEDLS-177 Section page basic info integration

kwunkiuma:
Integrate the front and back ends for basic info on the section page.

Replace hardcoded values in view with view model values.

Ran and passed all unit tests. Manually tested on Firefox and pages are displaying the correct information, including when opening a unenrolled section via URL.

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
@@ -21,6 +21,7 @@
         private ICourseContentService courseContentService;
         private ITutorialContentService tutorialContentService;
         private ISessionService sessionService;
+        private ISectionContentService sectionContentService;
         private ISession httpContextSession;
         private IConfiguration config;
         private const int CandidateId = 11;
@@ -37,6 +38,7 @@
             courseContentService = A.Fake<ICourseContentService>();
             tutorialContentService = A.Fake<ITutorialContentService>();
             sessionService = A.Fake<ISessionService>();
+            sectionContentService = A.Fake<ISectionContentService>();
 
             var user = new ClaimsPrincipal(new ClaimsIdentity(new[]
             {
@@ -45,7 +47,7 @@
             }, "mock"));
             httpContextSession = new MockHttpContextSession();
 
-            controller = new LearningMenuController(logger, config, courseContentService, tutorialContentService, sessionService)
+            controller = new LearningMenuController(logger, config, courseContentService, sectionContentService, tutorialContentService, sessionService)
             {
                 ControllerContext = new ControllerContext {
                     HttpContext = new DefaultHttpContext
@@ -245,15 +247,19 @@
             A.CallTo(() => sessionService.StartOrUpdateSession(A<int>._, A<int>._, A<ISession>._)).MustNotHaveHappened();
         }
 
-
         [Test]
-        public void Sections_should_StartOrUpdate_course_sessions()
+        public void Sections_should_StartOrUpdate_course_sessions_if_valid_section()
         {
             // Given
-            const int sectionId = 199;
+            const int progressId = 299;
+            var defaultSectionContent = SectionContentHelper.CreateDefaultSectionContent();
+            A.CallTo(() => sectionContentService.GetSectionContent(CustomisationId, CandidateId, SectionId))
+                .Returns(defaultSectionContent);
+            A.CallTo(() => courseContentService.GetOrCreateProgressId(CandidateId, CustomisationId, CentreId))
+                .Returns(progressId);
 
             // When
-            controller.Section(CustomisationId, sectionId);
+            controller.Section(CustomisationId, SectionId);
 
             // Then
             A.CallTo(() => sessionService.StartOrUpdateSession(CandidateId, CustomisationId, httpContextSession)).MustHaveHappenedOnceExactly();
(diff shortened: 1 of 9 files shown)


DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
```diff
@@ -45,7 +47,7 @@ public void SetUp()
             }, "mock"));
             httpContextSession = new MockHttpContextSession();
 
-            controller = new LearningMenuController(logger, config, courseContentService, tutorialContentService, sessionService)
+            controller = new LearningMenuController(logger, config, courseContentService, sectionContentService, tutorialContentService, sessionService)
```
FridaTveit, line 50: Minor but this should probably be split over several lines as it's getting a bit long 🙂

DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
```diff
@@ -263,6 +269,145 @@ public void Sections_should_StartOrUpdate_course_sessions()
...
+
+            // When
+            controller.Section(CustomisationId, SectionId);
+
+            // Then
+            A.CallTo(() => sessionService.StartOrUpdateSession(A<int>._, A<int>._, A<ISession>._)).MustNotHaveHappened();
+        }
+
+        [Test]
+        public void Sections_should_UpdateProgress_course_sessions_if_valid_section()
```
FridaTveit, line 304: Is the `course_sessions` part of this test name necessary? I'm not quite sure what it refers to? 🙂

DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
```diff
@@ -263,6 +269,145 @@ public void Sections_should_StartOrUpdate_course_sessions()
...
+            A.CallTo(() => courseContentService.GetOrCreateProgressId(A<int>._, A<int>._, A<int>._)).MustNotHaveHappened();
+            result.Should()
+                .BeRedirectToActionResult()
+                .WithControllerName("LearningSolutions")
+                .WithActionName("StatusCode")
+                .WithRouteValue("code", 404);
+        }
+
+        [Test]
+        public void Sections_should_500_if_failed_to_enrol()
```
FridaTveit, line 392: We've changed this to 404 on the tutorial and initial menu pages. I realised this could happen in a few "expected" cases, e.g. if the user centre doesn't match the customisation centre. In that case I think a 404 is more suitable 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
```diff
@@ -0,0 +1,117 @@
...
+
+    public class SectionContentViewModelTests
+    {
+        [Test]
+        public void Section_content_should_have_title()
+        {
+            // Given
+            const string applicationName = "Application name";
+            const string customisationName = "Customisation name";
+            const int customisationId = 5;
```
FridaTveit, line 16: Could have this as a variable on the class which they all reuse, if it's going to stay the same? 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
```diff
@@ -0,0 +1,117 @@
...
+            var sectionContent = SectionContentHelper.CreateDefaultSectionContent(
+                hasLearning: hasLearning,
+                percentComplete: percentComplete
+            );
+
+            // When
+            var sectionContentViewModel = new SectionContentViewModel(sectionContent, customisationId);
+
+            // Then
+            sectionContentViewModel.PercentComplete.Should().Be($"{percentComplete:f0}% Complete");
```
FridaTveit, line 62: I'd prefer for this to be written out explicitly instead of relying on `f0`. It makes the test a bit easier to read and means we're not relying on logic which could be wrong 🙂

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
```diff
@@ -76,8 +79,42 @@ public IActionResult Close()
         [Route("/LearningMenu/{customisationId:int}/{sectionId:int}")]
         public IActionResult Section(int customisationId, int sectionId)
         {
-            sessionService.StartOrUpdateSession(User.GetCandidateId(), customisationId, HttpContext.Session);
-            return View("Section/Section");
+            var candidateId = User.GetCandidateId();
+            var centreId = User.GetCentreId();
+
+            if (centreId == null)
```
FridaTveit, line 85: Could combine this with the check below as we have done with other actions 🙂

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
```diff
@@ -104,8 +141,8 @@ public IActionResult Tutorial(int customisationId, int sectionId, int tutorialId
                 return RedirectToAction("StatusCode", "LearningSolutions", new { code = 404 });
             }
 
-            courseContentService.UpdateProgress(progressId.Value);
             sessionService.StartOrUpdateSession(candidateId, customisationId, HttpContext.Session);
+            courseContentService.UpdateProgress(progressId.Value);
```
FridaTveit, line 145: Good spot! 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
kwunkiuma: It's been used in the tutorial page tests so I included it for consistency. I suppose it's more relevant to the `StartOrUpdate` tests since the `UpgradeProgress` method only uses the sessions and doesn't modify them.

Should I remove it from the tutorial tests too?

FridaTveit: I see! Yes, please remove them. It's only relevant to the StartOrUpdate tests 🙂

FridaTveit (approved):
Looks good 🙂

PR #82, opened 2020-12-11
Add ContentViewer front end

benjimarshall:
## Changes
- Add parameters to ContentViewer view model to replace hard coded values.
- Get data from TutorialContentService in the Learning Menu Controller
to pass into the ContentViewer view.
- Add some CSS to control the page content based on whether JS is
enabled.
- Add tests to these methods.

## Testing
Add some unit tests for controller and view model. Tested site using Chrome, 
Firefox, Edge and IE11, all with a screenreader, and with no-JS and high zoom in Chrome. 

## Questions
- There are scrollbars on some of the HTML content as well as some of the 
SCORM content. See the last screenshot. Is this ok?
- The height for the `iframe` was fixed in the proof of concept code, but this can lead to scrollbars (eg. 2nd screenshot of SCORM player) that could be avoided by allowing it to be longer. Should we change this?

### JavaScript disabled error message (and content `iframe` is hidden)
nojs_error

### A SCORM tutorial (note the scrollbars)
scorm_tutorial_scrollbar

### A HTML tutorial (note the scrollbar, is this ok?)
html_tutorial_scrollbar

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
@@ -571,7 +571,7 @@
         }
 
         [Test]
-        public void ContentViewer_should_StartOrUpdate_course_sessions_if_valid_tutorial()
+        public void ContentViewer_should_StartOrUpdate_session_if_valid_tutorial()
         {
             // Given
             var expectedTutorialContent = TutorialContentHelper.CreateDefaultTutorialContent();
@@ -595,7 +595,7 @@
         }
 
         [Test]
-        public void ContentViewer_should_not_StartOrUpdate_course_sessions_if_invalid_tutorial()
+        public void ContentViewer_should_not_StartOrUpdate_session_if_invalid_tutorial()
         {
             // Given
             const int progressId = 3;
@@ -614,7 +614,7 @@
         }
 
         [Test]
-        public void ContentViewer_should_not_StartOrUpdate_course_sessions_if_unable_to_enrol()
+        public void ContentViewer_should_not_StartOrUpdate_session_if_unable_to_enrol()
         {
             // Given
             var expectedTutorialContent = TutorialContentHelper.CreateDefaultTutorialContent();
@@ -633,7 +633,7 @@
         }
 
         [Test]
-        public void ContentViewer_should_UpdateProgress_course_sessions_if_valid_tutorial()
+        public void ContentViewer_should_UpdateProgress_if_valid_tutorial()
         {
             // Given
             var expectedTutorialContent = TutorialContentHelper.CreateDefaultTutorialContent();
@@ -655,7 +655,7 @@
         }
 
         [Test]
-        public void ContentViewer_should_not_UpdateProgress_course_sessions_if_invalid_tutorial()
+        public void ContentViewer_should_not_UpdateProgress_if_invalid_tutorial()
         {
             // Given
             const int progressId = 3;
@@ -673,7 +673,7 @@
         }
 
         [Test]
-        public void ContentViewer_should_not_UpdateProgress_course_sessions_if_unable_to_enrol()
+        public void ContentViewer_should_not_UpdateProgress_if_unable_to_enrol()
         {
             // Given
             var expectedTutorialContent = TutorialContentHelper.CreateDefaultTutorialContent();
--- a/DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
@@ -125,19 +125,6 @@
         }
 
         [Test]
-        public void ContentViewer_should_render_view()
-        {
-            // When
-            var result = controller.ContentViewer(CustomisationId, SectionId, TutorialId);
-
-            // Then
-            var expectedModel = new ContentViewerViewModel(config);
-
-            result.Should().BeViewResult()
-                .Model.Should().BeEquivalentTo(expectedModel);
-        }
-
-        [Test]
         public void Index_always_calls_get_course_content()
         {
             // Given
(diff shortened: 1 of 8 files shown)


FridaTveit:
> There are scrollbars on some of the HTML content as well as some of the SCORM content. See the last screenshot. Is this ok?

As per the discussion with Kevin during the demo I think this is fine to leave unless it's easy to fix 🙂 

> The height for the iframe was fixed in the proof of concept code, but this can lead to scrollbars (eg. 2nd screenshot of SCORM player) that could be avoided by allowing it to be longer. Should we change this?

That was just temporary for the proof of concept, so feel free to change to a better solution. I wasn't paying attention to any styling/UI things for the proof of concept, I was just checking I could get it to work 🙂

DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
```diff
@@ -583,6 +570,201 @@ public void Tutorials_should_return_404_if_unable_to_enrol()
...
+            // When
+            controller.ContentViewer(CustomisationId, SectionId, TutorialId);
+
+            // Then
+            A.CallTo(() => sessionService.StartOrUpdateSession(A<int>._, A<int>._, A<ISession>._))
+                .MustNotHaveHappened();
+        }
+
+        [Test]
+        public void ContentViewer_should_UpdateProgress_course_sessions_if_valid_tutorial()
```
FridaTveit, line 636: `course_sessions` doesn't really make sense in this test name? 🙂

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
```diff
@@ -143,7 +143,39 @@ public IActionResult Tutorial(int customisationId, int sectionId, int tutorialId
...
         public IActionResult ContentViewer(int customisationId, int sectionId, int tutorialId)
         {
-            var model = new ContentViewerViewModel(config);
+            var candidateId = User.GetCandidateId();
+            var centreId = User.GetCentreId();
+
+            var tutorialContent = tutorialContentService.GetTutorialContent(customisationId, sectionId, tutorialId);
+
+            if (tutorialContent?.TutorialPath == null || centreId == null)
+            {
```
FridaTveit, line 152: A log would be useful, as I think we have in the other actions? 🙂

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
```diff
@@ -143,7 +143,39 @@ public IActionResult Tutorial(int customisationId, int sectionId, int tutorialId
...
+            if (progressId == null)
+            {
+                logger.LogError(
+                    "Redirecting to 404 as no progress id was returned. " +
+                    $"Candidate id: {candidateId}, customisation id: {customisationId}, centre id: {centreId}");
+                return RedirectToAction("StatusCode", "LearningSolutions", new { code = 404 });
+            }
+
+            courseContentService.UpdateProgress(progressId.Value);
+            sessionService.StartOrUpdateSession(candidateId, customisationId, HttpContext.Session);
```
FridaTveit, line 167: You need to update the session before updating the progress. Because updating the progress relies on a new session having been created 🙂

DigitalLearningSolutions.Web/Styles/index.scss
```diff
@@ -16,6 +16,14 @@ ul > li > ul > li {
   display: none;
 }
 
+ .js-only-block {
```
FridaTveit, line 19: Indentation a little bit off?

DigitalLearningSolutions.Web/ViewModels/LearningMenu/ContentViewerViewModel.cs
```diff
@@ -1,50 +1,67 @@
...
+    using System.Text.RegularExpressions;
+    using DigitalLearningSolutions.Data.Models.TutorialContent;
     using DigitalLearningSolutions.Web.Helpers;
     using Microsoft.Extensions.Configuration;
 
     public class ContentViewerViewModel
     {
-        public string HtmlSource { get; }
-        public string ScormSource { get; }
+        public TutorialContent TutorialContent { get; }
```
FridaTveit, line 10: Do you need to store the entire tutorial content? 🙂

FridaTveit (commented):

benjimarshall:
> That was just temporary for the proof of concept, so feel free to change to a better solution. I wasn't paying attention to any styling/UI things for the proof of concept, I was just checking I could get it to work 🙂

I've had a look into this and it's quite hard to get cross-domain iframes to automatically resize. I know the iframes are not crossdomain in the actual system, but it would make our testing a lot harder so I'm tempted to leave it as it is

DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
benjimarshall: Yep, renamed in b073a6a 🙂

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
benjimarshall: Added in b073a6a 🙂

FridaTveit:
So with the current behaviour it would work there would just sometimes be vertical and horizontal scroll bars? I think that should be fine to leave in that case 🙂

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
benjimarshall: Thanks, I'm not sure how those got swapped over. Fixed in b073a6a 🙂

DigitalLearningSolutions.Web/Styles/index.scss
benjimarshall: Whoops, fixed in b073a6a 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/ContentViewerViewModel.cs
benjimarshall: I use all of the `TutorialContent` to make the source URL, but then don't use too much of it. I suppose I could take it in the constructor to make this URL, but then discard it afterwards if you like? 🙂

FridaTveit: Yes I think I'd prefer that. I think it's best if there's nothing publically accessible which isn't necessary. I think that makes the code a bit easier to understand 🙂

benjimarshall: Done in 141cbd5 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/ContentViewerViewModelTests.cs
```diff
@@ -19,32 +26,216 @@ public void SetUp()
...
+
+            // Then
+            contentViewerViewModel.ProgressId.Should().Be(ProgressId);
+        }
+
+        [Test]
+        public void Content_viewer_should_have_tutorialName()
+        {
+            // Given
+            var expectedTutorialContent = TutorialContentHelper.CreateDefaultTutorialContent();
```
FridaTveit, line 142: I'd suggest setting the name when creating this, as it's important for the test 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/ContentViewerViewModelTests.cs
```diff
@@ -19,32 +26,216 @@ public void SetUp()
...
+
+            // Then
+            contentViewerViewModel.TutorialName.Should().BeEquivalentTo(expectedTutorialContent.TutorialName);
+        }
+
+        [Test]
+        public void Content_viewer_should_have_courseTitle()
+        {
+            // Given
+            var expectedTutorialContent = TutorialContentHelper.CreateDefaultTutorialContent();
```
FridaTveit, line 164: Similarly I'd suggest setting the course title here 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/ContentViewerViewModel.cs
```diff
@@ -1,50 +1,77 @@
...
+            TutorialId = tutorialId;
+            CandidateId = candidateId;
+            ProgressId = progressId;
+
+            TutorialName = tutorialContent.TutorialName;
+            CourseTitle = tutorialContent.CourseTitle;
+
+            if (IsScormPath(tutorialContent.TutorialPath!))
+            {
+                ContentSource = GetScormSource(config, tutorialContent);
```
FridaTveit, line 45: Can be slightly simplified using ternary operator 🙂

FridaTveit (commented):
A few minor things, otherwise looks good 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/ContentViewerViewModelTests.cs
benjimarshall: Yep, done in 79d4b90 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/ContentViewerViewModelTests.cs
benjimarshall: Yep, done in 79d4b90 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/ContentViewerViewModel.cs
benjimarshall: Yep, done in 79d4b90 🙂

FridaTveit (approved):
Looks good 🙂

PR #83, opened 2020-12-11
HEEDLS-217 Section Page Tutorial And Assessments List Back End

SChandarana:
Back end portion of this ticket is done. Quite a big list of changes to decided to do the FE to BE integration in a different ticket.
Let me know if any tests are missing/would be good to have :D


kwunkiuma (commented):
Some tab indentation issues.

DigitalLearningSolutions.Data/Models/SectionContent/SectionContent.cs
```diff
@@ -1,30 +1,59 @@
...
         public double PercentComplete { get; }
+        public int DiagAttempts { get; }
+        public int SecScore { get; }
+        public int SecOutOf { get; }
+        public string DiagAssessPath { get; }
+        public string PLAssessPath { get; }
+        public int AttemptsPL { get; }
+        public int PLPassed { get; }
+        public bool DiagStatus { get; }
+        public bool IsAssessed { get; }
```
FridaTveit, line 21: We don't need to stick to the same names as in the old system, so I think some of these names could be improved 🙂

DigitalLearningSolutions.Data/Models/SectionContent/SectionTutorial.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace DigitalLearningSolutions.Data.Models.SectionContent
+{
+    public class SectionTutorial
+    {
+        public string TutorialName { get; }
+        public string CompletionStatus { get; }
+        public int TutorialTime { get; }
+        public int AverageTutorialTime { get; }
```
FridaTveit, line 8: Should have empty line before constructor 🙂

DigitalLearningSolutions.Web.Tests/TestHelpers/SectionContentHelper.cs
```diff
@@ -8,20 +8,38 @@ public static class SectionContentHelper
...
+            int timeMins = 1,
             int averageSectionTime = 2,
             bool hasLearning = true,
-            double percentComplete = 10
+            double percentComplete = 10,
+            int diagAttempts = 5,
+            int secScore = 10,
+            int secOutOf = 14,
+            string diagAssessPath = "Path to Diagnostic Assessment",
+            string plAssessPath = "Path to Post Learning Assessment",
```
FridaTveit, line 19: Could you use more realistic values for the paths? 🙂

DigitalLearningSolutions.Data/Services/SectionContentService.cs
```diff
@@ -24,56 +24,129 @@ public SectionContentService(IDbConnection connection, ILogger<SectionContentSer
...
+                            AND aspProgress.ProgressID = Progress.ProgressID
+                        LEFT JOIN TutStatus
+                            ON aspProgress.TutStat = TutStatus.TutStatusID
+                        WHERE CustomisationTutorials.CustomisationID = @customisationId
+                        ORDER BY CustomisationTutorials.TutorialID",
+                    (section, tutorial) =>
+                {
+                    sectionContent ??= section;
+
+                    sectionContent.Tutorials.Add(tutorial);
```
FridaTveit, line 144: So the reason we need to do the nested select is because we want to sum various info from the tutorials right? If so, could we just do that manually here? As we're having to go manually go through each entry to add the tutorials anyway. And then we probably wouldn't need the nested select and could simplify this query? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data/Services/SectionContentService.cs
SChandarana: Would that work with Dapper Multimapping? I'm not sure exactly how it works but I assumed all of the section information would need to be the same for the fields

FridaTveit: So if we take the TutTime as an example, I think something like this would work:
* Set it to aspProgress.TutTime in the query
* Here when adding the tutorial to the section content, also add the TutTime. You'd have to be a bit careful not to add the first one twice, so something like:
```
if (sectionContent == null)
{
  sectionContent = section;
}
else 
{
  sectionContent.TutTime += section.TutTime;
}
```

What do you think? 🙂

SChandarana: I think that works for most of them, there comes a problem when we go for Percent Complete as it isn't just a simple sum. For that would you recommend having a method within `SectionContent.cs` which takes the total percent complete (calculated in the same way as TutTime) and divides it by the total number of tutorials

SChandarana: I'll also just have to check what all the other coalesces/sum/max statements do and then replicate them within the multimap

FridaTveit: For percentage complete I'd recommend changing SectionContent so that it calculates that based on the other fields (by using a custom getter). That can use the length of the tutorials list + hasLearning to calculate it. Does that make sense? 🙂

DigitalLearningSolutions.Data/Services/SectionContentService.cs
```diff
@@ -57,23 +66,43 @@ ELSE CAST(SUM(aspProgress.TutStat) * 100 AS FLOAT) / (COUNT(Tutorials.TutorialID
                         LEFT JOIN aspProgress
                             ON aspProgress.TutorialID = CustomisationTutorials.TutorialID
                             and aspProgress.ProgressID = Progress.ProgressID
+                        LEFT JOIN (SELECT COUNT(AssessAttemptID) AS AttemptsPL,
+                            AssessAttempts.ProgressID,
+                            AssessAttempts.SectionNumber,
+                            COALESCE (MAX(ISNULL(CAST(AssessAttempts.Status AS Integer), 0)), 0) AS PLPassed
```
FridaTveit, line 72: What exactly are you trying to do here? 🙂

FridaTveit (commented):
A few comments, otherwise looks good. Good job on the slightly tricky refactor! 🙂

DigitalLearningSolutions.Data/Services/SectionContentService.cs
SChandarana: I'm not 100% sure on this one, I took the select statement from the stored procedure. Wasn't entirely sure how it worked

DigitalLearningSolutions.Data/Models/SectionContent/SectionContent.cs
```diff
@@ -1,30 +1,56 @@
...
-        public int? TimeMins { get; }
+        public int SectionTime { get; set; }
         public int AverageSectionTime { get; }
         public bool HasLearning { get; }
-        public double PercentComplete { get; }
+        public int DiagnosticAttempts { get; set; }
+        public int SectionScore { get; set; }
+        public int MaxSectionScore { get; set; }
+        public string DiagnosticAssessmentPath { get; }
+        public string PostLearningAssessmentPath { get; }
```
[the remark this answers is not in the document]
SChandarana: I'm not sure, I thought all tutorials had paths and so left all the data in the class. I was then planning on having all the logic for whether or not they should be used/displayed on the page will be done in the view model

DigitalLearningSolutions.Data/Models/SectionContent/SectionContent.cs
```diff
@@ -6,32 +6,30 @@ public class SectionContent
...
-        public int AttemptsPL { get; }
-        public int PLPassed { get; }
-        public bool DiagStatus { get; }
+        public int DiagnosticAttempts { get; set; }
+        public int SectionScore { get; set; }
+        public int MaxSectionScore { get; set; }
+        public string DiagnosticAssessmentPath { get; }
+        public string PostLearningAssessmentPath { get; }
+        public int PostLearningAttempts { get; }
+        public int PostLearningPassed { get; }
```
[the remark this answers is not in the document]
SChandarana: For some reason the type was coming out as an int rather than a boolean. I think this might have something to do with weird SELECT statement that was being used which has to CAST to an int to use MAX. Dapper wasn't accepting a bool in the constructor, only an int

DigitalLearningSolutions.Data/Models/SectionContent/SectionContent.cs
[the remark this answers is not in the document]
SChandarana: Just ran a query to check, this is nullable so will add that here :D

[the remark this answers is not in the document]
FridaTveit: If you look in the db you'll see that both fields can be null. We should make sure to check for that when writing queries so that we can correctly mark things as optional when necessary 🙂

DigitalLearningSolutions.Data/Services/SectionContentService.cs
FridaTveit: I see, I've found it now. So we're trying to assentially OR together all the Status fields (which are booleans). However that's sadly a pain to do in SQL 🙁 
But I think we can fix this and the type problem in one go:
1. Change this to count the number of passes instead of checking if there is one, e.g. `SUM(CAST(Status AS INT))`
2. Accept an int in the constructor of the section content, but make the constructor change it to a boolean (true if the sum is greater than 0).

Does that make sense? 🙂

DigitalLearningSolutions.Data/Models/SectionContent/SectionContent.cs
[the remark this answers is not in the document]
FridaTveit: Solution suggested in my other comment (on the SQL query) 🙂

DigitalLearningSolutions.Data/Models/SectionContent/SectionContent.cs
```diff
@@ -30,7 +30,7 @@ public class SectionContent
             int diagAttempts,
             int diagLast,
             int diagAssessOutOf,
-            string diagAssessPath,
+            string? diagAssessPath,
             string plAssessPath,
```
FridaTveit, line 34: Should also be optional 🙂

DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
```diff
@@ -316,5 +316,19 @@ public void Get_section_content_should_return_content_if_only_tutorial_status_is
             );
             result.Should().BeEquivalentTo(expectedSectionContent);
         }
+
+        [Test]
+        public void Get_section_content_post_learning_assessment_passed_should_convert_to_true()
```
FridaTveit, line 321: We should be able to test this when we use it in the view model? 🙂

DigitalLearningSolutions.Data/Services/SectionContentService.cs
```diff
@@ -24,23 +25,31 @@ public SectionContentService(IDbConnection connection, ILogger<SectionContentSer
...
-                        END) AS PercentComplete
-                    FROM
-                        Tutorials
+                        COALESCE (aspProgress.DiagAttempts, 0) AS DiagAttempts,
+                        COALESCE (aspProgress.DiagLast, 0) AS DiagLast,
+                        Tutorials.DiagAssessOutOf,
+                        Sections.DiagAssessPath, 
+                        Sections.PLAssessPath,
+                        COALESCE (Attempts.AttemptsPL,0) AS AttemptsPL,
+                        COALESCE (Attempts.PLPassed,0) AS PLPassed,
```
FridaTveit, line 44: This and Attempts.PLPassed should be called `..Passes` instead of `...Passed` because of how we've changed the calculation.

Also the spacing is inconsistent here again and in one of the lines below.

And I think we're coalescing this twice now, both here and when defining Assessed? Both shouldn't be necessary. I think this one is necessary and the one when defining Assessed is not. Because you're doing a left join so the Attempts fields could be null. But when summing the AssessAttempts Status entries none of them can be null as that's not nullable 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/SectionContentViewModel.cs
```diff
@@ -9,14 +11,25 @@ public class SectionContentViewModel
...
+            Tutorials = sectionContent.Tutorials.Select(tutorial => new TutorialCardViewModel(tutorial));
+        }
+
+        public double GetPercentComplete(SectionContent sectionContent)
+        {
+            var totalStatus = sectionContent.Tutorials.Sum(tutorial => tutorial.TutorialStatus);
+            return sectionContent.Tutorials.Count == 0 || !sectionContent.HasLearning
+                ? 0
+                : (totalStatus * 100) / (sectionContent.Tutorials.Count * 2);
+
```
FridaTveit, line 32: Unneeded empty line

DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
```diff
@@ -31,13 +31,30 @@ public void Get_section_content_should_return_section_content()
...
+                true
+            );
+            expectedSectionContent.Tutorials.AddRange(
+                new[]
+                {
+                    new SectionTutorial("Introduction to applications", 0, "Not started", 0, 17),
+                    new SectionTutorial("Common screen elements", 0, "Not started", 0, 11),
+                    new SectionTutorial("Using ribbon tabs", 0, "Not started", 0, 6),
+                    new SectionTutorial("Getting help", 0, "Not started", 0, 11)
+                }
```
FridaTveit, line 57: Do we have a test for getting the section content if there's no progress, i.e. the user isn't enrolled, like we do for the initial menu content? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
SChandarana: I thought it was sensible to have it here since the logic for it was in the data model. Do you think it would be better to have the test in the view model?

DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
SChandarana: Does the test for an unenrolled candidate count? As a candidate that isn't enrolled won't have a progressId for it?

FridaTveit: Yes, that's what I meant. Couldn't see it so just wanted to make sure we had it, as that should make sure that we have correct default values for when there's no progress 🙂

DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
FridaTveit: But it's not to do with the service so doesn't belong in the service tests 🙂 We could have a separate test for the section content itself, or we could test the logic as part of the view model. As the logic is so small I don't really mind which one we go for. But either way we need more than one test to check that it's working. I think testing 0, 1 and > 1 passes should do. You could do that using `[TestCase]` 🙂

SChandarana: Yep that's fine, I'll do it in the View Model test. Do you want me to add it as part of this PR or part of the integration PR as currently we aren't using any of these values in the view model?

FridaTveit: Fine to leave for now and do as part of the integration 🙂

FridaTveit (approved):
Looks good 🙂

PR #85, opened 2020-12-14
Add objectives to tutorial page

benjimarshall:
## Changes
- Install the Html Agility Pack NuGet package.
- Add tutorial objective parsing, to the view model, and display it in
the Tutorial view.
- Add tests for this view model parsing method.

## Testing
Add view model unit tests, test in Chrome, Firefox, Edge and IE11, using a screen reader and large zoom.

### Collapsed objectives
no_objectives

### Expanded objectives
objectives_expanded

### A tutorial without any objectives
objectives_collapsed

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/TutorialViewModelTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/TutorialViewModelTests.cs
@@ -160,5 +160,66 @@
             // Then
             tutorialViewModel.TutorialRecommendation.Should().Be("optional");
         }
+
+        [Test]
+        public void Tutorial_parsed_objectives_are_null_when_objectives_are_null()
+        {
+            // Given
+            var expectedTutorialInformation = TutorialContentHelper.CreateDefaultTutorialInformation(
+                objectives: null
+            );
+
+            // When
+            var tutorialViewModel = new TutorialViewModel(expectedTutorialInformation, CustomisationId, SectionId);
+
+            // Then
+            tutorialViewModel.Objectives.Should().BeNull();
+        }
+
+        [Test]
+        public void Tutorial_parses_inline_html_objectives()
+        {
+            // Given
+            const string objectives =
+                "Here are some example objectives: <ul> <li> objective 1 </li> <li> objective 2 </li> </ul>";
+            var expectedTutorialInformation = TutorialContentHelper.CreateDefaultTutorialInformation(
+                objectives: objectives
+            );
+
+            // When
+            var tutorialViewModel = new TutorialViewModel(expectedTutorialInformation, CustomisationId, SectionId);
+
+            // Then
+            tutorialViewModel.Objectives.Should().Be(objectives);
+        }
+
+        [Test]
+        public void Tutorial_parses_html_document_of_objectives()
+        {
+            // Given
+            const string objectives =
+                "<html> <head> <title>Tutorial Objective</title> </head>" +
+                "<body> In this tutorial you will learn to: " +
+                  "<ul>" +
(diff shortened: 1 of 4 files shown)


DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/TutorialViewModelTests.cs
```diff
@@ -160,5 +160,66 @@ public void Tutorial_should_not_recommend_a_tutorial_when_score_is_max()
             // Then
             tutorialViewModel.TutorialRecommendation.Should().Be("optional");
         }
+
+        [Test]
+        public void Tutorial_parsed_objectives_are_null_when_objectives_are_null()
```
FridaTveit, line 165: This test name doesn't quite read like a sentance. Maybe something like `Tutorial_handles_null_objectives`? 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/TutorialViewModel.cs
```diff
@@ -35,5 +40,21 @@ private static string ParseMinutes(int minutes)
...
+                return null;
+            }
+
+            var doc = new HtmlDocument();
+            doc.LoadHtml(objectives);
+
+            var body = doc.DocumentNode.SelectNodes("//body");
+
+            return body?.FirstOrDefault()?.InnerHtml
+                   ?? WebUtility.HtmlDecode(objectives);
```
FridaTveit, line 57: Sorry, why do we need to call `HtmlDecode` here? 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/TutorialViewModel.cs
```diff
@@ -35,5 +40,21 @@ private static string ParseMinutes(int minutes)
...
+            {
+                return null;
+            }
+
+            var doc = new HtmlDocument();
+            doc.LoadHtml(objectives);
+
+            var body = doc.DocumentNode.SelectNodes("//body");
+
+            return body?.FirstOrDefault()?.InnerHtml
```
FridaTveit, line 56: Why the `FirstOrDefault`? Don't we just want everything inside the body? I was imagining something like this https://html-agility-pack.net/from-string but with inner html instead of outer (I assume that will remove the body tags), although I haven't tried that 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web/ViewModels/LearningMenu/TutorialViewModel.cs
benjimarshall: So this produces a list of bodies, and I'm just taking the first body and returning its contents. Select Single node implements the `FirstOrDefault` for us so I've switched over to using that in 3ff95e1 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/TutorialViewModel.cs
benjimarshall: I'm not sure if I need this actually. It converts HTTP strings into normal character strings (eg. from `&gt;` to `>`) but now that I think of it I doubt it will be stored like that in the database (and I can't find any examples). Removed in 3ff95e1 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/TutorialViewModelTests.cs
benjimarshall: That's much better, thank you! Changed in 3ff95e1 🙂

FridaTveit (approved):
Looks good 🙂

PR #86, opened 2020-12-14
HEEDLS-239 Add backend to get Tutorial videos

benjimarshall:
## Changes
- Add a `TutorialVideo` data model
- Add a `GetTutorialVideo` method to the `TutorialContentService`
- Add tests for this method

## Testing
Add unit tests for the new method

### DIFF
--- a/DigitalLearningSolutions.Data/Models/TutorialContent/TutorialVideo.cs
+++ b/DigitalLearningSolutions.Data/Models/TutorialContent/TutorialVideo.cs
@@ -1,0 +1,21 @@
+﻿namespace DigitalLearningSolutions.Data.Models.TutorialContent
+{
+    public class TutorialVideo
+    {
+        public string TutorialName { get; }
+        public string CourseTitle { get; }
+        public string? VideoPath { get; }
+
+        public TutorialVideo(
+            string tutorialName,
+            string applicationName,
+            string customisationName,
+            string? videoPath
+        )
+        {
+            TutorialName = tutorialName;
+            CourseTitle = $"{applicationName} - {customisationName}";
+            VideoPath = videoPath;
+        }
+    }
+}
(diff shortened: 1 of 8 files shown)


DigitalLearningSolutions.Data/Models/TutorialContent/TutorialVideo.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace DigitalLearningSolutions.Data.Models.TutorialContent
+{
+    public class TutorialVideo
+    {
+        public string TutorialName { get; }
+        public string CourseTitle { get; }
+        public string? VideoPath { get; }
```
FridaTveit, line 7: Should this really be nullable? It won't be a valid tutorial video without a path, so we'll want to redirect to a 404 I think. In which case it might be better to get the service method to return null, if that makes sense? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data/Models/TutorialContent/TutorialVideo.cs
benjimarshall: I agree, but Dapper is still willing to make a TutorialVideo with a `null` `VideoPath` even if it isn't marked as nullable. Should I check for this in the `GetTutorialVideo` method? 🙂

FridaTveit: Yes, I think I'd do that 🙂

benjimarshall: Added in 7ce3621 🙂

FridaTveit: Wait actually we'd still need to mark it as optional then 🙁 Could we throw an exception in the contructor? I think we did that in a similar situation? 🙂

benjimarshall: Oh yeah, I had forgotten about that! Done in 41b5ea6 🙂

FridaTveit (approved):
Looks good 🙂

PR #87, opened 2020-12-15
HEEDLS-237 Redirect learning menu to 404 if customisation not active

kwunkiuma:
Change the query to filter out inactive customisations.

Some tests were using inactive customisations so new test cases had to be used.

Ran and passed all unit tests, manually tested on Firefox and checked logs to see the page was redirecting for the correct reason.


FridaTveit (approved):
Looks good 🙂

PR #88, opened 2020-12-15
HEEDLS-240 Launch video front end

benjimarshall:
## Changes
- Move ContentViewer to Tutorial folder
- Add an action to the LearningMenuController, and a tutorial video view
and view model
- Add tests for these objects and methods
- Add some CSS for the tutorial video page
- Link the "Play video" button on the tutorial overview to this action.

## Testing
Written some unit tests, tested in Chrome, Firefox, IE11 and Edge using a screen reader.
The "Your browser does not support videos" text is read by the screen reader in IE11 even when the video has rendered, but I think this is a known issue and is not a problem with the other browsers.

## Screenshot
video


DigitalLearningSolutions.Web/ViewModels/LearningMenu/TutorialVideoViewModel.cs
```diff
@@ -0,0 +1,45 @@
...
+            VideoPath = GetVideoPath(config, tutorialVideo.VideoPath);
+        }
+
+        private static string GetVideoPath(IConfiguration config, string videoPath)
+        {
+            if (Uri.IsWellFormedUriString(videoPath, UriKind.Absolute))
+            {
+                return videoPath;
+            }
+            if (Uri.IsWellFormedUriString("https://" + videoPath, UriKind.Absolute))
```
FridaTveit, line 38: Minor but could extract https + url to variable to avoid creating twice 🙂 And I'd generally use [string interpolation](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated) instead of `+`. But that's very minor in these small cases, feel free to stick to `+` if you prefer 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web/ViewModels/LearningMenu/TutorialVideoViewModel.cs
benjimarshall: Yeah I agree, done in 091daa2 🙂

FridaTveit (approved):
Looks good 🙂

PR #91, opened 2020-12-16
HEEDLS-241 Assessment page diagnostic front end

kwunkiuma:
Create the view using hardcoded values for the diagnostic assessment page.

Ran and passed all unit tests, screenshot taken in Firefox.

image

### DIFF
--- a/DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
@@ -110,7 +110,7 @@
             return View("Section/Section", model);
         }
 
-        [Route("/LearningMenu/1/1/Diagnostic")]
+        [Route("/LearningMenu/18438/993/Diagnostic")]
         public IActionResult Diagnostic()
         {
             return View("Diagnostic/Diagnostic");
--- a/DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
@@ -110,6 +110,12 @@
             return View("Section/Section", model);
         }
 
+        [Route("/LearningMenu/1/1/Diagnostic")]
+        public IActionResult Diagnostic()
+        {
+            return View("Diagnostic/Diagnostic");
+        }
+
         [Route("/LearningMenu/{customisationId:int}/{sectionId:int}/{tutorialId:int}")]
         public IActionResult Tutorial(int customisationId, int sectionId, int tutorialId)
         {
--- a/DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
@@ -110,8 +110,8 @@
             return View("Section/Section", model);
         }
 
-        [Route("/LearningMenu/18438/993/Diagnostic")]
-        public IActionResult Diagnostic()
+        [Route("/LearningMenu/{customisationId:int}/{sectionId:int}/Diagnostic")]
+        public IActionResult Diagnostic(int customisationId, int sectionId)
         {
             return View("Diagnostic/Diagnostic");
         }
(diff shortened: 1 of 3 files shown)


DigitalLearningSolutions.Web/Views/LearningMenu/Diagnostic/Diagnostic.cshtml
```diff
@@ -0,0 +1,86 @@
...
+    </div>
+
+    <button class="nhsuk-button nhsuk-u-margin-top-3 nhsuk-u-margin-bottom-0" type="submit">
+      Start assessment
+    </button>
+  </fieldset>
+</form>
+
+<div class="nhsuk-back-link">
+  <a class="nhsuk-back-link__link">
```
FridaTveit, line 80: Could set the link destination here, using hardcoded customisation and section ids 🙂

DigitalLearningSolutions.Web/Views/LearningMenu/Section/_DiagnosticAssessmentCard.cshtml
```diff
@@ -14,7 +14,7 @@
     </div>
     <div class="nhsuk-grid-row">
       <div class="nhsuk-grid-column-full">
-        <a class="nhsuk-button nhsuk-u-margin-bottom-0 nhsuk-u-margin-top-0">
+        <a class="nhsuk-button nhsuk-u-margin-bottom-0 nhsuk-u-margin-top-0" asp-action="Diagnostic">
```
FridaTveit, line 17: I noticed before when I was testing that this button is ignored by the screen reader. Is that still a problem now? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web/Views/LearningMenu/Section/_DiagnosticAssessmentCard.cshtml
kwunkiuma: The screen reader reads out "launch assessment" when hovering over the button (in Firefox). Is that the expected response?

FridaTveit: What specifically you want to test is moving through the screen using the navigation buttons and tab. Using tab should select all the clickable elements such as that button, and read it out as a button/link. There is information about how to test with a screen reader on our testing page: https://softwiretech.atlassian.net/wiki/spaces/HEE/pages/6703648740/Testing 🙂

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
```diff
@@ -110,7 +110,7 @@ public IActionResult Section(int customisationId, int sectionId)
             return View("Section/Section", model);
         }
 
-        [Route("/LearningMenu/1/1/Diagnostic")]
+        [Route("/LearningMenu/18438/993/Diagnostic")]
```
FridaTveit, line 113: I meant actually pass in values to this method, don't hardcode them. There's no reason to hardcode them as we should have this information available on the section page, right? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web/Views/LearningMenu/Section/_DiagnosticAssessmentCard.cshtml
kwunkiuma: I've tested the webpage with the screen reader (on Firefox) and it reads "link launch assessment" when using the navigation buttons and "launch assessment link" when using the tab key which I believe means it's working as intended.

FridaTveit: Great 🙂

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
kwunkiuma: `SectionID` doesn't seem to be available on the section page yet, Shiv is adding that now. I can change it so that it takes parameters but hardcode those values on the section page for now?

FridaTveit: Okay. @SChandarana the section id is available in the controller action so it's just a question of passing that from the action to the view model 🙂

FridaTveit (approved):
Looks good 🙂

PR #92, opened 2020-12-16
Add tutorial id to section tutorial

SChandarana:
Needed to add the tutorial idea to a section tutorial so we can direct to the correct tutorial page


FridaTveit (approved):
Looks great 🙂

PR #94, opened 2020-12-16
HEEDLS-244 Tutorial next button back end

benjimarshall:
## Changes
Amend `GetTutorialInformation` query and data model to get the next
tutorial ID, the next section ID, and the post-learning assessment
path.

## Testing
- Add test for course contents section test: test a course which has some sections where none of the section's
tutorials are in the CustomisationTutorial table (based on some checks I did this morning)
- Amend `GetTutorialInformation` tests, and add tests for edge cases like being the last tutorial
in a section, and check tutorials not in CustomisationTutorials are skipped.

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
@@ -445,6 +445,7 @@
         [TestCase(1, 15853, 101, 173218)]
         [TestCase(254480, 24224, 101, null)]
         [TestCase(22044, 10059, 121, 100467)]
+        [TestCase(11, 6226, 101, 89490)]
         public void Get_course_content_should_have_same_sections_as_stored_procedure(
             int candidateId,
             int customisationId,
--- a/DigitalLearningSolutions.Data.Tests/Services/TutorialContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/TutorialContentServiceTests.cs
@@ -129,6 +129,44 @@
         }
 
         [Test]
+        public void Get_tutorial_information_nextTutorial_can_return_smaller_tutorialId()
+        {
+            // Given
+            const int candidateId = 210934;
+            const int customisationId = 17731;
+            const int sectionId = 801;
+            const int tutorialId = 3332;
+
+            const int nextTutorialId = 3334;
+
+            // When
+            var tutorial = tutorialContentService.GetTutorialInformation(candidateId, customisationId, sectionId, tutorialId);
+
+            // Then
+            tutorial.Should().NotBeNull();
+            tutorial!.NextTutorialId.Should().Be(nextTutorialId);
+        }
+
+        [Test]
+        public void Get_tutorial_information_nextSection_can_return_smaller_sectionId()
+        {
+            // Given
+            const int candidateId = 210962;
+            const int customisationId = 24057;
+            const int sectionId = 2201;
+            const int tutorialId = 10184;
+
+            const int nextSectionId = 2193;
+
+            // When
+            var tutorial = tutorialContentService.GetTutorialInformation(candidateId, customisationId, sectionId, tutorialId);
+
+            // Then
+            tutorial.Should().NotBeNull();
+            tutorial!.NextSectionId.Should().Be(nextSectionId);
+        }
+
+        [Test]
         public void Get_tutorial_information_should_return_null_if_customisation_id_invalid()
         {
             // Given
--- a/DigitalLearningSolutions.Data.Tests/Services/TutorialContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/TutorialContentServiceTests.cs
@@ -135,9 +135,9 @@
             const int candidateId = 210934;
             const int customisationId = 17731;
             const int sectionId = 801;
-            const int tutorialId = 3332;
+            const int tutorialId = 3334;
 
-            const int nextTutorialId = 3334;
+            const int nextTutorialId = 3333;
 
             // When
             var tutorial = tutorialContentService.GetTutorialInformation(candidateId, customisationId, sectionId, tutorialId);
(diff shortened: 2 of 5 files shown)


DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
```diff
@@ -445,6 +445,7 @@ public void Get_course_content_should_only_return_sections_with_tutorials_in_cus
         [TestCase(1, 15853, 101, 173218)]
         [TestCase(254480, 24224, 101, null)]
         [TestCase(22044, 10059, 121, 100467)]
+        [TestCase(11, 6226, 101, 89490)]
```
FridaTveit, line 448: Great, thank you! Do keep adding any course you think might be useful to this test, the more the better! 🙂

DigitalLearningSolutions.Data.Tests/Services/TutorialContentServiceTests.cs
```diff
@@ -46,10 +46,88 @@ public void Get_tutorial_information_should_return_tutorial()
...
+        [Test]
+        public void Get_tutorial_information_nextTutorial_should_skip_tutorials_not_in_customisation()
+        {
+            // Given
+            const int candidateId = 210934;
+            const int customisationId = 18366;
+            const int sectionId = 973;
+            const int tutorialId = 4257;
+
+            // The next tutorial ID in this section is 4258, but the next tutorial selected in CustomisationTutorials is 4263
```
FridaTveit, line 99: Very useful comment, thanks! I think these tests where we're relying on the state of the db is a case where comments can be really useful 🙂

DigitalLearningSolutions.Data/Services/TutorialContentService.cs
```diff
@@ -49,17 +49,39 @@ int tutorialId
                          Tutorials.Objectives,
                          Tutorials.VideoPath,
                          Tutorials.TutorialPath,
-                         Tutorials.SupportingMatsPath AS SupportingMaterialPath
+                         Tutorials.SupportingMatsPath AS SupportingMaterialPath,
+                         Sections.PLAssessPath AS PostLearningAssessmentPath,
+                         MIN(SubsequentTutorials.TutorialID) AS NextTutorialID,
```
FridaTveit, line 54: Does this return the next tutorial according to the OrderByNumber? 🙂

DigitalLearningSolutions.Data/Services/TutorialContentService.cs
```diff
@@ -49,17 +49,39 @@ int tutorialId
                          Tutorials.Objectives,
                          Tutorials.VideoPath,
                          Tutorials.TutorialPath,
-                         Tutorials.SupportingMatsPath AS SupportingMaterialPath
+                         Tutorials.SupportingMatsPath AS SupportingMaterialPath,
+                         Sections.PLAssessPath AS PostLearningAssessmentPath,
+                         MIN(SubsequentTutorials.TutorialID) AS NextTutorialID,
+                         MIN(SubsequentSections.SectionID) AS NextSectionID
```
FridaTveit, line 55: Does this return the next section according to the SectionNumber? 🙂

DigitalLearningSolutions.Data/Services/TutorialContentService.cs
```diff
@@ -49,17 +49,39 @@ int tutorialId
...
+                         ON SubsequentCustomisationTutorials.CustomisationID = Customisations.CustomisationID
+                            AND SubsequentCustomisationTutorials.Status = 1
+
+                         LEFT JOIN Tutorials AS SubsequentTutorials
+                         ON SubsequentTutorials.SectionID = Sections.SectionID
+                            AND Tutorials.OrderByNumber < SubsequentTutorials.OrderByNumber
+                            AND SubsequentCustomisationTutorials.TutorialID = SubsequentTutorials.TutorialID
+
+                         LEFT JOIN Tutorials AS SubsequentSectionsTutorials
+                         ON  SubsequentCustomisationTutorials.TutorialID = SubsequentSectionsTutorials.TutorialID
```
FridaTveit, line 79: Why are we joining on Tutorials twice? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
benjimarshall: Will do 🙂

DigitalLearningSolutions.Data.Tests/Services/TutorialContentServiceTests.cs
benjimarshall: Yeah absolutely, and it was a bit ridiculous when I tried to encode all of this in the function title! 🙂

DigitalLearningSolutions.Data/Services/TutorialContentService.cs
benjimarshall: The `SubsequentTutorials` is used to join on the other tutorials in this section (that are in `CustomisationTutorials`) but only if it has a greater `OrderBy`. This is done in a left join because that will give us a single row with `null` in this `NextTutorialId`, instead of giving us an empty table if we join first and then filter later. There might be more than one tutorial that matches though, so we aggregate them into a single row with the `MIN` function.

The second join on tutorials, `SubsequentSectionsTutorials` gives us all of the tutorials in the customisation (which we then use to find all the sections that are in the customisation, as opposed to the customisations in the application), which we then join `SubsequentSections` to give us rows for each section in the customisation.

At one point I did just have the simpler second join, but I think it would make finding the next tutorial in the section more difficult, because if the tutorial is the last in the section we want this field to be `null`, and not just `WHERE` on having a greater `OrderBy` and getting an empty table. There might be another way to do this but I don't know about it.

Does that make sense? I'm not sure how clear that all was 🙂

DigitalLearningSolutions.Data/Services/TutorialContentService.cs
benjimarshall: Yes it does, I added that as a comment in bc4cce2 🙂

DigitalLearningSolutions.Data/Services/TutorialContentService.cs
benjimarshall: Yes it does, I added that as a comment in bc4cce2 🙂

DigitalLearningSolutions.Data/Services/TutorialContentService.cs
FridaTveit: >There might be more than one tutorial that matches though, so we aggregate them into a single row with the MIN function

So `SubsequentTutorials` gives all the subsequent tutorial in this section right? So if there's more than one tutorial after this one, how to we know that the MIN one is the next one? Shouldn't it be MIN based on the `OrderBy` instead of the `TutorialID`? 🙂 

The two joins on Tutorials make sense then, thanks 🙂

benjimarshall: Yep this is a problem, thanks for catching it! 🙂

benjimarshall: Fixed and regression tests added in 79d8087 🙂

benjimarshall:
Something else I just thought of is that there are a few sections with different tutorials sharing `OrderByNumber`s. I suspect there also are sections sharing numbers in a course. Currently this will arbitrarily pick one of them, but this means pressing next continuously from start to end of a course could lead to some sections being missed. 

Should we leave this (because tutorials sharing `OrderByNumber`s sounds strange, and only ~750/13000 tutorials do so), or try to account for this in some way (maybe ordering by tutorial ID as a tie break)? What do you think? 🙂

FridaTveit:
Good question... I'll check with Kevin if this is to be expected 🙂

FridaTveit (approved):
Looks good 🙂 Other than the concern about order by and section numbers being shared... But feel free to merge this for now and deal with that later if necessary 🙂

benjimarshall:
0a8424f adds support for shared `OrderByNumber`s and `SectionNumber`s. There's no tests for shared section numbers because there are no examples of this occurring in the database.

FridaTveit (approved):
Looks good 🙂

PR #96, opened 2020-12-17
HEEDLS-236-Section Page Assessments List Integration

SChandarana:
Integration all complete and tests added, let me know if there are any tests missing. As the PR was getting a little big I decided not to do the bug fix HEEDLS-250 and will instead do that separately :D


FridaTveit (approved):
Looks good 🙂

PR #98, opened 2020-12-17
HEEDLS-252 Add completion summary fields to course content

benjimarshall:
## Changes
Add completion summary data to the `CourseContent` model, update the `GetCourseContent` method

## Testing
Amend unit tests for `GetCourseContent` to cover these new fields


DigitalLearningSolutions.Data/Models/CourseContent/CourseContent.cs
```diff
@@ -12,6 +12,11 @@ public class CourseContent
         public string? BannerText { get; }
         public bool IncludeCertification { get; }
         public DateTime? Completed { get; }
+        public int AssessAttempts { get; }
```
FridaTveit, line 15: This is specifically post learning assessment attempts right? If so, can we make that clear in the name? 🙂 

Oh and it's also actually a max assessment attemps right? So worth making that clear in the name as well 🙂

FridaTveit (commented):
One comment about naming otherwise looks good 🙂

DigitalLearningSolutions.Data/Models/CourseContent/CourseContent.cs
benjimarshall: I think so, it's a field in `Applications`. Would something like `MaxPostLearningAssessmentAttempts` be good? 🙂

FridaTveit: Yes, that sounds good to me 🙂

benjimarshall: Done in 86103de 🙂

FridaTveit (approved):
Looks good 🙂

PR #99, opened 2020-12-17
HEEDLS-NONE- Add Tutorial ID to Order By

SChandarana:
Small bugfix to fix ordering of tutorials within the section page

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
@@ -316,5 +316,64 @@
             );
             result.Should().BeEquivalentTo(expectedSectionContent);
         }
+
+        [Test]
+        public void Get_section_content_should_order_by_tutorial_id_when_shared_order_by_number()
+        {
+            // When
+            const int candidateId = 1;
+            const int customisationId = 8194;
+            const int sectionId = 216;
+
+            // All in section 216
+            // Tutorial: 927  OrderByNumber 34
+            // Tutorial: 928  OrderByNumber 35
+            // Tutorial: 929  OrderByNumber 35
+            // Tutorial: 930  OrderByNumber 36
+            var result = sectionContentService.GetSectionContent(customisationId, candidateId, sectionId);
+
+            // Then
+            var expectedSectionContent = new SectionContent(
+                "Entry Level - Win 7, Office 2010",
+                "Testing",
+                "Working with applications",
+                11,
+                28,
+                true,
+                7,
+                35,
+                35,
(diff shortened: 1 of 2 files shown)


DigitalLearningSolutions.Data/Services/SectionContentService.cs
```diff
@@ -83,7 +83,7 @@ GROUP BY
                         AND Sections.SectionID = @sectionId
                         AND (Sections.ArchivedDate IS NULL)
                         AND (CustomisationTutorials.DiagStatus = 1 OR Customisations.IsAssessed = 1 OR CustomisationTutorials.Status = 1)
-                    ORDER BY Tutorials.OrderByNumber",
+                    ORDER BY Tutorials.OrderByNumber, Tutorials.TutorialID",
```
FridaTveit, line 86: Is there a test we can add for this? And this probably needs to be done on the initial menu page as well, for the sections? If so @benjimarshall it probably makes sense for you to do that as you're changing that page? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
```diff
@@ -316,5 +316,64 @@ public void Get_section_content_should_return_content_if_only_tutorial_status_is
...
+
+            // All in section 216
+            // Tutorial: 927  OrderByNumber 34
+            // Tutorial: 928  OrderByNumber 35
+            // Tutorial: 929  OrderByNumber 35
+            // Tutorial: 930  OrderByNumber 36
+            var result = sectionContentService.GetSectionContent(customisationId, candidateId, sectionId);
+
+            // Then
+            var expectedSectionContent = new SectionContent(
```
FridaTveit, line 336: Could maybe use some helper functions to create the section and tutorials here? As we only really care about the order, so only need to specify enough to make sure that's correct 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data/Services/SectionContentService.cs
FridaTveit: @benjimarshall? 🙂

FridaTveit (approved):
Looks good 🙂

DigitalLearningSolutions.Data/Services/SectionContentService.cs
benjimarshall: Yep I'll add that to HEEDLS-230. Sorry for missing these comments 🙂

FridaTveit: Great, thanks 🙂

PR #103, opened 2020-12-18
HEEDLS-249 Remove times from section cards in initial menu

benjimarshall:
## Changes
- Remove times from section cards in initial menu
- Tidy up the view model and tests.

## Testing
Tested in Chrome and IE11 using a high zoom, screen reader, and narrow viewport.

### Wide viewport
times_removed
### Narrow viewport
times_removed_narrow_viewport


FridaTveit (approved):
Looks good 🙂

PR #105, opened 2020-12-18
HEEDLS-246 - Section Page Improve Time Information

SChandarana:
What do we think about the "(s)", would we prefer to just add some logic to the view model to display the s if val != 1?

image

### DIFF
--- a/DigitalLearningSolutions.Web/Views/LearningMenu/Section/_TutorialCard.cshtml
+++ b/DigitalLearningSolutions.Web/Views/LearningMenu/Section/_TutorialCard.cshtml
@@ -8,7 +8,9 @@
           @Model.TutorialName
         </h3>
         <p class="nhsuk-u-secondary-text-color">
-          @Model.TimeInformation
+          @Model.TutorialTime minutes spent
+          <span class="visually-hidden">on this tutorial</span>
+          (average tutorial time @Model.AverageTutorialTime minutes)
         </p>
       </div>
       <div class="nhsuk-grid-column-one-quarter">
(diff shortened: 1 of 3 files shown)


SChandarana:
What do we think about the "(s)", would we prefer to just add some logic to the view model to display the s if val != 1?

DigitalLearningSolutions.Web/Views/LearningMenu/Section/_TutorialCard.cshtml
```diff
@@ -8,7 +8,9 @@
           @Model.TutorialName
         </h3>
         <p class="nhsuk-u-secondary-text-color">
-          @Model.TimeInformation
+          @Model.TutorialTime minutes spent
+          <span class="visually-hidden">on this tutorial</span>
+          (average tutorial time @Model.AverageTutorialTime minutes)
```
FridaTveit: Probably worth adding minute/minutes logic here. For that it would also probably be useful to move the logic of creating these strings back to the view model. For example having the tutorial time and average tutorial time being string fields being calculated in the view model and just slotting "on this tutorial" in the middle here would work well I think 🙂

FridaTveit (commented):

FridaTveit:
@SChandarana I think I accidentally reviewed an older version so sorry if my comment was a bit out of date! 🙁 Yes I think adding some logic to display either minute or minutes would be good. The same applies for the attempt/attempts testing issue on your other ticket actually 🙂

SChandarana:
Logic added so if Time Spent/ average time is 1 it will say "minute" rather than "minutes"
image

FridaTveit (approved):
Looks good 🙂

FridaTveit (approved):
Looks good 🙂

PR #106, opened 2020-12-18
HEEDLS-256 Add name to aria text of launch section link

benjimarshall:
## Changes
Add section id to heading text of section card, and use that for an `aria-describedby` on the Launch section button.

## Tested
Tested with a screen reader in Chrome, Firefox, Edge and IE11. No visual changes occured.


FridaTveit (approved):
Looks good 🙂

PR #107, opened 2020-12-21
HEEDLS-254 Content Closing page

benjimarshall:
I don't think this is ready to go yet, and I've got a questions section below. I thought the easiest way for me to explain what I had done so far was a PR, and I can add any changes needed to it

## Changes
- Add `closeMpe` function for Content Viewer in a new TypeScript file
- Add this script to the `ContentViewer` view

## Questions
- I've added no tests. How should I test this, in particular is there a good way of mocking up a `window` object? Also, where should I put these tests?
- I've exposed this as a global function using `window.closeMpe = closeMpe` to make sure it's still accessible after webpack runs. Is there a better way of doing this?
- The `closeMpe` function needs the `customisationId` and `sectionId` for it to redirect. Currently I'm extracting this from the URL using `window.location.href`, is this the best way of doing this or is there a better way of getting those IDs into the function?
- How can I test this in action? Do you happen to have a tutorial that you know of that calls this function?

### DIFF
--- a/DigitalLearningSolutions.Web/Scripts/learningMenu/contentViewer.ts
+++ b/DigitalLearningSolutions.Web/Scripts/learningMenu/contentViewer.ts
@@ -2,12 +2,11 @@ function closeMpe(): void {
   // Extract the customisationId and sectionId out of the URL
   const matches = window.location.href.match(/.*\/LearningMenu\/(\d+)\/(\d+)\/(\d+)\/Tutorial$/);
 
-  if (matches == null || matches.length < 3) {
+  if (!matches || matches.length < 4) {
     return;
   }
 
-  window.location.href = `/LearningMenu/${matches[1]}/${matches[2]}`;
+  window.location.href = `/LearningMenu/${matches[1]}/${matches[2]}/${matches[3]}`;
 }
 
 window.closeMpe = closeMpe;
-// export default closeMpe;
--- a/DigitalLearningSolutions.Web/Scripts/learningMenu/contentViewer.ts
+++ b/DigitalLearningSolutions.Web/Scripts/learningMenu/contentViewer.ts
@@ -1,5 +1,5 @@
 function closeMpe(): void {
-  // Extract the customisationId and sectionId out of the URL
+  // Extract the customisationId, sectionId and tutorialId out of the URL
   const matches = window.location.href.match(/.*\/LearningMenu\/(\d+)\/(\d+)\/(\d+)\/Tutorial$/);
 
   if (!matches || matches.length < 4) {
--- a/DigitalLearningSolutions.Web/Scripts/learningMenu/contentViewer.ts
+++ b/DigitalLearningSolutions.Web/Scripts/learningMenu/contentViewer.ts
@@ -1,0 +1,13 @@
+function closeMpe(): void {
+  // Extract the customisationId and sectionId out of the URL
+  const matches = window.location.href.match(/.*\/LearningMenu\/(\d+)\/(\d+)\/(\d+)\/Tutorial$/);
+
+  if (matches == null || matches.length < 3) {
+    return;
+  }
+
+  window.location.href = `/LearningMenu/${matches[1]}/${matches[2]}`;
+}
+
+window.closeMpe = closeMpe;
+// export default closeMpe;
--- a/DigitalLearningSolutions.Web/Scripts/spec/contentViewer.spec.ts
+++ b/DigitalLearningSolutions.Web/Scripts/spec/contentViewer.spec.ts
@@ -1,0 +1,42 @@
+/**
+ * @jest-environment jsdom
+ */
+
+import '../learningMenu/contentViewer';
+
+beforeAll(() => {
+  global.window = Object.create(window);
+  const url = 'https://example.com';
+  Object.defineProperty(global.window, 'location', {
+    value: {
+      href: url,
+    },
+  });
+});
+
+describe('closeMpe', () => {
+  it('should redirect to tutorial overview',
+    () => {
+      // Given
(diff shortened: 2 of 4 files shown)


FridaTveit:
> I've added no tests. How should I test this, in particular is there a good way of mocking up a window object? Also, where should I put these tests?

They do in Scripts/spec. Have a look at the tests already there to see if they help, if not let me know and I can try to help 🙂 

> I've exposed this as a global function using window.closeMpe = closeMpe to make sure it's still accessible after webpack runs. Is there a better way of doing this?

No, I think that's the best way. It's a bit of a strange setup, but they're planning to remove this content soon anyway.

> The closeMpe function needs the customisationId and sectionId for it to redirect. Currently I'm extracting this from the URL using window.location.href, is this the best way of doing this or is there a better way of getting those IDs into the function?
How can I test this in action? Do you happen to have a tutorial that you know of that calls this function?

Most content should. If you get to the end of html content and press the next button it should call it 🙂

benjimarshall:
I'm struggling to get any tests to work, because of the `window.closeMpe = closeMpe` line in the `contentViewer.ts` file. Maybe I could split these over two files, unless you know of a better way of fixing this? 🙂

FridaTveit (commented):
A few comments, and there's the test issue to resolve. Otherwise looks good 🙂

DigitalLearningSolutions.Web/Scripts/learningMenu/contentViewer.ts
```diff
@@ -0,0 +1,13 @@
+function closeMpe(): void {
+  // Extract the customisationId and sectionId out of the URL
+  const matches = window.location.href.match(/.*\/LearningMenu\/(\d+)\/(\d+)\/(\d+)\/Tutorial$/);
+
+  if (matches == null || matches.length < 3) {
```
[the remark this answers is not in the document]
benjimarshall: Yep, thank you! Done in 3d35a73 🙂

DigitalLearningSolutions.Web/Scripts/learningMenu/contentViewer.ts
```diff
@@ -0,0 +1,13 @@
...
+
+  if (matches == null || matches.length < 3) {
+    return;
+  }
+
+  window.location.href = `/LearningMenu/${matches[1]}/${matches[2]}`;
+}
+
+window.closeMpe = closeMpe;
+// export default closeMpe;
```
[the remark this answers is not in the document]
benjimarshall: Yeah sorry this was a bit rough in the first commit. Should be tidy in 3d35a73 🙂

benjimarshall:
Testing should be set up and working in 3d35a73 🙂

DigitalLearningSolutions.Web/Scripts/learningMenu/contentViewer.ts
```diff
@@ -2,12 +2,11 @@ function closeMpe(): void {
   // Extract the customisationId and sectionId out of the URL
```
FridaTveit, line 2: And tutorial id 🙂

DigitalLearningSolutions.Web/Scripts/spec/contentViewer.spec.ts
```diff
@@ -0,0 +1,42 @@
+/**
+ * @jest-environment jsdom
```
FridaTveit, line 2: What's this for? 🙂

FridaTveit (commented):
Nice, good work on the tests! Did you manage to test it manually as well? 🙂

DigitalLearningSolutions.Web/Scripts/spec/contentViewer.spec.ts
benjimarshall: In our current `testEnvironment` in the Jest config is `node`, and this leads to the `window` global being undefined and the `window.closeMpe = closeMpe` failing. One of the ways of fixing this problem is to change environment, according to [this](https://stackoverflow.com/questions/46274889/jest-test-fails-with-window-is-not-defined), but there may be others. Globally changing the environment breaks the other tests so this does it just for this file 🙂

FridaTveit: I see, sounds good 🙂

DigitalLearningSolutions.Web/Scripts/learningMenu/contentViewer.ts
benjimarshall: Ah sorry about that! Fixed in 88a0354 🙂

benjimarshall:
I've not managed to test this properly. I've not found many local courses, and it doesn't seem to be working when I test it. When I type `closeMpe()` into the developer console the function seems to work though, so I think it might work I've just not managed to do it myself. Do you know of any specific examples that should work? 🙂

FridaTveit:
@benjimarshall e.g. https://www.dls.nhs.uk/dev-my-learning-portal/LearningMenu/23829/1949/9311/Tutorial should close the window if you press the next button at the end. Make sure you're either running the old code or setting the current system base url to the UAT old system. If you're using UAT (which is probably more likely to work) then also make sure you're using content which is linked to the test user in UAT. If this seems like it's going to be a big pain though then we can merge it as is and I can test it on UAT 🙂

FridaTveit (approved):
Looks good 🙂

benjimarshall:
I've had a go but it hasn't worked. I think it's getting blocked due to cross-site scripting protections in my browser (I've got a load of errors relating to that in my developer console). I might have done something wrong, though I think it might be easier to just test this on UAT 🙂

FridaTveit:
@benjimarshall sounds sensible. Happy for this to be merged then 🙂

PR #108, opened 2020-12-21
Fix tutorial next link back end bug

benjimarshall:
*Bug description*:
If a course as a post learning assessment, but it has been disabled by
`IsAssessed`, then the next link should not link to that post learning
assessment.

## Changes
Amend query to discard post learning assessment path if `IsAssessed` is `false`

## Testing
Add a regression test for the example @FridaTveit found whilst testing.


FridaTveit (approved):
Looks good 🙂

PR #109, opened 2020-12-21
HEEDLS-247 Improve average duration display in initial menu

benjimarshall:
## Changes
Modify the query to get the raw number of minutes spent in a course,
rather than using the scalar-valued function `GetMinsForCustomisation`
which formats the times in a way that we would like to change.

Add time formatting to the Initial menu view model, and add tests
for this formatting.

## Testing
Update the `CourseContentService` tests, and add tests for the
initial menu view model. 

Tested UI in Chrome, Firefox, and IE11 using a screenreader and high zoom.

## Screenshots
coursetimes


DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -57,24 +72,26 @@ public CourseContentService(IDbConnection connection, ILogger<CourseContentServi
                             THEN 0
                             ELSE CAST(SUM(aspProgress.TutStat) * 100 AS FLOAT) / (COUNT(Tutorials.TutorialID) * 2)
                          END) AS PercentComplete
-                  FROM Applications
```
FridaTveit, line 60: From here downwards it's just spacing being changed right? No functional change? 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -35,10 +35,25 @@ public CourseContentService(IDbConnection connection, ILogger<CourseContentServi
 
             CourseContent? courseContent = null;
             return connection.Query<CourseContent, CourseSection, CourseContent>(
-                @"SELECT Customisations.CustomisationID AS id,
+                @"  WITH CustomisationDurations AS (
```
FridaTveit, line 38: Ah I didn't realise this would need changes as big as this! Just for the future it might have been a good idea to write a HLD for this or ping me a message to let me know this would require backend changes. Because in that case I might have suggested we push back on it or do a slightly different solution (e.g. replace `h` and `m` with hours and minutes with text replacement, possibly just in the voice over). We might very well have ended up with this solution anyway but would have been useful to consider 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
```diff
@@ -35,10 +35,25 @@ public CourseContentService(IDbConnection connection, ILogger<CourseContentServi
...
+                         SUM(AverageDuration) AS AverageDuration
+                    FROM (SELECT Customisations.CustomisationID,
+                                 CASE
+                                 WHEN Tutorials.OverrideTutorialMins > 0 THEN Tutorials.OverrideTutorialMins
+                                 ELSE Tutorials.AverageTutMins END AS AverageDuration
+                            FROM CustomisationTutorials
+                            INNER JOIN Customisations ON CustomisationTutorials.CustomisationID = Customisations.CustomisationID
+                            INNER JOIN Tutorials ON CustomisationTutorials.TutorialID = Tutorials.TutorialID
+                            WHERE CustomisationTutorials.CustomisationID = @customisationId
+                                  AND (CustomisationTutorials.Status = 1 OR CustomisationTutorials.DiagStatus = 1)
```
FridaTveit, line 49: Are you sure we should include tutorials with CT.Status != 1? Doesn't look like they're included in `GetMinsForCustomisation`? 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/InitialMenuViewModel.cs
```diff
@@ -23,7 +23,9 @@ public InitialMenuViewModel(CourseContent courseContent)
         {
             Id = courseContent.Id;
             Title = courseContent.Title;
-            AverageDuration = courseContent.AverageDuration;
+            //AverageDuration = courseContent.AverageDuration;
```
FridaTveit, line 26: Please remove commented out code 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/InitialMenuViewModel.cs
```diff
@@ -39,5 +41,21 @@ public InitialMenuViewModel(CourseContent courseContent)
...
+        private static string FormatDuration(int duration)
+        {
+            if (duration < 60)
+            {
+                return FormatTimeUnits(duration, "minute");
+            }
+
+            var formattedHours = FormatTimeUnits(duration / 60, "hour");
+            var formattedMinutes = FormatTimeUnits(duration % 60, "minute");
+            return duration % 60 == 0 ? formattedHours : formattedHours + formattedMinutes;
```
FridaTveit, line 54: Can extract `duration % 60` to a variable? 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/InitialMenuViewModel.cs
```diff
@@ -39,5 +41,21 @@ public InitialMenuViewModel(CourseContent courseContent)
...
+
+        }
+
+        private static string FormatTimeUnits(int duration, string unit) =>
+            duration == 1 ? $" {duration} {unit}" : $" {duration} {unit}s";
```
FridaTveit, line 59: Any reason why you've included the whitespace here? Not very important but I found it a bit surprising, so would have probably left it to be defined in the cshtml as before. It makes the average duration make more sense on its own 🙂

DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
```diff
@@ -420,7 +420,7 @@ public void Get_course_content_for_refreshed_course_should_return_new_course()
                     4169,
                     "Level 2 - Microsoft Excel 2010",
                     "MOS Excel 2010 CORE",
-                    "5h 49m",
+                    349,
```
FridaTveit, line 423: Could we add a few tests comparing the new value to the output of the old scalar function? E.g. a `TestCase` test taking customisation id, new value (int), old value (string) as parameters, if that makes sense? That would make me a bit more confident that we haven't broken anything 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data/Services/CourseContentService.cs
benjimarshall: Yeah that's right, I didn't spot that I had changed that but I think it's probably better 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
benjimarshall: Yeah neither did I! I spotted that the scalar function would need replacing early on but I had hoped I would get away with a simple `SUM` and then I kept realising how much more stuff I needed to add. Sorry I should have checked in earlier, but I didn't spend very much time on it if you did want to solve this problem a totally different way 🙂

FridaTveit: Okay, good to know 🙂 It would be a bit inconvenient to push back on right now as people are on holiday. So I think I'm happy to go with this solution, as long as we add some tests comparing it to the old function to make sure we haven't broken anything 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/InitialMenuViewModel.cs
benjimarshall: This was to make adding them together nice, but I agree it turned out to be a bit weird and I've changed it in 1341b0f 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/InitialMenuViewModel.cs
benjimarshall: Yep! Done in 1341b0f 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/InitialMenuViewModel.cs
benjimarshall: Whoops! Removed in 1341b0f 🙂

DigitalLearningSolutions.Data/Services/CourseContentService.cs
benjimarshall: Good point, change in 1341b0f 🙂

DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
benjimarshall: I've added some tests in 1c07aee 🙂

Whilst doing this I noticed that the duration can be null (for a course without tutorials). The function in the database would render this as "N/A", so course duration would read "Average Course Duration: N/A". This is what I've currently put into the new version, is this the behaviour you would like? Alternatively we could remove that line if there is no duration for the course 🙂

FridaTveit: Good question. I think leave it for now and I'll check with Kevin when he's back 🙂

FridaTveit (approved):
Looks good 🙂

PR #110, opened 2020-12-21
HEEDLS-178 Section Page Tutorials and Assessments List Bug Fixes

SChandarana:
Fixing the bugs on ticket HEEDLS-178
Todo:
- [x] Don't show tutorials with CustomisationTutorial.Status = 1
- [x] Change "attempts" to remove the s if attempts = 1 to make grammatical sense 
- [x] Change Launch buttons to say the tutorial/assessment name when using screen reader

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
@@ -1,6 +1,7 @@
 ﻿namespace DigitalLearningSolutions.Data.Tests.Services
 {
     using System.Linq;
+    using System.Transactions;
     using DigitalLearningSolutions.Data.Models.SectionContent;
     using DigitalLearningSolutions.Data.Services;
     using DigitalLearningSolutions.Data.Tests.Helpers;
@@ -12,6 +13,8 @@
     internal class SectionContentServiceTests
     {
         private SectionContentService sectionContentService;
+        private SectionContentTestHelper sectionContentTestHelper;
+        private CourseContentTestHelper courseContentTestHelper;
 
         [SetUp]
         public void Setup()
@@ -19,6 +22,8 @@
             var connection = ServiceTestHelper.GetDatabaseConnection();
             var logger = A.Fake<ILogger<SectionContentService>>();
             sectionContentService = new SectionContentService(connection, logger);
+            sectionContentTestHelper = new SectionContentTestHelper(connection);
+            courseContentTestHelper = new CourseContentTestHelper(connection);
         }
 
         [Test]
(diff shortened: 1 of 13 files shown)


DigitalLearningSolutions.Web/Views/LearningMenu/Section/_TutorialCard.cshtml
```diff
@@ -22,6 +22,7 @@
     <div class="nhsuk-grid-row">
       <div class="nhsuk-grid-column-full">
         <a class="nhsuk-button nhsuk-u-margin-bottom-0 nhsuk-u-margin-top-0"
+           aria-label="Launch tutorial @Model.TutorialName"
```
FridaTveit, line 25: This is not the same solution we're using elsewhere, e.g. in the course cards and the initial menu page. We should use `aria-describedby` to be consistent 🙂

DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
```diff
@@ -49,10 +49,10 @@ public void Get_section_content_should_return_section_content()
...
                 new[]
                 {
-                    new SectionTutorial("Introduction to applications", 0, "Not started", 0, 17, 1461),
-                    new SectionTutorial("Common screen elements", 0, "Not started", 0, 11, 1462),
-                    new SectionTutorial("Using ribbon tabs", 0, "Not started", 0, 6, 1463),
-                    new SectionTutorial("Getting help", 0, "Not started", 0, 11, 1464)
+                    new SectionTutorial("Introduction to applications", 0, "Not started", 0, 17, 1461, true),
+                    new SectionTutorial("Common screen elements", 0, "Not started", 0, 11, 1462, true),
+                    new SectionTutorial("Using ribbon tabs", 0, "Not started", 0, 6, 1463, true),
+                    new SectionTutorial("Getting help", 0, "Not started", 0, 11, 1464, true)
```
FridaTveit, line 55: As discussed could you add tests making sure we're returning the same tutorials as the old stored procedure, like the test we have for the initial menu page? We should test a few customisations for different users. In particular we should definitely test the courses which had this issue (customisation ids are on the ticket) 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
```diff
@@ -327,5 +317,37 @@ public void Get_section_content_should_order_by_tutorial_id_when_shared_order_by
...
+                    .Select(tutorial => tutorial.TutorialID);
+
+                // When
+                var tutorialIdsInSectionContent = sectionContentService
+                    .GetSectionContent(customisationId, candidateId, sectionId)?
+                    .Tutorials
+                    .Select(tutorial => tutorial.Id);
+
+                // Then
+                tutorialIdsInSectionContent?.Should().BeEquivalentTo(tutorialIdsReturnedFromOldStoredProcedure);
```
FridaTveit, line 349: Does this ignore order? If it does, can we change it so it will fail if the order is different? I think we did that for a different test? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
SChandarana: Ah yep you're right

FridaTveit (approved):
Looks good 🙂

PR #111, opened 2020-12-22
HEEDLS-254 Bugfix: use absolute, not relative URLs in content close script

benjimarshall:
### Changes
Change the URLs to be absolute and not relative, to fix the issue with redirects found in testing.

### Testing
Amend the TypeScript unit tests to reflect this change. Checked running the function works from Chrome developer console, integration testing has not been done yet because it is easiest to do this on UAT when there are no cross-site scripting concerns.


FridaTveit (approved):
Looks good 🙂

PR #112, opened 2020-12-22
HEEDLS-NONE Assessment page diagnostic tests

kwunkiuma:
Add some tests to compare results from the service class to results from stored procedures.

### DIFF
--- a/DigitalLearningSolutions.Data/Models/DiagnosticAssessment/OldTutorial.cs
+++ b/DigitalLearningSolutions.Data/Models/DiagnosticAssessment/OldTutorial.cs
@@ -1,0 +1,7 @@
+﻿namespace DigitalLearningSolutions.Data.Models.TutorialContent
+{
+    public class OldTutorial
+    {
+        public int TutorialId { get; set; }
+    }
+}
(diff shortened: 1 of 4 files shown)


DigitalLearningSolutions.Data/Models/DiagnosticAssessment/OldTutorial.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace DigitalLearningSolutions.Data.Models.DiagnosticAssessment
+{
+    public class OldTutorial
```
FridaTveit, line 3: Don't we already have a class like this? Or am I thinking of one @SChandarana has added in an open PR?

FridaTveit (commented):

DigitalLearningSolutions.Data/Models/DiagnosticAssessment/OldTutorial.cs
SChandarana: Yep I have "OldSectionTutorial.cs" in PR #110 which is the same as this

FridaTveit: In that case whichever one of you merges last should rebase and use the existing class. I'll leave it up to you to coordinate that 🙂 @SChandarana I think OldTutorial might be a better name 🙂

kwunkiuma: We've sorted out the order for this, are there any other comments for this PR?

FridaTveit (approved):
Looks good 🙂

PR #114, opened 2020-12-22
HEEDLS-NONE Rename a test procedure

benjimarshall:
Rename a test procedure added for HEEDLS-247. This procedure had a temporary name and ended up getting push up, sorry about that. I've renamed it to something more helpful here.


FridaTveit (approved):
Looks good 🙂

PR #116, opened 2020-12-23
HEEDLS-232 - Redirect To 404 if Customisation is not active

SChandarana:
As both HEEDLS-178 and this are accessing the same things I have based this one off the changes from 178


FridaTveit (approved):
Looks good 🙂

PR #117, opened 2020-12-23
HEEDLS-262 Assessment page post learning back end

kwunkiuma:
Add query for retrieving post learning information.

Add unit tests and compare results to stored procedure.

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Helpers/PostLearningAssessmentTestHelper.cs
+++ b/DigitalLearningSolutions.Data.Tests/Helpers/PostLearningAssessmentTestHelper.cs
@@ -16,10 +16,13 @@
             this.connection = connection;
         }
 
-        public IEnumerable<OldPostLearningAssessmentScores> ScoresFromOldStoredProcedure(int progressId, int sectionId)
+        public OldPostLearningAssessmentScores? ScoresFromOldStoredProcedure(int progressId, int sectionId)
         {
-            return connection.Query<OldPostLearningAssessmentScores>("uspReturnSectionsForCandCust_V2", new { progressId }, commandType: CommandType.StoredProcedure)
-                .Where(assessment => assessment.SectionID == sectionId);
+            return connection.Query<OldPostLearningAssessmentScores>(
+                "uspReturnSectionsForCandCust_V2",
+                new { progressId },
+                commandType: CommandType.StoredProcedure
+            ).FirstOrDefault(assessment => assessment.SectionID == sectionId);
         }
     }
 }
--- a/DigitalLearningSolutions.Data.Tests/Helpers/PostLearningAssessmentTestHelper.cs
+++ b/DigitalLearningSolutions.Data.Tests/Helpers/PostLearningAssessmentTestHelper.cs
@@ -1,6 +1,5 @@
 ﻿namespace DigitalLearningSolutions.Data.Tests.Helpers
 {
-    using System.Collections.Generic;
     using System.Data;
     using System.Linq;
     using Dapper;
--- a/DigitalLearningSolutions.Data.Tests/Helpers/PostLearningAssessmentTestHelper.cs
+++ b/DigitalLearningSolutions.Data.Tests/Helpers/PostLearningAssessmentTestHelper.cs
@@ -1,0 +1,25 @@
+﻿namespace DigitalLearningSolutions.Data.Tests.Helpers
+{
+    using System.Collections.Generic;
+    using System.Data;
+    using System.Linq;
+    using Dapper;
+    using DigitalLearningSolutions.Data.Models.PostLearningAssessment;
+    using Microsoft.Data.SqlClient;
+
+    public class PostLearningAssessmentTestHelper
+    {
+        private SqlConnection connection;
+
+        public PostLearningAssessmentTestHelper(SqlConnection connection)
+        {
+            this.connection = connection;
+        }
+
+        public IEnumerable<OldPostLearningAssessmentScores> ScoresFromOldStoredProcedure(int progressId, int sectionId)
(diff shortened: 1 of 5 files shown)


DigitalLearningSolutions.Data.Tests/Services/PostLearningAssessmentTests.cs
```diff
@@ -0,0 +1,178 @@
...
+
+            // When
+            var result = postLearningAssessmentService.GetPostLearningAssessment(customisationId, candidateId, sectionId);
+
+            // Then
+            result.Should().BeNull();
+        }
+
+        [Test]
+        public void Get_post_learning_assessment_should_return_null_if_archived_date_is_null()
```
FridaTveit, line 106: Should this be if archived date is not null? 🙂

DigitalLearningSolutions.Data.Tests/Services/PostLearningAssessmentTests.cs
```diff
@@ -0,0 +1,178 @@
...
+        [TestCase(160541, 11698, 198, 133094)]
+        [TestCase(208119, 4918, 136, 167637)]
+         public void Get_post_learning_assessment_should_have_same_scores_as_stored_procedure(
+            int candidateId,
+            int customisationId,
+            int sectionId,
+            int progressId
+        )
+        {
+            using (new TransactionScope())
```
FridaTveit, line 161: Does this need to be in transaction scope? For other tests like this I've added it since I made it enrol the user if they weren't enrolled. But it doesn't look like we're doing that here? 🙂

DigitalLearningSolutions.Data/Models/PostLearningAssessment/PostLearningAssessment.cs
```diff
@@ -0,0 +1,30 @@
+﻿namespace DigitalLearningSolutions.Data.Models.PostLearningAssessment
+{
+    public class PostLearningAssessment
+    {
+        public string CourseTitle { get; }
+        public string SectionName { get; }
+        public string PostLearningAssessmentPath { get; }
+        public int PostLearningScore { get; }
+        public int PostLearningAttempts { get; }
+        public bool PostLearningPassed { get; }
```
FridaTveit, line 10: Also need to know whether the post learning assessment is locked 🙂

DigitalLearningSolutions.Data/Services/PostLearningAssessmentService.cs
```diff
@@ -0,0 +1,71 @@
...
+                                    AssessAttempts.SectionNumber
+                            ) AS Attempts ON (Progress.ProgressID = Attempts.ProgressID) AND (Attempts.SectionNumber = Sections.SectionNumber)
+                    WHERE
+                        Customisations.CustomisationID = @customisationId
+                        AND Customisations.IsAssessed = 1
+                        AND Sections.SectionID = @sectionId
+                        AND Sections.ArchivedDate IS NULL
+                        AND Sections.PLAssessPath IS NOT NULL
+                    ORDER BY
+                        Sections.SectionNumber",
```
FridaTveit, line 66: Is this order by necessary? 🙂

DigitalLearningSolutions.Data/Services/PostLearningAssessmentService.cs
```diff
@@ -0,0 +1,71 @@
...
+                new { customisationId, candidateId, sectionId }
+            ).FirstOrDefault();
```
FridaTveit, line 68: Can use QueryFirstOrDefault instead? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data.Tests/Helpers/PostLearningAssessmentTestHelper.cs
```diff
@@ -0,0 +1,25 @@
...
+    public class PostLearningAssessmentTestHelper
+    {
+        private SqlConnection connection;
+
+        public PostLearningAssessmentTestHelper(SqlConnection connection)
+        {
+            this.connection = connection;
+        }
+
+        public IEnumerable<OldPostLearningAssessmentScores> ScoresFromOldStoredProcedure(int progressId, int sectionId)
```
FridaTveit, line 19: We're only expecting this to return one lot of scores right? So it would probably make more sense to call `FirstOrDefault` here, instead of `Where` 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data.Tests/Helpers/PostLearningAssessmentTestHelper.cs
kwunkiuma: The stored procedure takes a `ProgressID` and returns a row for each section of the customisation while the method in our service also takes the specific `SectionID` and returns a single row. The `Where` is being used to retrieve an individual section from the stored procedure in order to compare the values to the ones from the service.

kwunkiuma: Using `FirstOrDefault` would return the first row from the stored procedure, but it may not be the section that we want to compare.

FridaTveit: You can provide a comparison function to FirstOrDefault I believe, like you do with Where. So you should be able to use that instead of having to use FirstOrDefault later 🙂 https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.firstordefault?view=net-5.0

FridaTveit (approved):
Looks good 🙂

PR #118, opened 2021-01-04
HEEDLS-257 Completion Summary front end

benjimarshall:
## Changes
* Move button row CSS to more general location
* Add a completion page view and view model, and controller action
* Add link to completion summary page (involves importing the relevant
NHS JS to make clickable cards work properly)

## Testing
Added unit tests for new classes and methods. Tested in Chrome, Firefox and IE11, with narrow view ports and a screenreader.

Tested without JS as well, the NHS clickable cards (for the completion summary card in the initial menu) need JS to work properly, but still fail fairly gracefully without JS: clicking on any portion of the card has link-esque CSS, but the link itself only registers a `click()` when the actual link text is clicked. It behaves like a normal card with a link, but has the CSS styling of a clickable card. Other than this, everything else works without JS.

### Summary of a completed course using a wide viewport
complete_course_wide_viewport

### Summary of a completed course using a narrow viewport
complete_course_narrow_viewport

### Summary of a completed course that is not assessed
complete_course_evaluate

### Summary of an incomplete course
incomplete_course_summary


DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
```diff
@@ -1122,6 +1125,218 @@ public void TutorialVideo_should_return_404_if_unable_to_enrol()
...
+        public void CompletionSummary_should_return_404_if_course_does_not_include_certification()
+        {
+            // Given
+            var expectedCourseCompletion = CourseCompletionHelper.CreateDefaultCourseCompletion(
+                CustomisationId,
+                includeCertification: false
+            );
+            const int progressId = 101;
+
+            A.CallTo(() => courseCompletionService.GetCourseCompletion(CandidateId, CustomisationId))
```
FridaTveit, line 1324: It might be more consistent if `GetCourseCompletion` returns null in this case? 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/CourseCompletionViewModelTests.cs
```diff
@@ -0,0 +1,384 @@
...
+
+            // Then
+            courseCompletionViewModel.DiagnosticAttempts.Should().Be(diagnosticAttempts);
+        }
+
+        [Test]
+        public void CourseCompletion_should_have_percentageTutorialsCompleted()
+        {
+            // Given
+            const int percentageTutorialsCompleted = 55;
```
FridaTveit, line 96: Is any rounding necessary here? 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/CourseCompletionViewModelTests.cs
```diff
@@ -0,0 +1,384 @@
...
+
+            // When
+            var courseCompletionViewModel = new CourseCompletionViewModel(expectedCourseCompletion);
+
+            // Then
+            courseCompletionViewModel.FinaliseText.Should().Be("Evaluate");
+        }
+
+        [Test]
+        public void CourseCompletion_FinaliseAriaLabel_should_be_evaluate_when_course_is_not_assessed_and_not_evaluated()
```
FridaTveit, line 244: What about if the course is assessed and not evaluated? Do we have a test for that case? 🙂

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
```diff
@@ -264,5 +267,48 @@ public IActionResult TutorialVideo(int customisationId, int sectionId, int tutor
...
+            var candidateId = User.GetCandidateId();
+            var centreId = User.GetCentreId();
+
+            var courseCompletion = courseCompletionService.GetCourseCompletion(candidateId, customisationId);
+
+            if (courseCompletion == null || centreId == null)
+            {
+                logger.LogError(
+                    "Redirecting to 404 as customisation id was not found. " +
+                    $"Candidate id: {candidateId}, customisation id: {customisationId}, " +
```
FridaTveit, line 283: We should probably log the centre id here as well (and in the other similar logs), so that it's clear from the logs what's happened if the centre id is null 🙂

DigitalLearningSolutions.Web/Scripts/nhsuk.ts
```diff
@@ -18,4 +19,5 @@ document.addEventListener('DOMContentLoaded',
     SkipLink();
     Radios();
     Checkboxes();
+    Card();
```
FridaTveit, line 22: How come we didn't need this before now? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web/Scripts/nhsuk.ts
benjimarshall: The only clickable card that we have is for the course summary in the initial menu page. We need JS for the card to behave like a link properly, but I didn't notice this because the card was only given a real link (instead of a placeholder of `href="#"`) as part of cd088ea 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/CourseCompletionViewModelTests.cs
benjimarshall: This is getting truncated in the SQL query at the moment, and I should be rounding it, thanks for catching this! Should I round it in the SQL query, or the view model? 🙂

FridaTveit: View model would be more consistent with what we're doing on other pages as far as I remember 🙂

DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
benjimarshall: I agree but this would break the tests I currently have, because there are no courses in the test DB that have `IncludeCertification` as true. Either we need to change the data in the test DB, or we could leave these checks where they are in the learning menu controller. What do you think? 🙂

FridaTveit: We have a few scenarios like this already in the tests, where we want to test something we don't have the correct data for. In that case we've added a helper method to change the test db and just put that inside transaction scope in the test. We could also change the test db permanently but that would be a bit more hassle since we'd all need to do it on our copy of it and it means restoring the test db if it got into a bad state (which has happened a few times already) would be more hassle 🙂

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
benjimarshall: Yes I agree! Done in adf3f44 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/CourseCompletionViewModelTests.cs
benjimarshall: This corresponds to the "Complete evaluation and access certificate" button that I had in the design at one point, which now just says "Certificate", based on the implementation of this button in the Completed courses learning portal page. I've added two tests to cover this in adf3f44 🙂

DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
benjimarshall: I've added a method to add certification to the test records (in a transaction scope), and moved this check `IncludeCertification` into the data service, in 1c742e1 🙂

DigitalLearningSolutions.Data.Tests/Services/CourseCompletionServiceTests.cs
```diff
@@ -10,142 +11,175 @@
...
+                    0,
+                    0,
+                    6,
+                    10
+                ));
+            }
+        }
+
+        [Test]
+        public void Get_course_completion_should_return_null_when_course_does_not_exist()
```
FridaTveit, line 167: Should this name be swapped with the test below? 🙂

DigitalLearningSolutions.Data/Services/CourseCompletionService.cs
```diff
@@ -131,6 +131,8 @@ GROUP BY
                          PostLearningPasses.PostLearningPasses,
                          SectionCount.SectionCount;",
                 new { candidateId, customisationId });
+
+            return result?.IncludeCertification == true ? result : null;
```
FridaTveit, line 135: Could you add this to the `WHERE` instead? 🙂 The IncludeCertification shouldn't need to be part of the completion object we return, that would just be confusing because it would seem like it would be possible for it to be false in that case 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/CourseCompletionViewModelTests.cs
benjimarshall: Done in 6999d38 🙂

DigitalLearningSolutions.Data.Tests/Services/CourseCompletionServiceTests.cs
benjimarshall: Yes, sorry! Thanks for catching that, fixed in f072c1d 🙂

DigitalLearningSolutions.Data/Services/CourseCompletionService.cs
benjimarshall: Yeah that makes sense! Done in f072c1d 🙂

FridaTveit (approved):
Looks good 🙂

FridaTveit:
@benjimarshall one thing I just thought of, does this behave sensibly for courses/sections which have just assessments and no tutorials, e.g. customisation 5694? In particular does the percentage completion calculation still give a sensible output? 🙂

benjimarshall:
@FridaTveit I've just run it for that course (and candidate ID 1), the percentage complete for the tutorials is set to 0, which I think is reasonable? Calculating the completion of an empty set is a bit weird and arguably involves dividing by 0, so is this result ok? 🙂

benjimarshall:
This has the similar behaviour as the `Get_course_completion_of_course_should_return_when_all_tutorials_statuses_are_0` test which does have tutorials, but all their statuses are 0 so it gives a very similar effect (with percentage tutorial completion also being 0)  🙂

FridaTveit:
@benjimarshall I think it would be worth running this by Kevin, now that we know this does happen for some courses. Could you drop him a message and ask if in this case he'd prefer for us to show the percentage as 0, show the percentage as 100 or hide the percentage altogether. Unless you think the second and/or third options would be tricky to do? 🙂

benjimarshall:
I think it should just be a case of changing the default value in the coalesce. The old query returns `-1` in these cases, I'm not sure how it gets rendered on the old site though. I'll message him now 🙂

benjimarshall:
Actually as per our chat the old site doesn't displays it in this case, because it only displays this completion value if it is greater than 0, which -1 is not. My implementation has the same behaviour, so this is not an issue, but thank you for checking! 🙂

PR #119, opened 2021-01-05
HEEDLS-264 - Section Page Consolidation Exercise Frontend

SChandarana:
Added the front end for the consolidation button link as well as changes to view model to help with integration. Tests need to be changed/added once the backend is complete and these have been highlighted with comments.

image

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
@@ -2,13 +2,24 @@
 {
     using DigitalLearningSolutions.Web.Tests.TestHelpers;
     using DigitalLearningSolutions.Web.ViewModels.LearningMenu;
+    using FakeItEasy;
     using FluentAssertions;
+    using Microsoft.Extensions.Configuration;
     using NUnit.Framework;
 
     public class SectionContentViewModelTests
     {
+        private IConfiguration config;
         private const int CustomisationId = 5;
         private const int SectionId = 5;
+        private const string BaseUrl = "https://example.com";
+
+        [SetUp]
+        public void SetUp()
+        {
+            config = A.Fake<IConfiguration>();
+            A.CallTo(() => config["CurrentSystemBaseUrl"]).Returns(BaseUrl);
+        }
 
         [Test]
         public void Section_content_should_have_title()
@@ -22,7 +33,7 @@
             );
 
             // When
-            var sectionContentViewModel = new SectionContentViewModel(sectionContent, CustomisationId, SectionId);
+            var sectionContentViewModel = new SectionContentViewModel(config, sectionContent, CustomisationId, SectionId);
 
             // Then
             sectionContentViewModel.CourseTitle.Should().Be($"{applicationName} - {customisationName}");
@@ -38,7 +49,7 @@
             );
 
             // When
-            var sectionContentViewModel = new SectionContentViewModel(sectionContent, CustomisationId, SectionId);
+            var sectionContentViewModel = new SectionContentViewModel(config, sectionContent, CustomisationId, SectionId);
 
             // Then
             sectionContentViewModel.SectionName.Should().Be(sectionName);
@@ -59,7 +70,7 @@
             sectionContent.Tutorials.AddRange(tutorials);
 
             // When
-            var sectionContentViewModel = new SectionContentViewModel(sectionContent, CustomisationId, SectionId);
+            var sectionContentViewModel = new SectionContentViewModel(config, sectionContent, CustomisationId, SectionId);
 
             // Then
             sectionContentViewModel.PercentComplete.Should().Be("");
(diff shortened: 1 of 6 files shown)


DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
```diff
@@ -413,12 +424,14 @@ public void Diagnostic_assessment_separator_is_true_if_diagnostic_assessment_and
...
             // When
-            var sectionContentViewModel = new SectionContentViewModel(sectionContent, CustomisationId, SectionId);
+            var sectionContentViewModel = new SectionContentViewModel(config, sectionContent, CustomisationId, SectionId);
             
             // Then
             sectionContentViewModel.DisplayDiagnosticSeparator.Should().BeTrue();
         }
 
+        // This test will need to be changed once backend is complete and consolidation path can be changed, currently it will fail
+        /*
```
FridaTveit, line 434: I'd prefer not to have commented out tests. Please either fix it or remove it 🙂 

A few ways of fixing it would be:
* Add the consolidation path to the section content model, but give it a default value so the query doesn't need it. That way you could finish these tests now.
* Have a null consolidation path by default in the view model. That way all the old tests should still work. It means the consolidation exercise can't be demoed though, unless the backend is done by Friday, which would be a bit of a shame but not the end of the world 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
```diff
@@ -509,10 +525,17 @@ public void Diagnostic_assessment_completion_status_does_not_use_plural_if_attem
...
             // Then
             sectionContentViewModel.DiagnosticCompletionStatus.Should().Be($"{secScore}/{secOutOf} - {diagAttempts} attempt");
         }
+
+        // Test for diagnostic separator if only diagnostic assessment and consolidation present
+        // Test for tutorial separator if tutorials and consolidation present but no post learning assessment
+        // Test for post learning separator if post learning and consolidation present
+        // Test for no diagnostic separator if tutorial, post learning and consolidation are not shown
+        // Test for no tutorial separator if post learning and consolidation are not shown
+        // Test for no post learning separator if consolidation is not shown
```
FridaTveit, line 539: I'd prefer for this not to be committed and pushed. Feel free to keep it locally if it's useful 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/SectionContentViewModel.cs
```diff
@@ -30,9 +35,13 @@ public SectionContentViewModel(SectionContent sectionContent, int customisationI
             PostLearningStatus = GetPostLearningStatus(sectionContent);
             ShowDiagnostic = sectionContent.DiagnosticAssessmentPath != null && sectionContent.DiagnosticStatus;
             DiagnosticCompletionStatus = GetDiagnosticCompletionStatus(sectionContent);
+            ConsolidationExercisePath =
+                ContentUrlHelper.GetNullableContentPath(config, "sectionContent.SupportingMaterialPath (temporary placeholder)");
```
FridaTveit, line 39: Using a value from the db would be more useful I think? 🙂

DigitalLearningSolutions.Web/Views/LearningMenu/Section/_ConsolidationExerciseCard.cshtml
```diff
@@ -0,0 +1,24 @@
...
+    <div class="nhsuk-grid-row">
+      <div class="nhsuk-grid-column-full">
+        <h3 class="nhsuk-card__heading" id="consolidation-exercise-name">
+          Consolidation Exercise
+        </h3>
+      </div>
+    </div>
+    <div class="nhsuk-grid-row">
+      <div class="nhsuk-grid-column-full">
+        <a class="nhsuk-button nhsuk-button--secondary nhsuk-u-margin-bottom-0 nhsuk-u-margin-top-0"
```
FridaTveit, line 15: Any reason why this is a secondary button? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web/Views/LearningMenu/Section/_ConsolidationExerciseCard.cshtml
SChandarana: As we used a secondary button for the download supporting materials i thought it would be more consistent. I can change this to a regular button if you feel that's more appropriate

FridaTveit: We used a secondary button in that case because there could be more than one button. In this case this card will only have one button so I think it would be most consistent to make it a primary button 🙂

SChandarana:
I decided to have view model temporarily not show the consolidation exercise to make sure the tests pass. This makes it easier for me to work on the backend with no conflicts. The backend seems like it won't be too much of a problem and will likely get done before the demo.

updated photo
image

FridaTveit (approved):
Looks good 🙂

PR #120, opened 2021-01-05
HEEDLS-266 Add tutorial ID to tutorial content iframe URL

benjimarshall:
## Changes
Add a tutorial ID query parameter to the tutorial iframe URLs.

## Testing
Update the unit tests for this URL generation. I've not tested the changes in browser very thoroughly (it would be easier to do that on UAT I think) but it seems consistent


FridaTveit (approved):
Looks good 🙂

PR #121, opened 2021-01-05
HEEDLS-194 Add download summary link to button

benjimarshall:
## Changes
Add progress ID and app configuration to the view model parameters to
generate the URL for the summary link.

## Testing
Update the unit tests to reflect these changes. Tested in Google Chrome, but not with the old code running (I often have issues getting the old system running with SQL requests timing out). It should be easier to test on the test site, and I doubt there will be any issues with it considering how small the change is.

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/CourseCompletionViewModelTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/CourseCompletionViewModelTests.cs
@@ -12,7 +12,7 @@
     {
         private IConfiguration config;
         private const string BaseUrl = "https://example.com";
-        const int ProgressId = 23;
+        private const int ProgressId = 23;
 
         [SetUp]
         public void SetUp()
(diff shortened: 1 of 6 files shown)


DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/CourseCompletionViewModelTests.cs
```diff
@@ -3,11 +3,24 @@
...
+    using FakeItEasy;
     using FluentAssertions;
+    using Microsoft.Extensions.Configuration;
     using NUnit.Framework;
 
     class CourseCompletionViewModelTests
     {
+        private IConfiguration config;
+        private const string BaseUrl = "https://example.com";
+        const int ProgressId = 23;
```
FridaTveit, line 15: private? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/CourseCompletionViewModelTests.cs
benjimarshall: Yep, sorry! Fixed in 6a31d72 🙂

FridaTveit (approved):
Looks good 🙂

PR #122, opened 2021-01-05
HEEDLS-263 Assessment page post learning front end

kwunkiuma:
Create view for post learning page and add caption to diagnostic assessment page.

Integrate with back end by adding view model and controller logic.

Ran and passed all unit tests, screenshots taken in Firefox.

image

image

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/PostLearningAssessmentViewModelTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/PostLearningAssessmentViewModelTests.cs
@@ -112,7 +112,7 @@
                 new PostLearningAssessmentViewModel(postLearningAssessment, CustomisationId, SectionId);
 
             // Then
-            postLearningAssessmentViewModel.GetAssessmentStatus().Should().Be("Not attempted");
+            postLearningAssessmentViewModel.AssessmentStatus.Should().Be("Not attempted");
         }
 
         [Test]
@@ -132,7 +132,7 @@
                 new PostLearningAssessmentViewModel(postLearningAssessment, CustomisationId, SectionId);
 
             // Then
-            postLearningAssessmentViewModel.GetAssessmentStatus().Should().Be("Passed (90% - 4 attempts)");
+            postLearningAssessmentViewModel.AssessmentStatus.Should().Be("Passed (90% - 4 attempts)");
         }
 
         [Test]
@@ -152,7 +152,7 @@
                 new PostLearningAssessmentViewModel(postLearningAssessment, CustomisationId, SectionId);
 
             // Then
-            postLearningAssessmentViewModel.GetAssessmentStatus().Should().Be("Failed (10% - 5 attempts)");
+            postLearningAssessmentViewModel.AssessmentStatus.Should().Be("Failed (10% - 5 attempts)");
         }
 
         [Test]
--- a/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/PostLearningAssessmentViewModelTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/PostLearningAssessmentViewModelTests.cs
@@ -1,0 +1,188 @@
+﻿namespace DigitalLearningSolutions.Web.Tests.ViewModels.LearningMenu
+{
+    using DigitalLearningSolutions.Web.Tests.TestHelpers;
+    using DigitalLearningSolutions.Web.ViewModels.LearningMenu;
+    using FluentAssertions;
+    using NUnit.Framework;
+
+    public class PostLearningAssessmentViewModelTests
+    {
+        private const int CustomisationId = 5;
+        private const int SectionId = 5;
+
+        [Test]
+        public void Post_learning_assessment_should_have_title()
+        {
+            // Given
+            const string applicationName = "Application name";
+            const string customisationName = "Customisation name";
+            var postLearningAssessment = PostLearningAssessmentHelper.CreateDefaultPostLearningAssessment(
+                applicationName: applicationName,
+                customisationName: customisationName
+            );
+
+            // When
+            var postLearningAssessmentViewModel =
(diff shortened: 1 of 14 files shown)


DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/PostLearningAssessmentViewModelTests.cs
```diff
@@ -0,0 +1,188 @@
...
+                plPasses: postLearningPasses,
+                maxScorePl: 10
+            );
+
+            // When
+            var postLearningAssessmentViewModel =
+                new PostLearningAssessmentViewModel(postLearningAssessment, CustomisationId, SectionId);
+
+            // Then
+            postLearningAssessmentViewModel.AssessmentStatus.Should().Be("Failed (10% - 5 attempts)");
```
FridaTveit, line 155: What about if the percentage is not a whole number, rouding is probably required? We should have a test for that (and code to do the rounding if we don't already) 🙂

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
```diff
@@ -150,6 +153,40 @@ public IActionResult Diagnostic(int customisationId, int sectionId)
...
+            var centreId = User.GetCentreId();
+            var postLearningAssessment =
+                postLearningAssessmentService.GetPostLearningAssessment(customisationId, candidateId, sectionId);
+
+            if (postLearningAssessment == null || centreId == null)
+            {
+                logger.LogError(
+                    "Redirecting to 404 as section/centre id was not found. " +
+                    $"Candidate id: {candidateId}, customisation id: {customisationId}, " +
+                    $"section id: {sectionId}");
```
FridaTveit, line 169: We should include the centre id in this log 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/PostLearningAssessmentViewModel.cs
```diff
@@ -0,0 +1,41 @@
...
+        public int SectionId { get; }
+
+        public PostLearningAssessmentViewModel(PostLearningAssessment postLearningAssessment, int customisationId, int sectionId)
+        {
+            CourseTitle = postLearningAssessment.CourseTitle;
+            SectionName = postLearningAssessment.SectionName;
+            PostLearningAssessmentPath = postLearningAssessment.PostLearningAssessmentPath;
+            PostLearningLocked = postLearningAssessment.PostLearningLocked;
+
+            if (postLearningAssessment.PostLearningAttempts == 0)
```
FridaTveit, line 23: I'd prefer for this to be extracted to avoid reassigning to the same variable and making the constructor overly complicated 🙂

DigitalLearningSolutions.Web/Views/LearningMenu/PostLearning/PostLearning.cshtml
```diff
@@ -0,0 +1,55 @@
...
+      <span class="nhsuk-caption-l">Post learning assessment</span>
+      @Model.CourseTitle
+    </h1>
+    <h2 class="nhsuk-u-secondary-text-color">
+      @Model.AssessmentStatus
+    </h2>
+  </div>
+</div>
+
+<div class="nhsuk-inset-text nhsuk-u-margin-top-0 nhsuk-u-margin-bottom-4">
```
FridaTveit, line 31: If this is the same as on the diagnostic assessment then it's probably worth extracting to a partial? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/PostLearningAssessmentViewModelTests.cs
kwunkiuma: It looks like `AssessAttempts.Score` where we get the score from is stored as an int so no rounding should be necessary 🙂

FridaTveit: Ah I see. I think I was getting confused by the `maxScorePl` argument name. Could we rename this to more accurately reflect what it represents (their highest score so far I believe)? 🙂

kwunkiuma: Yes I agree the name is a bit misleading, I used it because it was the name of the column returned in the stored procedure. What do you think of `BestScore`?

FridaTveit: I like it 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/PostLearningAssessmentViewModel.cs
```diff
@@ -18,24 +19,30 @@ public PostLearningAssessmentViewModel(PostLearningAssessment postLearningAssess
...
+            CustomisationId = customisationId;
+            SectionId = sectionId;
+        }
 
-            if (postLearningAssessment.PostLearningAttempts == 0)
-            {
-                AssessmentStatus = "Not attempted";
-            }
-            else
+        public string GetAssessmentStatus()
```
FridaTveit, line 30: This doesn't need to be public. It can be a private method which you use to set the assessment status in the constructor. By giving it appropriate parameters you can store fewer variables while still keeping the constructor simple. Does that make sense? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web/ViewModels/LearningMenu/PostLearningAssessmentViewModel.cs
kwunkiuma: Ah I see, sorry for the misunderstanding 🤦‍♂️

FridaTveit (approved):
Looks good 🙂

PR #124, opened 2021-01-05
HEEDLS-267 Hide average duration on initial menu if n/a

benjimarshall:
## Changes
Hide average duration on initial menu if n/a. On Jira I've added my thoughts on why this is the only line that needs changing.

## Testing
Amend unit tests, and tested on course 6940 (one with an n/a duration).

### A course with an n/a duration
invalid_duration_hidden


FridaTveit (approved):
Looks good 🙂

PR #126, opened 2021-01-06
HEEDLS-265 - Section Page Consolidation Exercise Backend

SChandarana:
This PR includes the integration so the entire system works

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
@@ -49,7 +49,8 @@
                 1,
                 1,
                 true,
-                true
+                true,
+                null
             );
             expectedSectionContent.Tutorials.AddRange(
                 new[]
@@ -112,7 +113,8 @@
                 0,
                 0,
                 true,
-                true
+                true,
+                null
             );
             expectedSectionContent.Tutorials.AddRange(
                 new[]
@@ -155,7 +157,8 @@
                 0,
                 0,
                 true,
-                true
+                true,
+                null
             );
             expectedSectionContent.Tutorials.AddRange(
                 new[]
@@ -222,7 +225,8 @@
                 0,
                 0,
                 true,
-                false
+                false,
+                "https://www.dls.nhs.uk/tracking/MOST/Word07Core/cons/WC07-Exercise_1.zip"
             );
             // Will have no tutorials as CustomisationTutorial.Status is 0 for all tutorials in this section
 
@@ -252,7 +256,8 @@
                 0,
                 0,
                 false,
-                true
+                true,
+                "https://www.dls.nhs.uk/tracking/MOST/Word07Core/cons/WC07-Exercise_1.zip"
             );
             // Will have no tutorials as CustomisationTutorial.Status is 0 for all tutorials in this section
 
@@ -282,7 +287,8 @@
                 0,
                 0,
                 false,
-                false
+                false,
+                "https://www.dls.nhs.uk/tracking/MOST/Word07Core/cons/WC07-Exercise_1.zip"
             );
             expectedSectionContent.Tutorials.AddRange(
                 new[]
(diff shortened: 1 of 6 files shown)


DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
```diff
@@ -362,5 +368,58 @@ public void Get_section_content_returns_null_if_customisation_is_inactive()
...
+            // Then
+            const string expectedConsolidationPath = "https://www.dls.nhs.uk/tracking/MOST/Word07Core/cons/WC07-Exercise_1.zip";
+
+            result.ConsolidationPath.Should().Be(expectedConsolidationPath);
+        }
+
+        [Test]
+        public void Get_section_content_can_have_no_consolidation_path_but_still_return()
+        {
+            // When
```
FridaTveit, line 390: This could do with being split into given/when/then. I'd put the inputs and expected values in the given section 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
```diff
@@ -489,16 +490,20 @@ public void Tutorial_separator_is_false_if_no_tutorials_exist()
         }
 
         [Test]
-        public void Tutorial_separator_is_false_if_tutorials_exist_but_no_post_learning()
+        public void Tutorial_separator_is_false_if_tutorials_exist_but_no_post_learning_or_consolidation_exercise()
         {
             // Given
             const bool isAssessed = false;
+            const string? consolidationPath = null;
```
FridaTveit, line 497: Minor but I'd probably just inline this 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
```diff
@@ -526,5 +531,109 @@ public void Diagnostic_assessment_completion_status_does_not_use_plural_if_attem
...
+
+            // When
+            var sectionContentViewModel = new SectionContentViewModel(config, sectionContent, CustomisationId, SectionId);
+
+            // Then
+            sectionContentViewModel.DisplayPostLearningSeparator.Should().BeFalse();
+        }
+
+        [Test]
+        public void Post_learning_separator_is_false_if_post_learning_does_not_exists()
```
FridaTveit, line 626: I wonder if the tests for each of these separators would be better as one parameterised test each. As far as I can tell there are four conditions we care about for each:
1. Whether there's a diagnostic assessment
2. Whether there are tutorials
3. Whether there's a post learning assessment
4. Whether there's a consolidation exercise

So if you have one parameterised test for the diagnostic separator, one for the tutorial separator and one for the post learning separator. For each have inputs `showDiagnostic`, `showTutorials`, `showPostLearning`, `showConsolidation` and `expectedShowSeparator`. And in the test set it up according to the first four inputs and then compare the separator visibility to the last input. Then you can easily test all possible permutations of the first four inputs. What do you think? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
SChandarana: That seems good! I'll have to have a look at how paramterised testing works and how exactly I want to do it but that seems like a nice solution

FridaTveit: We have quite a few examples in the codebase already, e.g. the tests for checking our output matches the stored procedure in CourseContentServiceTests 🙂

SChandarana:
I only changed the given/when things for the current tests. After looking through almost all of the tests in this set don't follow the given/when/then. I can change this in a separate PR maybe? Or just keep it in mind for future tests and change the previous tests later once most of the other work is complete

FridaTveit:
My thoughts about given/when/then:
* I think it's worth following, even for short tests
* For long tests I'll usually insist on following given/when/then because otherwise the tests are hard to read
* For shorter tests I would prefer given/when/then but will accept just when/then

So no need to change all the tests, but feel free to change any you come across and to keep it in mind for future tests. That's a good programming principle in general, to try to improve any code you touch in little ways if possible so that you leave the code better than you found it. That lets you improve the code in small manageable ways, without starting a big refactor which can be time consuming and difficult. Does that make sense? 🙂

SChandarana:
Yep okay no worries! I'll keep that in mind :D

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
```diff
@@ -532,108 +402,104 @@ public void Diagnostic_assessment_completion_status_does_not_use_plural_if_attem
             sectionContentViewModel.DiagnosticCompletionStatus.Should().Be($"{secScore}/{secOutOf} - {diagAttempts} attempt");
         }
 
-        [Test]
-        public void Diagnostic_assessment_separator_is_false_if_no_diagnostic_assessment()
+        [TestCase(true, true, true, true, true)]
+        [TestCase(true, false, true, true, true)]
+        [TestCase(true, false, false, true, true)]
+        [TestCase(true, false, false, false, false)]
+        [TestCase(false, true, true, true, false)]
```
FridaTveit, line 409: I think you're missing some cases here. We want to test all the permutations of the four booleans, which is why a parameterised test is so useful 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
```diff
@@ -532,108 +402,104 @@ public void Diagnostic_assessment_completion_status_does_not_use_plural_if_attem
...
-            sectionContentViewModel.DisplayDiagnosticSeparator.Should().BeFalse();
+            sectionContentViewModel.DisplayDiagnosticSeparator.Should().Be(separatorExpected);
         }
 
-        [Test]
-        public void Diagnostic_assessment_separator_is_true_if_only_diagnostic_assessment_and_consolidation_exercise_is_present()
+        [TestCase(true, true, true, true)]
+        [TestCase(true, false, true, true)]
+        [TestCase(true, false, false, false)]
+        [TestCase(false, true, true, false)]
```
FridaTveit, line 447: Missing some cases here as well 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
```diff
@@ -532,108 +402,104 @@ public void Diagnostic_assessment_completion_status_does_not_use_plural_if_attem
...
-
-            // Then
-            sectionContentViewModel.DisplayPostLearningSeparator.Should().BeFalse();
-        }
-
-        [Test]
-        public void Post_learning_separator_is_false_if_post_learning_does_not_exists()
+        [TestCase(true, true, true)]
+        [TestCase(true, false, false)]
+        [TestCase(false, true, false)]
```
FridaTveit, line 482: Missing one case here (false, false) 🙂

FridaTveit (commented):
A few comments, otherwise looks good. Great job adding the parameterised tests! 🙂

FridaTveit (approved):
Looks good 🙂

PR #128, opened 2021-01-06
HEEDLS-273 Initial Menu page customisation back end

benjimarshall:
## Changes
Add CourseSettings field to course content service

## Testing
Amend tests


DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
```diff
@@ -47,7 +47,8 @@ public void Get_course_content_of_partially_complete_course_should_return_course
                 true,
                 85,
                 90,
-                0
+                0,
+                null
```
FridaTveit, line 51: Probably worth having at least one test with a course which has settings, either by changing which course is used for one of these tests or by adding a new test? 🙂

FridaTveit (commented):
One comment and as discussed on slack it might be worth doing the parsing in the model. Otherwise looks good 🙂

DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
benjimarshall: Yeah I agree. I'm going to add that parsing now and put tests for that here 🙂

benjimarshall: There are no active courses that have a non-null course settings for the fields used by the initial menu. Instead I've added a test that modifies a course to have some interesting settings, and then test that the `GetCourseContent` method gets the correct settings back 🙂

benjimarshall:
The course settings JSON parser has been added in 06dfe77 🙂

DigitalLearningSolutions.Data/Models/CourseSettings.cs
```diff
@@ -0,0 +1,38 @@
+﻿namespace DigitalLearningSolutions.Data.Models
+{
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+
+    public class CourseSettings
+    {
+        public readonly bool ShowPercentage = true;
```
FridaTveit, line 8: Minor but using a property with getter only is more consistent with other classes I think 🙂

FridaTveit (commented):
One minor comment otherwise looks good 🙂

DigitalLearningSolutions.Data/Models/CourseSettings.cs
benjimarshall: Yep, sorry! Fixed in ea68e57 🙂

FridaTveit (approved):
Looks good 🙂

PR #133, opened 2021-01-07
HEEDLS-280 Add section customisation front end

benjimarshall:
## Changes
Add show percentage, show learning status, and consolidation exercise
settings to section page.

## Testing
Add some unit tests, tested in Google Chrome

### A section with show percentage and show time disabled
show_percentage_and_show_time_disabled

### A section with learn status disabled
learn_status_removed

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
@@ -75,22 +75,34 @@
         }
 
         [Test]
-        public void Section_content_should_have_setting_consolidation_path_when_courseSetting_exercise_is_not_null()
+        public void Section_content_should_have_consolidationExerciseLabel_from_courseSetting()
         {
             // Given
-            const string consolidationPath = "consolidation/path";
-            const string consolidationPathFromSettings = "consolidation/path/fromsetting";
+            const string consolidationText = "Different consolidation description";
             var sectionContent = SectionContentHelper.CreateDefaultSectionContent(
-                consolidationPath: consolidationPath,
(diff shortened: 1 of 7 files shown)


DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
```diff
@@ -56,24 +57,112 @@ public void Section_content_should_have_section_name()
...
+            // Then
+            sectionContentViewModel.ConsolidationExercisePath.Should().Be(expectedConsolidationPath);
+        }
+
+        [Test]
+        public void Section_content_should_have_setting_consolidation_path_when_courseSetting_exercise_is_not_null()
+        {
+            // Given
+            const string consolidationPath = "consolidation/path";
+            const string consolidationPathFromSettings = "consolidation/path/fromsetting";
```
FridaTveit, line 82: Ah no I don't think this is what this setting is for. It's meant to be the text you show on the link to the consolidation exercise. It should be "Consolidation exercise" by default, but it could be set to something else such as "Extra exercise" or something. Does that make sense? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
benjimarshall: Ah I had misread the ticket, sorry about that! Fixed in a94e772

benjimarshall:
### Here is a screenshot of the fixed consolidation exercise custom text
custom_consolidation_label

FridaTveit (approved):
Looks good 🙂

PR #136, opened 2021-01-08
HEEDLS-220 - Section Page Next Button Backend

SChandarana:
Backend complete and tests added. Only missing test case is when sectionNumber is shared by 2 sections as that case doesn't exist in the DB. The tests could potentially be parameterized but i'm not sure if it would have as good readability so let me know what you prefer :D

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
@@ -490,6 +490,7 @@
             const int candidateId = 1;
             const int sectionId = 382;
             const int expectedNextSectionId = 383;
+
             // When
             var result = sectionContentService.GetSectionContent(customisationId, candidateId, sectionId);
 
@@ -505,6 +506,7 @@
             const int candidateId = 1;
             const int sectionId = 383;
             const int expectedNextSectionId = 384;
+
             // When
             var result = sectionContentService.GetSectionContent(customisationId, candidateId, sectionId);
 
@@ -519,6 +521,7 @@
             const int customisationId = 15853;
             const int candidateId = 1;
             const int sectionId = 386;
+
             // When
             var result = sectionContentService.GetSectionContent(customisationId, candidateId, sectionId);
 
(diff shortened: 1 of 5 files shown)


DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
```diff
@@ -474,5 +481,84 @@ public void Get_section_content_should_have_default_course_settings_when_json_is
...
         }
+
+        [Test]
+        public void Get_section_content_next_section_id_should_return_id_if_at_start_of_list()
+        {
+            // Given
+            const int customisationId = 15853;
+            const int candidateId = 1;
+            const int sectionId = 382;
+            const int expectedNextSectionId = 383;
```
FridaTveit, line 492: Empty line needed before `when` section, here and in other tests 🙂

DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
```diff
@@ -474,5 +481,84 @@ public void Get_section_content_should_have_default_course_settings_when_json_is
...
+        public void Get_section_content_next_section_id_should_skip_empty_section()
+        {
+            // Given
+            const int customisationId = 18366;
+            const int candidateId = 210934;
+            const int sectionId = 974;
+
+            // The next section ID in this Application is 975, but the next section with a tutorial selected in
+            // CustomisationTutorials is 978
+            const int expectedNextSectionId = 978;
```
FridaTveit, line 539: What about if the next section only has a diagnostic or post learning assessment? 🙂

FridaTveit (commented):
A couple of comments, otherwise looks good. Probably also worth @benjimarshall reviewing as he did the next button on the tutorial page 🙂

benjimarshall:
Looks good! There are some other edge cases that aren't being tested in the tutorial next button (that probably should be) that you could add here. I don't think there examples in the test DB of these edge cases, so I would just edit some rows (in a transaction scope) to create these cases, and then test that the next section id is as expected. The cases I can think of (but there might be more) are based on multiple sections sharing a section number (which is a case Kevin has asked us to cover):
* For two sections sharing a section number, check that the one with the smaller ID has the one with the larger ID as the next section
* Similarly the one with the larger ID has the one with the next section number as the next ID
* For the previous section number, the next section ID should have the smaller of the two section IDS (who share a section number)
* When the two sections sharing a section number are the last two sections, the one with the larger ID has a null next section

I think your query would satisfy these tests, but the best way to check is just to write them 🙂

FridaTveit (approved):
Looks good 🙂

PR #139, opened 2021-01-08
HEEDLS-248 Make time summary partial for tutorial page

benjimarshall:
## Changes
Take time summary used by tutorial card and put it in a partial, for
use in tutorial card and also tutorial page.

## Testing
Amend tests for these view models, move time summary tests into the
partial's view model tests.

Pages tested in Chrome, Firefox, Edge, and IE11, using a screen reader and a narrow viewport.

## Screenshot
tutorial_time_improvement

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/Helpers/DurationFormattingHelperTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/Helpers/DurationFormattingHelperTests.cs
@@ -1,0 +1,106 @@
+﻿namespace DigitalLearningSolutions.Web.Tests.Helpers
+{
+    using DigitalLearningSolutions.Web.Helpers;
+    using DigitalLearningSolutions.Web.Tests.TestHelpers;
+    using DigitalLearningSolutions.Web.ViewModels.LearningMenu;
+    using FluentAssertions;
+    using NUnit.Framework;
+
+    class DurationFormattingHelperTests
+    {
+        [Test]
+        public void FormatDuration_should_have_averageDuration_for_0_minutes()
+        {
+            // Given
+            const int averageDuration = 0;
+
+            // When
+            var result = DurationFormattingHelper.FormatDuration(averageDuration);
+
+            // Then
+            result.Should().Be("0 minutes");
+        }
+
+        [Test]
+        public void FormatDuration_should_have_averageDuration_for_1_minute()
+        {
+            // Given
+            const int averageDuration = 1;
+
+            // When
+            var result = DurationFormattingHelper.FormatDuration(averageDuration);
+
+            // Then
+            result.Should().Be("1 minute");
+        }
+
+        [Test]
+        public void FormatDuration_should_have_averageDuration_for_under_an_hour()
+        {
+            // Given
+            const int averageDuration = 30;
+
+            // When
+            var result = DurationFormattingHelper.FormatDuration(averageDuration);
+
+            // Then
+            result.Should().Be("30 minutes");
+        }
+
+        [Test]
+        public void FormatDuration_should_have_averageDuration_for_whole_number_of_hours()
+        {
+            // Given
+            const int averageDuration = 120;
+
+            // When
+            var result = DurationFormattingHelper.FormatDuration(averageDuration);
+
+            // Then
+            result.Should().Be("2 hours");
+        }
+
+        [Test]
+        public void FormatDuration_should_have_averageDuration_for_one_hour_one_minute()
+        {
+            // Given
+            const int averageDuration = 61;
+
+            // When
+            var result = DurationFormattingHelper.FormatDuration(averageDuration);
+
+            // Then
+            result.Should().Be("1 hour 1 minute");
+        }
+
+
+        [Test]
+        public void FormatDuration_should_have_averageDuration_for_multiple_hours()
+        {
+            // Given
+            const int averageDuration = 195;
+
+            // When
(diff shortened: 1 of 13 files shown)


DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/TutorialTimeSummaryViewModelTests.cs
```diff
@@ -0,0 +1,144 @@
...
+    using FluentAssertions;
+    using NUnit.Framework;
+
+    class TutorialTimeSummaryViewModelTests
+    {
+        private const bool ShowTime = true;
+        private const bool ShowLearnStatus = true;
+
+        [Test]
+        public void Tutorial_card_average_time_spent_string_should_have_plural_if_value_is_not_one()
```
FridaTveit, line 13: Probably shouldn't say `card`? 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/TutorialViewModel.cs
```diff
@@ -73,21 +78,6 @@ private string GetScoreSummary(int currentScore, int possibleScore)
             return CanShowProgress ? $"(score: {currentScore} out of {possibleScore})" : "";
         }
 
-        private string GetTimeSummary(int timeSpent, int averageTutorialDuration)
-        {
-            return $"{ParseMinutes(timeSpent)}" +
-                   $" (average time {ParseMinutes(averageTutorialDuration)})";
-        }
-
-        private static string ParseMinutes(int minutes)
```
FridaTveit, line 82: Instead of not doing this on either page, I think we should do it on both. It definitely looks better when the time is longer than an hour 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/TutorialViewModelTests.cs
```diff
@@ -219,13 +219,18 @@ public void Tutorial_should_have_nextLinkViewModel()
...
+            const int averageTutorialDuration = 73;
+            const int timeSpent = 41;
+
             var expectedTutorialInformation = TutorialContentHelper.CreateDefaultTutorialInformation(
-                averageTutorialDuration: 73,
-                timeSpent: 41
+                timeSpent: timeSpent,
+                averageTutorialDuration: averageTutorialDuration
             );
+            // TODO: Test different customisations settings when found as part of HEEDLS-196
```
FridaTveit, line 232: Minor but is this comment necessary? TODO comments are very easy to forget and leave in after the ticket it refers to is done, so I'd probably leave it out 🙂

FridaTveit (commented):
A few comments, otherwise looks good. Nice extraction of shared code 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/TutorialTimeSummaryViewModelTests.cs
benjimarshall: Yep, sorry! Fixed in 18eef65 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/TutorialViewModel.cs
benjimarshall: Yeah I agree. I remembered some code I wrote for the initial menu did this well, so I've refactored that and used it in this case as well 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/TutorialViewModelTests.cs
benjimarshall: Yeah it's not necessary, I should remember this when I get around to this ticket so I've_ removed that in 18eef65 🙂

DigitalLearningSolutions.Web.Tests/Helpers/DurationFormattingHelperTests.cs
```diff
@@ -0,0 +1,106 @@
...
+    using DigitalLearningSolutions.Web.Helpers;
+    using DigitalLearningSolutions.Web.Tests.TestHelpers;
+    using DigitalLearningSolutions.Web.ViewModels.LearningMenu;
+    using FluentAssertions;
+    using NUnit.Framework;
+
+    class DurationFormattingHelperTests
+    {
+        [Test]
+        public void FormatDuration_should_have_averageDuration_for_0_minutes()
```
FridaTveit, line 12: Very minor but this formatting doesn't really have anything to do with average as it could be used for any duration, and we possibly do use it for the time spent as well? Either way I think it would be better to avoid talking about average duration here as it could be confusing 🙂

FridaTveit (commented):
One small comment about naming, otherwise looks good 🙂

DigitalLearningSolutions.Web.Tests/Helpers/DurationFormattingHelperTests.cs
benjimarshall: Yes I agree, that was left over from the refactoring and I didn't think rename them. Fixed in 4476c83 🙂

FridaTveit (approved):
Looks good 🙂

PR #140, opened 2021-01-08
HEEDLS-199 Section Page Consolidation Exercise Bug Fix

SChandarana:


DigitalLearningSolutions.Web/Helpers/ConfigHelper.cs
```diff
@@ -52,5 +52,10 @@ public static string GetDownloadSummaryUrl(this IConfiguration config, int progr
         {
             return $"{config[CurrentSystemBaseUrlName]}/tracking/summary?ProgressID={progressId}";
         }
+
+        public static string GetConsolidationPathUrl(this IConfiguration config, string consolidationPath)
+        {
+            return config["CurrentSystemBaseUrl"] + "/tracking/dlconsolidation?client=" + consolidationPath;
```
FridaTveit, line 58: Should use string template to be consistent with the other methods 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
```diff
@@ -65,7 +65,7 @@ public void Section_content_should_have_consolidation_path_when_courseSetting_ex
                 consolidationPath: consolidationPath,
                 courseSettings: null
             );
-            var expectedConsolidationPath = ContentUrlHelper.GetContentPath(config, consolidationPath);
+            var expectedConsolidationPath = config.GetConsolidationPathUrl(consolidationPath);
```
FridaTveit, line 68: Please put the actual path in here, instead of relying on the implementation of the config extension method. Otherwise we're not actually testing that the value is correct 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
SChandarana: Should I still use config["CurrentSystemBaseUrl"] in the test?

FridaTveit: My usual process is:
* Store the mock base url in a variable 
* Set config to return that base url as the CurrentSystemBaseUrl
* Check that the result is the mock base url + the path you expect
Does that make sense? For example see `Content_viewer_should_parse_scorm_source` in `ContentViewerViewModelTests`  🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
```diff
@@ -65,7 +65,7 @@ public void Section_content_should_have_consolidation_path_when_courseSetting_ex
                 consolidationPath: consolidationPath,
                 courseSettings: null
             );
-            var expectedConsolidationPath = config.GetConsolidationPathUrl(consolidationPath);
+            var expectedConsolidationPath = $"{config["CurrentSystemBaseUrl"]}/tracking/dlconsolidation?client={consolidationPath}";
```
FridaTveit, line 68: This should use the BaseUrl variable. It shouldn't use the config at all. We want it to not rely on any external methods. Does that make sense? 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
```diff
@@ -618,5 +618,20 @@ bool separatorExpected
             // Then
             sectionContentViewModel.DisplayPostLearningSeparator.Should().Be(separatorExpected);
         }
+
+        [Test]
+        public void Section_content_should_properly_convert_consolidation_path()
```
FridaTveit, line 623: How is this test different to the other test? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
SChandarana: Ah okay yep that makes sense

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
SChandarana: The other test is doing some sort of check with "courseSettings" to make sure it has a consolidation path. I added this test because it would test that the consolidation path is being properly converted so if the courseSettings thing fails, we know its due to something else and not the conversion. We could change the courseSettings test to just check if consolidation path is not null? So that these tests are both separate. Does that make sense?

FridaTveit: The course settings is irrelevant to the consolidation path. So we only need one of these tests 🙂

SChandarana: Ah okay yep looks like a the coursesettings test was changed/added by mistake, will fix that now

FridaTveit (approved):
Looks good 🙂

PR #142, opened 2021-01-11
HEEDLS-277 Diagnostic assessment launch selection

kwunkiuma:
Create view for diagnostic content.

Integrate with back end and add logic for checkboxes on assessment page.

Ran and passed all unit tests. Screenshot taken in Firefox.
image


DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
```diff
@@ -574,6 +576,193 @@ public void Diagnostic_assessment_should_404_if_failed_to_enrol()
...
+            A.CallTo(() => diagnosticAssessmentService.GetDiagnosticContent(CustomisationId, SectionId))
+                .Returns(defaultDiagnosticContent);
+            A.CallTo(() => courseContentService.GetOrCreateProgressId(CandidateId, CustomisationId, CentreId)).
+                Returns(progressId);
+
+            // When
+            var result = controller.DiagnosticContent(CustomisationId, SectionId, EmptyTutorialSelection);
+
+            // Then
+            var expectedModel = new DiagnosticContentViewModel(config, defaultDiagnosticContent, selectedTutorials, CustomisationId, CentreId, SectionId, progressId, CandidateId);
```
FridaTveit, line 697: Minor but I think this is worth splitting over multiple lines 🙂

DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
```diff
@@ -574,6 +576,193 @@ public void Diagnostic_assessment_should_404_if_failed_to_enrol()
...
+            // Then
+            result.Should()
+                .BeRedirectToActionResult()
+                .WithControllerName("LearningSolutions")
+                .WithActionName("StatusCode")
+                .WithRouteValue("code", 404);
+        }
+
+        [Test]
+        public void Diagnostic_content_should_404_if_tutorial_id_not_found()
```
FridaTveit, line 742: This test name doesn't really explain the scenario we're trying to test 🙂

DigitalLearningSolutions.Web.Tests/TestHelpers/DiagnosticAssessmentHelper.cs
```diff
@@ -26,5 +26,26 @@ public static class DiagnosticAssessmentHelper
...
         }
+
+        public static DiagnosticContent CreateDefaultDiagnosticContent(
+            string applicationName = "application name",
+            string customisationName = "customisation name",
+            string sectionName = "section name",
+            string diagAssessPath = "https://www.dls.nhs.uk/tracking/MOST/Word10Core/Assess/L2_Word_2010_Diag_1.dcr",
+            bool diagObjSelect = true,
+            int plaPassThreshold = 50,
+            int currentVersion = 1
```
FridaTveit, line 37: I think some of these could do with more descriptive names? 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/DiagnosticContentViewModelTests.cs
```diff
@@ -0,0 +1,240 @@
...
+        private const int SectionId = 5;
+        private const int CandidateId = 5;
+        private const int ProgressId = 5;
+
+        [SetUp]
+        public void SetUp()
+        {
+            config = A.Fake<IConfiguration>();
+            A.CallTo(() => config["CurrentSystemBaseUrl"]).Returns(BaseUrl);
+            SelectedTutorials = new List<int>();
```
FridaTveit, line 27: I'm not sure this should be a class variable, as it's not used for every test. I would either:
* Inline it and define it for each test
* Define it as a constant named `EmptySelectedTutorials` to make it clear how it's meant to be used. 

Does that make sense? 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/DiagnosticContentViewModelTests.cs
```diff
@@ -0,0 +1,240 @@
...
+        }
+
+        [Test]
+        public void Diagnostic_content_should_parse_html_url()
+        {
+            // Given
+            const int centreId = 11;
+            const int customisationId = 22;
+            const int candidateId = 33;
+            const int sectionId = 44;
```
FridaTveit, line 135: Any reason why you're defining these here when you've got them as class constants? 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/DiagnosticContentViewModelTests.cs
```diff
@@ -0,0 +1,240 @@
...
+            const int currentVersion = 55;
+            const int progressId = 66;
+            const int plaPassThreshold = 77;
+            const string diagAssessPath = "https://www.dls.nhs.uk/tracking/RFM/L1_Word10/Assess/L1_2.03_Diag.dcr";
```
FridaTveit, line 139: Not very important, but since we don't support dcr I think it would be best to use a different example. Just to avoid confusion 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/DiagnosticContentViewModelTests.cs
```diff
@@ -0,0 +1,240 @@
...
+            // Then
+            diagnosticContentViewModel.ContentSource.Should().Be(
+                "https://www.dls.nhs.uk/tracking/RFM/L1_Word10/Assess/L1_2.03_Diag.dcr" +
+                "?CentreID=11&CustomisationID=22&CandidateID=33&SectionID=44&Version=55&ProgressID=66" +
+                $"&type=diag&TrackURL={BaseUrl}/tracking/tracker&objlist=[1,2,3]&plathresh=77"
+            );
+        }
+
+        [Test]
+        public void Diagnostic_content_should_have_tutorials()
```
FridaTveit, line 168: This test doesn't match the name. In fact it looks like a copy of the test above? 🙂

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
```diff
@@ -153,6 +155,58 @@ public IActionResult Diagnostic(int customisationId, int sectionId)
...
+        public IActionResult DiagnosticContent(int customisationId, int sectionId, List<int> checkedTutorials)
+        {
+            var candidateId = User.GetCandidateId();
+            var centreId = User.GetCentreId();
+            var diagnosticContent = diagnosticAssessmentService.GetDiagnosticContent(customisationId, sectionId);
+
+            if (diagnosticContent == null || centreId == null)
+            {
+                logger.LogError(
+                    "Redirecting to 404 as section/centre id was not found. " +
```
FridaTveit, line 168: Could also be customisation id 🙂

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
```diff
@@ -153,6 +155,58 @@ public IActionResult Diagnostic(int customisationId, int sectionId)
...
+                    $"Candidate id: {candidateId}, customisation id: {customisationId}, " +
+                    $"centre id: {centreId?.ToString() ?? "null"}, section id: {sectionId}");
+                return RedirectToAction("StatusCode", "LearningSolutions", new { code = 404 });
+            }
+
+            if (checkedTutorials.Except(diagnosticContent.Tutorials).Any())
+            {
+                logger.LogError(
+                    "Redirecting to 404 as tutorial id was not found. " +
```
FridaTveit, line 177: This error message doesn't really match what's happened. The issue is that the checked tutorials don't match the tutorials in the diagnostic assessment 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/DiagnosticContentViewModel.cs
```diff
@@ -0,0 +1,79 @@
...
+    public class DiagnosticContentViewModel
+    {
+        public int CustomisationId { get; }
+        public int SectionId { get; }
+        public string CourseTitle { get; }
+        public string SectionName { get; }
+        public string ContentSource { get; }
+        public List<int> Tutorials { get; }
+        private static readonly Regex ScormRegex = new Regex(@".*imsmanifest\.xml$");
+        private static bool IsScormPath(string path) => ScormRegex.IsMatch(path);
```
FridaTveit, line 18: Don't we have this in another class as well? If so, probably worth extracting so it can be shared 🙂

DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
```diff
@@ -574,6 +576,193 @@ public void Diagnostic_assessment_should_404_if_failed_to_enrol()
...
+        {
+            // Given
+            const int progressId = 299;
+            var availableTutorials = new List<int>(new[] { 1, 2, 3 });
+            var selectedTutorials = new List<int>(new[] { 4, 5, 6 });
```
FridaTveit, line 747: I think we should also test the scenario where there's some overlap between the two. And do we have a test where the selected tutorials are a subset of the available tutorials (which should be allowed)? 🙂

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
```diff
@@ -153,6 +155,58 @@ public IActionResult Diagnostic(int customisationId, int sectionId)
...
+            if (diagnosticContent == null || centreId == null)
+            {
+                logger.LogError(
+                    "Redirecting to 404 as section/centre id was not found. " +
+                    $"Candidate id: {candidateId}, customisation id: {customisationId}, " +
+                    $"centre id: {centreId?.ToString() ?? "null"}, section id: {sectionId}");
+                return RedirectToAction("StatusCode", "LearningSolutions", new { code = 404 });
+            }
+
+            if (checkedTutorials.Except(diagnosticContent.Tutorials).Any())
```
FridaTveit, line 174: I think this belongs in a service, not in the controller. There's also another check we need; if diagObjSelect is not true then the selected tutorials should be ignored and all tutorials used. I don't think you have any handling of that at the moment? I think splitting the `DiagnosticAssessmentService` into two services; `DiagnosticAssessmentDataService` which handles just fetching the data like the current service does, and `DiagnosticAssessmentService` which handles any processing of the data like this. What do you think? 🙂

DigitalLearningSolutions.Web/Views/LearningMenu/Diagnostic/Diagnostic.cshtml
```diff
@@ -30,7 +30,7 @@
 
 <partial name="Shared/_AssessmentPageImportantInformation" />
 
-<form class="nhsuk-form-group">
+<form class="nhsuk-form-group" method="get" asp-action="DiagnosticContent" asp-route-customisationId="@Model.CustomisationId" asp-route-sectionId="@Model.SectionId">
```
FridaTveit, line 33: Worth splitting over multiple lines

DigitalLearningSolutions.Web/Views/LearningMenu/Diagnostic/DiagnosticContent.cshtml
```diff
@@ -0,0 +1,56 @@
...
+}
+
+@section scripts {
+  <script src="@Url.Content("~/js/learningMenu/contentViewer.js")" asp-append-version="true"></script>
+}
+
+<div class="nhsuk-grid-row">
+  <div class="nhsuk-grid-column-full">
+    <h1 id="page-heading" class="nhsuk-heading-xl">@Model.SectionName</h1>
+    <div class="nhsuk-error-summary non-js-only" aria-labelledby="error-summary-title" role="alert" tabindex="-1">
```
FridaTveit, line 42: This is used on the other content viewer too right? If so, worth extracting to a partial 🙂

DigitalLearningSolutions.Web/Views/Shared/_Layout.cshtml
```diff
@@ -106,6 +106,8 @@
       </nav>
       <div class="visual-separator"></div>
     </header>
+    
+    @RenderSection("NavBreadcrumbs", required: false)
```
FridaTveit, line 110: This should be merged to master really. But as it's such a small change it should be fine to handle the merge conflicts of doing it on both branches. FYI @benjimarshall (as putting it on master is part of your ticket).

FridaTveit (commented):

DigitalLearningSolutions.Web/Views/Shared/_Layout.cshtml
benjimarshall: Thanks, I'll see where we are when I've finished HEEDLS-286 🙂

DigitalLearningSolutions.Web/Views/LearningMenu/Diagnostic/DiagnosticContent.cshtml
```diff
@@ -0,0 +1,56 @@
...
+    <div class="nhsuk-width-container">
+      <ol class="nhsuk-breadcrumb__list">
+        <li class="nhsuk-breadcrumb__item"><a class="nhsuk-breadcrumb__link" asp-controller="LearningPortal" asp-action="Current">Current activities</a></li>
+        <li class="nhsuk-breadcrumb__item"><a class="nhsuk-breadcrumb__link" asp-action="Index" asp-route-customisationId="@Model.CustomisationId">@Model.CourseTitle</a></li>
+        <li class="nhsuk-breadcrumb__item"><a class="nhsuk-breadcrumb__link" asp-action="Section" asp-route-customisationId="@Model.CustomisationId" asp-route-sectionId="@Model.SectionId">@Model.SectionName</a></li>
+        <li class="nhsuk-breadcrumb__item"><a class="nhsuk-breadcrumb__link" asp-action="Diagnostic" asp-route-customisationId="@Model.CustomisationId" asp-route-sectionId="@Model.SectionId">Diagnostic assessment</a></li>
+        <li class="nhsuk-breadcrumb__item">Content</li>
+      </ol>
+      <p class="nhsuk-breadcrumb__back"><a class="nhsuk-breadcrumb__backlink" asp-action="Section" asp-route-customisationId="@Model.CustomisationId" asp-route-sectionId="@Model.SectionId">Return to section</a></p>
```
FridaTveit, line 30: This should be return to diagnostic assessment, not return to section 🙂

DigitalLearningSolutions.Web/Views/LearningMenu/Diagnostic/DiagnosticContent.cshtml
```diff
@@ -0,0 +1,56 @@
...
+}
+@section NavMenuItems {
+  <partial name="Shared/_NavMenuItems" />
+}
+
+@section NavBreadcrumbs {
+  <nav class="nhsuk-breadcrumb" aria-label="Breadcrumb">
+    <div class="nhsuk-width-container">
+      <ol class="nhsuk-breadcrumb__list">
+        <li class="nhsuk-breadcrumb__item"><a class="nhsuk-breadcrumb__link" asp-controller="LearningPortal" asp-action="Current">Current activities</a></li>
```
FridaTveit, line 24: Sorry one more thing, this should use our close action instead of Current, so that it will stop the session 🙂

benjimarshall, line 24: Current activities should redirect to the close action, not to the learning portals directly to close the session tracking. See the back link on the initial menu page for an example of this link) 🙂

benjimarshall: Haha Frida beat me to it!

DigitalLearningSolutions.Web/Controllers/LearningMenuController/LearningMenuController.cs
kwunkiuma: I'm not sure what we want to do in the new `DiagnosticAssessmentService`, would it have a method taking the `checkedTutorials` and `diagnosticContent` then return a `bool`? How about using a helper class?

FridaTveit: The new service should be a wrapper around the old one. So it should have a method to get the diagnostic assessment and one to get the diagnostic assessment content, like the current service does. The first method will just call the same method on the data service. But the second will take the checked tutorials and do the checks described after fetching the assessment content. Does that make sense? 🙂

DigitalLearningSolutions.Web/Views/LearningMenu/Diagnostic/DiagnosticContent.cshtml
kwunkiuma: The two partial views will need a different error message, how should I pass this in?

FridaTveit: Ah good point. You could pass it in through a model... Yes you could probably make the model just a string, the string being either "tutorial" or "assessment" (I assume that's the only difference?). That would probably be fine. Although if that doesn't work then it's fine to keep them separate, it's only a small thing after all 🙂

DigitalLearningSolutions.Data.Tests/Helpers/DiagnosticAssessmentTestHelper.cs
```diff
@@ -28,5 +28,49 @@ public void UpdateDiagAssessOutOf(int tutorialId, int diagAssessOutOf)
...
         }
+
+        public DiagnosticAssessment CreateDefaultDiagnosticAssessment(
+            string applicationName = "application name",
+            string customisationName = "customisation name",
+            string sectionName = "section name",
+            int diagnosticAttempts = 1,
+            int sectionScore = 2,
+            int maxSectionScore = 3,
+            string diagnosticAssessmentPath = "https://www.dls.nhs.uk/tracking/MOST/Word10Core/Assess/L2_Word_2010_Diag_1.dcr",
```
FridaTveit, line 39: Please use the html path, like in the method below 🙂

DigitalLearningSolutions.Data.Tests/Services/DiagnosticAssessmentDataServiceTests.cs
```diff
@@ -0,0 +1,466 @@
+﻿namespace DigitalLearningSolutions.Data.Tests.Services
```
FridaTveit, line 1: I assume this has just been renamed from DiagnosticAssessmentServiceTests and the content hasn't changed? 🙂

DigitalLearningSolutions.Data/Services/DiagnosticAssessmentDataService.cs
```diff
@@ -0,0 +1,156 @@
+﻿namespace DigitalLearningSolutions.Data.Services
```
FridaTveit, line 1: Have any changes been made to this file or has it just been renamed? 🙂

DigitalLearningSolutions.Web.Tests/TestHelpers/DiagnosticAssessmentHelper.cs
```diff
@@ -26,5 +26,26 @@ public static class DiagnosticAssessmentHelper
                 diagObjSelect
             );
         }
+
+        public static DiagnosticContent CreateDefaultDiagnosticContent(
```
FridaTveit, line 30: Don't you have this defined in another file as well? 🙂

DigitalLearningSolutions.Web/Helpers/ContentViewerHelper.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace DigitalLearningSolutions.Web.Helpers
+{
+    using System.Text.RegularExpressions;
+
+    public static class ContentViewerHelper
+    {
+        public static readonly Regex ScormRegex = new Regex(@".*imsmanifest\.xml$");
```
FridaTveit, line 7: Does this need to be public? 🙂

DigitalLearningSolutions.Web/Views/LearningMenu/Diagnostic/DiagnosticContent.cshtml
```diff
@@ -0,0 +1,49 @@
...
+        <li class="nhsuk-breadcrumb__item"><a class="nhsuk-breadcrumb__link" asp-action="Diagnostic" asp-route-customisationId="@Model.CustomisationId" asp-route-sectionId="@Model.SectionId">Diagnostic assessment</a></li>
+        <li class="nhsuk-breadcrumb__item">Content</li>
+      </ol>
+      <p class="nhsuk-breadcrumb__back"><a class="nhsuk-breadcrumb__backlink" asp-action="Diagnostic" asp-route-customisationId="@Model.CustomisationId" asp-route-sectionId="@Model.SectionId">Return to diagnostic assessment</a></p>
+    </div>
+  </nav>
+}
+
+@section scripts {
+  <script src="@Url.Content("~/js/learningMenu/contentViewer.js")" asp-append-version="true"></script>
```
FridaTveit, line 37: I don't think you want to load this as it's not quite the right script 🙂

DigitalLearningSolutions.Web/Views/LearningMenu/Tutorial/ContentViewer.cshtml
```diff
@@ -10,6 +10,7 @@
   ViewData["CustomisationId"] = Model.CustomisationId;
   ViewData["HeaderPath"] = $"{Configuration["AppRootPath"]}/LearningMenu/{Model.CustomisationId}";
   ViewData["HeaderPathName"] = Model.CourseTitle;
+  const string contentType = "tutorial";
```
FridaTveit, line 13: I wouldn't expect this to be defined here. You should be able to define it when you pass it to the partial. This section is just for setting view data 🙂

DigitalLearningSolutions.Web/Views/LearningMenu/Diagnostic/DiagnosticContent.cshtml
```diff
@@ -0,0 +1,49 @@
...
+<link rel="stylesheet" href="@Url.Content("~/css/learningMenu/index.css")" asp-append-version="true">
+
+@{
+  ViewData["HeaderPrefix"] = "";
+  ViewData["Application"] = Model.CourseTitle;
+  ViewData["Title"] = Model.CourseTitle;
+  ViewData["CustomisationId"] = Model.CustomisationId;
+  ViewData["HeaderPath"] = $"{Configuration["AppRootPath"]}/LearningMenu/{Model.CustomisationId}";
+  ViewData["HeaderPathName"] = Model.CourseTitle;
+  const string contentType = "assessment";
```
FridaTveit, line 15: I don't think this should be defined here 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data/Services/DiagnosticAssessmentDataService.cs
kwunkiuma: This was renamed from `DiagnosticAssessmentService`.

DigitalLearningSolutions.Web.Tests/TestHelpers/DiagnosticAssessmentHelper.cs
kwunkiuma: It's defined in a web test helper which I don't think we should be referencing from data tests?

kwunkiuma: Actually that might not be the case

FridaTveit: Should be fine to reuse helpers across test projects anyway 🙂

kwunkiuma: Is it okay to use data test helpers in the web tests? If so I can change this to use the data test helper

FridaTveit: See my comment above 🙂

DigitalLearningSolutions.Data.Tests/Services/DiagnosticAssessmentDataServiceTests.cs
kwunkiuma: Correct, just a file rename

DigitalLearningSolutions.Data/Services/DiagnosticAssessmentDataService.cs
kwunkiuma: Just a file rename

DigitalLearningSolutions.Web/Views/LearningMenu/Tutorial/ContentViewer.cshtml
kwunkiuma: I had some troubles because passing the model requires the variable name to be enclosed in quotation marks. How should I try to pass a string to the partial?

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/DiagnosticAssessmentViewModelTests.cs
```diff
@@ -11,13 +11,18 @@ public class DiagnosticAssessmentViewModelTests
         private const int CustomisationId = 5;
         private const int SectionId = 5;
 
+        [SetUp]
```
FridaTveit, line 14: Could be removed? 🙂

FridaTveit (commented):
One small thing, otherwise looks good 🙂

FridaTveit (approved):
Looks good 🙂

PR #146, opened 2021-01-11
HEEDLS-NONE Improve useful queries in readme

FridaTveit:
The old query would just get the day in the date, e.g. 11 today. So it would return any date with day 11, for example 11th Nov last year. Which was not exactly what I wanted... 🤦


benjimarshall (approved):
Looks good to me 👍

PR #151, opened 2021-01-12
HEEDLS 221 Section Page Next Button Frontend

SChandarana:
image

image

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
@@ -617,5 +617,36 @@
             // Then
             sectionContentViewModel.DisplayPostLearningSeparator.Should().Be(separatorExpected);
         }
+
+        [Test]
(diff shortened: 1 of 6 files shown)


DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
```diff
@@ -617,5 +617,36 @@ bool separatorExpected
             // Then
             sectionContentViewModel.DisplayPostLearningSeparator.Should().Be(separatorExpected);
         }
+
+        [Test]
+        public void Section_content_should_have_next_section_id_if_there_is_one()
```
FridaTveit, line 622: The `section_content` part of this test name isn't really necessary I think. In fact it makes it a bit confusing as it sounds like you're testing the section content instead of the view model. Just `Should_have_next_section_id_if_there_is_one` is fine I think 🙂

FridaTveit (commented):
One minor comment about naming, otherwise looks good 🙂

FridaTveit (approved):
Looks good 🙂

PR #152, opened 2021-01-12
HEEDLS-288 Ignore archived tutorials in data services

benjimarshall:
### Changes
Ignore archived tutorials in data services by adding filters
to `WHERE` clauses in various queries.

### Testing
* Test that archived tutorials are ignored (both in lists of tutorials, and when aggregating scores over tutorials).
* Rewrite the tests comparing to the old stored procedures because these services now diverge from the stored procedures (they would include archived tutorials).

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
@@ -502,63 +502,146 @@
             }
         }
 
-        [TestCase(254480, 12589, 101, 285054)]
-        [TestCase(1, 15853, 101, 173218)]
-        [TestCase(254480, 24224, 101, null)]
-        [TestCase(22044, 10059, 121, 100467)]
-        [TestCase(11, 6226, 101, 89490)]
-        public void Get_course_content_should_have_same_sections_as_stored_procedure(
-            int candidateId,
-            int customisationId,
-            int centreId,
-            int? progressId
-        )
-        {
-            using (new TransactionScope())
-            {
-                // Given
-                var validProgressId = progressId ?? courseContentTestHelper.CreateProgressId(customisationId, candidateId, centreId);
-
-                var sectionIdsReturnedFromOldStoredProcedure = courseContentTestHelper
-                    .SectionsFromOldStoredProcedure(validProgressId)
-                    .Select(section => section.SectionID);
-
-                // When
-                var sectionIdsInCourseContent = courseContentService
-                    .GetCourseContent(candidateId, customisationId)?
-                    .Sections
-                    .Select(section => section.Id);
-
-                // Then
-                sectionIdsInCourseContent.Should().BeEquivalentTo(sectionIdsReturnedFromOldStoredProcedure);
-            }
-        }
-
-        [TestCase(254480, 12589)]
-        [TestCase(1, 15853)]
-        [TestCase(254480, 24224)]
-        [TestCase(22044, 10059)]
-        [TestCase(11, 6226)]
-        [TestCase(207900, 274400)]
-        [TestCase(213382, 4339)]
-        [TestCase(286788, 23638)]
-        public void Get_course_content_should_have_same_duration_as_stored_procedure(
-            int candidateId,
-            int customisationId
-        )
+        [Test]
+        public void Get_course_content_should_not_return_archived_sections()
         {
             // Given
-            var durationFromOldStoredFunction =
-                courseContentTestHelper.GetCustomisationDurationFromOldProcedure(customisationId);
+            const int candidateId = 23031;
+            const int customisationId = 14212;
+
+            // When
+            var sectionIdsInCourseContent = courseContentService
+                .GetCourseContent(candidateId, customisationId)?
+                .Sections
+                .Select(section => section.Id);
+
+            // Then
+            sectionIdsInCourseContent.Should().Equal(249, 250, 251, 252);
+        }
+
+        [Test]
+        public void Get_course_content_should_only_return_sections_with_tutorials_with_status_or_diagStatus_1_when_not_assessed()
+        {
+            // Given
+            const int candidateId = 22044;
+            const int customisationId = 10059;
+
+            // When
+            var sectionIdsInCourseContent = courseContentService
+                .GetCourseContent(candidateId, customisationId)?
+                .Sections
+                .Select(section => section.Id);
+
+            // Then
+            sectionIdsInCourseContent.Should().Equal(212, 213, 215, 219, 221);
+        }
+
+        [Test]
+        public void Get_course_content_returns_null_when_there_are_no_customisationTutorials()
+        {
+            // Given
+            const int customisationId = 58;
+            const int candidateId = 4;
 
             // When
             var result = courseContentService.GetCourseContent(candidateId, customisationId);
 
             // Then
-            var formattedResult = courseContentTestHelper.FormatDurationLikeOldProcedure(
-                result?.AverageDuration
-            );
-            formattedResult.Should().Be(durationFromOldStoredFunction);
+            result.Should().BeNull();
+        }
+
+        [Test]
+        public void Get_course_content_returns_section_when_all_tutorial_status_are_0_but_isAssessed()
+        {
+            // Given
+            const int customisationId = 9736;
+            const int candidateId = 2;
+
+            // When
+            var sectionIdsInCourseContent = courseContentService
+                .GetCourseContent(candidateId, customisationId)?
+                .Sections
+                .Select(section => section.Id);
+
+            // Then
+            sectionIdsInCourseContent.Should().Contain(103);
+        }
+
+        [Test]
+        public void Get_course_content_uses_overrideTutorialMins_when_calculating_duration()
+        {
+            // Given
+            const int customisationId = 23638;
+            const int candidateId = 286788;
+
+            // When
+            var result = courseContentService.GetCourseContent(candidateId, customisationId);
+
+            // Then
+            result.Should().NotBeNull();
+            result!.AverageDuration.Should().Be(75);
+        }
+
+        [Test]
+        public void Get_course_content_should_not_use_archived_tutorials_in_percentComplete()
+        {
+            // Given
+            const int candidateId = 11;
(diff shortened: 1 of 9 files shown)


DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
```diff
@@ -502,63 +502,146 @@ public void Get_course_content_should_only_return_sections_with_tutorials_in_cus
...
+                .GetCourseContent(candidateId, customisationId)?
+                .Sections
+                .Select(section => section.Id);
+
+            // Then
+            sectionIdsInCourseContent.Should().Contain(103);
+        }
+
+        [Test]
+        public void Get_course_content_uses_overrideTutorialMins_when_calculating_duration()
```
FridaTveit, line 571: There are more changes to these tests than I expected. Is that due to missing out cases like this the first time? Or have these tests been moved from somewhere else? 🙂

DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
```diff
@@ -502,63 +502,146 @@ public void Get_course_content_should_only_return_sections_with_tutorials_in_cus
...
+            const int candidateId = 11;
+            const int customisationId = 15937;
+
+            // When
+            var result = courseContentService.GetCourseContent(candidateId, customisationId);
+
+            // Then
+            result.Should().NotBeNull();
+            result!.Sections.First(section => section.Id == 392)
+                .PercentComplete.Should().Be(100.0 * 3 / 14);
```
FridaTveit, line 598: Quite a few magic numbers here. Could you extract them to some well named variables please? 🙂

DigitalLearningSolutions.Data.Tests/Services/DiagnosticAssessmentServiceTests.cs
```diff
@@ -398,7 +398,7 @@ public void Get_diagnostic_content_should_not_return_tutorials_where_diagAssess_
         }
 
         [Test]
-        public void Get_diagnostic_content_should_not_return_tutorials_where_archived_date_is_null()
+        public void Get_diagnostic_content_should_not_return_archived_tutorials()
```
FridaTveit, line 401: Might have small merge conflict with @kwunkiuma

DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
```diff
@@ -331,36 +344,41 @@ public void Get_section_content_should_order_by_tutorial_id_when_shared_order_by
...
-        [TestCase(254480, 24224, 101, 1, null)]
-        public void Get_section_content_should_have_same_tutorials_as_stored_procedure(
-            int candidateId,
-            int customisationId,
-            int centreId,
-            int sectionId,
-            int? progressId
-        )
+        [Test]
+        public void Get_section_content_should_sort_using_orderByNumber()
```
FridaTveit, line 348: This seems unrelated to the current change. Has this test just been moved?

DigitalLearningSolutions.Data/Services/SectionContentService.cs
```diff
@@ -82,9 +82,10 @@ GROUP BY
                     WHERE
                         CustomisationTutorials.CustomisationID = @customisationId
                         AND Sections.SectionID = @sectionId
-                        AND (Sections.ArchivedDate IS NULL)
+                        AND Sections.ArchivedDate IS NULL
```
FridaTveit, line 85: What about the id for the next section, are we checking that isn't archived?

DigitalLearningSolutions.Data/Services/TutorialContentService.cs
```diff
@@ -176,7 +178,8 @@ FROM CustomisationTutorials
                          AND Sections.SectionID = @sectionId
                          AND Tutorials.TutorialId = @tutorialId
                          AND Customisations.Active = 1
-                         AND CustomisationTutorials.Status = 1;",
+                         AND CustomisationTutorials.Status = 1
+                         AND Tutorials.ArchivedDate IS NULL;",
```
FridaTveit, line 182: Are we checking the section isn't archived?

FridaTveit (commented):
This contains changes to a lot of queries so is a bit hard to review. It would have been useful to split it up to avoid conflicts. Have you checked with @kwunkiuma and @SChandarana that there are no conflicts? A HLD would also have been useful 🙂

DigitalLearningSolutions.Data/Services/TutorialContentService.cs
```diff
@@ -54,6 +54,7 @@ FROM Tutorials
 
                          LEFT JOIN Tutorials AS NextTutorials
                          ON NextTutorials.SectionID = Tutorials.SectionID
+                            AND NextTutorials.ArchivedDate IS NULL
```
FridaTveit, line 57: No need to check archived date of the tutorial in the join below? Or of the section in the two joins on the sections table?

DigitalLearningSolutions.Data/Services/TutorialContentService.cs
```diff
@@ -147,6 +148,7 @@ FROM Tutorials
                      AND Tutorials.TutorialID = @tutorialId
                      AND Customisations.Active = 1
                      AND CustomisationTutorials.Status = 1
+                     AND Tutorials.ArchivedDate IS NULL
```
FridaTveit, line 151: What about if the section is achived?

DigitalLearningSolutions.Data/Services/TutorialContentService.cs
```diff
@@ -206,7 +209,8 @@ FROM CustomisationTutorials
                          AND Sections.SectionID = @sectionId
                          AND Tutorials.TutorialId = @tutorialId
                          AND Customisations.Active = 1
-                         AND CustomisationTutorials.Status = 1;",
+                         AND CustomisationTutorials.Status = 1
+                         AND Tutorials.ArchivedDate IS NULL;",
```
FridaTveit, line 213: What about if the section is archived?

FridaTveit (commented):
A few additional comments. I intentionally haven't been thinking about whether there are any more queries we should be checking/changing, because if that is the case then I think we should do that in a new PR after getting this merged. Will have a look through the queries tomorrow 🙂

benjimarshall:
@FridaTveit I totally agree that this should have been split up over a couple of PRs, each doing a few of the services. I still can do that (maybe after sorting this round of review comments) if that would be useful. 

The actual SQL changes are all fairly minor, but because we are now diverging from the old stored procedure's behaviour I thought it was best to extract the different cases that the "service is like old stored procedure" tests were covering into named tests that don't rely on the old stored procedures. I went through each of the cases for each of these tests and tried to work out which cases parts of the query they were testing to then put into tests, though I admit I got a bit carried away and there might have been a better way of doing this. Sorry for not checking with you! 🙂

DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
benjimarshall: This was one of the things being checked using the stored procedure `Get_section_content_should_have_same_tutorials_as_stored_procedure`, so I added this when I removed that test 🙂

DigitalLearningSolutions.Data/Services/SectionContentService.cs
benjimarshall: This line is just some brackets getting removed for consistency I think. The query isn't getting the next section ID here, because I haven't merged in Shiv's latest changes. I'll add those in soon 🙂

FridaTveit:
Yes, I know the SQL changes are small but making sure they're correct is still not straight forward. Similarly for the test changes. So for next time I would recommend splitting it up 🙂 

For right now, it depends how much work you think it would be to split it up. If it won't be too much work then I think it would be useful. It would make it easier to see how the test changes match up with SQL query changes, and means the more straight forward queries could get merged quickly. But if it's going to be difficult then it's okay to keep it as it is, just remember for next time 🙂

DigitalLearningSolutions.Data.Tests/Services/CourseContentServiceTests.cs
benjimarshall: Yep! Fixed in an upcoming PR 🙂

DigitalLearningSolutions.Data.Tests/Services/DiagnosticAssessmentServiceTests.cs
benjimarshall: This didn't turn out be too bad, but thanks for the heads up 🙂

DigitalLearningSolutions.Data/Services/TutorialContentService.cs
benjimarshall: No, added this change in #154 🙂

DigitalLearningSolutions.Data/Services/TutorialContentService.cs
benjimarshall: Yes we need to check the archived dates for that `NextSectionTutorial`. The first section join is the current section, so is checked with the change in the previous markup. The next section is now being checked, thank you for that! All done in #154 🙂

DigitalLearningSolutions.Data/Services/TutorialContentService.cs
benjimarshall: Yep, I hadn't thought of that! Fixed in #154 🙂

DigitalLearningSolutions.Data/Services/TutorialContentService.cs
benjimarshall: Yep, I hadn't thought of that! Fixed in #154 🙂

DigitalLearningSolutions.Data/Services/SectionContentService.cs
benjimarshall: Added in #157 🙂

benjimarshall:
All review comments have been resolved in subsequent PRs, and PRs #154, #156 and #157 make all of the changes proposed in this PR

PR #154, opened 2021-01-13
HEEDLS 294 Fix archived tutorials in course and tutorials services

benjimarshall:
### Changes
Ignore archived tutorials and sections in data services by adding filters to `WHERE` clauses in various queries. Service changed for this ticket:
* Tutorial content
* Course content
* Course completion

This includes changes from review comments in #152.

### Testing
Remove tests that depend on the old stored procedures, and replace them with new tests. Add tests to check correcting ignoring of archived sections and tutorials.

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Helpers/TutorialContentTestHelper.cs
+++ b/DigitalLearningSolutions.Data.Tests/Helpers/TutorialContentTestHelper.cs
@@ -1,0 +1,26 @@
+﻿namespace DigitalLearningSolutions.Data.Tests.Helpers
+{
+    using Dapper;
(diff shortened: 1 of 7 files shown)


FridaTveit (approved):
Looks good 🙂 Just a note for next time, if you were starting this from scratch now I would have recommended splitting this up even further. I would have at least split it up into one PR per service, if not one PR per service method. That's just because although the changes are small, the queries themselves are rather complex. So thinking through whether we're checking archived in the right places and testing all the relevant scenarios is tricky, and gets harder the more queries you're having to consider at the same time. I hope that makes sense 🙂

PR #157, opened 2021-01-13
HEEDLS-295 Ignore archived tutorials in section service

benjimarshall:
### Changes
Ignore archived sections in the tutorial list of the section page,
and account for this when looking for the next section. 

This includes changes from review comments in #152.

### Testing
Amend tests to not use incorrect old stored procedure, apply this
change for both the sections, and the next section as well.

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Helpers/SectionContentTestHelper.cs
+++ b/DigitalLearningSolutions.Data.Tests/Helpers/SectionContentTestHelper.cs
@@ -15,11 +15,6 @@
             this.connection = connection;
         }
 
-        public IEnumerable<OldTutorial> TutorialsFromOldStoredProcedure(int progressId, int sectionId)
-        {
-            return connection.Query<OldTutorial>("uspReturnProgressDetail_V3", new { progressId, sectionId }, commandType: CommandType.StoredProcedure);
-        }
-
         public void UpdateSectionNumber(int sectionId, int sectionNumber)
         {
             connection.Execute(
@@ -27,6 +22,16 @@
                         SET SectionNumber = @sectionNumber
                         WHERE SectionID = @sectionId",
                 new { sectionId, sectionNumber }
+            );
+        }
+
+        public void UpdateDiagAttempts(int tutorialId, int progressId, int diagAttempts)
+        {
+            connection.Execute(
+                @"UPDATE aspProgress
+                        SET DiagAttempts = @diagAttempts
+                        WHERE TutorialID = @tutorialId AND ProgressID = @progressId",
+                new { tutorialId, progressId, diagAttempts }
             );
         }
     }
--- a/DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
@@ -403,7 +403,7 @@
         }
 
         [Test]
-        public void Get_section_content_should_not_use_archived_tutorials_in_scores()
+        public void Get_section_content_should_not_use_archived_tutorials_in_maxScore()
         {
             // Given
             const int customisationId = 22416;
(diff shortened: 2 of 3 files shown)


DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
```diff
@@ -382,6 +402,62 @@ public void Get_section_content_returns_null_if_customisation_is_inactive()
...
+
+            // When
+            var result = sectionContentService.GetSectionContent(customisationId, candidateId, sectionId);
+
+            // Then
+            result.Should().NotBeNull();
+            result!.SectionScore.Should().Be(0);
+            result!.MaxSectionScore.Should().Be(2); // Not 3 as uspReturnSectionsForCandCust_V2 returns because
+                                                    // it counts archived tutorial 9366
+            result!.DiagnosticAttempts.Should().Be(1);
```
FridaTveit, line 421: Is this relevant to the test? 🙂

FridaTveit (commented):
One minor comment, otherwise looks good 🙂

DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
benjimarshall: Which bit? The comment isn't strictly necessary but I thought it was helpful to precisely locate the discrepancy for whoever uses this test in the future 🙂

FridaTveit: Sorry I meant checking the diagnostic attempts. It didn't seem relevant to what was being tested? I like the comment, I agree it's useful 🙂

benjimarshall: So the diagnostic attempts and the section score do both depend on whether archived tutorials are included or not (the section score being the sum of the tutorial scores, and the section diagnostic attempts being the maximum over the diagnostic attempt values for each tutorials in the section). As it happens both of these agree with the old stored procedure for this case, but they might not for every case. I think it's helpful to include because it checks the three values that including archived tutorials can effect, even if there only is a discrepancy in one value 🙂

benjimarshall: Alternatively I could temporarily edit the database to make all three of these values depend on whether we are including the archived tutorials or not? 🙂

FridaTveit: Ah I see. I think changing the db (in transaction scope) so they all depend on archived tutorials would make it a more useful test 🙂 I'd also be tempted to split it into two tests; one for the max score and one for the current score 🙂

benjimarshall: Yes I agree, added in 187795d 🙂

DigitalLearningSolutions.Data.Tests/Helpers/SectionContentTestHelper.cs
```diff
@@ -15,11 +15,6 @@ public SectionContentTestHelper(SqlConnection connection)
             this.connection = connection;
         }
 
-        public IEnumerable<OldTutorial> TutorialsFromOldStoredProcedure(int progressId, int sectionId)
```
FridaTveit, line 18: Are there other helper methods like this which we can remove, as I think we removed most if not all of the tests comparing our results with the old stored procedures now? 🙂

DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
```diff
@@ -415,10 +415,37 @@ public void Get_section_content_should_not_use_archived_tutorials_in_scores()
...
+                // This will should not change the overall DiagnosticAttempts because the tutorial has now been archived
+                sectionContentTestHelper.UpdateDiagAttempts(tutorialToArchive, progressId, 2);
+
+                // When
+                var result = sectionContentService.GetSectionContent(customisationId, candidateId, sectionId);
+
+                // Then
+                result.Should().NotBeNull();
+                result!.SectionScore.Should().Be(7);
+                result!.MaxSectionScore.Should().Be(10);
```
FridaTveit, line 446: I don't think we should be testing the max score here, as that's covered by the other test? 🙂

FridaTveit (commented):
A couple of comments otherwise looks good 🙂

DigitalLearningSolutions.Data.Tests/Services/SectionContentServiceTests.cs
benjimarshall: Yep! Removed in 0f6b296 🙂

FridaTveit (approved):
Looks good 🙂

PR #159, opened 2021-01-13
HEEDLS-282 Fix the tutorial page next button back end

benjimarshall:
### Changes
Get next section IDs when sections just have diagnostic assessments,
or just have post learning assessments (previously it was only displayed
if there was tutorial learning for that next section).

### Testing
Add some unit tests to cover these cases.

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Services/TutorialContentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/TutorialContentServiceTests.cs
@@ -243,14 +243,14 @@
             const int sectionId = 150;
             const int tutorialId = 634;
 
-            const int nextSectionId = 151; // All tutorials are CustomisationTutorials.Status = 0, though some DiagStatus = 1
+            const int expectedNextSectionId = 151; // All tutorials are CustomisationTutorials.Status = 0, though some DiagStatus = 1
 
             // When
             var tutorial = tutorialContentService.GetTutorialInformation(candidateId, customisationId, sectionId, tutorialId);
 
             // Then
             tutorial.Should().NotBeNull();
-            tutorial!.NextSectionId.Should().Be(nextSectionId);
+            tutorial!.NextSectionId.Should().Be(expectedNextSectionId);
         }
 
         [Test]
@@ -262,8 +262,8 @@
             const int sectionId = 104;
             const int tutorialId = 331;
 
-            const int expectedNextSectionId = 105;
-
+            const int expectedNextSectionId = 105; // All tutorials are CustomisationTutorials.Status and DiagStatus = 0
+                                                   // Customisations.IsAssessed = 1 and Sections.PLAssessPath is not null
             // When
             var tutorial = tutorialContentService.GetTutorialInformation(candidateId, customisationId, sectionId, tutorialId);
 
@@ -272,7 +272,7 @@
         }
 
         [Test]
-        public void Get_tutorial_information_nextSection_skips_section_with_only_post_learning_assessment_but_no_path()
+        public void Get_tutorial_information_nextSection_skips_assessed_section_with_no_assessment_path()
         {
             using (new TransactionScope())
             {
@@ -282,7 +282,8 @@
                 const int sectionId = 104;
                 const int tutorialId = 331;
 
-                const int originalNextSectionId = 105;
+                const int originalNextSectionId = 105; // All tutorials are CustomisationTutorials.Status and DiagStatus = 0
+                                                       // Customisations.IsAssessed = 1
                 tutorialContentTestHelper.UpdatePLAssessPath(originalNextSectionId, null);
                 const int expectedNextSectionId = 106;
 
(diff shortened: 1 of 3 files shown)


DigitalLearningSolutions.Data.Tests/Services/TutorialContentServiceTests.cs
```diff
@@ -234,6 +234,66 @@ public void Get_tutorial_information_nextSection_skips_sections_full_of_archived
...
+        [Test]
+        public void Get_tutorial_information_nextSection_returns_section_with_only_diagnostic_assessment()
+        {
+            // Given
+            const int candidateId = 74411;
+            const int customisationId = 5852;
+            const int sectionId = 150;
+            const int tutorialId = 634;
+
+            const int nextSectionId = 151; // All tutorials are CustomisationTutorials.Status = 0, though some DiagStatus = 1
```
FridaTveit, line 246: Very minor but I would name this `expectedNextSectionId` 🙂

DigitalLearningSolutions.Data.Tests/Services/TutorialContentServiceTests.cs
```diff
@@ -234,6 +234,66 @@ public void Get_tutorial_information_nextSection_skips_sections_full_of_archived
...
+        [Test]
+        public void Get_tutorial_information_nextSection_returns_section_with_only_post_learning_assessment()
+        {
+            // Given
+            const int customisationId = 10820;
+            const int candidateId = 1;
+            const int sectionId = 104;
+            const int tutorialId = 331;
+
+            const int expectedNextSectionId = 105;
```
FridaTveit, line 265: Similar comment to the one in the previous test would be useful I think? In general I prefer making code more clear over adding comments, but I think our data tests are one case where we could probably be more generous with comments. As many of the rely on knowing the state of the db, and knowing how rather complicated bits of the queries work together which makes it harder to understand the tests. Comments like the one you added above make that a lot easier 🙂

DigitalLearningSolutions.Data.Tests/Services/TutorialContentServiceTests.cs
```diff
@@ -234,6 +234,66 @@ public void Get_tutorial_information_nextSection_skips_sections_full_of_archived
...
+
+            // When
+            var tutorial = tutorialContentService.GetTutorialInformation(candidateId, customisationId, sectionId, tutorialId);
+
+            //Then
+            tutorial.NextSectionId.Should().Be(expectedNextSectionId);
+        }
+
+        [Test]
+        public void Get_tutorial_information_nextSection_skips_section_with_only_post_learning_assessment_but_no_path()
```
FridaTveit, line 275: Might be more accurate to say it's assessed but has no path? Because it doesn't really have a post learning assessment if there's no path 🙂

DigitalLearningSolutions.Data/Services/TutorialContentService.cs
```diff
@@ -62,8 +65,16 @@ FROM Tutorials
                             )
                             AND NextCustomisationTutorials.TutorialID = NextTutorials.TutorialID
 
+                         LEFT JOIN CustomisationTutorials AS NextSectionCustomisationTutorials
```
FridaTveit, line 68: Why do we need a new join on CustomisationTutorials instead of a adding to the one we have? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data/Services/TutorialContentService.cs
benjimarshall: The previous join only selected `Status = 1` tutorials; tutorials whose pages can be visited for the `Next tutorial` button.

This join used to find the next section has a slightly different purpose: to find sections that either have tutorials that can be viewed (`Status = 1`), have a diagnostic assessment (`DiagStatus = 1`) or have a post learning assessment (`IsAssessed = 1`). In each of these cases the section page can be visited (and if none are true then they can be skipped). I don't think it is possible to cover all of these cases with just one join, but I might be missing something 🙂

DigitalLearningSolutions.Data.Tests/Services/TutorialContentServiceTests.cs
benjimarshall: Yep, sorry! Fixed in 08afd86 🙂

DigitalLearningSolutions.Data/Services/TutorialContentService.cs
FridaTveit: Okay, fair enough 🙂

FridaTveit (approved):
Looks good 🙂

DigitalLearningSolutions.Data.Tests/Services/TutorialContentServiceTests.cs
benjimarshall: Yeah absolutely, added in 08afd86 🙂

DigitalLearningSolutions.Data.Tests/Services/TutorialContentServiceTests.cs
benjimarshall: Thanks, that's much better! Changed in 08afd86 🙂

PR #161, opened 2021-01-13
HEELDS-297 Post learning assessment launch view models

kwunkiuma:
Implement view model and tests.

### DIFF
--- a/DigitalLearningSolutions.Web/ViewModels/LearningMenu/DiagnosticContentViewModel.cs
+++ b/DigitalLearningSolutions.Web/ViewModels/LearningMenu/DiagnosticContentViewModel.cs
@@ -12,7 +12,8 @@
         public string CourseTitle { get; }
         public string SectionName { get; }
         public string ContentSource { get; }
-        public List<int> Tutorials { get; }
+        private const string type = "diag";
+        private List<int> Tutorials { get; }
         private int CentreId { get; }
         private int CandidateId { get; }
         private int ProgressId { get; }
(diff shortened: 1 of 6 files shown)


DigitalLearningSolutions.Web/ViewModels/LearningMenu/PostLearningContentViewModel.cs
```diff
@@ -0,0 +1,75 @@
...
+            PostLearningContent postLearningContent)
+        {
+            return $"{config.GetScormPlayerUrl()}" +
+                   $"?CentreID={CentreId}" +
+                   $"&CustomisationID={CustomisationId}" +
+                   $"&CandidateID={CandidateId}" +
+                   $"&SectionID={SectionId}" +
+                   $"&Version={postLearningContent.Version}" +
+                   $"&tutpath={postLearningContent.PostLearningAssessmentPath}" +
+                   "&type=pl";
```
FridaTveit, line 72: Aren't these two very similar for diagnostic and post learning assessments? If so, worth extracting so the code can be shared? 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/PostLearningContentViewModel.cs
```diff
@@ -0,0 +1,75 @@
...
+    using Microsoft.Extensions.Configuration;
+
+    public class PostLearningContentViewModel
+    {
+        public int CustomisationId { get; }
+        public int SectionId { get; }
+        public string CourseTitle { get; }
+        public string SectionName { get; }
+        public string ContentSource { get; }
+        public List<int> Tutorials { get; }
```
FridaTveit, line 15: Does this need to be public?

FridaTveit (commented):

DigitalLearningSolutions.Web/ViewModels/LearningMenu/PostLearningContentViewModel.cs
kwunkiuma: If we extract it there would still be a really big constructor in each class. If that's still an improvement then I'm happy to do so

FridaTveit: That's a good point. It would mainly be to ensure the two are consistent and the code to construct the url isn't duplicated. It's only two statements (one for scorm, one for non-scorm), but each is pretty complicated. I don't feel very strongly about it though, happy for you to keep it here if you prefer 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/PostLearningContentViewModel.cs
kwunkiuma: Looks like this was used in a test. Since we're indirectly testing the tutorials when we test HTML assessments, I could remove the specific tutorial list test?

FridaTveit: Yes please. Tests should only be testing the public part of a class 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/DiagnosticContentViewModel.cs
```diff
@@ -12,7 +12,8 @@ public class DiagnosticContentViewModel
         public string CourseTitle { get; }
         public string SectionName { get; }
         public string ContentSource { get; }
-        public List<int> Tutorials { get; }
+        private const string type = "diag";
+        private List<int> Tutorials { get; }
         private int CentreId { get; }
         private int CandidateId { get; }
         private int ProgressId { get; }
```
FridaTveit, line 19: Actually you don't need any of these private variables do you? You can just pass them to the methods which need them in the constructor. No need to store something in the class if you're just going to use it in the constructor 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web/ViewModels/LearningMenu/DiagnosticContentViewModel.cs
kwunkiuma: In that case should I remove those two methods in the class and just call the helper methods instead?

FridaTveit: You could do 🙂

FridaTveit (approved):
Looks good 🙂

PR #162, opened 2021-01-13
HEEDLS-298 Add tutorial customisation back end

benjimarshall:
### Changes
* Add a "Supporting information" field to the course settings object
* Get course settings when getting tutorial information

### Testing
Add some more unit tests to the course settings and tutorial service,
tweak some existing ones to make constructor signatures match.

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Models/CourseSettingsTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Models/CourseSettingsTests.cs
@@ -190,5 +190,34 @@
             // Then
             courseSettings.ConsolidationExercise.Should().BeNull();
         }
+        [Test]
+        public void CourseSettings_should_take_SupportingInformation_from_json()
+        {
+            // When
+            var courseSettings = new CourseSettings("{\"lm:si\":\"Supporting information\"}");
+
+            // Then
+            courseSettings.SupportingInformation.Should().Be("Supporting information");
+        }
+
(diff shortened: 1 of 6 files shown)


DigitalLearningSolutions.Data.Tests/Models/CourseSettingsTests.cs
```diff
@@ -190,5 +190,34 @@ public void CourseSettings_should_have_default_ConsolidationExercise_when_wrong_
...
+        {
+            // When
+            var courseSettings = new CourseSettings("{\"lm:si\":\"Supporting information\"}");
+
+            // Then
+            courseSettings.SupportingInformation.Should().Be("Supporting information");
+        }
+
+        [Test]
+        public void CourseSettings_should_have_default_SupportingInformation_when_not_in_json()
```
FridaTveit, line 204: One thing I hadn't considered before, for string values in the course setting we should probably not use them if they're empty 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data.Tests/Models/CourseSettingsTests.cs
benjimarshall: Yeah neither had I! Fixed in 72ce2a9 🙂

FridaTveit (approved):
Looks good 🙂

PR #165, opened 2021-01-14
HEEDLS-289 - Section Page Fix Diagnostic Assessment Scores

SChandarana:
Some of these tests will also need to be done on the Diagnostic Assessment tests which I will do in a separate PR

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Helpers/SectionContentTestHelper.cs
+++ b/DigitalLearningSolutions.Data.Tests/Helpers/SectionContentTestHelper.cs
@@ -25,7 +25,7 @@
             );
         }
 
-        public void UpdateDiagAttempts(int tutorialId, int progressId, int diagAttempts)
+        public void UpdateDiagnosticAttempts(int tutorialId, int progressId, int diagAttempts)
         {
             connection.Execute(
                 @"UPDATE aspProgress
@@ -35,7 +35,7 @@
             );
         }
 
-        public void UpdateDiagScore(int tutorialId, int progressId, int score)
+        public void UpdateDiagnosticScore(int tutorialId, int progressId, int score)
         {
             connection.Execute(
                 @"UPDATE aspProgress
--- a/DigitalLearningSolutions.Data.Tests/Helpers/SectionContentTestHelper.cs
+++ b/DigitalLearningSolutions.Data.Tests/Helpers/SectionContentTestHelper.cs
@@ -34,5 +34,15 @@
                 new { tutorialId, progressId, diagAttempts }
             );
         }
+
+        public void UpdateDiagScore(int tutorialId, int progressId, int score)
+        {
+            connection.Execute(
+                @"UPDATE aspProgress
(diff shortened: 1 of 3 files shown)


DigitalLearningSolutions.Data.Tests/Helpers/SectionContentTestHelper.cs
```diff
@@ -34,5 +34,15 @@ public void UpdateDiagAttempts(int tutorialId, int progressId, int diagAttempts)
                 new { tutorialId, progressId, diagAttempts }
             );
         }
+
+        public void UpdateDiagScore(int tutorialId, int progressId, int score)
```
FridaTveit, line 38: Please avoid abbreviations 🙂

DigitalLearningSolutions.Data/Services/SectionContentService.cs
```diff
@@ -128,8 +128,14 @@ GROUP BY
                     if (sectionContent == null)
                     {
                         sectionContent = section;
+                        if (!section.DiagnosticStatus)
```
FridaTveit, line 131: Hm, I'd prefer to avoid this nested if statement. I think I'd prefer using ternary operators:
```
sectionContent.DiagnosticAttempts = section.DiagnosticStatus ? sectionContent.DiagnosticAttempts : 0;
...
```
What do you think? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data/Services/SectionContentService.cs
SChandarana: Yeah when I wrote this I was trying to think of a better way to do it. I thought that ternary operators would be too much repetition of the same thing and it felt less clear that the 3 statements were linked by the same condition. I'm happy to change to ternary operators if you think that's better?

FridaTveit: Yes I think this is a case where neither is ideal and it's more a case if personal preference. I think I'd find the ternary operators slightly more readable 🙂

DigitalLearningSolutions.Data.Tests/Helpers/SectionContentTestHelper.cs
SChandarana: The other method in this helper is titled "UpdateDiagAttempts" as it is updating the diagAttempts column. Woud you prefer I change this to "UpdateDiagLast" for consistency or "UpdateDiagnosticScore" for readability in the test? (I can also change "UpdateDiagAttempts" to "UpdateDiagnosticAttempts" if that's preferred too)

FridaTveit: Change to `UpdateDiagnosticScore` and `UpdateDiagnosticAttempts` please. I think we should try to avoid referring to the names in the db and they can often be confusing 🙂

FridaTveit (approved):
Looks good 🙂

PR #170, opened 2021-01-14
HEEDLS-290 - Section Page Fix Diag Status

SChandarana:


DigitalLearningSolutions.Data/Services/SectionContentService.cs
```diff
@@ -134,6 +134,7 @@ GROUP BY
                     }
                     else if (section.DiagnosticStatus)
                     {
+                        sectionContent.DiagnosticStatus = section.DiagnosticStatus;
```
FridaTveit, line 137: I don't think this will work. This way the diagnostic status will reflect the diagnstic status of the last tutorial in the section which is not what you want. In addition to fixing this we should include a test which would fail with this implementation (i.e. one where the last diag status is 0 but at least one other tutorial has diag status 1) 🙂

FridaTveit (commented):
Smal tweak needed 🙂

DigitalLearningSolutions.Data/Services/SectionContentService.cs
SChandarana: That test case does currently exist, I've tested where all statuses are 0 except the second tutorial (out of 4 tutorials). As this is an `else if (section.DiagnosticStatus)` it means that if the diagnostic status is being overwritten, it's only being overwritten with "true"

FridaTveit: Ah I see! I hadn't considered that the `if` was checking the same variable. That seems fine then 🙂

FridaTveit (approved):
Looks good 🙂

PR #172, opened 2021-01-15
HEEDLS-NONE Remove interface access modifiers

kwunkiuma:
Remove `public` from some interface methods.


FridaTveit (approved):
Looks good, thanks! 🙂

PR #175, opened 2021-01-15
HEEDLS-301 Extract completion summary card into partial

benjimarshall:
## Changes
Refactor the completion summary card out of the initial menu view and
view model, into its own partial view.

## Testing
Rearrange some tests to reflect the refactoring. Tested in Chrome, UI should appear the same as before the PR.

### Course with a completion summary
course_with_completion_summary
### Course without a completion summary
course_without_completion_summary

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/CompletionSummaryCardViewModelTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/CompletionSummaryCardViewModelTests.cs
@@ -1,0 +1,199 @@
+﻿namespace DigitalLearningSolutions.Web.Tests.ViewModels.LearningMenu
+{
+    using System;
+    using DigitalLearningSolutions.Web.ViewModels.LearningMenu;
+    using FluentAssertions;
+    using NUnit.Framework;
+
+    class CompletionSummaryCardViewModelTests
+    {
+        private const int CustomisationId = 1;
+        private readonly DateTime? Completed = null;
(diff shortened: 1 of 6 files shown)


DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/CompletionSummaryCardViewModelTests.cs
```diff
@@ -0,0 +1,199 @@
...
+{
+    using System;
+    using DigitalLearningSolutions.Web.ViewModels.LearningMenu;
+    using FluentAssertions;
+    using NUnit.Framework;
+
+    class CompletionSummaryCardViewModelTests
+    {
+        private const int CustomisationId = 1;
+        private readonly DateTime? Completed = null;
```
FridaTveit, line 11: Is it really useful to have this as a class variable? I'd only use class variables for tests if they:
* Will be used for every test, or nearly every test
* Need to be defined as variables. For example random numbers are useful to define as variables because otherwise it's unclear what they represent. But `null`, `true` or `false` are often better to inline 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/CompletionSummaryCardViewModelTests.cs
benjimarshall: Ah thank you for clearing that up! Sorted in 9948a2d 🙂

FridaTveit (approved):
Looks good 🙂

PR #176, opened 2021-01-18
HEEDLS-275 - Use Floor Instead Of Round When Calculatating Percentages

SChandarana:


DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/SectionContentViewModelTests.cs
```diff
@@ -648,5 +648,28 @@ public void Should_have_null_next_section_id_if_there_is_no_next_section()
             // Then
             sectionContentViewModel.NextSectionId.Should().BeNull();
         }
+
+        [Test]
+        public void Percent_complete_should_correctly_be_floored()
```
FridaTveit, line 653: There are already tests about this further up in this file I believe. This test should be grouped with them 🙂

FridaTveit (commented):

FridaTveit (approved):
Looks good 🙂

PR #177, opened 2021-01-18
HEEDLS-302 Add single section front end

benjimarshall:
## Changes 
* Amend initial menu action to redirect to the section page if the course has only one section.
* Add completion data to section view model, to be shown if it is the only section in the course, and to flag to record whether this is the only section or not. Also hide the next button in this case.

## Testing
Amend tests to reflect these changes to the initial menu action, and these additions to the section content view model. 

Tested in Chrome, IE11, using a screenreader. The tags are a bit weird with a screenreader in IE11 (the opening and closing `<strong>` tags get read), but this problem is present on the NHS design system website as well.

### Single section page, with a completion summary
one_section_with_completion
### Single section page, without a completion summary
one_section_without_completion
### One section among others, with a completion summary available (but not displayed because that is shown on the initial menu)
one_section_of_multiple_with_completion

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
@@ -99,6 +99,58 @@
         }
 
         [Test]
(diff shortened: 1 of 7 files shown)


FridaTveit (approved):
Looks good 🙂

PR #178, opened 2021-01-18
HEEDLS-307-Section Page Show DIagnostic Recommendation Status Frontend

SChandarana:
Regular view
image

Mobile view
image

Had to use float: right as the normal methods had some problems with text overflowing during resizing. I checked with kevin and david and they were happy with how it looked

### DIFF
--- a/DigitalLearningSolutions.Web/Views/LearningMenu/Section/_DiagnosticAssessmentCard.cshtml
+++ b/DigitalLearningSolutions.Web/Views/LearningMenu/Section/_DiagnosticAssessmentCard.cshtml
@@ -9,7 +9,7 @@
         </h3>
       </div>
       <div class="nhsuk-grid-column-one-quarter">
-        <h3 class="nhsuk-heading-xs nhsuk-u-secondary-text-color">
+        <h3 class="nhsuk-heading-xs nhsuk-u-secondary-text-color float-right-additional-information">
           @Model.DiagnosticCompletionStatus
         </h3>
       </div>
--- a/DigitalLearningSolutions.Web/Views/LearningMenu/Section/_TutorialCard.cshtml
+++ b/DigitalLearningSolutions.Web/Views/LearningMenu/Section/_TutorialCard.cshtml
@@ -3,15 +3,15 @@
 <div class="nhsuk-card nhsuk-u-margin-bottom-3 learning-menu-card">
   <div class="nhsuk-card__content">
     <div class="nhsuk-grid-row">
-      <div class="nhsuk-grid-column-three-quarters">
-        <h3 class="nhsuk-card__heading completion-card-heading" id="@Model.Id-name">
+      <div class="nhsuk-grid-column-two-thirds">
+        <h3 class="nhsuk-card__heading card-with-status-heading" id="@Model.Id-name">
           @Model.TutorialName
         </h3>
         <div class="nhsuk-u-margin-bottom-4">
           <partial name="Shared/_TutorialTimeSummary" model="@Model.TimeSummary"/>
         </div>
       </div>
-      <div class="nhsuk-grid-column-one-quarter">
+      <div class="nhsuk-grid-column-one-third">
         <strong class="status-tag nhsuk-tag--orange float-right-additional-information">
           Recommended
         </strong>
--- a/DigitalLearningSolutions.Web/Views/LearningMenu/Section/_TutorialCard.cshtml
+++ b/DigitalLearningSolutions.Web/Views/LearningMenu/Section/_TutorialCard.cshtml
@@ -15,13 +15,14 @@
         <strong class="status-tag nhsuk-tag--orange float-right-additional-information">
           Recommended
         </strong>
-        @if (Model.ShowLearnStatus) {
+        @if (Model.ShowLearnStatus)
+        {
           <h3 class="nhsuk-heading-xs nhsuk-u-secondary-text-color float-right-additional-information">
             @Model.CompletionStatus
           </h3>
         }
       </div>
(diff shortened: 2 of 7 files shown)


DigitalLearningSolutions.Web/Views/LearningMenu/Section/_TutorialCard.cshtml
```diff
@@ -3,22 +3,25 @@
 <div class="nhsuk-card nhsuk-u-margin-bottom-3 learning-menu-card">
   <div class="nhsuk-card__content">
     <div class="nhsuk-grid-row">
-      <div class="@(Model.ShowLearnStatus ? "nhsuk-grid-column-three-quarters" : "nhsuk-grid-column-full")">
```
FridaTveit, line 6: When you integrate with the backend you probably need to change this again as we won't always have a recommended/optional tag to show. But fine to leave like this for now 🙂

DigitalLearningSolutions.Web/Views/LearningMenu/Section/_TutorialCard.cshtml
```diff
@@ -3,22 +3,25 @@
...
-      {
-        <div class="nhsuk-grid-column-one-quarter">
-          <h3 class="nhsuk-heading-xs nhsuk-u-secondary-text-color">
+      <div class="nhsuk-grid-column-one-quarter">
+        <strong class="status-tag nhsuk-tag--orange float-right-additional-information">
+          Recommended
+        </strong>
+        @if (Model.ShowLearnStatus)
+        {
+          <h3 class="nhsuk-heading-xs nhsuk-u-secondary-text-color float-right-additional-information">
```
FridaTveit, line 20: Why do we need the float right styling here if we didn't need to for the completion summary card? 🙂

FridaTveit (commented):
A few comments, otherwise looks good. Might also be useful for @benjimarshall to have a look as he did the styling for the completion summary card (if I remember correctly) 🙂

DigitalLearningSolutions.Web/Views/LearningMenu/Section/_DiagnosticAssessmentCard.cshtml
```diff
@@ -9,7 +9,7 @@
         </h3>
       </div>
       <div class="nhsuk-grid-column-one-quarter">
-        <h3 class="nhsuk-heading-xs nhsuk-u-secondary-text-color">
+        <h3 class="nhsuk-heading-xs nhsuk-u-secondary-text-color float-right-additional-information">
```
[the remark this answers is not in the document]
SChandarana: As the tutorial cards are changing to float right, it makes the "not started" etc align slightly differently to these cards so to keep them all consistent they all had to float right

DigitalLearningSolutions.Web/Views/LearningMenu/Section/_TutorialCard.cshtml
```diff
@@ -3,22 +3,25 @@
 <div class="nhsuk-card nhsuk-u-margin-bottom-3 learning-menu-card">
   <div class="nhsuk-card__content">
     <div class="nhsuk-grid-row">
-      <div class="@(Model.ShowLearnStatus ? "nhsuk-grid-column-three-quarters" : "nhsuk-grid-column-full")">
-        <h3 class="nhsuk-card__heading" id="@Model.Id-name">
+      <div class="nhsuk-grid-column-three-quarters">
+        <h3 class="nhsuk-card__heading completion-card-heading" id="@Model.Id-name">
```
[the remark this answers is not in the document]
SChandarana: This was something that benji wrote for the completion card and i needed to use the same styling. He was at the time working on extracting the partial so if I made some changes they would cause a lot of conflicts. I should be able to merge learning-menu-master into this now and make those changes as he is now done with that

DigitalLearningSolutions.Web/Views/LearningMenu/Section/_DiagnosticAssessmentCard.cshtml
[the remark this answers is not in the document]
FridaTveit: Ah I see, that makes sense. Which browsers have you tried it in? I know float right can be a bit of a pain in different browsers so probably worth giving it a quick check in a few different ones 🙂

DigitalLearningSolutions.Web/Views/LearningMenu/Section/_TutorialCard.cshtml
SChandarana: The reason is because the word "Recommended" is slightly too long and it means that during a resize, the tag overflows over the edge of the card 5-10 pixels before the responsive styling thing kicks in to put it on a new line

FridaTveit: Sorry, I'm not sure I understand how floating right fixes "Recommended" being too long? Based on your other comment I was assuming you were using float right since we've now got two fields on the right and it looks better if they're aligned with the right edge rather than the start of the column? 🙂

SChandarana: Yeah I wasn't too sure either. I initially did the float right thing to fix alignment, then whilst testing out some non-float right solutions (as I realised that using float right meant I'd need to edit all the other cards and also cause complications with the completion summary card) I noticed the problem with "Recommended" being too long only happened during the non-float right solutions and float right somehow fixed it

FridaTveit (approved):
Looks good 🙂

PR #179, opened 2021-01-19
HEEDLS-312 Add single tutorial backend

benjimarshall:
### Changes
* Add single tutorial in section field, and single section in course field to tutorial information. This uses a similar approach to the single section checker in the section content query, making a list of other tutorials and then from this extracting different bits of information:
    * Whether there are other tutorials in the section
    * Whether there other sections in the course 
    * The next tutorial ID
    * The next section ID
* Add completion information to tutorial information.

### Testing
Amend existing tests and add more unit tests for these additional fields.

PR #180, opened 2021-01-19
HEEDLS-310 Diagnostic assessment next button back end

kwunkiuma:
Modify query for retrieving diagnostic assessment and add tests.


kwunkiuma:
Just noticed I also need to fetch if the section has a post learning assessment, will add that now.

DigitalLearningSolutions.Data.Tests/Services/DiagnosticAssessmentDataServiceTests.cs
```diff
@@ -316,6 +326,325 @@ public void Get_diagnostic_assessment_should_get_tutorialIds_ordered_by_orderBy(
...
+
+                // When
+                var result = diagnosticAssessmentDataService.GetDiagnosticAssessment(customisationId, candidateId, sectionId);
+
+                // Then
+                result.Should().NotBeNull();
+                result!.NextTutorialId.Should().Be(nextTutorialId);
+            }
+        }
+
```
FridaTveit, line 647: This file is getting quite long, might be worth splitting it up, e.g. by moving all the tests for next tutorial/section into their own test file. You can use a partial class so you can still share the setup, like the tests for the learning portal controller 🙂

DigitalLearningSolutions.Data/Services/DiagnosticAssessmentDataService.cs
```diff
@@ -29,6 +29,60 @@ public DiagnosticAssessmentDataService(IDbConnection connection, ILogger<Diagnos
...
+                                ON NextSections.SectionID = NextSectionTutorials.SectionID
+                                AND NextSections.ArchivedDate IS NULL
+                                AND (
+                                    Sections.SectionNumber < NextSections.SectionNumber
+                                    OR Sections.SectionID < NextSections.SectionID
+                                )
+                                AND (
+                                    NextSectionCustomisationTutorials.Status = 1
+                                    OR NextSectionCustomisationTutorials.DiagStatus = 1
+                                    OR NextSections.PLAssessPath IS NOT NULL
```
FridaTveit, line 57: For it to have a diagnostic assessment it also needs a non-null diag path. For it to have a post learning assessment it also needs to be assessed 🙂

FridaTveit (commented):
A few comments, otherwise looks good. Probably also worth @benjimarshall having a look as he worked on the similar next query for tutorials 🙂

DigitalLearningSolutions.Data/Services/DiagnosticAssessmentDataService.cs
kwunkiuma: These conditions are based on the `GetTutorialInformation` query and are only being used to filter out invalid tutorials.
The diagnostic assessment checks are are being done in the main query below, and the post learning checks will be done in the view model.

FridaTveit: Sorry how is it meant to filter out invalid tutorials? Post learning assessment path doesn't have anything to do with invalid tutorials as far as I can tell? Do you mean filter out invalid sections? 

You're right that it's also used in the tutorial query so @benjimarshall can maybe explain why those checks aren't needed? 🙂

benjimarshall: Yeah I was beginning to respond to this, and this was an issue that I recently spotted in the tutorial content service (which has been fixed on my rewrite in #179), sorry for not passing it on to you sooner. Line 56 needs to be:
`OR (NextSectionCustomisationTutorials.DiagStatus = 1 AND NextSections.DiagAssessPath IS NOT NULL`
because for a section with just a diagnostic assessment that path needs to be present, and there needs to be a `DiagStatus=1` tutorial. If one of these were not true (and there wasn't any section to be visited, nor any post learning assessment) then this section should be skipped. Does that make sense? 🙂

kwunkiuma: Ah my bad, I was somewhat confused because I thought this was the part of the query dealing with the next tutorial. I'll make the necessary changes now.

benjimarshall:
I've had a look through this and I think it looks good, though will have another read through tomorrow. One thing I've just thought of is that I think this is going to need rewriting when I do HEEDLS-269, in much the same way that the other next tutorial/section queries have been rewritten recently, though I don't know if that changes anything for this ticket

FridaTveit (approved):
I'm happy if Benji's happy after he's had another look tomorrow 🙂

DigitalLearningSolutions.Data/Services/DiagnosticAssessmentDataService.cs
```diff
@@ -29,6 +29,61 @@ public DiagnosticAssessmentDataService(IDbConnection connection, ILogger<Diagnos
...
+                            LEFT JOIN Tutorials AS NextSectionTutorials
+                                ON NextSectionTutorials.TutorialID = NextSectionCustomisationTutorials.TutorialID
+                                AND NextSectionTutorials.ArchivedDate IS NULL
+                            LEFT JOIN Sections AS NextSections
+                                ON NextSections.SectionID = NextSectionTutorials.SectionID
+                                AND NextSections.ArchivedDate IS NULL
+                                AND (
+                                    Sections.SectionNumber < NextSections.SectionNumber
+                                    OR Sections.SectionID < NextSections.SectionID
+                                )
```
benjimarshall, line 53: We also need to check that the next section number is `<=` the current section, else this will give us any sections with a smaller section number if they happen to have a larger section ID. In no case should we get a next section that has a smaller section number to the current one, it should either have a larger section number (and any ID it likes) or have the same section number, and a larger ID 🙂

benjimarshall (changes requested):
One straightforward thing to change, but otherwise it looks good to me 👍

benjimarshall (approved):
Perfect, looks good!

PR #182, opened 2021-01-19
HEEDLS-309 Implement fullscreen on tutorial content viewer

FridaTveit:
Added support for moving to and from fullscreen on the tutorial content viewer page. It will only work with js enabled.

Tested on Chrome, Firefox and IE11. Tested with voice over. Tested on mobile size screen.

Adding it to the assessment pages should be pretty simple, just calling the fullscreen setup function and adding the buttons in the cshtml files.

image

image


kwunkiuma (approved):
Looks good to me 🙂

benjimarshall (approved):
Also looks good to me 👍

PR #183, opened 2021-01-20
HEEDLS-313 Add single tutorial frontend

benjimarshall:
## Changes
* Amend section action to redirect to the tutorial page if the course has only one tutorial.
* Add completion data to tutorial view model, to be shown if it is the only tutorial in the course, and flags to record whether this is the only section, and only item in the section. Also hide the next button in this case.

## Testing
Amend tests to reflect these changes to the section action, and these
additions to the tutorial view model.

Tested in Chrome, IE11, using a screenreader. The tags are a bit weird with a screenreader in IE11 (the opening and closing <strong> tags get read), but this problem is present on the NHS design system website as well.

### Single tutorial in section without completion summary
single_tutorial_without_consolidation
### Single tutorial in section with a completion summary
single_tutorial_with_consolidation
### Section page with single tutorial and consolidation exercise
section_with_just_consolidation_exercise
### Section page with single tutorial and post learning assessment
single_tutorial_but_assessments


DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
```diff
@@ -430,6 +430,300 @@ public void Sections_should_render_view()
                 .Model.Should().BeEquivalentTo(expectedModel);
         }
 
+        [Test]
+        public void Sections_should_redirect_to_section_page_if_one_tutorial_in_section()
```
FridaTveit, line 434: Should redirect to tutorial page? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
benjimarshall: Yes, sorry those were named wrong! Fixed in 0a65197 👍

FridaTveit (approved):
Looks good 🙂

PR #185, opened 2021-01-20
HEEDLS-NONE Remove some redundant test helper methods

benjimarshall:
## Changes
As promised in the previous PR: rearrange the helper functions in the various data test helpers so each helper relates to a specific concept (tutorial, course, etc.) and remove a redundant helper which had two implementations.

## Testing
All tests pass


DigitalLearningSolutions.Data.Tests/Helpers/TutorialContentTestHelper.cs
```diff
@@ -65,5 +95,14 @@ public void UpdateTutorialStatus(int tutorialId, int customisationId, int status
                 new { tutorialId, customisationId, status }
             );
         }
```
FridaTveit, line 97: Empty line between functions please. And might be worth moving this up to the other diagnostic functions? 🙂

FridaTveit (commented):
One small thing, otherwise looks good 🙂

DigitalLearningSolutions.Data.Tests/Helpers/TutorialContentTestHelper.cs
benjimarshall: Yep sorry! Rearranged in 60d7021 🙂

FridaTveit (approved):
Looks good 🙂

PR #186, opened 2021-01-20
HEEDLS-307 - Section Page Show Diagnostic Recommendation Status Fix

SChandarana:
Fixing a bug where the completion status and recommendation tag were on the same line if recommendation status was "Optional"

Normal view:

image

Mobile view:

image


FridaTveit (approved):
Looks good 🙂

PR #188, opened 2021-01-20
HEEDLS-304 - Customisation Fix Typo In Keys

SChandarana:


DigitalLearningSolutions.Data/Models/CourseSettings.cs
```diff
@@ -38,22 +38,22 @@ public CourseSettings(string? settingsText)
...
 
-                if (settings.ContainsKey("lm.sl") && settings["lm.sl"] is bool)
+                if (settings.ContainsKey("lm.sl") && settings["lm.sl"] is bool showLearnStatus)
                 {
-                    ShowLearnStatus = (bool)settings["lm.sl"];
+                    ShowLearnStatus = showLearnStatus;
                 }
 
-                if (settings.ContainsKey("lm:ce") && settings["lm:ce"] is string)
+                if (settings.ContainsKey("lm.ce") && settings["lm.ce"] is string)
```
FridaTveit, line 46: It might be worth storing these strings as class constants, both to avoid having to type out constants multiple times and to avoid inconsistencies like this again. Maybe something like:
```
private const string BaseKey = "lm.";
private const string ConsolidationKey = $"{BaseKey}ce";
...
```
what do you think? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Data/Models/CourseSettings.cs
SChandarana: Yeah i think that seems sensible

FridaTveit (approved):
Looks good 🙂

PR #197, opened 2021-01-21
HEEDLS-320 Add single diagnostic assessment front end

benjimarshall:
## Changes
* Amend section action to redirect to the diagnostic assessment page if the course only has a diagnostic assessment.
* Add completion data to diagnostic assessment view model, to be shown if it is the only item in the course, and flags to record whether this is the only section, and only item in the section.
* Hide the next button if this is the only item in the section.
* Modify the mobile-only back breadcrumb to have a more helpful name, depending on if there are other sections in the course or not.

## Testing
Add unit tests, tested in Google Chrome

### Diagnostic, only item in section, among other sections, wide viewport
single_diagnostic_section_among_other_sections
### Diagnostic, only item in section, among other sections, narrow viewport
single_diagnostic_section_among_other_sections_narrow
### Diagnostic, only item in course, wide viewport
single_diagnostic_course
### Diagnostic, only item in course, narrow viewport
single_diagnostic_course_narrow
### Diagnostic, only item in course, wide viewport, with completion summary
single_diagnostic_course_with_completion

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
@@ -473,11 +473,11 @@
             var tutorial = SectionContentHelper.CreateDefaultSectionTutorial(id: tutorialId);
             var expectedSectionContent = SectionContentHelper.CreateDefaultSectionContent(
                 diagAssessPath: "some/diagnostic/path.html",
+                diagStatus: false,
                 plAssessPath: null,
                 consolidationPath: null,
                 otherSectionsExist: true
             );
-            expectedSectionContent.DiagnosticStatus = false;
             expectedSectionContent.Tutorials.Add(tutorial);
 
             A.CallTo(() => sectionContentService.GetSectionContent(customisationId, CandidateId, sectionId))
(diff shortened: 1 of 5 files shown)


DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
```diff
@@ -626,6 +626,101 @@ public void Sections_should_redirect_to_post_learning_assessment_if_there_is_dia
...
+            // Given
+            const int customisationId = 123;
+            const int sectionId = 456;
+            var expectedSectionContent = SectionContentHelper.CreateDefaultSectionContent(
+                diagAssessPath: "some/diagnostic/path.html",
+                plAssessPath: null,
+                consolidationPath: null,
+                otherSectionsExist: true
+            );
+            expectedSectionContent.DiagnosticStatus = true;
```
FridaTveit, line 641: Why not set this as part of creating the section content? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web.Tests/Controllers/LearningMenu/LearningMenuControllerTests.cs
benjimarshall: This value is not exposed as part of the constructor, it is set as the list of tutorials is processed in the multi-map part of the query call in the service. I could amend this helper method to set these values after content has been created if you think that would make it clearer? 🙂

FridaTveit: I'm pretty sure it is part of the constructor, it just gets updated as part of the multimap? 🙂

benjimarshall: Yes it is, sorry about that! Fixed in ee9c443 🙂

FridaTveit (approved):
Looks good 🙂

PR #198, opened 2021-01-21
HEEDLS-322 Tutorial content viewer close not working after fullscreen

kwunkiuma:
Small modification to the regex to account for URL after enabling fullscreen.

Add test for URL after enabling fullscreen.

Ran and passed all unit tests, manually tested on Firefox by using `closeMpe()` in fullscreen and after exiting fullscreen.


FridaTveit (approved):
Looks good 🙂

PR #199, opened 2021-01-21
HEEDLS-284 - Section Page Show Diagnostic Recommendation Status Fix

SChandarana:
No new tests have been added, all the diffs in the test files are to fix existing tests under the new section tutorial constructor. 

Diagnostic Status and diagnostic attempts are now using the per tutorial values rather than per section values which means the status tag will be displayed/removed in the correct situations. 

For the view, changes have been made so that in the case where there is no time information and no recommendation status, the learn status will appear on the same row as the title. Unfortunately due to the nature of float right it would require a lot of messy logic to have the learn status replace the recommendation status (if recommendation status is not shown and time information is) so in that case there will be white space where the recommendation status would have been.

Something I realised is that the recommendation status is based on if the learn status is true, this means that all the whitespace problems could be fixed by swapping the location of the recommendation status and the learn status (as if the learn status isn't visible, then neither will the recommendation status meaning no top white space). There would be some formatting challenges with the status tag that would require some extra css but it would be plausible. Let me know if you think this is a change we want to do or if we're happy leaving it with the white space in that scenario

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/TutorialCardViewModelTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/TutorialCardViewModelTests.cs
@@ -45,9 +45,7 @@
                 showTime,
                 showLearnStatus,
                 CustomisationId,
-                SectionId,
-                DiagnosticStatus,
-                DiagnosticAttempts
+                SectionId
             );
 
             // Then
@@ -66,9 +64,7 @@
                 showTime: true,
                 showLearnStatus: true,
                 CustomisationId,
-                SectionId,
-                DiagnosticStatus,
-                DiagnosticAttempts
+                SectionId
             );
 
             // Then
@@ -87,9 +83,7 @@
                 showTime: true,
                 showLearnStatus: false,
                 CustomisationId,
-                SectionId,
-                DiagnosticStatus,
-                DiagnosticAttempts
+                SectionId
             );
 
             // Then
@@ -108,9 +102,7 @@
                 ShowTime,
                 ShowLearnStatus,
                 CustomisationId,
-                SectionId,
-                DiagnosticStatus,
-                DiagnosticAttempts
+                SectionId
             );
 
             // Then
@@ -129,9 +121,7 @@
                 ShowTime,
                 ShowLearnStatus,
                 CustomisationId,
-                SectionId,
-                DiagnosticStatus,
-                DiagnosticAttempts
+                SectionId
             );
 
             // Then
@@ -142,7 +132,10 @@
         public void Tutorial_card_should_show_recommendation_status_if_all_conditions_are_met()
         {
             // Given
-            var sectionTutorial = SectionTutorialHelper.CreateDefaultSectionTutorial();
+            var sectionTutorial = SectionTutorialHelper.CreateDefaultSectionTutorial(
+                tutorialDiagnosticStatus: true,
+                tutorialDiagnosticAttempts: 2
+            );
 
             // When
             var tutorialCardViewModel = new TutorialCardViewModel(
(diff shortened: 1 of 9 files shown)


FridaTveit:
@SChandarana screenshots? 🙂

SChandarana:
Yep let me just get some, I think they might require some local DB changes/changing the centre and candidate ids in the login controller to get the cases

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/TutorialCardViewModelTests.cs
```diff
@@ -163,17 +154,17 @@ public void Tutorial_card_should_show_recommendation_status_if_all_conditions_ar
         public void Tutorial_card_should_not_show_recommendation_status_if_diagnostic_status_is_false()
         {
             // Given
-            var sectionTutorial = SectionTutorialHelper.CreateDefaultSectionTutorial();
+            var sectionTutorial = SectionTutorialHelper.CreateDefaultSectionTutorial(
+                tutorialDiagnosticStatus: false
```
FridaTveit, line 158: Should set attempts to something > 0 here 🙂

DigitalLearningSolutions.Web.Tests/ViewModels/LearningMenu/TutorialCardViewModelTests.cs
```diff
@@ -205,17 +194,17 @@ public void Tutorial_card_should_not_show_recommendation_status_if_show_learn_st
         public void Tutorial_card_should_not_show_recommendation_status_if_diagnostic_attempts_is_zero()
         {
             // Given
-            var sectionTutorial = SectionTutorialHelper.CreateDefaultSectionTutorial();
+            var sectionTutorial = SectionTutorialHelper.CreateDefaultSectionTutorial(
+                tutorialDiagnosticAttempts: 0
```
FridaTveit, line 198: Should set status to true here 🙂

DigitalLearningSolutions.Web/Views/LearningMenu/Section/_TutorialCard.cshtml
```diff
@@ -3,34 +3,33 @@
...
+      }
+      @if (Model.ShowLearnStatus) {
+        <div class="nhsuk-grid-column-one-third">
           <h3 class="nhsuk-heading-xs nhsuk-u-secondary-text-color float-right-additional-information">
             @Model.CompletionStatus
           </h3>
-        }
-      </div>
+        </div>
+      }
```
FridaTveit, line 32: Sorry, I'm confused. If both ShowRecommendationStatus and ShowLearnStatus are true, won't all these column sizes add up to more than 1? 🙂

FridaTveit (commented):
A few comments, and some screenshots would be good 🙂

DigitalLearningSolutions.Web/Views/LearningMenu/Section/_TutorialCard.cshtml
SChandarana: From the way the grid system works, It's fine to have columns add up to more than 1 as it will just start a new row as soon as one row is full. In the case where everything is true, there will be two-thirds for the title, one third for the recommendation status, two-thirds for the time and one third for the learn status. The time and the learn status will automatically drop onto a new row.
By not explicitly defining a new row for the time information and the learn status, it means that in the case were the recommendation status and time information are not displayed, the learn status will appear on the same row as the title and there won't be the white space that was found when testing

SChandarana:
Recommended, Optional and Not shown:

image

Recommendation status and time information not shown, learn status appears on same row as title:

image

FridaTveit (approved):
Looks good 🙂

PR #203, opened 2021-01-22
HEEDLS-325 Change post learning page status text to green

benjimarshall:
## Changes
Change post learning status text on the post learning page to green
when the assessment has been completed successfully.

The text size does not need to be changed, already it is 24px on the
narrowest of displays.

## Testing
Add some unit tests, tested in Chrome, Firefox, Edge and IE11, and using a screenreader.

### Not attempted post learning assessment
not_attempted_post_learning
### Failed post learning assessment
failed_post_learning_assessment
### Completed post learning assessment
green_post_learning_status

### DIFF
--- a/DigitalLearningSolutions.Web/ViewModels/LearningMenu/PostLearningAssessmentViewModel.cs
+++ b/DigitalLearningSolutions.Web/ViewModels/LearningMenu/PostLearningAssessmentViewModel.cs
@@ -7,6 +7,7 @@
         public string CourseTitle { get; }
         public string SectionName { get; }
         public string AssessmentStatus { get; }
+        public string AssessmentStatusStyling { get; }
         public string? ScoreInformation { get; }
         public bool PostLearningLocked { get; }
         public int CustomisationId { get; }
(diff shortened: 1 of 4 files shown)


DigitalLearningSolutions.Web/ViewModels/LearningMenu/PostLearningAssessmentViewModel.cs
```diff
@@ -29,10 +30,12 @@ public PostLearningAssessmentViewModel(PostLearningAssessment postLearningAssess
             if (postLearningAssessment.PostLearningAttempts == 0)
             {
                 AssessmentStatus = "Not attempted";
+                AssessmentStatusStyling = "not-passed-text";
             }
             else
             {
                 AssessmentStatus = GetPassStatus(postLearningAssessment);
+                AssessmentStatusStyling = GetPassStatusStyling(postLearningAssessment);
```
FridaTveit, line 38: You could assign this outside the if? Saves needing to type out the not passed css class twice (which means there's a chance they could become inconsistent). And I think assigning to a variable just once makes the code easier to read (instead of assigning to it in the if branch and the else) 🙂

FridaTveit (commented):
One small suggestion otherwise looks good 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/PostLearningAssessmentViewModel.cs
benjimarshall: Yeah that looks better! Done in 42a3e2f 🙂

FridaTveit (approved):
Looks good 🙂

PR #204, opened 2021-01-22
HEEDLS-326 - Tutorial Page - Change Button Text On Restart

SChandarana:
image

### DIFF
--- a/DigitalLearningSolutions.Web/ViewModels/LearningMenu/TutorialViewModel.cs
+++ b/DigitalLearningSolutions.Web/ViewModels/LearningMenu/TutorialViewModel.cs
@@ -29,6 +29,8 @@
         public bool OnlyItemInThisSection { get; }
         public bool ShowCompletionSummary { get; }
         public CompletionSummaryCardViewModel CompletionSummaryCardViewModel { get; }
+        public string TutorialStartButtonColour { get; }
+        public string TutorialStartButtonText { get; }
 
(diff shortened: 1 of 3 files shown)


DigitalLearningSolutions.Web/ViewModels/LearningMenu/TutorialViewModel.cs
```diff
@@ -89,6 +91,8 @@ int sectionId
                 tutorialInformation.DiagnosticAssessmentCompletionThreshold,
                 tutorialInformation.TutorialsCompletionThreshold
             );
+            TutorialStartButtonColour = tutorialInformation.Status == "Complete" ? "nhsuk-button--secondary" : "";
```
FridaTveit, line 94: I think `TutorialStartButtonStyling` or even `TutorialStartButtonAdditionalStyling` might be a more descriptive name? 🙂

FridaTveit (commented):
One small suggestion, otherwise looks good 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/TutorialViewModel.cs
SChandarana: Thanks! I was fighting with a variable name for that and couldn't for the life of me think of one aha

FridaTveit (approved):
Looks good 🙂

PR #207, opened 2021-01-25
Change post learning start button text

benjimarshall:
## Changes
Change post learning assessment start button to say restart instead of start when the assessment has already been attempted.

## Testing
Add some more unit tests. Tested in Chrome, Firefox, Edge and IE11, using high zoom and a screen reader.

### Not attempted assessment page
not_attempted_assessment_button
### Passed assessment page
passed_assessment_button
### Failed assessment page
failed_assessment_button

### DIFF
--- a/DigitalLearningSolutions.Web/ViewModels/LearningMenu/PostLearningAssessmentViewModel.cs
+++ b/DigitalLearningSolutions.Web/ViewModels/LearningMenu/PostLearningAssessmentViewModel.cs
@@ -8,6 +8,8 @@
         public string SectionName { get; }
         public string AssessmentStatus { get; }
         public string AssessmentStatusStyling { get; }
+        public string StartButtonText { get; }
+        public string StartButtonAdditionalStyling { get; }
         public string? ScoreInformation { get; }
         public bool PostLearningLocked { get; }
         public int CustomisationId { get; }
@@ -30,11 +32,15 @@
             if (postLearningAssessment.PostLearningAttempts == 0)
             {
                 AssessmentStatus = "Not attempted";
+                StartButtonText = "Start assessment";
+                StartButtonAdditionalStyling = "";
             }
             else
             {
                 AssessmentStatus = GetPassStatus(postLearningAssessment);
                 ScoreInformation = GetScoreInformation(postLearningAssessment);
+                StartButtonText = "Restart assessment";
+                StartButtonAdditionalStyling = "nhsuk-button--secondary";
             }
             AssessmentStatusStyling = GetPassStatusStyling(postLearningAssessment);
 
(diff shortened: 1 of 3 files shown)


DigitalLearningSolutions.Web/ViewModels/LearningMenu/PostLearningAssessmentViewModel.cs
```diff
@@ -30,11 +32,15 @@ public PostLearningAssessmentViewModel(PostLearningAssessment postLearningAssess
...
                 AssessmentStatus = "Not attempted";
+                StartButtonText = "Start assessment";
+                StartButtonAdditionalStyling = "";
             }
             else
             {
                 AssessmentStatus = GetPassStatus(postLearningAssessment);
                 ScoreInformation = GetScoreInformation(postLearningAssessment);
+                StartButtonText = "Restart assessment";
+                StartButtonAdditionalStyling = "nhsuk-button--secondary";
```
FridaTveit, line 43: I would probably assign to these outside the if, using a ternary statement. However I don't think one is really better than the other, as using ternaries means you'll do the same check several times. So up to you, feel free to keep as it is if you prefer 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/PostLearningAssessmentViewModel.cs
```diff
@@ -30,11 +32,15 @@ public PostLearningAssessmentViewModel(PostLearningAssessment postLearningAssess
             if (postLearningAssessment.PostLearningAttempts == 0)
             {
                 AssessmentStatus = "Not attempted";
+                StartButtonText = "Start assessment";
+                StartButtonAdditionalStyling = "";
```
FridaTveit, line 36: Isn't this the default value? 🙂

FridaTveit (commented):

DigitalLearningSolutions.Web/ViewModels/LearningMenu/PostLearningAssessmentViewModel.cs
benjimarshall: I'm not quite sure, what you mean, maybe the empty string on line 36? I think so but it was giving a "value might not be assigned" warning without me explicitly setting it to 0 here 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/PostLearningAssessmentViewModel.cs
benjimarshall: Yep, sure! Changed in 5bf55d4 🙂

FridaTveit (approved):
Looks good 🙂

DigitalLearningSolutions.Web/ViewModels/LearningMenu/PostLearningAssessmentViewModel.cs
FridaTveit: Ah in which case assigning it to empty string when defining the variable would probably make the error go away. But since you changed it to use a ternary operator it's fine as it is 🙂

PR #216, opened 2021-01-26
DLSV2-151 Reads min and max values from database in service method

kevwhitt-hee:
And tweaks tests and calls accordingly

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Services/SelfAssessmentServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/SelfAssessmentServiceTests.cs
@@ -162,14 +162,12 @@
             const int competencyId = 2;
             const int assessmentQuestionId = 2;
             const int result = 5;
-            const int minValue = 0;
-            const int maxValue = 10;
 
             using (new TransactionScope())
             {
                 // When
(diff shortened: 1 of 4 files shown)


FridaTveit:
@kevwhitt-hee looks like this breaks the data tests with the error `Invalid column name 'CompetencyGroupID'`. Do you know why that is? Has that column been added to the db? Do the data tests work locally for you? 🙂

kevwhitt-hee:
Argh! Trying to fix a bug with the self-assessments and need to tweak a service call to make it work with my dev database (which has changes relating to another branch) and mistakenly checked it in. Fixing now - my bad :(

FridaTveit:
I see, no worries 🙂

FridaTveit:
@kevwhitt-hee just so you're aware I'm still seeing failing tests on master. Specifically `SetResultForCompetency_does_not_set_result_for_invalid_question`, `SetResultForCompetency_does_not_set_result_for_invalid_result` and `SetResultForCompetency_does_not_set_result_for_negative_result`. They all give a `NullReferenceException`. Looks like it's from trying to access `assessmentQuestion.MinValue` in the `SetResultForCompetency` method. I assume because the select returns `NULL` as the assessment question id is invalid 🙂

kevwhitt-hee:
Sorry, @FridaTveit . Had a nightmare with this one. Should be good, now. :)

FridaTveit:
Yes all appears to be working now, thanks! 🙂

PR #227, opened 2021-03-17
HEEDLS-356 - Created Migration that adds new ResetPassword table and …

DanBloxham-sw:
…adds new ResetPasswordID columns to AdminUsers table and Candidates table.

Tested the Up Migration through the usual means of running the Web project - everything that needed to be added to the database was added successfully.
Tested the Down Migration through the process described in the "Reversing the Migration" section in the readme - all the new columns and the new table were deleted successfully.

### DIFF
--- a/DigitalLearningSolutions.Data.Migrations/202103161535_AddResetPasswordTable.cs
+++ b/DigitalLearningSolutions.Data.Migrations/202103161535_AddResetPasswordTable.cs
@@ -9,22 +9,24 @@
         {
             Create.Table("ResetPassword")
                 .WithColumn("ID").AsInt32().NotNullable().PrimaryKey().Identity()
-                .WithColumn("ResetPasswordHash").AsCustom("text").NotNullable()
-                .WithColumn("PasswordResetDateTime").AsCustom("timestamp").NotNullable();
+                .WithColumn("ResetPasswordHash").AsString(64).NotNullable()
+                .WithColumn("PasswordResetDateTime").AsDateTime().NotNullable().WithDefault(SystemMethods.CurrentUTCDateTime); ;
 
             Alter.Table("AdminUsers")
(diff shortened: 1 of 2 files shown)


DigitalLearningSolutions.Data.Migrations/202103161535_AddResetPasswordTable.cs
```diff
@@ -0,0 +1,30 @@
...
+
+    [Migration(202103161535)]
+    public class AddResetPasswordTable : Migration
+    {
+        public override void Up()
+        {
+            Create.Table("ResetPassword")
+                .WithColumn("ID").AsInt32().NotNullable().PrimaryKey().Identity()
+                .WithColumn("ResetPasswordHash").AsCustom("text").NotNullable()
+                .WithColumn("PasswordResetDateTime").AsCustom("timestamp").NotNullable();
```
stellake, line 13: Ah, I just realised that I wasn't super specific on the ticket, but I think we password reset hash is max 64 characters, so we can create is as string, and given that DLS is only in UK, we should probably use AsDateTime for consistency with other tables/columns.

stellake (commented):
Just one comment about the data types, otherwise looks good! :)

DigitalLearningSolutions.Data.Migrations/202103161535_AddResetPasswordTable.cs
```diff
@@ -0,0 +1,30 @@
...
+
+            Alter.Table("AdminUsers")
+                .AddColumn("ResetPasswordID").AsInt32().Nullable();
+
+            Alter.Table("Candidates")
+                .AddColumn("ResetPasswordID").AsInt32().Nullable();
```
stellake, line 19: Ah, I think these should be marked as foreign keys (which means the order of deletions should be different too)

stellake (approved):

PR #230, opened 2021-03-22
HEEDLS-353 - Added a basic form page for resetting passwords, and a g…

DanBloxham-sw:
…eneric confirmation page.

A simple page at /ForgotPassword has been created to take in an email address for resetting (actual work for the resetting of a password will be completed in a different task). When the Reset button is hit, a confirmation page is shown. The confirmation page is generic and can be reused, the first instance of its use is this task - a ConfirmationViewModel needs to be passed to it containing a title to be used as the page title and in the h1, and an array of strings can also be passed with each one forming a new paragraph. 

Tested with:
- No Javascript
- IE 11, Edge, Chrome, Firefox
- Chrome Mobile view down to 320px wide
- Screen reader
- All Unit tests in Web.Tests
- Made sure it was accessible when "logged out" by using incognito windows.
All above tests worked fine.

Screenshots of new pages:
ForgotPassword Confirmation
ForgotPassword


DigitalLearningSolutions.Web/Controllers/ForgotPasswordController.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    using Microsoft.AspNetCore.Mvc;
+    using DigitalLearningSolutions.Web.ViewModels.ForgotPassword;
+    using DigitalLearningSolutions.Web.ViewModels.LearningSolutions;
+
+    public class ForgotPasswordController : Controller
+    {
+
+        [Route("/ForgotPassword")]
+        public IActionResult ForgotPassword()
```
stellake, line 11: The Index method on the controller should have a "/ForgotPassword" route by default, so I think it'll be fine to name this Index. Also, it'll make it really obvious that this is the main forgot password page (both in controller and view folders). :)

DigitalLearningSolutions.Web/Controllers/ForgotPasswordController.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            var model = new ForgotPasswordViewModel();
+            return View(model);
+        }
+
+        [HttpPost]
+        public IActionResult SubmitEmail(string emailAddress)
```
stellake, line 18: Hmm.. I'd probably name this RecoverPassword as we'll also be doing some db changes and validation here. :)

Also, do you think it'd be worth redirecting to a /ForgotPassword/Confirm url when everything here is successful?

DigitalLearningSolutions.Web/Controllers/ForgotPasswordController.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            //Do some validation
```
stellake, line 20: We usually add a ticket reference and a TODO for upcoming work, as:
* There are a lot of tools that detect comments with a string "TODO" in them and summarise outstanding TODOs in the codebase.
* Adding a JIRA ticket reference helps with tracking work (e.g. whether ticket is done and someone has forgotten to remove a todo, or maybe if something has got prioritised the JIRA ticket will have more reasons on it about why it has happened)

In this case, this line could be:
// TODO: HEEDLS-354 - form submission logic

DigitalLearningSolutions.Web/Controllers/ForgotPasswordController.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            string pageTitle = "Password Reset Email Sent";
+            string[] pageDescription =
+            {
+                "An email has been sent to you giving details of how to reset your password.",
+                "The link provided in that email will expire in two hours.",
+                "If you have not received an email, please contact your centre administrator."
+            };
+
+            return new ConfirmationViewModel(pageTitle, pageDescription);
```
stellake, line 36: Hmm.. My initial thought was that we’ll have lots of confirmation pages with just a title and description that might be worth commonising. However, given that there are several paragraphs that we might need to add here and there, pulling the content away from the HTML files doesn’t seem to be worth the commonisation. What are your thoughts? Would you be happy to change this to just forgot password specific confirmation page and reverting the commonisation?

DigitalLearningSolutions.Web/Views/ForgotPassword/ForgotPassword.cshtml
```diff
@@ -0,0 +1,20 @@
...
+<div class="nhsuk-grid-row">
+  <div class="nhsuk-grid-column-full">
+    <h1 id="page-heading">Reset your password</h1>
+
+    <form method="post" role="form" asp-action="SubmitEmail">
+      <label class="nhsuk-label nhsuk-u-margin-bottom-5" for="emailAddress">
+        Enter your email address and we will send you password reset instructions.
+      </label>
+      <input class="nhsuk-input nhsuk-u-margin-bottom-5" type="email" id="emailAddress" required aria-required="true" asp-for="EmailAddress" value="@Model.EmailAddress"/>
+      <input class="nhsuk-button" type="submit" value="Reset" />
```
stellake, line 16: I think we usually use tags in the forms as they should be more flexible than input type="submit"

DigitalLearningSolutions.Web/Views/LearningSolutions/Confirmation.cshtml
```diff
@@ -0,0 +1,15 @@
+﻿@using DigitalLearningSolutions.Web.ViewModels.LearningSolutions
+@model ConfirmationViewModel
+@{
+  ViewData["Title"] = "Digital Learning Solutions - " + Model.Title;
+}
+
+<div class="nhsuk-grid-row">
+  <div class="nhsuk-grid-column-two-thirds nhsuk-u-reading-width">
```
stellake, line 8: Is there a reason this is two-thirds?

DigitalLearningSolutions.Web/Controllers/ForgotPasswordController.cs
DanBloxham-sw: I am happy to revert the commonisation and just use regular HTML files for this and future confirmation pages. I think that the impact will be very minor either way. Probably depends on how many confirmation pages we think we will need, how complex they will be, and how often they will need modifying, but I don't currently have a wide understanding across the whole project of these factors. Do we anticipate lots of confirmations that aren't just a couple of lines?

DigitalLearningSolutions.Web/Views/ForgotPassword/ForgotPassword.cshtml
```diff
@@ -0,0 +1,20 @@
...
+@model ForgotPasswordViewModel
+@{
+  ViewData["Title"] = "Digital Learning Solutions - Reset Your Password";
+}
+
+<div class="nhsuk-grid-row">
+  <div class="nhsuk-grid-column-full">
+    <h1 id="page-heading">Reset your password</h1>
+
+    <form method="post" role="form" asp-action="SubmitEmail">
```
stellake, line 11: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Form_Role seem to imply that role="form" is for div elements that resemble a form. Is there a reason to add it to the form tag?

stellake:
UI wise, I think it would be nicer to have slightly more spacing between the title and description than between the description and the input form. 
Could you also make the input slightly less wide (unless it's bad for accessibility for some reason)?

DigitalLearningSolutions.Web/Controllers/ForgotPasswordController.cs
stellake: Yeah, I think the registration will have a much more complex confirmation page and sign in page won't need one. We might need one when the password has been reset successfully, but I don't think there's too much of it. Maybe we should revert for now and reconsider later if we feel like there's a lot of repetition across the project? :)

stellake (approved):
Looks good! Just super minor things that you can resolve and then merge:
- I think the "RecoverEmail" method should be "RecoverPassword" (or "ResetPassword" if you prefer that.
- Super minor, but the "Password Reset Email Sent" heading capitalisation should probably be consistent with the main Forgot password page - "Password reset email sent"
- Just wanted to double check that the one-half width tag on the input works OK on mobile? Otherwise, we could probably just define a max-width on the CSS.

PR #233, opened 2021-03-26
HEEDLS-354 - Reset Password email sending implemented.

DanBloxham-sw:
ForgotPassword controller has the post functionality implemented with some basic validation. This is hooked up to a new PasswordResetService that does most of the logic. A new basic UserService for getting AdminUsers or Delegates (Candidates) was introduced, as has an EmailService. 

Note that the NotificationService has been modified to now use the EmailService and has been sanity checked, but not properly tested.

Succeeded tests of:
- Manually tested IE11, Edge, Chrome, Firefox*
- Manually tested Chrome mobile view
- Manually tested No JS
- Tested with screen reader
- Ran all unit tests successfully.
*Firefox has a very old bug (see: https://bugzilla.mozilla.org/show_bug.cgi?id=308064 ) that prevents the error summary link from focusing on the input textbox as required by the NHS error summary description - discussion in the Show & Tell on 26th March 2021 confirmed that no workaround should be performed for this.

UI Change (/ForgotPassword page with error messages):
ForgotPassword with errors

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Helpers/EmailTestHelper.cs
+++ b/DigitalLearningSolutions.Data.Tests/Helpers/EmailTestHelper.cs
@@ -1,0 +1,37 @@
+﻿namespace DigitalLearningSolutions.Data.Tests.Helpers
+{
+    using DigitalLearningSolutions.Data.Models.Email;
+    using MimeKit;
+
+    public static class EmailTestHelper
+    {
+        public const string DefaultHtmlBody = "<body style= 'font - family: Calibri; font - size: small;'>\r\n" +
+                                              "                                    <p>Test Body</p>\r\n" +
+                                              "                                </body>";
+
+        public static Email GetDefaultEmail
+        (
+            string[]? to = null,
+            string[]? cc = null,
+            string[]? bcc = null,
+            string subject = "Test Subject Line",
+            BodyBuilder? body = null
+        )
+        {
+            return new Email
+            (
+                to: to ?? new string[1] { "recipient@example.com" },
+                cc: cc ?? new string[1] { "cc@example.com" },
+                bcc: bcc ?? new string[1] { "bcc@example.com" },
+                subject: subject,
+                body: body ?? new BodyBuilder
+                {
+                    TextBody = "Test body",
+                    HtmlBody = @"<body style= 'font - family: Calibri; font - size: small;'>
+                                    <p>Test Body</p>
+                                </body>"
+                }
+            );
+        }
+    }
+}
--- a/DigitalLearningSolutions.Data.Tests/Helpers/EmailTestHelper.cs
+++ b/DigitalLearningSolutions.Data.Tests/Helpers/EmailTestHelper.cs
@@ -5,9 +5,9 @@
 
     public static class EmailTestHelper
     {
-        public const string DefaultHtmlBody = "<body style= 'font - family: Calibri; font - size: small;'>\r\n" +
-                                              "                                    <p>Test Body</p>\r\n" +
-                                              "                                </body>";
+        public const string DefaultHtmlBody = "<body style= 'font - family: Calibri; font - size: small;'\r\n>" +
+                                              "   <p>Test Body</p>\r\n" +
+                                              "</body>";
 
         public static Email GetDefaultEmail
         (
@@ -27,9 +27,7 @@
                 body: body ?? new BodyBuilder
                 {
                     TextBody = "Test body",
-                    HtmlBody = @"<body style= 'font - family: Calibri; font - size: small;'>
-                                    <p>Test Body</p>
-                                </body>"
+                    HtmlBody = DefaultHtmlBody
                 }
             );
         }
--- a/DigitalLearningSolutions.Data.Tests/Services/EmailServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/EmailServiceTests.cs
@@ -1,0 +1,296 @@
+﻿namespace DigitalLearningSolutions.Data.Tests.Services
+{
+    using System.Threading;
+    using DigitalLearningSolutions.Data.Factories;
+    using DigitalLearningSolutions.Data.Models.Email;
+    using DigitalLearningSolutions.Data.Services;
+    using FakeItEasy;
+    using MailKit.Net.Smtp;
+    using MailKit.Security;
+    using MimeKit;
+    using NUnit.Framework;
+
+    public class EmailServiceTests
+    {
+        private EmailService emailService;
+        private IConfigService configService;
+        private ISmtpClient smtpClient;
+        private Email testEmail;
+        private Email multipleAddresseesTestEmail;
+
+        [SetUp]
+        public void Setup()
+        {
+            BodyBuilder emailTestBody = new BodyBuilder
+            {
+                TextBody = "Test body",
+                HtmlBody = @"<body style= 'font - family: Calibri; font - size: small;'>
+                                    <p>Test Body</p>
+                                </body>"
+            };
+
+            testEmail = new Email(
+                "recipient@example.com",
+                "Test Subject Line",
+                emailTestBody,
+                "cc@example.com",
+                "bcc@example.com");
+
+            multipleAddresseesTestEmail = new Email(
+                new string[2]{"recipient1@example.com", "recipient2@example.com"},
+                "Test Subject Line",
+                emailTestBody,
+                new string[2]{"cc1@example.com", "cc2@example.com"},
+                new string[2]{"bcc1@example.com","bcc2@example.com"});
+
+            configService = A.Fake<IConfigService>();
+            var smtpClientFactory = A.Fake<ISmtpClientFactory>();
+            smtpClient = A.Fake<ISmtpClient>();
+            A.CallTo(() => smtpClientFactory.GetSmtpClient()).Returns(smtpClient);
+
+            A.CallTo(() => configService.GetConfigValue(ConfigService.MailPort)).Returns("25");
+            A.CallTo(() => configService.GetConfigValue(ConfigService.MailUsername)).Returns("username");
+            A.CallTo(() => configService.GetConfigValue(ConfigService.MailPassword)).Returns("password");
+            A.CallTo(() => configService.GetConfigValue(ConfigService.MailServer)).Returns("smtp.example.com");
+            A.CallTo(() => configService.GetConfigValue(ConfigService.MailFromAddress)).Returns("test@example.com");
+
+            emailService = new EmailService(configService, smtpClientFactory);
+            // Going to want to copy the others over, but also perform it with string lists in the to/cc/bcc etc.
+        }
+
+        [Test]
+        public void Trying_to_send_mail_with_null_config_values_should_throw_an_exception()
+        {
+            // Given
+            A.CallTo(() => configService.GetConfigValue(ConfigService.MailPassword)).Returns(null);
+
+            // Then
+            Assert.Throws<ConfigValueMissingException>(() => emailService.SendEmail(testEmail));
+        }
+
+        [Test]
+        public void The_server_credentials_are_correct()
+        {
+            // When
+            emailService.SendEmail(testEmail);
+
+            // Then
+            A.CallTo(() =>
+                    smtpClient.Authenticate("username", "password", default(CancellationToken))
+                )
+                .MustHaveHappened();
+        }
+
+        [Test]
+        public void The_server_details_are_correct()
+        {
+            // When
+            emailService.SendEmail(testEmail);
+
+            // Then
+            A.CallTo(() =>
+                    smtpClient.Connect(
+                        "smtp.example.com",
+                        25,
+                        SecureSocketOptions.Auto,
+                        default(CancellationToken)
+                    )
+                )
+                .MustHaveHappened();
+        }
+
+        [Test]
+        public void The_sender_email_address_is_correct()
+        {
+            // When
+            emailService.SendEmail(testEmail);
+
+            // Then
+            A.CallTo(() =>
+                    smtpClient.Send(
+                        A<MimeMessage>.That.Matches(m =>
+                            m.From.ToString() == "test@example.com"
+                        ),
+                        default(CancellationToken),
+                        null
+                    )
+                )
+                .MustHaveHappened();
+        }
+
+        [Test]
+        public void The_email_subject_line_is_correct()
+        {
+            // When
+            emailService.SendEmail(testEmail);
+
+            // Then
+            A.CallTo(() =>
+                    smtpClient.Send(
+                        A<MimeMessage>.That.Matches(m =>
+                            m.Subject.ToString() == "Test Subject Line"
+                        ),
+                        default(CancellationToken),
+                        null
+                    )
+                )
+                .MustHaveHappened();
+        }
+
+        [Test]
+        public void The_email_text_body_is_correct()
+        {
+            // When
+            emailService.SendEmail(testEmail);
+
+            // Then
+            A.CallTo(() =>
+                    smtpClient.Send(
+                        A<MimeMessage>.That.Matches(m =>
+                            m.TextBody.ToString() == "Test body"
+                        ),
+                        default(CancellationToken),
+                        null
+                    )
+                )
+                .MustHaveHappened();
+        }
+
+        [Test]
+        public void The_email_HTML_body_is_correct()
+        {
+            // When
+            emailService.SendEmail(testEmail);
+
+            // Then
+            A.CallTo(() =>
+                    smtpClient.Send(
+                        A<MimeMessage>.That.Matches(m =>
+                            m.HtmlBody.ToString() == "<body style= 'font - family: Calibri; font - size: small;'>\r\n" +
+                            "                                    <p>Test Body</p>\r\n" +
+                            "                                </body>"
+                        
+                        ),
+                        default(CancellationToken),
+                        null
+                    )
+                )
+                .MustHaveHappened();
+        }
+
+        [Test]
+        public void The_recipient_email_address_is_correct()
+        {
+            // When
+            emailService.SendEmail(testEmail);
+
+            // Then
+            A.CallTo(() =>
+                    smtpClient.Send(
+                        A<MimeMessage>.That.Matches(m =>
+                            m.To.ToString() == "recipient@example.com"
+                        ),
+                        default(CancellationToken),
+                        null
+                    )
+                )
+                .MustHaveHappened();
+        }
+
+        [Test]
+        public void The_recipient_email_addresses_are_correct()
+        {
+            // When
+            emailService.SendEmail(multipleAddresseesTestEmail);
+
+            // Then
+            A.CallTo(() =>
+                    smtpClient.Send(
+                        A<MimeMessage>.That.Matches(m =>
+                            m.To.ToString() == "recipient1@example.com, recipient2@example.com"
+                        ),
+                        default(CancellationToken),
+                        null
+                    )
+                )
+                .MustHaveHappened();
+        }
+
+        [Test]
+        public void The_cc_email_address_is_correct()
+        {
+            // When
+            emailService.SendEmail(testEmail);
+
+            // Then
+            A.CallTo(() =>
+                    smtpClient.Send(
+                        A<MimeMessage>.That.Matches(m =>
+                            m.Cc.ToString() == "cc@example.com"
+                        ),
+                        default(CancellationToken),
+                        null
+                    )
+                )
+                .MustHaveHappened();
+        }
+
+        [Test]
+        public void The_cc_email_addresses_are_correct()
+        {
+            // When
+            emailService.SendEmail(multipleAddresseesTestEmail);
+
+            // Then
+            A.CallTo(() =>
+                    smtpClient.Send(
+                        A<MimeMessage>.That.Matches(m =>
+                            m.Cc.ToString() == "cc1@example.com, cc2@example.com"
+                        ),
+                        default(CancellationToken),
+                        null
+                    )
+                )
+                .MustHaveHappened();
+        }
+
+        [Test]
+        public void The_bcc_email_address_is_correct()
+        {
+            // When
+            emailService.SendEmail(testEmail);
+
+            // Then
+            A.CallTo(() =>
+                    smtpClient.Send(
+                        A<MimeMessage>.That.Matches(m =>
+                            m.Bcc.ToString() == "bcc@example.com"
+                        ),
+                        default(CancellationToken),
+                        null
+                    )
+                )
+                .MustHaveHappened();
+        }
+
+        [Test]
+        public void The_bcc_email_addresses_are_correct()
+        {
+            // When
+            emailService.SendEmail(multipleAddresseesTestEmail);
+
+            // Then
+            A.CallTo(() =>
+                    smtpClient.Send(
+                        A<MimeMessage>.That.Matches(m =>
+                            m.Bcc.ToString() == "bcc1@example.com, bcc2@example.com"
+                        ),
+                        default(CancellationToken),
+                        null
+                    )
+                )
+                .MustHaveHappened();
+        }
+
+    }
+}
--- a/DigitalLearningSolutions.Data.Tests/Services/EmailServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/EmailServiceTests.cs
@@ -1,9 +1,8 @@
 ﻿namespace DigitalLearningSolutions.Data.Tests.Services
 {
-    using System.Threading;
     using DigitalLearningSolutions.Data.Factories;
-    using DigitalLearningSolutions.Data.Models.Email;
     using DigitalLearningSolutions.Data.Services;
+    using DigitalLearningSolutions.Data.Tests.Helpers;
     using FakeItEasy;
     using MailKit.Net.Smtp;
     using MailKit.Security;
@@ -12,37 +11,13 @@
 
     public class EmailServiceTests
     {
-        private EmailService emailService;
         private IConfigService configService;
+        private EmailService emailService;
         private ISmtpClient smtpClient;
-        private Email testEmail;
-        private Email multipleAddresseesTestEmail;
 
         [SetUp]
         public void Setup()
         {
-            BodyBuilder emailTestBody = new BodyBuilder
-            {
-                TextBody = "Test body",
-                HtmlBody = @"<body style= 'font - family: Calibri; font - size: small;'>
-                                    <p>Test Body</p>
-                                </body>"
-            };
-
-            testEmail = new Email(
-                "recipient@example.com",
-                "Test Subject Line",
-                emailTestBody,
-                "cc@example.com",
-                "bcc@example.com");
-
-            multipleAddresseesTestEmail = new Email(
-                new string[2]{"recipient1@example.com", "recipient2@example.com"},
-                "Test Subject Line",
-                emailTestBody,
-                new string[2]{"cc1@example.com", "cc2@example.com"},
-                new string[2]{"bcc1@example.com","bcc2@example.com"});
-
             configService = A.Fake<IConfigService>();
             var smtpClientFactory = A.Fake<ISmtpClientFactory>();
             smtpClient = A.Fake<ISmtpClient>();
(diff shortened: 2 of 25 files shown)


DigitalLearningSolutions.Data.Tests/Services/EmailServiceTests.cs
```diff
@@ -0,0 +1,296 @@
...
+            A.CallTo(() => smtpClientFactory.GetSmtpClient()).Returns(smtpClient);
+
+            A.CallTo(() => configService.GetConfigValue(ConfigService.MailPort)).Returns("25");
+            A.CallTo(() => configService.GetConfigValue(ConfigService.MailUsername)).Returns("username");
+            A.CallTo(() => configService.GetConfigValue(ConfigService.MailPassword)).Returns("password");
+            A.CallTo(() => configService.GetConfigValue(ConfigService.MailServer)).Returns("smtp.example.com");
+            A.CallTo(() => configService.GetConfigValue(ConfigService.MailFromAddress)).Returns("test@example.com");
+
+            emailService = new EmailService(configService, smtpClientFactory);
+            // Going to want to copy the others over, but also perform it with string lists in the to/cc/bcc etc.
```
stellake, line 58: I'm not sure what this comment is referring to?

DanBloxham-sw: That was a note-to-self that I clearly forgot to get rid of. I'll remove that.

DigitalLearningSolutions.Data.Tests/Services/EmailServiceTests.cs
```diff
@@ -0,0 +1,296 @@
...
+            };
+
+            testEmail = new Email(
+                "recipient@example.com",
+                "Test Subject Line",
+                emailTestBody,
+                "cc@example.com",
+                "bcc@example.com");
+
+            multipleAddresseesTestEmail = new Email(
```
stellake, line 39: I'm not 100% sure about defining these emails here. I feel like multiple addresses email should be created in relevant tests (e.g. multiple recipients, multiplePeopleCcd etc). Potentially having some sort of defaultEmail to use in tests that don't care which email is used is worth it though (maybe the current testEmail renamed?)

DigitalLearningSolutions.Data.Tests/Services/EmailServiceTests.cs
```diff
@@ -0,0 +1,296 @@
...
+            A.CallTo(() => configService.GetConfigValue(ConfigService.MailPassword)).Returns("password");
+            A.CallTo(() => configService.GetConfigValue(ConfigService.MailServer)).Returns("smtp.example.com");
+            A.CallTo(() => configService.GetConfigValue(ConfigService.MailFromAddress)).Returns("test@example.com");
+
+            emailService = new EmailService(configService, smtpClientFactory);
+            // Going to want to copy the others over, but also perform it with string lists in the to/cc/bcc etc.
+        }
+
+        [Test]
+        public void Trying_to_send_mail_with_null_config_values_should_throw_an_exception()
```
stellake, line 62: Isn't this testing just the MailPassword being missing? Could we potentially use a parameterised test here (I know that in xUnit we could use [Theory], but not sure about nUnit)?

stellake: Ah, looks like nUnit has [TestCase] syntax

DigitalLearningSolutions.Data.Tests/Services/EmailServiceTests.cs
```diff
@@ -0,0 +1,296 @@
...
+        {
+            // When
+            emailService.SendEmail(testEmail);
+
+            // Then
+            A.CallTo(() =>
+                    smtpClient.Send(
+                        A<MimeMessage>.That.Matches(m =>
+                            m.HtmlBody.ToString() == "<body style= 'font - family: Calibri; font - size: small;'>\r\n" +
+                            "                                    <p>Test Body</p>\r\n" +
```
stellake, line 170: Hmm.. This is a bit odd (having all that spacing hardcoded) - could we pull the HTML string into a private variable that can be referenced here too?

DigitalLearningSolutions.Data.Tests/Services/PasswordResetServiceTests.cs
```diff
@@ -0,0 +1,81 @@
...
+    using DigitalLearningSolutions.Data.Exceptions;
+    using DigitalLearningSolutions.Data.Models;
+    using DigitalLearningSolutions.Data.Models.Email;
+    using DigitalLearningSolutions.Data.Models.User;
+    using DigitalLearningSolutions.Data.Services;
+    using DigitalLearningSolutions.Data.Tests.Helpers;
+    using FakeItEasy;
+    using Microsoft.Extensions.Logging;
+    using NUnit.Framework;
+
```
stellake, line 14: Looks like there's some unused imports - I recommend running ReSharper's CleanUpCode on the files you touched to remove unused imports & use specific formatting (I'll ignore any potential formatting issues for the rest of the review) :)

DigitalLearningSolutions.Data.Tests/Services/PasswordResetServiceTests.cs
```diff
@@ -0,0 +1,81 @@
...
+            connection = ServiceTestHelper.GetDatabaseConnection();
+
+            A.CallTo(() => userService.GetUserByEmailAddress(A<string>._)).Returns(new DelegateUser
+            {
+                Id = 1,
+                FirstName = "Forename",
+                Surname = "Surname",
+                EmailAddress = "recipient@example.com",
+                ResetPasswordId = null
+            });
```
stellake, line 40: Some of the Web.Tests project tests have builders under TestHelpers. Not sure if it's really necessary for one user, but just FYI for the future if you want to pull out some of the model constructions. :)

DigitalLearningSolutions.Data.Tests/Services/PasswordResetServiceTests.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            // Given
+            A.CallTo(() => userService.GetUserByEmailAddress(A<string>._)).Returns(null);
+
+            // Then
+            Assert.Throws<EmailAddressNotFoundException>(() => passwordResetService.SendResetPasswordEmail("recipient@example.com"));
+        }
+
+        [Test]
+        public void Trying_to_send_reset_password_email_with_null_config_values_should_throw_an_exception()
```
stellake, line 59: Isn't null config value tested as part of the email tests or is this just checking the bare url?

DigitalLearningSolutions.Data.Tests/Services/PasswordResetServiceTests.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        [Test]
+        public void Trying_to_send_password_reset_sends_email()
+        {
+            // When
+            passwordResetService.SendResetPasswordEmail("recipient@example.com");
+
+            // Then
+            A.CallTo(() =>
+                    emailService.SendEmail(A<Email>._)
```
stellake, line 76: Could we also check that it's called with correct params?

DigitalLearningSolutions.Data.Tests/Services/UnlockServiceTests.cs
```diff
@@ -37,109 +31,39 @@ public void Setup()
...
-            A.CallTo(() => configService.GetConfigValue(ConfigService.MailServer)).Returns("smtp.example.com");
-            A.CallTo(() => configService.GetConfigValue(ConfigService.MailFromAddress)).Returns("test@example.com");
 
-            notificationService = new NotificationService(notificationDataService, configService, smtpClientFactory);
+            notificationService = new NotificationService(notificationDataService, configService, emailService);
         }
 
         [Test]
-        public void Trying_to_send_mail_with_null_config_values_should_throw_an_exception()
+        public void Trying_get_null_unlock_data_should_throw_an_exception()
```
stellake, line 39: What does it mean to get null unlock data?

DigitalLearningSolutions.Data.Tests/Services/PasswordResetServiceTests.cs
DanBloxham-sw: The BaseURL is only retrieved in the PasswordResetService to build the password reset link used in the email. It is all the same logic, but the exception would be thrown in a different place to the EmailService. A similar "duplication" appears in the NotificationsService.

DigitalLearningSolutions.Data.Tests/Services/UnlockServiceTests.cs
```diff
@@ -37,109 +31,39 @@ public void Setup()
...
+            A.CallTo(() => notificationDataService.GetUnlockData(A<int>._)).Returns(null);
 
             // Then
-            Assert.Throws<ConfigValueMissingException>(() => notificationService.SendUnlockRequest(1));
+            Assert.Throws<NotificationDataException>(() => notificationService.SendUnlockRequest(1));
         }
 
         [Test]
-        public void The_sender_email_address_is_correct()
+        public void Trying_to_send_unlock_request_with_null_config_values_should_throw_an_exception()
```
stellake, line 49: maybe throws an exception if tracking system base url is null?

DigitalLearningSolutions.Data/Models/Email/Email.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    using MimeKit;
+
+    public class Email
+    {
+        public string[] To, Cc, Bcc;
+        public string Subject;
+        public BodyBuilder Body;
+
+        public Email()
```
stellake, line 15: I feel like the subject, body and recipient should be required fields if you want to create an Email, so maybe we should only have two constructors:
1. string subject, BodyBuilder body, string to, string[]? to, string[]? bcc
2. string[] subjects, BodyBuilder body, string to, string[]? to, string[]? bcc

DigitalLearningSolutions.Data/Models/Email/Email.cs
```diff
@@ -0,0 +1,42 @@
...
+        public Email(string? to, string? subject, BodyBuilder? body, string? cc = null, string? bcc = null)
+        {
+            this.To = (to is null) ? new string[0] : new string[1] {to};
+            this.Cc = (cc is null) ? new string[0] : new string[1] {cc};
+            this.Bcc = (bcc is null) ? new string[0] : new string[1] {bcc};
+            this.Subject = subject ?? string.Empty;
+            this.Body = body ?? new BodyBuilder();
+        }
+
+        public Email(string[]? to, string? subject, BodyBuilder? body, string[]? cc = null, string[]? bcc = null)
```
stellake, line 33: Do you think it would be worth using default values of empty string array over null?

DigitalLearningSolutions.Data/Services/EmailService.cs
```diff
@@ -0,0 +1,127 @@
...
+{
+    using System;
+    using DigitalLearningSolutions.Data.Factories;
+    using DigitalLearningSolutions.Data.Models.Email;
+    using MimeKit;
+
+    public interface IEmailService
+    {
+        void SendEmail(Email email);
+        // make class of email details with To, CC, (BCC?) as string lists, and body as BodyBuilder
```
stellake, line 11: Ah, is this left over from dev? :)

DigitalLearningSolutions.Data/Services/EmailService.cs
```diff
@@ -0,0 +1,127 @@
...
+                using var client = smtpClientFactory.GetSmtpClient();
+                client.Timeout = 10000;
+                client.Connect(mailServerAddress, mailServerPort);
+
+                client.Authenticate(mailServerUsername, mailServerPassword);
+
+                client.Send(message);
+                client.Disconnect(true);
+            }
+            catch
```
stellake, line 69: Are we going to keep the "catch" here and swallow any errors?

DigitalLearningSolutions.Data/Services/EmailService.cs
```diff
@@ -0,0 +1,127 @@
...
+            {
+                return "Encountered an error while trying to send an email: The value of mailServerAddress is null";
+            }
+
+            if (mailServerPortString == null)
+            {
+                return "Encountered an error while trying to send an email: The value of mailServerPortString is null";
+            }
+
+            return "Encountered an error while trying to send an email: The value of mailSenderAddress is null";
```
stellake, line 123: I appreciate this is copied from somewhere else, but do you think it's worth having a private method for generating this error string that takes in a name of the value that's null? Then, if we ever want to modify the error message, we only need to do it in one place. :)

DigitalLearningSolutions.Data/Services/NotificationService.cs
```diff
@@ -83,12 +37,12 @@ public void SendUnlockRequest(int progressId)
             }
 
             unlockData.ContactForename = unlockData.ContactForename == "" ? "Colleague" : unlockData.ContactForename;
-
-
-            var trackingSystemBaseUrl = configService.GetConfigValue(ConfigService.TrackingSystemBaseUrl);
+            var trackingSystemBaseUrl = configService.GetConfigValue(ConfigService.TrackingSystemBaseUrl) ??
```
stellake, line 40: Is this error handled anywhere/what happens if this is thrown?

DigitalLearningSolutions.Data/Services/NotificationService.cs
```diff
@@ -125,41 +83,12 @@ public void SendFrameworkCollaboratorInvite(int adminId, int frameworkId, int in
...
-            }
-
-            if (mailSenderAddress == null)
-            {
-                return "Encountered an error while trying to send an unlock request email: The value of mailSenderAddress is null";
-            }
+            emailService.SendEmail(new Email(collaboratorNotification.Email, emailSubjectLine, builder, collaboratorNotification.InvitedByEmail));
+        }
 
+        private static string GenerateConfigValueMissingMessage()
```
stellake, line 90: Should this be named something specific to trackingSystemBaseUrl?

DigitalLearningSolutions.Data/Services/PasswordResetService.cs
```diff
@@ -0,0 +1,121 @@
...
+            IEmailService emailService)
+        {
+            this.userService = userService;
+            this.connection = connection;
+            this.logger = logger;
+            this.configService = configService;
+            this.emailService = emailService;
+        }
+
+        public void SendResetPasswordEmail(string emailAddress)
```
stellake, line 44: I feel like this method is actually doing more than just sending an email. Maybe name is ResetUsersPassword or ResetPassword?

DigitalLearningSolutions.Data/Services/EmailService.cs
DanBloxham-sw: The empty catch was implemented in the EmailService at the same place it was in the SendSMTPMessage() in the NotificationService - it was brought over so that whatever originally used SendSMTPMessage() would experience the same functionality and would not break if an unexpected exception was thrown. I think it would be best to not swallow the errors, but it might require additional handling in the NotificationService (the PasswordResetService works as expected when there is an exception thrown there, showing the user the "A problem has occurred" error page). Would that additional handling be something to cover as part of this ticket?

DigitalLearningSolutions.Data/Services/PasswordResetService.cs
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        private Email GeneratePasswordResetEmail(string emailAddress, string resetHash, string firstName)
+        {
+            string baseUrl =
+                configService.GetConfigValue(ConfigService.BaseUrl) ??
+                throw new ConfigValueMissingException("Encountered an error while trying to generate a Reset Password email: " +
+                                                      "The value of baseURL is null ");
+            UriBuilder resetPasswordUrl = new UriBuilder(baseUrl);
+            resetPasswordUrl.Path += "ResetPassword";
```
stellake, line 97: Just to check - is the baseUrl for old or refactored site? (We will need the refactored site)

DigitalLearningSolutions.Data/Services/PasswordResetService.cs
DanBloxham-sw: I feel like "ResetUsersPassword" and "ResetPassword" might be a misnomer as it does not actually do any resetting of passwords, just sending an email about it. Would something like "GenerateResetPasswordLink" be more suitable?

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -0,0 +1,70 @@
...
+                @"SELECT TOP (1)
+                        AdminID,
+                        Forename,
+                        Surname,
+                        Email,
+                        ResetPasswordID
+                    FROM AdminUsers
+                    WHERE (Email = @emailAddress)",
+                new { emailAddress }
+            ).FirstOrDefault();
```
stellake, line 48: Do we need both FirstOrDefault and SELECT TOP (1)? Also, just heads up for next phases, but we'll need to get all delegate records in the future :P 

Another question I had was about getting admin user and getting delegate user being public methods - are these used anywhere else?

DigitalLearningSolutions.Data/Services/PasswordResetService.cs
DanBloxham-sw: This is drawn from the BaseURL (ConfigID=21) located in the Config table. In the backed-up version of the database, it is recorded as "https://www.dls.nhs.uk/". Do you know if we still plan on using that config table for the baseURL, or should there be a different source/value for testing?

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -0,0 +1,70 @@
...
+            this.connection = connection;
+        }
+
+        public User? GetUserByEmailAddress(string emailAddress)
+        {
+            // Assume user is a delegate (in Candidates Table)
+            User? user = GetDelegateUserByEmailAddress(emailAddress);
+
+            // If no Candidates entry found, look to AdminUsers
+            user ??= GetAdminUserByEmailAddress(emailAddress);
```
stellake, line 31: I actually had a minor preference doing it the other way around (admin first, then delegate), because admin feels like the users "main" account, but tbh there will be more delegates so I actually think your approach is better! :)

DigitalLearningSolutions.Web.Tests/Controllers/ForgotPassword/ForgotPasswordControllerTests.cs
```diff
@@ -0,0 +1,75 @@
...
+            A.CallTo(() => passwordResetService.SendResetPasswordEmail(A<string>._)).MustHaveHappened();
+        }
+
+        [Test]
+        public void Bad_email_submission_should_render_basic_form_with_error()
+        {
+            string errorMessage = "No user account could be found with the specified email address";
+
+            // Given
+            A.CallTo(() => passwordResetService.SendResetPasswordEmail(A<string>._)).Throws(new EmailAddressNotFoundException(errorMessage));
```
stellake, line 65: Should we add an equivalent test for the db error case? Just because in theory we could handle these errors differently?

DigitalLearningSolutions.Web.Tests/Controllers/ForgotPassword/ForgotPasswordControllerTests.cs
```diff
@@ -0,0 +1,75 @@
...
+            passwordResetService = A.Fake<IPasswordResetService>();
+
+            controller = new ForgotPasswordController(passwordResetService);
+        }
+
+        [Test]
+        public void No_model_should_render_basic_form()
+        {
+            // When
+            var result = controller.Index(model: null);
```
stellake, line 30: Could this just be Index()? Also, I'm trying to think of a more intuitive test name, but my mind is blank haha! Maybe just Renders_basic_form, as "No model" can be a bit confusing if you just scan over the test names?

DigitalLearningSolutions.Web/ViewModels/ForgotPassword/ForgotPasswordViewModel.cs
```diff
@@ -3,5 +3,18 @@
     public class ForgotPasswordViewModel
     {
         public string? EmailAddress;
+
+        public bool ErrorHasOccurred;
+
+        public string? EmailErrorMessage;
```
stellake, line 9: Is ErrorHasOccurred field necessary? Can it be that error occurs and we don't have an error message or vice versa?

DigitalLearningSolutions.Data/Services/PasswordResetService.cs
stellake: That's a fair point! I feel like GenerateResetPasswordLink would describe the first half of what the service does though. Maybe "HandlePasswordReset" (fairly generic) or "GenerateAndSendPasswordResetLink" (slightly clunky but fairly accurate)?

DigitalLearningSolutions.Data/Services/PasswordResetService.cs
stellake: Ah - I think you can get the url from the config for the refactored site:
Configuration["AppRootPath"]

It looks like the old site is also stored in config:
Configuration["CurrentSystemBaseUrl"]

DigitalLearningSolutions.Data.Tests/Services/EmailServiceTests.cs
DanBloxham-sw: Just to clarify, are you suggesting we should have 
- a single default Email instance defined in the SetUp that is used when testing things that aren't specifically about the addresses
- and each test about the addresses should define their own email instance?

stellake: Yes, that's correct, although you could potentially create something similar to CurrentCourseHelper.CreateDefaultCurrentCourse, where you create the Email with defaults and only override the values that are needed for the test when calling CreateDefaultEmail.

DigitalLearningSolutions.Data.Tests/Services/PasswordResetServiceTests.cs
stellake: Got it! Just a minor thing, can we rename the test to "null_base_url_value" rather than "config_values" to be more specific?

DigitalLearningSolutions.Data/Services/EmailService.cs
stellake: Yes, we did discuss this! I will create another ticket for this error handling! :)

DigitalLearningSolutions.Data/Models/Email/Email.cs
DanBloxham-sw: I think it would, but after doing some research it looks like the only valid option is null as object references can't be used as compile-time constants (see: https://stackoverflow.com/questions/3480382/passing-an-empty-array-as-default-value-of-an-optional-parameter )

DigitalLearningSolutions.Web.Tests/Controllers/ForgotPassword/ForgotPasswordControllerTests.cs
DanBloxham-sw: It looks like the Index(model: null) is necessary to specify between the HttpGet Index() action and the HttpPost Index() action as they both take a single nullable input, otherwise NUnit doesn't know which one to test when nothing is passed in. I have renamed it "Render_basic_form_when_there_is_no_error_model" which should help with the name and explain the null model, if that is satisfactory?

DigitalLearningSolutions.Data/Services/NotificationService.cs
DanBloxham-sw: No changes have been made as to how the NotificationService handles exceptions. Both times that exception is thrown in the NotificationService (and similarly when an exception is thrown in the EmailService that isn't explicitly handled within the function/controller) the generic "Something went wrong" page is shown. I think that is suitable for these - if not, do let me know.

DigitalLearningSolutions.Web.Tests/Controllers/ForgotPassword/ForgotPasswordControllerTests.cs
DanBloxham-sw: Sorry, which DB error case specifically? Currently, any exception that isn't explicitly handled inside the controller/services causes the user to get redirected to the UnknownError page, the testing for which is done in LearningSolutionsControllerTests. As there isn't much useful information we could supply a user for this error, I can't think of a more specific way to handle it.

DigitalLearningSolutions.Data/Models/Email/Email.cs
stellake: Ah, fair! I think I've done too much Kotlin recently! 😄

DigitalLearningSolutions.Web.Tests/Controllers/ForgotPassword/ForgotPasswordControllerTests.cs
stellake: Sounds good to me! :)

DigitalLearningSolutions.Web.Tests/Controllers/ForgotPassword/ForgotPasswordControllerTests.cs
stellake: Oh, I meant the ResetPasswordInsertException, so an equivalent test would just be:
A.CallTo(() => passwordResetService.SendResetPasswordEmail(A<string>._)).Throws(new ResetPasswordInsertException(errorMessage));

DigitalLearningSolutions.Data/Services/NotificationService.cs
stellake: Yep, that sounds reasonable, thanks!

DigitalLearningSolutions.Data.Tests/Helpers/EmailTestHelper.cs
```diff
@@ -0,0 +1,37 @@
+﻿namespace DigitalLearningSolutions.Data.Tests.Helpers
+{
+    using DigitalLearningSolutions.Data.Models.Email;
+    using MimeKit;
+
+    public static class EmailTestHelper
+    {
+        public const string DefaultHtmlBody = "<body style= 'font - family: Calibri; font - size: small;'>\r\n" +
+                                              "                                    <p>Test Body</p>\r\n" +
```
stellake, line 9: Is the big spacing needed between `"` and `<p>` ?

DigitalLearningSolutions.Data.Tests/Services/EmailServiceTests.cs
```diff
@@ -0,0 +1,269 @@
...
+            A.CallTo(() => configService.GetConfigValue(ConfigService.MailFromAddress)).Returns("test@example.com");
+
+            emailService = new EmailService(configService, smtpClientFactory);
+        }
+
+        [TestCase(ConfigService.MailPort)]
+        [TestCase(ConfigService.MailUsername)]
+        [TestCase(ConfigService.MailPassword)]
+        [TestCase(ConfigService.MailServer)]
+        [TestCase(ConfigService.MailFromAddress)]
```
stellake, line 39: Nice! :)

DigitalLearningSolutions.Data.Tests/Helpers/EmailTestHelper.cs
DanBloxham-sw: It seems that the spacing needs to exactly match what is in the string. The actual string in the GetDefaultEmail method that this is compared against was copied over, formatting intact, from the previous email sending methods, which had those large gaps in. Would we prefer a slightly more sensible sensible layout?

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            List<AdminUser> adminUsers = GetAdminUsersByEmailAddress(emailAddress);
+            List<DelegateUser> delegateUsers = GetDelegateUsersByEmailAddress(emailAddress);
+
+            return (adminUsers, delegateUsers);
+        }
+
+        private List<AdminUser> GetAdminUsersByEmailAddress(string emailAddress)
+        {
+            List<AdminUser> user = connection.Query<AdminUser>(
```
stellake, line 33: Minor, but this should be "users"

DigitalLearningSolutions.Web/Controllers/ForgotPasswordController.cs
```diff
@@ -1,21 +1,55 @@
...
+            }
+
+            string baseUrl = ConfigHelper.GetAppConfig()["AppRootPath"];
+
+            try
+            {
+                passwordResetService.GenerateAndSendPasswordResetLink
+                (
+                    emailAddress,
+                    baseUrl
```
stellake, line 41: Is there a reason we define the baseUrl here and pass it as a param, rather than define it when it needs to be used?

DigitalLearningSolutions.Web/Views/ForgotPassword/Index.cshtml
```diff
@@ -1,19 +1,43 @@
 ﻿@using DigitalLearningSolutions.Web.ViewModels.ForgotPassword
 @model ForgotPasswordViewModel
 @{
-  ViewData["Title"] = "Digital Learning Solutions - Reset Your Password";
+  ViewData["Title"] = "Reset Your Password";
+  var inputErrorClass = !String.IsNullOrWhiteSpace(Model.EmailErrorMessage) ? "nhsuk-input--error" : "";
+  var formErrorClass = !String.IsNullOrWhiteSpace(Model.EmailErrorMessage) ? "nhsuk-form-group--error" : "";
```
stellake, line 6: Would be nice to pull the !String.IsNullOrWhiteSpace() into a variable :)

DigitalLearningSolutions.Web/Controllers/ForgotPasswordController.cs
DanBloxham-sw: As part of the previous set of corrections, the baseUrl is now retrieved from the appsettings under the web project. The data project does not have access to this. We could refactor the service so that the controller makes multiple calls and builds the link itself before passing that to the email sender?

stellake (approved):

PR #234, opened 2021-03-29
HEEDLS-362: Integration tests

stellake:
Set up basic integration tests:
1. Spin up a test server against which we can run some tests.
2. Use test database when running the tests.
3. Add some basic tests around authentication.
4. Make LearningSolution pages (accessibility help, terms) unauthenticated.

Further work:
Add running these tests to CI/CD pipeline.

Note that the build won't pass atm as some tests are broken due to the clocks change (fixing this is next thing on my list)


DanBloxham-sw (approved):
Looks all good to me.

PR #235, opened 2021-03-29
HEEDLS-NONE: Fix broken tests - use UTC time.

stellake:
Previously, default session was created not using UTC time. Hence, the clock changes made the tests failed because the expected time was +1 hour from UTC (British Summer Time).


DanBloxham-sw (approved):
Looks good to me.

PR #236, opened 2021-03-29
HEEDLS-336 - Added empty SignIn and Register pages with links from th…

DanBloxham-sw:
…e home page.

The new sign in and register pages have been made. The /Home landing page has buttons that take the user to these pages (unless they are already logged in, in which case it just reloads /Home again). The old code that automatically logged in as Kevin has been shifted to the SignIn button on the /Login index page so it can still be used while we don't have a proper log in system.

Manually tested IE11, Edge, Chrome, Firefox
Manually tested Chrome mobile view
Manually tested No JS
Tested with screen reader
Ran all unit tests successfully.

UI screen shots:
Home
Login
Register

### DIFF
--- a/DigitalLearningSolutions.Web/Controllers/LoginController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/LoginController.cs
@@ -13,6 +13,18 @@
     {
         public IActionResult Index()
         {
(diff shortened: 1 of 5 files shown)


stellake (approved):
Looks good! 👍

PR #238, opened 2021-03-30
HEEDLS-NONE: Render custom logo only if user is linked to a centre.

stellake:
Workaround on test environment to prevent app from crashing. This change is also required for the future, where a user could be authenticated using OpenID, but not be associated with any centres.


DanBloxham-sw (approved):
All looks good to me.

PR #240, opened 2021-03-31
HEEDLS-363 - Basic login form with error display created

DanBloxham-sw:
The page at /Login now has a basic form for entering details, however the actual logging in functionality does not exist. The error display also has been implemented. Entering the username as "TestUsernameError" or "TestPasswordError" will prevent the automatic log in and allows the UI to display the appropriate errors. The "Remember Me" checkbox is also present, but is non-functional as that will also be dealt with in a different ticket.

Tested with:
- IE11, Edge, Chrome, Firefox
- Chrome mobile view
- No javascript
- With Screen Reader
- Ran all unit tests

There are a couple of things worth noting: 
- Keyboard Focus is not set to the Error Summary when it is first displayed.
- When using NV Access screen reader in Browse Mode, the visual/keyboard focus can differ from normal when encountering errors.

UI screen shots below, one as the basic Login page, one with sample errors displayed (page is partially cropped to focus on the errors).

Login

Login With errors

### DIFF
--- a/DigitalLearningSolutions.Web/Controllers/LoginController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/LoginController.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq;
     using System.Security.Claims;
     using DigitalLearningSolutions.Web.Helpers;
     using DigitalLearningSolutions.Web.ViewModels.Login;
@@ -23,8 +24,7 @@
         }
 
         [HttpPost]
-        [Route("/Login")]
-        public IActionResult Login(LoginViewModel model)
+        public IActionResult Index(LoginViewModel model)
         {
             // TODO: HEEDLS-364 - Overwrite this old code for automatic sign in with new code that signs in the user after validation
             if (model.Username?.ToLower() == "testusernameerror")
--- a/DigitalLearningSolutions.Web/Controllers/LoginController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/LoginController.cs
@@ -4,6 +4,7 @@
     using System.Collections.Generic;
     using System.Security.Claims;
     using DigitalLearningSolutions.Web.Helpers;
+    using DigitalLearningSolutions.Web.ViewModels.Login;
     using Microsoft.AspNetCore.Authentication;
     using Microsoft.AspNetCore.Mvc;
     using Microsoft.FeatureManagement.Mvc;
@@ -21,9 +22,25 @@
             return View();
         }
 
-        public IActionResult SignIn()
+        [HttpPost]
+        [Route("/Login")]
+        public IActionResult Login(LoginViewModel model)
         {
-            // TODO: HEEDLS-364 - Overwrite this old code for automatic sign in with new code that signs in the user
+            // TODO: HEEDLS-364 - Overwrite this old code for automatic sign in with new code that signs in the user after validation
+            if (model.Username?.ToLower() == "testusernameerror")
+            {
+                ModelState.AddModelError("username", "There is a username error.");
+            }
+
+            if (model.Username?.ToLower() == "testpassworderror")
+            {
+                ModelState.AddModelError("password", "There is a password error.");
+            }
+
+            if (!ModelState.IsValid)
+            {
+                return View("Index", model);
+            }
 
             var claims = new List<Claim>
             {
--- a/DigitalLearningSolutions.Web/ViewModels/Login/LoginViewModel.cs
+++ b/DigitalLearningSolutions.Web/ViewModels/Login/LoginViewModel.cs
@@ -1,0 +1,23 @@
+﻿namespace DigitalLearningSolutions.Web.ViewModels.Login
+{
+    using System.ComponentModel.DataAnnotations;
+
+    public class LoginViewModel
+    {
+        public LoginViewModel()
+        {
+            Username = string.Empty;
+            Password = string.Empty;
+            RememberMe = false;
+        }
+
+        [Required]
+        public string Username { get; set; }
+
+        [Required]
+        [DataType(DataType.Password)]
+        public string Password { get; set; }
+
+        public bool RememberMe { get; set; }
+    }
+}
--- a/DigitalLearningSolutions.Web/ViewModels/Login/LoginViewModel.cs
+++ b/DigitalLearningSolutions.Web/ViewModels/Login/LoginViewModel.cs
@@ -11,16 +11,13 @@
             RememberMe = false;
         }
 
-        [Display(Order = -1)]
         [Required(ErrorMessage = "Please enter a username to log in.")]
-        public string Username { get; set; }
+        public string? Username { get; set; }
 
-        [Display(Order = 2)]
         [Required (ErrorMessage = "Please enter a password to log in.")]
         [DataType(DataType.Password)]
-        public string Password { get; set; }
+        public string? Password { get; set; }
 
-        [Display(Order = 3)]
         public bool RememberMe { get; set; }
     }
 }
--- a/DigitalLearningSolutions.Web/ViewModels/Login/LoginViewModel.cs
+++ b/DigitalLearningSolutions.Web/ViewModels/Login/LoginViewModel.cs
@@ -11,13 +11,16 @@
             RememberMe = false;
         }
 
-        [Required]
+        [Display(Order = -1)]
+        [Required(ErrorMessage = "Please enter a username to log in.")]
         public string Username { get; set; }
 
-        [Required]
+        [Display(Order = 2)]
+        [Required (ErrorMessage = "Please enter a password to log in.")]
         [DataType(DataType.Password)]
         public string Password { get; set; }
 
+        [Display(Order = 3)]
         public bool RememberMe { get; set; }
     }
 }
--- a/DigitalLearningSolutions.Web/Views/Login/Index.cshtml
+++ b/DigitalLearningSolutions.Web/Views/Login/Index.cshtml
@@ -1,12 +1,82 @@
-﻿@{
+﻿@using DigitalLearningSolutions.Web.ViewModels.Login
+@model LoginViewModel 
+@{
   ViewData["Title"] = "Login";
+  var errorHasOccurred = !ViewData.ModelState.IsValid;
+  var usernameError = ViewData.ModelState["Username"]?.Errors?.Count > 0;
+  var passwordError = ViewData.ModelState["Password"]?.Errors?.Count > 0;
+  var passwordInputErrorClass = passwordError ? "nhsuk-input--error" : "";
+  var passwordFormErrorClass = passwordError ? "nhsuk-form-group--error" : "";
+  var usernameInputErrorClass = usernameError ? "nhsuk-input--error" : "";
+  var usernameFormGroupError = usernameError ? "nhsuk-form-group--error" : "";
 }
 
 <div class="nhsuk-grid-row">
   <div class="nhsuk-grid-column-full">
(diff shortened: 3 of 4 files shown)


DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -21,9 +22,25 @@ public IActionResult Index()
             return View();
         }
 
-        public IActionResult SignIn()
+        [HttpPost]
+        [Route("/Login")]
```
stellake, line 26: Do we need the route?

DigitalLearningSolutions.Web/ViewModels/Login/LoginViewModel.cs
```diff
@@ -0,0 +1,23 @@
...
+    public class LoginViewModel
+    {
+        public LoginViewModel()
+        {
+            Username = string.Empty;
+            Password = string.Empty;
+            RememberMe = false;
+        }
+
+        [Required]
```
stellake, line 14: Could we potentially modify the error messages that the user gets to be something like "Please enter a username to log in" rather than using "XXX field"?

DigitalLearningSolutions.Web/Views/Login/Index.cshtml
```diff
@@ -1,12 +1,82 @@
...
+      <h1 id="page-heading">Login</h1>
+      <p>Enter your details to log into your account.</p>
+
+      <div class="nhsuk-form-group @usernameFormGroupError">
+        <label class="nhsuk-label" for="Username">
+          Email or user ID
+        </label>
+        @if (usernameError) {
+          <span class="nhsuk-error-message" id="username-error">
+            @foreach (var error in ViewData.ModelState["Username"].Errors) {
```
stellake, line 31: Feels weird to get several username errors. The username either is not entered or is entered and doesn't exist in the database. Am I missing any error cases, or is this to satisfy error summary model?

DigitalLearningSolutions.Web/Views/Login/Index.cshtml
```diff
@@ -1,12 +1,82 @@
...
+              <span class="nhsuk-u-visually-hidden">Error:</span> @error.ErrorMessage
+            }
+          </span>
+        }
+        <input class="nhsuk-input nhsuk-u-width-one-half nhsuk-u-margin-bottom-4 @usernameInputErrorClass" type="text" id="Username" spellcheck="false" autocomplete="email" aria-describedby="username-error" asp-for="Username" />
```
stellake, line 36: I think you forgot the email autocomplete here

DigitalLearningSolutions.Web/Views/Login/Index.cshtml
```diff
@@ -1,12 +1,82 @@
...
+            @foreach (var error in ViewData.ModelState["Password"].Errors) {
+              <span class="nhsuk-u-visually-hidden">Error:</span> @error.ErrorMessage
+            }
+          </span>
+        }
+        <input class="nhsuk-input nhsuk-u-width-one-half nhsuk-u-margin-bottom-4 @passwordInputErrorClass" type="password" id="Password" spellcheck="false" aria-describedby="password-error" asp-for="Password" />
+      </div>
+
+      <div class="nhsuk-checkboxes__item">
+        @Html.CheckBoxFor(model => model.RememberMe, new { @class = "nhsuk-checkboxes__input", id = "RememberMe" })
```
stellake, line 54: Hmm.. Could we not use the HTML input element here?

DigitalLearningSolutions.Web/Views/Login/Index.cshtml
```diff
@@ -1,12 +1,82 @@
...
+        <label class="nhsuk-label nhsuk-checkboxes__label" for="RememberMe">
+          Remember Me
+        </label>
+      </div>
+
+      <div class="nhsuk-inset-text">
+        <span class="nhsuk-u-visually-hidden">Information: </span>
+        <p>
+          By logging in, I consent to my details being stored and processed in line with your
+          <a class="" href="https://www.hee.nhs.uk/about/privacy-notice">Privacy Policy</a>
```
stellake, line 64: Do we need the empty class attribute?

stellake:
This is looking good. Styling-wise, there are a few minor things I noticed from the screenshot:
- It's odd that the spacing between the form elements is bigger than the spacing between the description and email input. This makes it look like title, description and email input are the first section and everything else is a second section, whereas I'd expect the sections to be title/description, then the form as a whole (especially the email and password inputs), then the "Alternatively, ..." section.
- I'd expect the order of errors in the summary to be the same as order of inputs from UX perspective.
- Login should be Log in everywhere (verb)
- Log in and Forgot password buttons have quite a lot of spacing. Could we bring them slightly closer together?
- The "Alternatively" text at the bottom is actually part of a different ticket (I think it wasn't in the "Scope" and it should have been in "Further work" section of the JIRA"), but given that you've added it (which is great!), could we make it clear that it's not part of the form, and add some spacing between the form buttons and that last section?

DigitalLearningSolutions.Web/Views/Login/Index.cshtml
```diff
@@ -1,12 +1,82 @@
...
+          and agree to use the system according to
+          <a class="" asp-controller="LearningSolutions" asp-action="Terms">Terms of Use</a>
+        </p>
+      </div>
+
+      <button class="nhsuk-button" type="submit">Login</button>
+      <a class="nhsuk-button nhsuk-button--secondary nhsuk-u-margin-left-5" asp-controller="ForgotPassword" asp-action="Index">
+        Forgot Password
```
stellake, line 72: Really minor, but could we have "password" with a small letter here?

DigitalLearningSolutions.Web/Controllers/LoginController.cs
DanBloxham-sw: I thought it did in order to get rid of the error summary anchors (e.g. #Username) from the url when resubmitting after clicking one of those links. After retesting, it looks like it is unnecessary (provided the method is renamed Index so the url is not /Login/Login. So I shall remove the route and change it the name to Index.

DigitalLearningSolutions.Web/Views/Login/Index.cshtml
DanBloxham-sw: There should only ever be one username error, but I was trying to keep the error message production as general as possible. The error summary should not affect this. I can either :
- cut it back to just show the first error
- or move everything inside the "if" statement into a new partial view, keeping the "foreach" in case future use of the error message requires multiple errors to be displayed.

DigitalLearningSolutions.Web/Views/Login/Index.cshtml
DanBloxham-sw: The ticket mentions that the username input is "Email/UserId text input" - would the chance of input not being an email address (e.g. UserId) mean that the autocomplete="email" should be removed?

DanBloxham-sw:
Couple of things @stellake :
> * I'd expect the order of errors in the summary to be the same as order of inputs from UX perspective.

It looks like the _ErrorSummary partial view orders the links alphabetically based on the ModelState keys the errors are tied to (the property names in this case: Password, RememberMe, Username). The alphabetical ordering could be removed, but then we are left with the random ordering in the dictionary which still puts password above username. I might be an idiot, but I can't see any easy way to change the OrderBy to use the order of the properties as written in the view model. The best way around I can think of is to make the partial view take in an optional variable that is an array of the key names manually placed in the correct order, and sorts against that (or alphabetically if that variable is null/empty). Either that or rewrite the partial view completely to use HTML helpers that might be able to use the Display(Order) data annotation.

> * The "Alternatively" text at the bottom is actually part of a different ticket (I think it wasn't in the "Scope" and it should have been in "Further work" section of the JIRA"), but given that you've added it (which is great!), could we make it clear that it's not part of the form, and add some spacing between the form buttons and that last section?

Other than greater spacing between the form buttons and the "Alternatively..." section, is there anything else we want to do to differentiate the two sections?

DigitalLearningSolutions.Web/Views/Login/Index.cshtml
stellake: I think unless the component is shared and needs being generic, I wouldn't make the logic more complicated than it needs to be as it will most likely be YAGNI. Hence, I think if it's straight-forward, we should only have one username error.

stellake:
Regarding the order of errors, if this is the case across the site, I will raise a new bug ticket for handling that, so don't worry about it now and add a comment to the ticket for testing.
Regarding the spacing, I think keeping it similar to the spacing on the wireframes should be sufficient.

Hope this makes sense. :)

DigitalLearningSolutions.Web/ViewModels/Login/LoginViewModel.cs
```diff
@@ -0,0 +1,26 @@
...
+    public class LoginViewModel
+    {
+        public LoginViewModel()
+        {
+            Username = string.Empty;
+            Password = string.Empty;
+            RememberMe = false;
+        }
+
+        [Display(Order = -1)]
```
stellake, line 14: Does this actually fix the ordering issue/how does it work when combined with the errors added in controller?

DigitalLearningSolutions.Web/ViewModels/Login/LoginViewModel.cs
```diff
@@ -0,0 +1,26 @@
...
+        [Required(ErrorMessage = "Please enter a username to log in.")]
+        public string Username { get; set; }
```
stellake, line 16: Given that we're doing null checks for this field in the controller, should it be a nullable string?

DigitalLearningSolutions.Web/ViewModels/Login/LoginViewModel.cs
DanBloxham-sw: Short answer: it doesn't do anything to the error summary, and I forgot to remove it.

DigitalLearningSolutions.Web/Views/Login/Index.cshtml
```diff
@@ -1,12 +1,80 @@
-﻿@{
-  ViewData["Title"] = "Login";
+﻿@using DigitalLearningSolutions.Web.ViewModels.Login
+@model LoginViewModel 
+@{
+  ViewData["Title"] = "Log in";
```
stellake, line 4: Should this have "Error" in front when we display the error summary?

DigitalLearningSolutions.Web/Views/Login/Index.cshtml
```diff
@@ -1,12 +1,80 @@
...
+  //ViewBag["ErrorSummaryOrder"] = new [] { "Username", "Password", "RememberMe" };
```
stellake, line 5: Is this left over from dev?

DigitalLearningSolutions.Web/Views/Login/Index.cshtml
```diff
@@ -1,12 +1,79 @@
...
+        }
+        <input class="nhsuk-input nhsuk-u-width-one-half @usernameInputErrorClass" type="text" id="Username" spellcheck="false" autocomplete="email" aria-describedby="username-error" asp-for="Username" />
+      </div>
+
+      <div class="nhsuk-form-group @passwordFormErrorClass">
+        <label class="nhsuk-label" for="Password">
+          Password
+        </label>
+        @if (passwordError) {
+      <span class="nhsuk-error-message" id="password-error">
```
stellake, line 43: The formatting here is a little bit odd - could you run the formatter on this? 🙏

stellake (approved):
Looks good! Feel free to merge after running the formatter :)

PR #242, opened 2021-03-31
HEEDLS-365 - Forgot Password confirm page now links to old /FindYourC…

DanBloxham-sw:
…entre

Tested manually on Chrome.
Ran all unit tests
Tested with screen reader.

UI screenshot with new link:
Confirm


stellake:
Super tiny, but could we have the full stop without the underline? Otherwise looks good to me!

stellake (approved):
Super tiny, but could we have the full stop without the underline? Otherwise looks good to me!

PR #243, opened 2021-03-31
docs: Update readme with Rider and migrations troubleshooting

SteveJacksonSoft:
Rider works! 🎉 
Also left a note that migrations sometimes don't all run the first time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 - [Visual Studio Professional 2019](https://visualstudio.microsoft.com/downloads/)
     - Make sure you have the [NPM Task Runner](https://marketplace.visualstudio.com/items?itemName=MadsKristensen.NPMTaskRunner) extension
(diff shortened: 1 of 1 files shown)


DanBloxham-sw (approved):
Looks good to me!

README.md
```diff
@@ -2,6 +2,7 @@
 
 - [Visual Studio Professional 2019](https://visualstudio.microsoft.com/downloads/)
     - Make sure you have the [NPM Task Runner](https://marketplace.visualstudio.com/items?itemName=MadsKristensen.NPMTaskRunner) extension
+    - JetBrains Rider can work as an alternative. Follow the setup steps laid out below. In addition, you'll have to run `npm run dev` manually to build the JS and SASS.
```
stellake, line 5: Do you think it would be useful to:
- Add a link to Rider (like we have for VS)
- Specify that it's an alternative to VS rather than alternative to NPM Task Runner to be extra clear.

SteveJacksonSoft: Good points - done.

stellake (approved):

PR #244, opened 2021-04-01
HEEDLS-337 - Forgot Password bug fixes and refactoring

DanBloxham-sw:
Added new content and fixed all bugs mentioned in the comments of https://softwiretech.atlassian.net/browse/HEEDLS-337 except for the "Skip to content" link skipping past the error summary - this will require further investigation, probably in another ticket. The ForgotPassword controller and Index page were refactored to use more generic components.

Tested manually in Chrome and ran all Unit tests.

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/Controllers/ForgotPassword/ForgotPasswordControllerTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/Controllers/ForgotPassword/ForgotPasswordControllerTests.cs
@@ -5,7 +5,6 @@
     using DigitalLearningSolutions.Web.Controllers;
     using DigitalLearningSolutions.Web.ViewModels.ForgotPassword;
     using FakeItEasy;
-    using FluentAssertions;
     using FluentAssertions.AspNetCore.Mvc;
     using NUnit.Framework;
 
@@ -23,23 +22,13 @@
         }
 
         [Test]
-        public void Render_basic_form_when_there_is_no_error_model()
-        {
-            // When
-            var result = controller.Index(model: null);
-
-            // Then
-            result.Should().BeViewResult().WithDefaultViewName();
-        }
-
-        [Test]
         public void Successful_email_submission_should_render_confirm_page()
         {
             // Given
             A.CallTo(() => passwordResetService.GenerateAndSendPasswordResetLink(A<string>._, A<string>._)).DoesNothing();
 
             // When
-            var result = controller.Index("recipient@example.com");
+            var result = controller.Index(new ForgotPasswordViewModel("recipient@example.com"));
 
             // Then
             result.Should().BeRedirectToActionResult().WithActionName("Confirm");
@@ -49,7 +38,7 @@
         public void Email_submission_should_call_password_reset_service()
         {
             // When
-            controller.Index("recipient@example.com");
+            controller.Index(new ForgotPasswordViewModel("recipient@example.com"));
 
             // Then
             A.CallTo(() => passwordResetService.GenerateAndSendPasswordResetLink(A<string>._, A<string>._)).MustHaveHappened();
@@ -64,11 +53,12 @@
             A.CallTo(() => passwordResetService.GenerateAndSendPasswordResetLink(A<string>._, A<string>._)).Throws(new EmailAddressNotFoundException(errorMessage));
 
             // When
-            var result = controller.Index("recipient@example.com");
+            var result = controller.Index(new ForgotPasswordViewModel("recipient@example.com"));
 
             // Then
             result.Should().BeViewResult().WithDefaultViewName()
-                .ModelAs<ForgotPasswordViewModel>().EmailErrorMessage.Should().NotBeNullOrWhiteSpace();
+                .ModelAs<ForgotPasswordViewModel>();
+            Assert.IsTrue(controller.ModelState.IsValid is false);
         }
 
         [Test]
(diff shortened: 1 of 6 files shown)


DigitalLearningSolutions.Web.Tests/Controllers/ForgotPassword/ForgotPasswordControllerTests.cs
```diff
@@ -22,24 +21,14 @@ public void SetUp()
             controller = new ForgotPasswordController(passwordResetService);
         }
 
-        [Test]
-        public void Render_basic_form_when_there_is_no_error_model()
```
stellake, line 26: I might be missing something but why are we removing this test?

DigitalLearningSolutions.Web.Tests/Controllers/ForgotPassword/ForgotPasswordControllerTests.cs
```diff
@@ -64,11 +53,12 @@ public void Bad_email_submission_should_render_basic_form_with_error()
...
 
             // When
-            var result = controller.Index("recipient@example.com");
+            var result = controller.Index(new ForgotPasswordViewModel("recipient@example.com"));
 
             // Then
             result.Should().BeViewResult().WithDefaultViewName()
-                .ModelAs<ForgotPasswordViewModel>().EmailErrorMessage.Should().NotBeNullOrWhiteSpace();
+                .ModelAs<ForgotPasswordViewModel>();
+            Assert.IsTrue(controller.ModelState.IsValid is false);
```
stellake, line 61: Could this be simplified with IsFalse?

DigitalLearningSolutions.Web/Controllers/ForgotPasswordController.cs
```diff
@@ -16,19 +16,22 @@ public ForgotPasswordController(IPasswordResetService passwordResetService)
         }
 
         [HttpGet]
-        public IActionResult Index(ForgotPasswordViewModel? model)
+        public IActionResult Index()
         {
-            model ??= new ForgotPasswordViewModel();
-            return View(model);
+            if (User.Identity.IsAuthenticated)
```
stellake, line 21: Can the User or Identity be null (I remember us handling nulls in LogoViewComponent previously)?

DigitalLearningSolutions.Web.Tests/Controllers/ForgotPassword/ForgotPasswordControllerTests.cs
DanBloxham-sw: I refactored the code around to use ModelState for errors rather than a custom error property (to keep things cleaner, and so that the generic _ErrorSummary partial view worked correctly). That test was making sure that when the custom error property was null, the basic form was shown. Without that property, that test would effectively become "the Index action returns the index page" which I did not think was really worth keeping in. Do we still want that (or something similar in its place)?

DigitalLearningSolutions.Web/Controllers/ForgotPasswordController.cs
```diff
@@ -37,14 +40,15 @@ public IActionResult Index(string? emailAddress)
...
-                    emailAddress,
+                    model.EmailAddress,
                     baseUrl
                 );
             }
             catch (EmailAddressNotFoundException)
             {
-                ForgotPasswordViewModel model = GetEmailNotFoundErrorModel(emailAddress);
-                return Index(model);
+                ModelState.AddModelError("EmailAddress","User with this email address does not exist. Please try again.");
```
stellake, line 49: I think this will need a space between , and " - probably worth running the formatter quickly.

DigitalLearningSolutions.Web/Controllers/ForgotPasswordController.cs
DanBloxham-sw: VS seems to be fine with this, no warnings are present to say it could be null. I suspect this is because it is within a controller action rather than a ViewComponent as LogoViewComponent was, but I might need to do some more reading to confirm that as the exact reason. Either way, its probably safe as is.

DigitalLearningSolutions.Web.Tests/Controllers/ForgotPassword/ForgotPasswordControllerTests.cs
stellake: Yes, I think it's worth keeping, especially with the redirect logic in place for logged in users.

DigitalLearningSolutions.Web.Tests/Controllers/ForgotPassword/ForgotPasswordControllerTests.cs
```diff
@@ -1,45 +1,87 @@
...
 
             // Then
             result.Should().BeViewResult().WithDefaultViewName();
         }
 
+        [Test]
+        public void Index_should_redirect_if_user_is_authenticated()
+        {
+            // Given user is authenticated
+            var user = new ClaimsPrincipal(new ClaimsIdentity("mock"));
```
stellake, line 56: Sometimes when commenting is required in code to explain what the code is doing, a small restructure of the code can help. In this case, it might be clearer to have a function like getControllerWithMockUser(), so the test is more to the point. Then, if anyone wants to see how the mock user is set up, they can simply navigate into the private function. :)
e.g.

`// Given
controller = getControllerWithMockUser()

// When
var result = controller.Index();

// Then
result.Should().BeRedirectToActionResult().WithControllerName("Home").WithActionName("Index");`

DigitalLearningSolutions.Web.Tests/Controllers/ForgotPassword/ForgotPasswordControllerTests.cs
```diff
@@ -1,45 +1,87 @@
...
-    using FluentAssertions;
     using FluentAssertions.AspNetCore.Mvc;
+    using Microsoft.AspNetCore.Http;
+    using Microsoft.AspNetCore.Mvc;
     using NUnit.Framework;
 
     internal class ForgotPasswordControllerTests
     {
         private ForgotPasswordController controller;
+        private ISession httpContextSession;
```
stellake, line 18: Minor but we could just do `Session = new MockHttpContextSession() ` inline when defining the controller (or define the session variable in the scope of a function where we need it. :)

stellake (approved):
Looks good - just a minor comment about the unit test.

PR #245, opened 2021-04-06
Heedls 338 homepage

SteveJacksonSoft:
I think this looks fairly decent.
I've messed with breakpoints a bit to get something alright-looking on mobile by putting the logos for the learning content items above the text and centering them: 
image

Though I have two concerns with this.

1. Because the second learning content item (reasonable adjustment flag) has the icon to the right of the text on desktop-size screen, a screenreader will go to the logo after the text here. The `aria-flowto` attribute would allow us to re-order the flow, but I don't know if we could make it screen-width-dependent.
2. (A lesser worry) The Product items look better left-aligned because they have bullet points, but the learning content looks better centered because of the images. So they're not very consistent. 
image


DigitalLearningSolutions.Web/Views/Home/Index.cshtml
```diff
@@ -1,16 +1,111 @@
...
+    <h1 id="page-heading">A platform for digital learning delivery</h1>
+
+    <p id="page-subheading">Enabling health and care organisations to deliver digital learning to staff and students.</p>
+
+    <a class="nhsuk-button auth-button auth-button--left" asp-controller="Login" asp-action="Index">
+      Sign in
+    </a>
+    <a class="nhsuk-button nhsuk-button--secondary auth-button" asp-controller="Register" asp-action="Index">
+      Register
+    </a>
```
DanBloxham-sw, line 18: The scope in the ticket mentions "If the user is logged in, the Login and Registration buttons should be replaced with an “Application selector” button that takes the user to the application selector on the old site." Is that going to be handled in another ticket?

DigitalLearningSolutions.Web/Views/Home/Index.cshtml
```diff
@@ -1,16 +1,111 @@
...
+    <h2 id="learningcontent">Learning Content</h2>
+    <ul class="content-area-list">
+      <li>
+        <div class="content-area--with-image--image-before content-area--with-image content-area nhsuk-card nhsuk-card--clickable">
+          <div class="content-area__image-box content-area__image-box--left">
+            <img class="content-area__image" src="images/itSkillsPathwayLogo.png" alt="It Skills Pathway logo"/>
+          </div>
+          <div>
+            <h3 class="learning-content__item-text learning-content__item-header">
+              <a href="https://www.dls.nhs.uk/Learning?brand=ITSkillsPathway">
```
DanBloxham-sw, line 66: I know in some of the work that I have done I have referenced the old site using the CurrentSystemBaseUrl config value in the appsettings.json files (e.g. on the /ForgotPassword/Confirm.cshtml page for the FindYourCentre link). @stellake do you know if we want to use the config value in this case, or is the href fine as it is?

DanBloxham-sw (commented):
When I've been doing my testing, I've used a minimum width of 320px as mentioned in the testing document on the swiki https://softwiretech.atlassian.net/wiki/spaces/HEE/pages/6703648740/Testing - not sure if we still want to keep this as the minimum, but it might help keep things looking slightly nicer when looking at the display.

For concern 1., off the top of my head the only way you might be able to get aria-flowto to work based on screen width is by implementing some javascript to get the screen width and setting the value of aria-flowto as appropriate. However, it would mean some dependence on JS is necessary otherwise the same issue still arises. My personal vote would be to change the design around so that the images are all on the left hand side - but then my aesthetic taste is awful and would want confirmation from someone with a better eye than mine that it would be a fine change. 

For concern 2., I think the consistency across those two sections doesn't really matter that much when it comes to the alignment. If different alignments look best, then I'd say go with different alignments.

DigitalLearningSolutions.Web/Views/Home/Index.cshtml
stellake: I think we should use the config value :)

stellake:
The probability of getting a user with a screen width less than 320px is extremely low and most of the users will be on desktop (as the old site doesn't even support mobile). Hence, I think keeping 320px as minimum makes sense (plus it's agreed with the client!). This is interesting for references on screen sizes: https://ux.stackexchange.com/questions/74798/are-there-devices-narrower-than-320px-and-data-on-their-usage-for-web-browsing 

Regarding the issues around the design - it's looking more likely that we'll be using the Mini-Hub page design, so things are likely to move around a bit. I don't want to dig too deep into the issues before we get that decision confirmed, but in general HEE is quite flexible with moving the designs around if it's well justified. :)

DigitalLearningSolutions.Web/Views/Home/Index.cshtml
SteveJacksonSoft: Done.

DigitalLearningSolutions.Web.IntegrationTests/AuthenticationTests.cs
```diff
@@ -1,36 +1,53 @@
...
+
+            // Act
+            var response = await client.GetAsync(url);
+
+            // Assert
+            response.EnsureSuccessStatusCode();
+        }
+
+        [Fact]
+        public async Task BaseUrlRedirectsToWelcomePage()
```
SteveJacksonSoft, line 40: I don't think this belongs in AuthenticationTests, but I don't know whether to change the class name or move this test into a separate class. Thoughts?

DigitalLearningSolutions.Web/Controllers/HomeController.cs
```diff
@@ -1,12 +1,69 @@
...
+        }
+
+        [HttpGet]
+        [Route("Home/Section/LearningContent")]
+        public IActionResult LearningContent()
+        {
+            return View(GetLandingPageViewModel(2));
+        }
+
+        private LandingPageViewModel GetLandingPageViewModel(int sectionIndex)
```
SteveJacksonSoft, line 55: I'm not massively happy about code which calls this having to know what order the sections are in.
I think I could change MiniHubNavigationModel.Sections to from a list to a Dictionary<T, MiniHubSection> where T : Enumeration, where Enumeration is a base enumeration class as in https://docs.microsoft.com/en-us/dotnet/architecture/microservices/microservice-ddd-cqrs-patterns/enumeration-classes-over-enum-types.

That may be me getting a bit paranoid and excited, so let me know what you think.

DigitalLearningSolutions.Web.IntegrationTests/AuthenticationTests.cs
stellake: For now, I think I'd keep it here and change the class name as it is all testing the Microsoft Identity stuff. If this class starts getting too big, we can separate it out. :)

.editorconfig
```diff
@@ -6,6 +6,9 @@ indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
 
+# ReSharper properties
+resharper_csharp_wrap_lines = true
```
stellake, line 10: 🤩

DigitalLearningSolutions.Web/Views/Home/LearningContent.cshtml
```diff
@@ -0,0 +1,73 @@
+@using DigitalLearningSolutions.Web.ViewComponents
+@using DigitalLearningSolutions.Web.ViewModels.Home
+@model LandingPageViewModel
+
+<link rel="stylesheet" href="@Url.Content("~/css/home/learningContent.css")"/>
+
+<div class="nhsuk-grid-row">
+  <div id="column" class="nhsuk-grid-column-full">
+
+    <h2>Learning Content</h2>
```
stellake, line 10: I think we might want these to be h1s as per https://service-manual.nhs.uk/design-system/patterns/mini-hub
Also, could we add the font style explicitly here - https://service-manual.nhs.uk/design-system/styles/typography#headings

DigitalLearningSolutions.Web/Views/Home/Products.cshtml
```diff
@@ -0,0 +1,48 @@
...
+
+    @await Component.InvokeAsync(
+      typeof(ContentsListViewComponent),
+      new { miniHubNavigationModel = Model.MiniHubNavigationModel })
+
+        <div class="content-area">
+          <h3>Tracking System</h3>
+          <p>Distribute and track digital learning</p>
+            <ul class="__sub-list">
+            <li>Provide your staff or students with access to quality assured digital learning content</li>
```
stellake, line 19: This indentation looks a bit off - maybe worth running the formatter quickly :)

DigitalLearningSolutions.Web/Views/Home/Welcome.cshtml
```diff
@@ -0,0 +1,38 @@
...
+      </a>
+    }
+
+    @await Component.InvokeAsync(
+      typeof(ContentsListViewComponent),
+      new { miniHubNavigationModel = Model.MiniHubNavigationModel })
+
+    <p>Our products and services can help you manage digital learning delivery for the benefit of the people in your organisation.</p>
+
+    @await Component.InvokeAsync(typeof(PaginationViewComponent), new { miniHubNavigationModel = Model.MiniHubNavigationModel })
```
stellake, line 36: In general, I think the tag-helper syntax for vcs is much nicer: https://docs.microsoft.com/en-us/aspnet/core/mvc/views/view-components?view=aspnetcore-3.1

I think we already use it for back button in the LearningContent page. :)

DigitalLearningSolutions.Web/Views/Shared/Components/Pagination/Default.cshtml
```diff
@@ -0,0 +1,33 @@
...
+          <svg class="nhsuk-icon nhsuk-icon__arrow-left" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" aria-hidden="true">
+            <path d="M4.1 12.3l2.7 3c.2.2.5.2.7 0 .1-.1.1-.2.1-.3v-2h11c.6 0 1-.4 1-1s-.4-1-1-1h-11V9c0-.2-.1-.4-.3-.5h-.2c-.1 0-.3.1-.4.2l-2.7 3c0 .2 0 .4.1.6z"></path>
+          </svg>
+        </a>
+      </li>
+    }
+    @if (Model.CurrentSectionIndex < Model.Sections.Count - 1) {
+      <li class="nhsuk-pagination-item--next">
+        @{ var nextSection = Model.Sections[Model.CurrentSectionIndex + 1]; }
+        <a class="nhsuk-pagination__link nhsuk-pagination__link--next" href="@Url.Action(nextSection.ActionName, nextSection.ControllerName)">
```
stellake, line 22: Is there a benefit to using href over MVC asp-action/asp-controller attributes?

DigitalLearningSolutions.Web/Controllers/HomeController.cs
```diff
@@ -1,12 +1,69 @@
...
+            this.configuration = configuration;
+        }
+
+        public IActionResult Index()
+        {
+            return RedirectToAction("Welcome");
+        }
+
+        [HttpGet]
+        [Route("Home/Section/Welcome")]
```
stellake, line 35: Is there a reason we're not using the default routes (Home/Welcome etc)?

DigitalLearningSolutions.Web/Controllers/HomeController.cs
stellake: This sounds really cool, but I think that because of the scope change of the home page, we shouldn't spend ages trying to perfect everything as I think the code is readable/understandable as is.
If you feel like you have some time in hand and that it'll be worth in the long run, feel free to timebox implementing it - I trust your judgement :)

DigitalLearningSolutions.Web/ViewModels/Common/MiniHub/MiniHubSection.cs
```diff
@@ -0,0 +1,9 @@
+namespace DigitalLearningSolutions.Web.ViewModels.Common.MiniHub
+{
+    public class MiniHubSection
+    {
+        public string ControllerName { get; set; }
```
stellake, line 5: Should we have these as get-only properties and set them in a constructor (or are we expecting to change them after setting them in the first place)?

DigitalLearningSolutions.Web/Views/Home/Products.cshtml
```diff
@@ -0,0 +1,48 @@
+@using DigitalLearningSolutions.Web.ViewComponents
+@using DigitalLearningSolutions.Web.ViewModels.Home
+@model LandingPageViewModel
+
+<link rel="stylesheet" href="@Url.Content("~/css/home/products.css")" />
+
```
stellake, line 6: Ah - we probably want to set the page title here & in the Learning Content pages :)

stellake (commented):
Looks good in general - main comments around using the tag helper syntax and the MVC syntax asp-action/asp-controller on the links. :)

DigitalLearningSolutions.Web/Styles/home/learningContent.scss
```diff
@@ -0,0 +1,35 @@
...
+
+.learning-content-item__text {
+  @include govuk-media-query($until: tablet) {
+    margin-left: auto;
+    margin-right: auto;
+    text-align: center;
+  }
+}
+
+.learning-content-item__details-container {
```
stellake, line 18: Just wanted to double check that none of these NHS components do this already?
https://service-manual.nhs.uk/design-system/styles/layout

DigitalLearningSolutions.Web/Views/Home/LearningContent.cshtml
SteveJacksonSoft: Good point. I'm also thinking we should add a sub-heading to the Products and LearningContent sections so it's clear the user's in the landing page. Not sure what it should be though. "Welcome"? Feels a bit weak...

stellake: If unsure, you could check in the external HEE channel as they're in charge of the content anyway. :)

DigitalLearningSolutions.Web/Controllers/HomeController.cs
SteveJacksonSoft: Not really. I was following the NHS Pagination example, but it seems I was following it wrong and its backend was doing some fancy stuff. I'll remove the `Section/` and use the defaults.

DigitalLearningSolutions.Web/Controllers/HomeController.cs
SteveJacksonSoft: May not be worth it if we only ever use a mini-hub pattern here actually. Tech debt ticket? Or you could just keep it in your head in case another mini-hub comes up in the future.

DigitalLearningSolutions.Web/Styles/home/learningContent.scss
SteveJacksonSoft: Nothing there seems perfectly suited to making a column fit the width of the image, but I may have missed something...

stellake: Yeah, looks like they're mainly using grid layout for this, but it might not be suitable for this case: https://service-manual.nhs.uk/design-system/styles/layout

DigitalLearningSolutions.Web/Controllers/HomeController.cs
```diff
@@ -1,12 +1,69 @@
...
+        {
+            this.configuration = configuration;
+        }
+
+        public IActionResult Index()
+        {
+            return RedirectToAction("Welcome");
+        }
+
+        [HttpGet]
```
stellake, line 37: Is this attribute necessary?

DigitalLearningSolutions.Web/Controllers/HomeController.cs
```diff
@@ -1,12 +1,69 @@
...
+            return View(GetLandingPageViewModel(2));
+        }
+
+        private LandingPageViewModel GetLandingPageViewModel(int sectionIndex)
+        {
+            return new LandingPageViewModel
+            {
+                MiniHubNavigationModel = new MiniHubNavigationModel(LandingPageMiniHubName)
+                {
+                    Sections = sections,
```
stellake, line 61: Is there a reason these are not set in the constructor, but the page name is?

DigitalLearningSolutions.Web/Styles/home/homeShared.scss
```diff
@@ -0,0 +1,34 @@
+@import "~nhsuk-frontend/packages/core/all";
+
+.landing-page__body-text {
+  @extend .nhsuk-body-m;
```
stellake, line 4: If this is just an extension, is there a reason we're not using the nhsuk-body-m class directly to avoid an unnecessary layer of nesting? [same question for content-area__heading]

DigitalLearningSolutions.Web/Views/Home/LearningContent.cshtml
```diff
@@ -0,0 +1,75 @@
...
+@model LandingPageViewModel
+
+@{
+  ViewData["Title"] = "Welcome - Learning Content";
+}
+
+<link rel="stylesheet" href="@Url.Content("~/css/home/learningContent.css")"/>
+
+<div class="nhsuk-grid-row">
+  <div id="column" class="nhsuk-grid-column-full">
```
stellake, line 12: Out of curiosity, what are these ids used for?

DigitalLearningSolutions.Web/Views/Home/LearningContent.cshtml
```diff
@@ -0,0 +1,75 @@
...
+
+    <vc:mini-hub-heading mini-hub-navigation-model="@Model.MiniHubNavigationModel" />
```
stellake, line 14: Just a thought - are some of these VCs simple enough to be partial views instead? (an example can be found on LearningPortal/Current page)

DigitalLearningSolutions.Web/Views/Home/LearningContent.cshtml
```diff
@@ -0,0 +1,75 @@
...
+        </div>
+        <div class="learning-content-item__text-container">
+          <p class="learning-content-item__text">Learning resources relating to Clinical and Pharmacy Terminology and Clinical Classifications products, including ICD, OPCS, SNOMED CT and dm+d. These resources will help you to understand and make best use of these products.</p>
+          <a class="learning-content-item__text" asp-controller="LearningContent" asp-action="TerminologyAndClassificationsDeliveryService">See more</a>
+        </div>
+      </div>
+
+    </div>
+
+    @await Component.InvokeAsync(typeof(PaginationViewComponent), new { miniHubNavigationModel = Model.MiniHubNavigationModel })
```
stellake, line 72: Can we use the tag helper syntax here too?

DigitalLearningSolutions.Web/Views/Home/Welcome.cshtml
```diff
@@ -0,0 +1,36 @@
...
+
+<div class="nhsuk-grid-row">
+  <div id="column" class="nhsuk-grid-column-full">
+    <h1 class="nhsuk-heading-xl">A platform for digital learning delivery</h1>
+
+    <p class="nhsuk-u-font-size-32">Enabling health and care organisations to deliver digital learning to staff and students.</p>
+
+    @if (Model.UserIsLoggedIn) {
+      <a class="nhsuk-button auth-button auth-button--left" href="@Model.CurrentSiteBaseUrl/Home?action=appselect">
+        Application&nbsp;selector
```
stellake, line 19: Should we change this to "Switch application" based on the feedback from Show & Tell? :)

DigitalLearningSolutions.Web/Views/Shared/Components/ContentsList/Default.cshtml
```diff
@@ -0,0 +1,21 @@
...
+<nav class="nhsuk-contents-list" role="navigation" aria-label="Pages in this guide">
+  <h2 class="nhsuk-u-visually-hidden">Contents</h2>
+  <ol class="nhsuk-contents-list__list">
+    @for (var pageIndex = 0; pageIndex < @Model.Sections.Count; pageIndex++) {
+      <li class="nhsuk-contents-list__item" aria-current="page">
+        @{ var sectionToRender = Model.Sections[pageIndex]; }
+        @if (Model.CurrentSectionIndex == pageIndex) {
+          <span class="nhsuk-contents-list__current">@sectionToRender.SectionTitle</span>
+        } else {
+          @Html.ActionLink(sectionToRender.SectionTitle,
```
stellake, line 13: Could that just be an anchor element?

DigitalLearningSolutions.Web/Views/Shared/Components/ContentsList/Default.cshtml
```diff
@@ -0,0 +1,21 @@
+@using DigitalLearningSolutions.Web.ViewModels.Common.MiniHub
```
stellake, line 1: Would it be worth putting the Mini-Hub related shared components into a separate minihub foldeR?

DigitalLearningSolutions.Web/Controllers/HomeController.cs
SteveJacksonSoft: Possibly not, though I find it helps make the code quicker to read / understand, and I don't know of any drawbacks - do you think it's better to remove it.?

DigitalLearningSolutions.Web/Styles/home/homeShared.scss
SteveJacksonSoft: Consistency was my reasoning - I guess we want the body text to be the same across the landing page, so if we should decide to change the size we only need to make one change rather than finding all the nhsuk-body-m sized text across the 3 views.

DigitalLearningSolutions.Web/Views/Home/LearningContent.cshtml
SteveJacksonSoft: Not sure. That was copied from the welcome page I was working from. Want me to check?

DigitalLearningSolutions.Web/Views/Home/LearningContent.cshtml
SteveJacksonSoft: D'oh

DigitalLearningSolutions.Web/Views/Shared/Components/ContentsList/Default.cshtml
SteveJacksonSoft: Yeah, you're probably right.

DigitalLearningSolutions.Web/Controllers/HomeController.cs
stellake: To me, it makes the code a bit clunkier to read, but it's so marginal that if it helps you, I'm happy for it to stay! :)

DigitalLearningSolutions.Web/Views/Home/LearningContent.cshtml
stellake: Yes please, it doesn't look like they're used for styling and we don't need the ids in code.

DigitalLearningSolutions.Web/Views/Home/Welcome.cshtml
```diff
@@ -0,0 +1,35 @@
...
+
+<div class="nhsuk-grid-row">
+  <div id="column" class="nhsuk-grid-column-full">
+    <h1 class="nhsuk-heading-xl">A platform for digital learning delivery</h1>
+
+    <p class="nhsuk-u-font-size-32">Enabling health and care organisations to deliver digital learning to staff and students.</p>
+
+    @if (Model.UserIsLoggedIn) {
+      <a class="nhsuk-button auth-button auth-button--left" href="@Model.CurrentSiteBaseUrl/Home?action=appselect">
+        Switch&nbsp;application
```
stellake, line 18: I think we should handle it just like "Log in" and other buttons, using a normal space, unless the NHS buttons automatically tries to have a new line here.

DigitalLearningSolutions.Web/Styles/home/homeShared.scss
stellake: Hmm.. I feel like the pages are quite different in terms of layout, so I don't agree with the "consistency" aspect here. I can see someone wanting to modify the Welcome page body text and accidentally affecting the other pages too. However, I can see how the consistency matters within the Learning Content cards for example or the common elements that you've put into a partial. ⭐

stellake (approved):
Looks good - a few minor comments, but happy for you to merge it without rereview after resolving these 👍

PR #248, opened 2021-04-09
HEEDLS-342 Added new links to the main page navigation bar

davidm-m:


DigitalLearningSolutions.Web/Views/Shared/_Layout.cshtml
```diff
@@ -78,35 +79,113 @@
...
+            @{
+              string helpPage;
+              if (User.HasClaim(CustomClaimTypes.UserUserAdmin, "True")) {
+                helpPage = "/help/Introduction.html";
+              } else if (User.Identity.IsAuthenticated) {
+                helpPage = "/learningportal/help/LearningPortal.html";
+              } else {
+                helpPage = "/help/Introduction.html";
+              }
+            }
```
davidm-m, line 130: I'm not sure if there's a more idiomatic way of setting this, my C# is a bit rusty.

DigitalLearningSolutions.Web/Views/Shared/_Layout.cshtml
```diff
@@ -44,8 +45,8 @@
       <div class="nhsuk-width-container nhsuk-header__container">
         <div class="nhsuk-header__logo">
           <a class="nhsuk-header__link nhsuk-header__link--service"
-             href="@(ViewData["HeaderPath"] ?? @Configuration["CurrentSystemBaseUrl"] + "/Home?action=appselect")"
-             aria-label="NHS Digital Learning Solutions - @(ViewData["HeaderPathName"] ?? "Switch Application")">
+             href="@(ViewData["HeaderPath"] ?? @Configuration["CurrentSystemBaseUrl"] + "/Home")"
```
DanBloxham-sw, line 48: The Configuration["CurrentSystemBaseUrl"] is used to point to the old site. Configuration["AppRootPath"] points to the new site, and I think that should be used instead for most of these links. In theory, these should end up the same, but in the dev environment, we probably want AppRootPath so we can navigate between the different links without running the old code. The ticket mentions which links "currently navigates to old site" but I'm not sure if that means the new links should point there as well. I reckon the Welcome link should use AppRootPath but FindYourCentre should use CurrentSystemBaseUrl, but I think we'd need some clarity on that @stellake

DigitalLearningSolutions.Web/Views/Shared/_Layout.cshtml
DanBloxham-sw: I think this is the best way we have. Might be able to check if the UserUserAdmin is False and display the learningportal.html, otherwise should introduction, but I am going to need to double check that the log in system actually works such that that would be feasible.

DanBloxham-sw:
Minor thing I've also noticed is that the highlighting of link tabs looks a bit funny. When logged in, some of the link text expands on to multiple lines, widening the nav bar. When hovering over a multiple line link, the entire section of the navbar is highlighted, but when hovering over a single line link, only a smaller portion is highlighted. Not sure what the best way to fix this would be, if at all.
image
image

davidm-m:
> 
> 
> When logged in, some of the link text expands on to multiple lines, widening the nav bar. 

That's my bad, I noticed this and discussed it with Stella already. I meant to get rid of the Products and Learning content links which 
only link to different parts of the homepage anyway.

DigitalLearningSolutions.Web/Views/Shared/_Layout.cshtml
```diff
@@ -78,35 +79,113 @@
...
+                Welcome
                 <svg class="nhsuk-icon nhsuk-icon__chevron-right" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" aria-hidden="true">
                   <path d="M15.5 12a1 1 0 0 1-.29.71l-5 5a1 1 0 0 1-1.42-1.42l4.3-4.29-4.3-4.29a1 1 0 0 1 1.42-1.42l5 5a1 1 0 0 1 .29.71z"></path>
                 </svg>
               </a>
             </li>
             <li class="nhsuk-header__navigation-item">
-              <a class="nhsuk-header__navigation-link" href="@Configuration["CurrentSystemBaseUrl"]/learningportal/help/LearningPortal.html" target="_blank">
-                Help
+              <a class="nhsuk-header__navigation-link" href="@Configuration["CurrentSystemBaseUrl"]/Home#products">
```
stellake: This is a last minute scope change (so sorry), but can we remove the products & learning content? 🙏

DigitalLearningSolutions.Web/Views/Shared/_Layout.cshtml
stellake: "Currently navigates to the old site" means that we should navigate to old site, but this will be changed in the future - sorry about the slightly confusing wording! 🙏 

I think doing something special in dev is not necessary - I'd just use AppRootPath for new pages & CurrentSystemBaseUrl for old pages. :)

DigitalLearningSolutions.Web/Views/Shared/_Layout.cshtml
DanBloxham-sw: I've done some extra digging on this - it looks like the only guaranteed way to check if a user is an admin is to check if the User has a claim of CustomClaimTypes.UserAdminId - if the user is NOT an admin then this claim will never be added. 

Other than that, I think the code is fine. I don't think there is a way to make those checks any better due to the way that the log in system mashes the different claims together if the user has both admin and delegate accounts.

DigitalLearningSolutions.Web/Views/Shared/_Layout.cshtml
```diff
@@ -1,4 +1,5 @@
-﻿@using Microsoft.Extensions.Configuration
+﻿@using DigitalLearningSolutions.Web.Helpers
```
stellake, line 1: Just double checking whether these changes affect LearningPortal (e.g. /LearningPortal/Current) - the Learning Portal's navigation bar should not be getting all the new tabs (except for maybe My Account, which seems reasonable). Note that the LP user will never be signed out so including the sign in/registration buttons in the common layout won't be an issue!

DigitalLearningSolutions.Web/Views/Shared/_Layout.cshtml
stellake: Checking for admin id sounds good - regarding the if/else if/else - I think the logic is good as is, because HEE will come back to us with a "not logged in" resource at some point and it will make the change easy.

DigitalLearningSolutions.Web/Views/Shared/_Layout.cshtml
davidm-m: Ah, this is affecting the LP nav bar - I'll see if I can find a way around that.

davidm-m: Whenever custom nav bar options are provided now, we won't show the Welcome, Find your centre, Pricing or Help links. Login and register/My account, switch application and Log out will still be shown depending on whether they're logged in.
Is that what you were looking for? The LearningPortal nav bar now looks like this:
image

stellake: Yes, this is exactly what I was looking for (mainly just not affecting the LP functionality)! I think you're right in highlighting keeping the Help, because it was there before. 👍

DigitalLearningSolutions.Web/Views/Shared/_Layout.cshtml
stellake, line 48: I think you've forgotten to change the base url here :)

davidm-m: Nice spot 😄

stellake (approved):

PR #249, opened 2021-04-09
HEEDLS-344: Add 'My Account' page without details expandable

JBsoftwire:
This creates the MyAccount page as specified on HEEDLS-344. It doesn't create any way to get to that page except by entering the URL directly, as one wasn't specified on the ticket and I assumed it was to be added later.

### DIFF
--- a/DigitalLearningSolutions.Data/Services/CentresService.cs
+++ b/DigitalLearningSolutions.Data/Services/CentresService.cs
@@ -1,7 +1,7 @@
 ﻿namespace DigitalLearningSolutions.Data.Services
 {
-    using System.Data;
     using Dapper;
+    using System.Data;
 
     public interface ICentresService
     {
--- a/DigitalLearningSolutions.Data/Services/CentresService.cs
+++ b/DigitalLearningSolutions.Data/Services/CentresService.cs
@@ -6,6 +6,7 @@
     public interface ICentresService
     {
         string? GetBannerText(int centreId);
+        string? GetCentreName(int centreId);
     }
 
     public class CentresService : ICentresService
@@ -21,6 +22,16 @@
         {
             return connection.QueryFirstOrDefault<string?>(
                 @"SELECT BannerText
+                        FROM Centres
+                        WHERE CentreID = @centreId",
+                new { centreId }
+            );
+        }
+
+        public string? GetCentreName(int centreId)
+        {
+            return connection.QueryFirstOrDefault<string?>(
+                @"SELECT CentreName
                         FROM Centres
                         WHERE CentreID = @centreId",
                 new { centreId }
--- a/DigitalLearningSolutions.Data/Services/CentresService.cs
+++ b/DigitalLearningSolutions.Data/Services/CentresService.cs
@@ -45,6 +45,8 @@
                     $"No centre found for centre id {centreId}"
                 );
             }
+
+            return name;
         }
     }
 }
--- a/DigitalLearningSolutions.Data/Services/CentresService.cs
+++ b/DigitalLearningSolutions.Data/Services/CentresService.cs
@@ -13,9 +13,9 @@
     public class CentresService : ICentresService
     {
         private readonly IDbConnection connection;
-        private readonly ILogger<CourseService> logger;
+        private readonly ILogger<CentresService> logger;
 
-        public CentresService(IDbConnection connection, ILogger<CourseService> logger)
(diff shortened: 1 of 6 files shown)


DigitalLearningSolutions.Data/Services/CentresService.cs
```diff
@@ -26,5 +27,15 @@ FROM Centres
                 new { centreId }
             );
         }
+
+        public string? GetCentreName(int centreId)
```
stellake, line 31: Hmm.. I think we'd never expect the centre name to be null.. Maybe we could log it if it's null as it is unexpected?

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
```
stellake, line 4: Are these used? It might be worth doing a quick Cleanup Code for auto-formatting the files you've touched :) (we're currently not running any formatting checks on pipeline)

DigitalLearningSolutions.Data/Services/CentresService.cs
```diff
@@ -6,6 +6,7 @@
     public interface ICentresService
```
stellake, line 6: Could we add some unit tests for the new method? :)

DigitalLearningSolutions.Web/Views/MyAccount/Index.cshtml
```diff
@@ -0,0 +1,67 @@
...
+              @Model.Centre
+            </dd>
+          </div>
+
+          <div class="nhsuk-summary-list__row">
+            <dt class="nhsuk-summary-list__key">
+              User:
+            </dt>
+            <dd class="nhsuk-summary-list__value">
+              @Model.User
```
stellake, line 32: Can we also check if the Model.User exists, before presenting this. In the live system, I don't see this field when I log in with my delegate account (I do see it with my admin account), but I'm not sure why it's not set on the claims. I will investigate, but I think the best thing to do is to handle the case where it's null :)

DigitalLearningSolutions.Web/Views/MyAccount/Index.cshtml
```diff
@@ -0,0 +1,67 @@
+﻿@using DigitalLearningSolutions.Web.ViewModels.MyAccount
```
stellake, line 1: Really minor thing - we usually add screenshots to the MR to make it easier to visualise the changes (just FYI for the next time).
I think one thing I'd change is removing the bottom border from the last row in the summary list to match the wireframes. I would also add a little bit of margin under the main title of the page.

DigitalLearningSolutions.Data/Services/CentresService.cs
JBsoftwire: It could only be null if there's no centre with that ID, as the field isn't nullable. I'll add a log for that.

DigitalLearningSolutions.Web/Styles/myAccount/myAccount.scss
```diff
@@ -0,0 +1,24 @@
+﻿@import "~nhsuk-frontend/packages/core/all";
+
+$nhs-dark-grey: #425563;
+
+.account-card-content {
+  padding-top: 16px;
```
stellake, line 6: Super minor things and more of a FYI:
* In CSS you can specify padding like this: padding: 16px 32px 16px 32px; (this is in order of top right bottom left). If you leave out a particular value, it default to the opposite side, e.g. padding: 16px 32px; would be equivalent to the previous one. The only exception is padding: 16px; which would set 16px padding on all sides.
I personally prefer that syntax as it's more concise and keeps the padding properties in one line.
* 0 doesn't need units so I'd probably keep it consistent across the file.
* I'd have a preference for "border: none" as it's more intuitive, but I think 0 is fine too! :)

DigitalLearningSolutions.Web/Styles/myAccount/myAccount.scss
```diff
@@ -0,0 +1,24 @@
+﻿@import "~nhsuk-frontend/packages/core/all";
+
+$nhs-dark-grey: #425563;
```
stellake, line 3: What's this used for?

stellake (approved):

PR #250, opened 2021-04-09
HEEDLS-352 Add Sign In/Register links between login pages

AlexJacksonDS:
Added the Sign In and Register links between the Register and ForgotPassword pages and checked the existing ones on the Login page.

Checked that they display and work OK in Chrome/Firefox/IE and on Edge with no JS, and that there is no layout strangeness at mobile sizes.

ForgotPassword:
image

Register:
image


stellake:
Looks good - one minor thing - could you make this align on registration page:
image

stellake (approved):

PR #254, opened 2021-04-12
HEEDLS-364 - Added sign in logic, hashed password comparison, and Rem…

DanBloxham-sw:
…emberMe functionality from HEEDLS-340.

Tested:
- Ran all unit tests
- Manually tested log in process with multiple different accounts/types in chrome
- Looked at new error page in google chrome and in chrome mobile view

New error page view for when a user only has a delegate account that is not approved:
AccountNotApproved


DigitalLearningSolutions.Data.Tests/Services/PasswordResetServiceTests.cs
```diff
@@ -14,7 +14,6 @@
 
     public class PasswordResetServiceTests
     {
-        private IConfigService configService;
```
DanBloxham-sw, line 17: configService was left in after previous changes and was no longer being used, so I removed it here.

DigitalLearningSolutions.Data.Tests/Services/PasswordResetServiceTests.cs
```diff
@@ -46,7 +44,7 @@ public void Trying_get_null_user_should_throw_an_exception()
             A.CallTo(() => userService.GetUsersByEmailAddress(A<string>._)).Returns((new List<AdminUser>(), new List<DelegateUser>()));
 
             // Then
-            Assert.Throws<EmailAddressNotFoundException>(() => passwordResetService.GenerateAndSendPasswordResetLink("recipient@example.com", "example.com"));
+            Assert.Throws<UserAccountNotFoundException>(() => passwordResetService.GenerateAndSendPasswordResetLink("recipient@example.com", "example.com"));
```
DanBloxham-sw, line 47: Renamed EmailAddressNotFoundException to UserAccountNotFoundException to make it more general, which caused some knock-on changes to the ForgotPassword service/controller to be required to make sure the correctly name exception was being used.

stellake:
Really minor things I'd amend in the JIRA ticket scope now that I see the screenshot:
* Could we remove the "Error: " from the beginning of the title?
* Could we make "Approved" to lower case to be consistent with other titles?

DigitalLearningSolutions.Data.Tests/Services/LoginServiceTests.cs
```diff
@@ -0,0 +1,59 @@
...
+            // Then
+            Assert.Throws<IncorrectPasswordLoginException>(() => loginService.VerifyUserDetailsAndGetClaims("username", "password"));
+        }
+
+        [Test]
+        public void Exception_thrown_when_user_is_unapproved_delegate_and_not_admin()
+        {
+            // Given
+            A.CallTo(() => userService.GetAdminUsersByUsername(A<string>._)).Returns(new List<AdminUser>());
+            A.CallTo(() => userService.GetDelegateUsersByUsername(A<string>._)).Returns(new List<DelegateUser> { UserTestHelper.GetDefaultDelegateUser() });
```
stellake, line 53: Really minor, but I wouldn't expect a "Default delegate user" to be unapproved (we'll probably use it in other places too in the future tests)

DigitalLearningSolutions.Data/Models/User/AdminUser.cs
```diff
@@ -2,5 +2,26 @@
 {
     public class AdminUser : User
     {
+        public bool CentreAdmin { get; set; }
```
stellake, line 5: Hmm.. I'm wondering whether it's worth making the name consistent at least at a class level - e.g. isCentreManager, isCentreAdmin etc

DigitalLearningSolutions.Data/Services/LoginService.cs
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        public (AdminUser?, DelegateUser?) VerifyUserDetailsAndGetClaims(string username, string password)
+        {
+            List<AdminUser> adminUsers = userService.GetAdminUsersByUsername(username);
+            List<DelegateUser> delegateUsers = userService.GetDelegateUsersByUsername(username);
+
+            if (adminUsers.Count == 0 && delegateUsers.Count == 0)
+            {
+                throw new UserAccountNotFoundException("No account with that email address or user ID could be found.");
```
stellake, line 31: I think throwing exceptions to control the system flow is generally considered an anti-pattern. I'm wondering if we can split this into various functions and handle the cases of no user found, no password match and not approved on a controller level?

DigitalLearningSolutions.Data/Services/LoginService.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        public LoginService(IUserService userService, ICryptoService cryptoService)
+        {
+            this.userService = userService;
+            this.cryptoService = cryptoService;
+        }
+
+        public (AdminUser?, DelegateUser?) VerifyUserDetailsAndGetClaims(string username, string password)
+        {
+            List<AdminUser> adminUsers = userService.GetAdminUsersByUsername(username);
```
stellake, line 26: I think we should never expect more than one admin user per email or user id.

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
```diff
@@ -0,0 +1,122 @@
...
+
+            // Then
+            result.Should().BeRedirectToActionResult().WithControllerName("Home").WithActionName("Index");
+        }
+
+        [Test]
+        public void Successful_sign_in_should_render_home_page()
+        {
+            // Given
+            A.CallTo(() => loginService.VerifyUserDetailsAndGetClaims(A<string>._, A<string>._)).Returns((UserTestHelper.GetDefaultAdminUser(), UserTestHelper.GetDefaultDelegateUser()));
```
stellake, line 52: Really minor thing across the MR - some of the lines do get quite long with the mocks, so it might make it slightly easier to read if the "Returns" is put on a new line :)

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -24,54 +33,96 @@ public IActionResult Index()
...
                 AllowRefresh = true,
-                IsPersistent = true,
+                IsPersistent = model.RememberMe,
                 IssuedUtc = DateTime.UtcNow
             };
             HttpContext.SignInAsync("Identity.Application", new ClaimsPrincipal(claimsIdentity), authProperties);
             return RedirectToAction("Index", "Home");
         }
+
+        private List<Claim> GetClaimsForSignIn(AdminUser? adminUser, DelegateUser? delegateUser)
```
stellake, line 75: I wonder if it would make sense to pull this out into a service?

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -24,54 +33,96 @@ public IActionResult Index()
...
+            {
+                claims.Add(new Claim(CustomClaimTypes.UserForename, firstName));
+            }
+
+            if (surname != null)
+            {
+                claims.Add(new Claim(CustomClaimTypes.UserSurname, surname));
+            }
+
+            if (centreName != null)
```
stellake, line 110: I don't think this can ever be null (if the centre id exists, which it should)

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
DanBloxham-sw: I thought I had done that but it turns out that Resharper's Cleanup Code puts them back by default. I'll manually change them back.

DigitalLearningSolutions.Web/Controllers/LoginController.cs
DanBloxham-sw: Adding it into the LoginService should be easy enough to do.

DigitalLearningSolutions.Data/Services/LoginService.cs
DanBloxham-sw: Should be feasible enough. I think splitting it up into several functions was probably going to have be the first thing I was going to do in 361 - choose a centre.

DigitalLearningSolutions.Data/Models/User/AdminUser.cs
DanBloxham-sw: I was trying to go for consistency with the database column names, so they match as closely as possible to the relevant table (excluding shared columns like Id etc.). Do we still want to change the bools to "is..." when they don't match the table?

DigitalLearningSolutions.Web/Controllers/LoginController.cs
DanBloxham-sw: Actually, never mind. The function uses CustomClaimTypes, which is defined in Web.Helpers. so is not accessible in the data project services to avoid circular references. We could redefine it in the data project, but I think it would be more sensible to leave it in the Web project as claims aren't going to be used in any other project.

DigitalLearningSolutions.Data/Services/CryptoService.cs
```diff
@@ -0,0 +1,45 @@
...
+        public bool VerifyHashedPassword(string? hashedPassword, string password)
+        {
+            if (string.IsNullOrWhiteSpace(hashedPassword))
+            {
+                return false;
+            }
+
+            var hashedPasswordBytes = Convert.FromBase64String(hashedPassword);
+
+            if (hashedPasswordBytes.Length != 1 + SaltSize + PBKDF2SubkeyLength || hashedPasswordBytes[0] != 0x00)
```
stellake, line 27: Could this be pulled into a variable(s)? This will make the statement easier to read and remove the need for a comment.

DigitalLearningSolutions.Web/Controllers/LoginController.cs
stellake: Ah, that's fair!

DigitalLearningSolutions.Data/Services/LoginService.cs
DanBloxham-sw: Actually, I think the entirety of the LoginService can just be moved into the controller instead. Pretty much all the LoginService is doing is grabbing some data from the UserService, and calling a CryptoService function, and doing some null checks on them. I'm not seeing anything that would be out of the ordinary in a controller, so I think I will get rid of the LoginService.

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -24,54 +35,107 @@ public IActionResult Index()
...
+                new Claim(CustomClaimTypes.LearnCandidateId, delegateUser?.Id.ToString() ?? "False"),
+                new Claim(CustomClaimTypes.LearnUserAuthenticated, (delegateUser != null).ToString()),
+                new Claim(CustomClaimTypes.AdminCategoryId, adminUser?.CategoryId.ToString() ?? "False"),
+                new Claim(CustomClaimTypes.IsSupervisor, adminUser?.IsSupervisor.ToString() ?? "False"),
+                new Claim(CustomClaimTypes.IsTrainer, adminUser?.IsTrainer.ToString() ?? "False"),
+                new Claim(CustomClaimTypes.IsFrameworkDeveloper, adminUser?.IsFrameworkDeveloper.ToString() ?? "False"),
+                new Claim(CustomClaimTypes.UserCentreName, adminUser?.CentreName ?? delegateUser?.CentreName)
+            };
+
+            var firstName = delegateUser?.FirstName ?? adminUser?.FirstName;
```
AlexJacksonDS, line 115: Just wanted to check that these are in the correct order given most of the above are admin ?? delegate

DigitalLearningSolutions.Web/Views/Login/AccountNotApproved.cshtml
```diff
@@ -0,0 +1,12 @@
+﻿@{
+  ViewData["Title"] = "Registration not yet approved";
+}
+
+<div class="nhsuk-grid-row">
+  <div class="nhsuk-grid-column-full">
+    <h1 id="page-heading">Registration not yet approved</h1>
```
AlexJacksonDS, line 7: We should be using an appropriate [nhsuk-heading class](https://service-manual.nhs.uk/design-system/styles/typography#headings) on all of the hX elements in the solution, rather than relying on other styling defined ourselves/default styling. Some places are using these, some aren't. We probably need to go back though and make sure they are consistent @stellake

AlexJacksonDS (commented):
A couple of comments around consistency/things to check

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -20,6 +22,60 @@ public UserService(IDbConnection connection)
...
+                        cd.CentreID,
+                        cd.EmailAddress,
+                        cd.FirstName,
+                        cd.LastName,
+                        cd.Password,
+                        cd.Approved
+                    FROM Candidates AS cd
+                    INNER JOIN Centres AS ct ON ct.CentreID = cd.CentreID 
+                    WHERE cd.Active = 1 AND
+                         (cd.CandidateNumber = @username OR cd.EmailAddress = @username OR cd.AliasID = @username)",
```
stellake, line 72: Ah, is the alias used in the old system? (I might ask Kevin about where it's used across the system and how it works if so - to make sure we don't get into issues when handling this)

DigitalLearningSolutions.Data/Services/LoginService.cs
AlexJacksonDS: I don't really agree here with the approach of pulling more logic into controllers, it makes things harder to unit test. Particularly as we now have a pretty significant chunk of logic around getting users and checking passwords here.

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
```diff
@@ -0,0 +1,153 @@
...
+            // Then
+            result.Should().BeViewResult().WithDefaultViewName();
+        }
+
+        [Test]
+        public void Index_should_redirect_if_user_is_authenticated()
+        {
+            // Given
+            var controllerWithAuthenticatedUser =
+                LoginTestHelper.GetLoginControllerWithAuthenticatedUser(userService, cryptoService);
```
stellake, line 51: Ah, I think it might be nicer to add the authentication case to Integration tests rather than here (sorry for not noticing before). We will have auth related stuff there anyway because we can't test the [Authorize] annotations otherwise, so grouping the auth tests together makes sense and I think will make these unit tests tidier too! :)

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
```diff
@@ -0,0 +1,153 @@
...
+            // When
+            var result = controllerWithAuthenticatedUser.Index();
+
+            // Then
+            result.Should().BeRedirectToActionResult()
+                .WithControllerName("Home").WithActionName("Index");
+        }
+
+        [Test]
+        public void Successful_sign_in_should_render_home_page()
```
stellake, line 62: Could we set up the test such that the user service & crypto service mocks return what's expected (e.g. approved user and true)

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -24,54 +35,107 @@ public IActionResult Index()
...
             }
 
-            if (model.Username?.ToLower() == "testpassworderror")
+            AdminUser? adminUser = userService.GetAdminUserByUsername(model.Username);
+            List<DelegateUser> delegateUsers = userService.GetDelegateUsersByUsername(model.Username);
+
+            if (adminUser == null && delegateUsers.Count == 0)
             {
-                ModelState.AddModelError("password", "There is a password error.");
+                ModelState.AddModelError("Username","No account with that email address or user ID could be found.");
```
stellake, line 48: I can see a cheeky no space after "," 😄 Might be worth running the formatter quickly!

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -24,54 +35,107 @@ public IActionResult Index()
...
+            delegateUsers = delegateUsers.Where(du => cryptoService.VerifyHashedPassword(du.Password, model.Password)).ToList();
+
+            if (adminUser == null && delegateUsers.Count == 0)
             {
+                ModelState.AddModelError("Password", "The password you have entered is incorrect.");
                 return View("Index", model);
             }
 
-            var claims = new List<Claim>
+            if (adminUser == null)
```
stellake, line 61: All of this is quite long and could potentially be put into a service or a method. Also, we should probably follow this logic:
// Get first approved delegate
// If approvedDelegate == null && adminUser == null, then Go to account not approved page (note that if the user logs in with their delegate id and have an admin account associated with the same email address as the delegate account has, we should still go to the unapproved page, because they're trying to access the delegate account specifically and need communication about why their sign in isn't working)
// Else try and find an admin account associated with the approved & verified delegate account and use that to set some claims.

DigitalLearningSolutions.Web/Views/Login/AccountNotApproved.cshtml
stellake: Hmm.. I think the h1 etc get the heading class style automatically in the compiled css, but I don't mind deciding on a rule of consistency and following it
image

AlexJacksonDS: I agree they probably are getting automatically applied based on that, and it is probably fine. I just think that we should probably be using these components as they are specified in the nhs design system, someone could easily go and change stuff on #page-heading affecting multiple pages.
image

DigitalLearningSolutions.Data/Services/UserService.cs
DanBloxham-sw: It is indeed used in the old system, used identically as it is here to get delegate records by userid/email.

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
DanBloxham-sw: In the SetUp, I set up the mocks to return what's expected for most of the tests, and then when those tests differ in requirements I change the mock return in the tests themselves. Would we prefer them being defined in every test then?

DigitalLearningSolutions.Web/Controllers/LoginController.cs
DanBloxham-sw: The order of these doesn't really matter as far as I am aware, as the old code's order for these will change depending on which account is found first. But I might as well make it consistent with the rest.

DigitalLearningSolutions.Web/Views/Login/AccountNotApproved.cshtml
stellake: Oh I see what you mean - you think we should replace the id="page-heading" with the NHS class, but I thought the initial comment was about whether we should be explicit about the NHS style or whether being implicit is ok.

**The explicit vs implicit** - I think my preference would be explicit, which means that I'm happy to set the standard as "use the typography classes explicitly".

**The use of "page-heading" vs NHS typography classes** - I don't think it should be X or Y as the id is specified on some of the pages to identify the main header (e.g. it's used in full screen learning menu etc) and to specify some styling in addition to the NHS styling, which is something HEE have asked us to do. I'd prefer the header styling to be consistent across the site, so I don't think we should remove the id, but if we go for explicit styles, we should add the class in addition to the id.

DigitalLearningSolutions.Data.Tests/Services/LoginServiceTests.cs
```diff
@@ -0,0 +1,56 @@
...
+        public void Setup()
+        {
+            userService = A.Fake<IUserService>();
+            cryptoService = A.Fake<ICryptoService>();
+
+            loginService = new LoginService(userService, cryptoService);
+        }
+
+        [Test]
+        public void Return_no_delegate_users_when_delegate_stored_password_is_empty()
```
stellake, line 26: I feel like we probably want a few more tests here. A full set would probably look something like that:
-- GetUsersByUsername
* Returns admin and delegate users by username (to make sure what's returned by user service is passed back)

-- VerifyUsers
* Returns verified admin user
* Does not return unverified admin user
* Returns verified delegates
* Filters out unverified delegates
* Does not return any delegates if all unverified

-- Get admin user associated with delegate
* Returns nothing if delegate emial is empty
* Return nothing if no admin associated with delegate account
* Returns admin account associated with delegate

DigitalLearningSolutions.Data/Services/LoginService.cs
```diff
@@ -0,0 +1,66 @@
...
+                    ? unverifiedAdminUser
+                    : null;
+            var verifiedDelegateUsers =
+                unverifiedDelegateUsers.Where(du => cryptoService.VerifyHashedPassword(du.Password, password))
+                    .ToList();
+
+            return (verifiedAdminUser, verifiedDelegateUsers);
+        }
+
+        public AdminUser? GetVerifiedAdminUserAssociatedWithApprovedDelegateUser(DelegateUser approvedDelegateUser,
```
stellake, line 51: Could this just be GetVerifiedAdminUserAssociatedWithDelegateUser? (as the approval of delegates is unrelated to this method)

DigitalLearningSolutions.Data/Services/LoginService.cs
```diff
@@ -0,0 +1,66 @@
+﻿namespace DigitalLearningSolutions.Data.Services
+{
+    using System.Collections.Generic;
+    using System.Linq;
+    using DigitalLearningSolutions.Data.Models.User;
+
+    public interface ILoginService
+    {
+        public (AdminUser?, List<DelegateUser>) GetUsersByUsername(string username);
```
AlexJacksonDS, line 9: I think this should probably be in the UserService instead since it doesn't strictly deal with Login/Out (and would potentially be useful on the MyAccount page which shouldn't be using a LoginService for example)

Similarly GetVerifiedAdminUserAssociatedWithApprovedDelegateUser?

DigitalLearningSolutions.Web/Views/ForgotPassword/Confirm.cshtml
```diff
@@ -6,7 +6,7 @@
 
 <div class="nhsuk-grid-row">
   <div class="nhsuk-grid-column-full">
-    <h1 id="page-heading" class="nhsuk-u-margin-bottom-6">Password reset email sent</h1>
+    <h1 id="page-heading" class="nhsuk-heading-xl nhsuk-u-margin-bottom-6">Password reset email sent</h1>
```
AlexJacksonDS, line 9: 😄

AlexJacksonDS (commented):
Just a comment on the location of methods in the services

DigitalLearningSolutions.Data/Services/LoginService.cs
DanBloxham-sw: I'd argue that GetVerifiedAdminUserAssociatedWithApprovedDelegateUser should probably stay in LoginService as it also takes in an input password to verify against the account. But GetUsersByUsername can most certainly be put in the UserService.

AlexJacksonDS: Yeah that makes sense, didn't notice it took a password

stellake (approved):

PR #255, opened 2021-04-12
HEEDLS-NONE: Fix app root path.

stellake:
Should not have "/" in the end to match other envs.


DanBloxham-sw (commented):
Shouldn't be a problem. I did just check this by copying it into appsettings.development.json and trying it out. The reset password link without the query strings became: https://hee-dls-test.softwire.com:443/ResetPassword? with a port of 443 on the end. Do we remember if that was expected on the test environment before?

stellake:
@DanBloxham-sw looks like the Outlook has a lot of emails with the port in the url from last week so this seems unrelated, but is worth investigating.
I'll look into it to see whether it's also an issue in UAT and raise a new ticket for it.

DanBloxham-sw (approved):

PR #257, opened 2021-04-13
HEEDLS-387: Use consistent wording for logging in and out across site.

stellake:
Replace "Sign in" with "Log in" and "Sign out" with "Log out" across the site.
I did not change the "Login" into "Log in" on the landing page, because this will be handled as part of HEEDLS-338.


DanBloxham-sw (approved):
I noticed there is a button on /Home Index page that still says "Login" rather than "Log in" that might be worth changing but I imagine that will be changed in the Landing page tickets. 
EDIT: Ignore the above, didn't realise there was already a comment addressing this.

PR #260, opened 2021-04-13
Heedls 341 old content in iframes

AlexJacksonDS:
Added several new controllers to display content from the old site in iframes. Added a view component for the nhsuk back link.

I've tested the responsive sizing of the iframe across a few browsers and with the chrome devtools devices. There is currently an issue that the size of the main nhsuk-width-container is 960px, the exact breakpoint for the old site, so we get the tablet(?) layouts in the iframe which have some slightly odd layout behaviour as you resize the window. We also currently have double margins (the main layout plus the margins in the iframe) which would ease some of this layout behaviour.

The FindYourCentre iframe content doesn't fully work with javascript off (tested in Edge), the GoogleMaps map doesn't load.

The screen reader gets into the iframe correctly, but we are then limited by the accessibility of the content we are displaying within that area. It is also possible that it will start reading from the top, when it is already scrolled to the bottom of the content. I was able to trigger this behaviour by tabbing through to the Skip to Main Content link, which came after all the content in the iframe, and so the iframe was scrolled to the last tab item it contained.

image

### DIFF
--- a/DigitalLearningSolutions.Web/Controllers/FindYourCentreController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/FindYourCentreController.cs
@@ -1,0 +1,13 @@
+﻿namespace DigitalLearningSolutions.Web.Controllers
+{
+    using Microsoft.AspNetCore.Mvc;
+
+    [Route("FindYourCentre")]
+    public class FindYourCentreController : Controller
+    {
+        public IActionResult Index()
+        {
+            return View();
+        }
+    }
+}
--- a/DigitalLearningSolutions.Web/Controllers/FindYourCentreController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/FindYourCentreController.cs
@@ -2,7 +2,6 @@
 {
     using Microsoft.AspNetCore.Mvc;
 
-    [Route("FindYourCentre")]
     public class FindYourCentreController : Controller
     {
         public IActionResult Index()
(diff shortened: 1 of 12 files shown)


AlexJacksonDS:
Also note, I've added a bit of extra css to the maincontentwrapper and maincontent (nhsuk-main-wrapper) elements (from _Layout.cshtml surrounding the RenderBody()) to allow the page content to resize correctly. I've had a quick check around the other pages and it doesn't appear to break any layout.

Without this added css, I think we would have to set heights on the iframe/container more explicitly.

DigitalLearningSolutions.Web/Controllers/FindYourCentreController.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace DigitalLearningSolutions.Web.Controllers
+{
+    using Microsoft.AspNetCore.Mvc;
+
+    [Route("FindYourCentre")]
```
stellake, line 5: I think this should be the route by default (and for controllers below)

stellake (approved):
Looks good - just a comment about Routes! 

Regarding the styling, we might want to remove the wrapper padding when the backbutton is not present such that the iframe starts where the nav bar ends. However, I think we should present the current version to HEE at the Show and Tell on Friday and see what they think. If they're happy with the current version, I'll be happy :)

AlexJacksonDS:
> Looks good - just a comment about Routes!
> 
> Regarding the styling, we might want to remove the wrapper padding when the backbutton is not present such that the iframe starts where the nav bar ends. However, I think we should present the current version to HEE at the Show and Tell on Friday and see what they think. If they're happy with the current version, I'll be happy :)

The top padding comes from the nhsuk-main-wrapper (in the _Layout.cshtml) so there might not be an easy way to remove it in a sensible way.

PR #261, opened 2021-04-13
HEEDLS-352 Add the role button to links styled as buttons

AlexJacksonDS:
Added the button role to the links styled as buttons added between the Login, Register and Forgot Password screens. This makes screen readers read them out as buttons rather than links, matching their styling. Checked that they were read out as expected.


stellake (approved):
Looks good!

PR #263, opened 2021-04-14
Heedls 391 add basic myaccount details

AlexJacksonDS:
Added an expandable container with basic account details in it populating them the same way as the existing fields from the basic version.

Checked that the contents of the page display/resize correctly in Chrome and with Chrome devices, and checked that it worked fine with Edge no javascript. Also checked that the screen reader was able to read out the contents of the expander.

Changed the way the email was being retrieved as the existing version didn't seem to work.

image

image

### DIFF
--- a/DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
@@ -1,5 +1,6 @@
 ﻿namespace DigitalLearningSolutions.Web.Controllers
 {
+    using System.Security.Claims;
     using DigitalLearningSolutions.Data.Services;
     using DigitalLearningSolutions.Web.Helpers;
     using DigitalLearningSolutions.Web.ViewModels.MyAccount;
@@ -21,12 +22,14 @@
                 return RedirectToAction("Index", "Login");
             }
 
-            var userEmail = User.GetEmail();
+            var userEmail = User.GetCustomClaim(ClaimTypes.Email);
             var delegateId = User.GetCustomClaim(CustomClaimTypes.LearnCandidateNumber);
             var centreId = User.GetCentreId();
+            var firstName = User.GetCustomClaim(CustomClaimTypes.UserForename);
+            var surname = User.GetCustomClaim(CustomClaimTypes.UserSurname);
             var centreName = centresService.GetCentreName(centreId);
 
-            var model = new MyAccountViewModel(centreName, userEmail, delegateId);
+            var model = new MyAccountViewModel(centreName, userEmail, delegateId, firstName, surname);
             return View(model);
         }
     }
--- a/DigitalLearningSolutions.Web/Views/MyAccount/Index.cshtml
+++ b/DigitalLearningSolutions.Web/Views/MyAccount/Index.cshtml
@@ -56,7 +56,7 @@
 </div>
 <div class="nhsuk-grid-row">
   <div class="nhsuk-grid-column-full">
-    <details class="nhsuk-details nhsuk-expander">
+    <details class="nhsuk-details nhsuk-expander bottom-margin-nhsuk-card">
       <summary class="nhsuk-details__summary">
         <span class="nhsuk-details__summary-text">
           My Details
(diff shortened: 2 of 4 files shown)


stellake:
Nice! One minor thing in terms of styling - could we make the spacing between the expander and "Manage notifications" buttons more symmetric with the spacing between the account summary and "My details" sections?

image

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -21,12 +22,14 @@ public IActionResult Index()
                 return RedirectToAction("Index", "Login");
             }
 
-            var userEmail = User.GetEmail();
+            var userEmail = User.GetCustomClaim(ClaimTypes.Email);
             var delegateId = User.GetCustomClaim(CustomClaimTypes.LearnCandidateNumber);
             var centreId = User.GetCentreId();
+            var firstName = User.GetCustomClaim(CustomClaimTypes.UserForename);
+            var surname = User.GetCustomClaim(CustomClaimTypes.UserSurname);
```
stellake, line 29: Thinking about future features, I think we'll need to take these values from the database, because I don't think it makes sense to update the claims on the cookie when the user edits these details on a different page.

I will need to check what the old code does and whether it's something we need to handle in old pages.

stellake: [note that the delegate id and centre id are not editable to my knowledge]

DigitalLearningSolutions.Web/Views/MyAccount/Index.cshtml
```diff
@@ -12,7 +12,7 @@
 
 <div class="nhsuk-grid-row">
   <div class="nhsuk-grid-column-full">
-    <h1 id="app-page-heading">My account</h1>
+    <h1 class="nhsuk-heading-xl" id="app-page-heading">My account</h1>
```
stellake, line 15: Nice :)

DigitalLearningSolutions.Web/Views/MyAccount/Index.cshtml
```diff
@@ -56,14 +56,67 @@
...
+          <div class="nhsuk-summary-list__row account-summarylist_row">
+            <dt class="nhsuk-summary-list__key">
+              Last name:
+            </dt>
+            <dd class="nhsuk-summary-list__value">
+              @Model.Surname
+            </dd>
+          </div>
+
+          @if (Model.User?.Length > 0)
```
stellake, line 85: I think we probably want to show blank here if the email address doesn't exist so the user knows what they have an option to set it up

DigitalLearningSolutions.Web/Views/MyAccount/Index.cshtml
```diff
@@ -56,14 +56,67 @@
...
+  <div class="nhsuk-grid-column-full">
+    <a class="nhsuk-button" role="button">
       Manage notifications
     </a>
   </div>
 </div>
 <div class="nhsuk-grid-row">
   <div class="nhsuk-grid-column-full">
-    <a class="nhsuk-button nhsuk-button--secondary" href="@Configuration["CurrentSystemBaseUrl"]/logout">
+    <a class="nhsuk-button nhsuk-button--secondary" href="@Configuration["CurrentSystemBaseUrl"]/logout" role="button">
```
stellake, line 119: 👍

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
AlexJacksonDS: Yeah I figured that might be the case, I just followed what was already there for now. I need to add some database access for the other 2 parts of the ticket.

stellake: Cool - I'm happy for you to change this to use the data from db as a separate subtask of the story. :)

AlexJacksonDS:
> Nice! One minor thing in terms of styling - could we make the spacing between the expander and "Manage notifications" buttons more symmetric with the spacing between the account summary and "My details" sections?

This was due to the differing spacing between the nhsuk-card and the nhsuk-details/expander components. Added some extra css to match the single card, but we'll probably need more variant of this as we find other random combinations of components that don't quite match up

stellake (approved):
Looks good!

PR #269, opened 2021-04-16
HEEDLS-346: get notification preferences in the backend

JBsoftwire:
Part 2, covering backend fetching of notification data

### DIFF
--- a/DigitalLearningSolutions.Data/Models/NotificationPreference.cs
+++ b/DigitalLearningSolutions.Data/Models/NotificationPreference.cs
@@ -1,0 +1,9 @@
+﻿namespace DigitalLearningSolutions.Data.Models
+{
+    public class NotificationPreference
+    {
+        public string? NotificationName { get; set; }
+        public string? Description { get; set; }
+        public bool Accepted { get; set; }
+    }
+}
--- a/DigitalLearningSolutions.Data/Models/NotificationPreference.cs
+++ b/DigitalLearningSolutions.Data/Models/NotificationPreference.cs
@@ -2,7 +2,8 @@
 {
     public class NotificationPreference
     {
-        public string? NotificationName { get; set; }
+        public int NotificationId { get; set; }
+        public string NotificationName { get; set; }
         public string? Description { get; set; }
         public bool Accepted { get; set; }
     }
--- a/DigitalLearningSolutions.Data/Services/NotificationPreferenceService.cs
+++ b/DigitalLearningSolutions.Data/Services/NotificationPreferenceService.cs
@@ -1,0 +1,58 @@
+﻿namespace DigitalLearningSolutions.Data.Services
+{
+    using System.Collections.Generic;
+    using System.Data;
+    using Dapper;
+    using DigitalLearningSolutions.Data.Models;
+
+    public interface INotificationPreferenceService
+    {
+        IEnumerable<NotificationPreference> GetNotificationPreferencesForAdmin(int adminId);
+        IEnumerable<NotificationPreference> GetNotificationPreferencesForDelegate(int delegateId);
+    }
+
+    public class NotificationPreferenceService : INotificationPreferenceService
+    {
+        private readonly IDbConnection connection;
+
+        public NotificationPreferenceService(IDbConnection connection)
+        {
+            this.connection = connection;
+        }
+
+        public IEnumerable<NotificationPreference> GetNotificationPreferencesForAdmin(int adminId)
+        {
+            return connection.Query<NotificationPreference>(
+                @"DECLARE @CentreAdmin BIT, @IsCentreManager BIT, @ContentManager BIT, @ContentCreator BIT, @Supervisor BIT, @Trainer BIT
+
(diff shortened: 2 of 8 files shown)


DigitalLearningSolutions.Data/Models/NotificationPreference.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace DigitalLearningSolutions.Data.Models
+{
+    public class NotificationPreference
+    {
+        public string? NotificationName { get; set; }
```
stellake, line 5: I think this should never be null (as per not null constraint in db)

DigitalLearningSolutions.Data/Services/NotificationPreferenceService.cs
```diff
@@ -0,0 +1,58 @@
+﻿namespace DigitalLearningSolutions.Data.Services
+{
```
stellake, line 2: Could we call this "NotificationPreferenceDataService"? I think we should try to go towards a pattern where all repository equivalent services are called "data service" to distinguish between the two. I'm in the process of thinking how to best separate the services and data services/repositories.

DigitalLearningSolutions.Data/Services/NotificationPreferenceService.cs
```diff
@@ -0,0 +1,58 @@
...
+    public class NotificationPreferenceService : INotificationPreferenceService
+    {
+        private readonly IDbConnection connection;
+
+        public NotificationPreferenceService(IDbConnection connection)
+        {
+            this.connection = connection;
+        }
+
+        public IEnumerable<NotificationPreference> GetNotificationPreferencesForAdmin(int adminId)
```
stellake, line 23: I'd probably add a few data tests for these new methods :)

DigitalLearningSolutions.Web/Controllers/NotificationPreferencesController.cs
```diff
@@ -14,9 +25,20 @@ public IActionResult Index()
...
+
             var delegateId = User.GetCustomClaimAsInt(CustomClaimTypes.LearnCandidateId);
+            var delegateNotifications = new List<NotificationPreference>();
+            if (delegateId.HasValue)
+            {
+                delegateNotifications = notificationPreferenceService.GetNotificationPreferencesForDelegate(delegateId.Value).ToList();
+            }
 
-            var model = new NotificationPreferencesViewModel(adminId, delegateId);
+            var model = new NotificationPreferencesViewModel(adminId, delegateId, adminNotifications, delegateNotifications);
```
stellake, line 41: 1. Does the ViewModel need the admin and delegateIds anymore? (as I think we can just check the notifications list size, which will be equivalent)
2. Could we move GetDelegateNotifications (and admin) into a service (maybe NotificationPreferenceService if the data service is renamed to "DataService"), which can accept null id values and converts the preferences to a list, just to keep the controller as logic free as possible? Alternatively, I'd be happy for the data service to do the null check on id and return an empty collection if that's the case. :)

DigitalLearningSolutions.Web/ViewModels/MyAccount/NotificationPreferencesViewModel.cs
```diff
@@ -1,15 +1,28 @@
...
+    using System.Collections.Generic;
+    using DigitalLearningSolutions.Data.Models;
+
     public class NotificationPreferencesViewModel
     {
-        public NotificationPreferencesViewModel(int? adminId, int? delegateId)
+        public NotificationPreferencesViewModel(
+            int? adminId,
+            int? delegateId,
+            IEnumerable<NotificationPreference> adminNotifications,
```
stellake, line 11: Should this be a list now (given the conversion to list in the controller)?

JBsoftwire: I used that conversion to quiet down a multiple enumeration complaint, but the new logic doesn't have that issue so I left it all as Enumerable.

DigitalLearningSolutions.Data/Services/NotificationPreferencesDataService.cs
```diff
@@ -0,0 +1,76 @@
...
+            }
+
+            return connection.Query<NotificationPreference>(
+                @"DECLARE @CentreAdmin BIT, @IsCentreManager BIT, @ContentManager BIT, @ContentCreator BIT, @Supervisor BIT, @Trainer BIT
+
+                    SELECT @CentreAdmin = CentreAdmin, @IsCentreManager = IsCentreManager, @ContentManager = ContentManager, @ContentCreator = ContentCreator, @Supervisor = Supervisor, @Trainer = Trainer
+	                    FROM AdminUsers
+	                    WHERE AdminID = @adminId
+
+                    DECLARE @Whitespace VARCHAR(3)
```
JBsoftwire, line 37: The notification names have unhelpful whitespace and new lines at the end, which caused some problems when setting up the tests. Stripping them to make things easier downstream and leaves the question of how to display the name entirely in the hands of the View.

stellake: Double checked with Kevin and he said that it's probably just some bad handling form the old system so they're fine to remove! :)

stellake (approved):
Looks good :)

PR #270, opened 2021-04-19
HEEDLS-339 - Fixed log in links and admin user id login.

DanBloxham-sw:
Made changes to the Log in Index page so that the links to Privacy and Terms of Use now open in new tabs.
Also modified the log in controller to grab delegate accounts associated with an admin account if only the admin account specific log in was used. Similar logic was already set up for grabbing admin accounts from delegate specific login.

Tested:
- Ran all unit tests
- Manually tested different log ins
- Tested log in and link opening in chrome.

### DIFF
--- a/DigitalLearningSolutions.Data/Services/LoginService.cs
+++ b/DigitalLearningSolutions.Data/Services/LoginService.cs
@@ -9,7 +9,7 @@
         public (AdminUser?, List<DelegateUser>) VerifyUsers(
             string password, AdminUser? unverifiedAdminUser, List<DelegateUser> unverifiedDelegateUsers);
 
-        public AdminUser? GetVerifiedAdminUserAssociatedWithDelegateUser(DelegateUser approvedDelegateUser,
+        public AdminUser? GetVerifiedAdminUserAssociatedWithDelegateUser(DelegateUser delegateUser,
             string password);
 
         public List<DelegateUser> GetVerifiedDelegateUsersAssociatedWithAdminUser(AdminUser adminUser, string password);
--- a/DigitalLearningSolutions.Data/Services/LoginService.cs
+++ b/DigitalLearningSolutions.Data/Services/LoginService.cs
@@ -11,8 +11,6 @@
 
         public AdminUser? GetVerifiedAdminUserAssociatedWithDelegateUser(DelegateUser delegateUser,
             string password);
-
-        public List<DelegateUser> GetVerifiedDelegateUsersAssociatedWithAdminUser(AdminUser adminUser, string password);
     }
 
     public class LoginService : ILoginService
@@ -52,21 +50,6 @@
             return cryptoService.VerifyHashedPassword(adminUserAssociatedWithDelegate?.Password, password)
                 ? adminUserAssociatedWithDelegate
                 : null;
-        }
-
-        public List<DelegateUser> GetVerifiedDelegateUsersAssociatedWithAdminUser(AdminUser adminUser, string password)
-        {
-            var delegateUsersAssociatedWithAdmin = new List<DelegateUser>();
-
-            if (string.IsNullOrWhiteSpace(adminUser.EmailAddress))
-            {
-                return delegateUsersAssociatedWithAdmin;
-            }
-
-            delegateUsersAssociatedWithAdmin = userService.GetDelegateUsersByUsername(adminUser.EmailAddress);
-
-            return delegateUsersAssociatedWithAdmin
-                .Where(du => cryptoService.VerifyHashedPassword(du.Password, password)).ToList();
         }
     }
 }
--- a/DigitalLearningSolutions.Data/Services/LoginService.cs
+++ b/DigitalLearningSolutions.Data/Services/LoginService.cs
@@ -11,6 +11,8 @@
 
         public AdminUser? GetVerifiedAdminUserAssociatedWithDelegateUser(DelegateUser approvedDelegateUser,
             string password);
+
+        public List<DelegateUser> GetVerifiedDelegateUsersAssociatedWithAdminUser(AdminUser adminUser, string password);
     }
 
     public class LoginService : ILoginService
@@ -38,19 +40,33 @@
             return (verifiedAdminUser, verifiedDelegateUsers);
         }
 
-        public AdminUser? GetVerifiedAdminUserAssociatedWithDelegateUser(DelegateUser approvedDelegateUser,
-            string password)
+        public AdminUser? GetVerifiedAdminUserAssociatedWithDelegateUser(DelegateUser delegateUser, string password)
         {
-            if (string.IsNullOrWhiteSpace(approvedDelegateUser.EmailAddress))
+            if (string.IsNullOrWhiteSpace(delegateUser.EmailAddress))
             {
                 return null;
             }
 
             var adminUserAssociatedWithDelegate =
-                userService.GetAdminUserByUsername(approvedDelegateUser.EmailAddress);
+                userService.GetAdminUserByUsername(delegateUser.EmailAddress);
             return cryptoService.VerifyHashedPassword(adminUserAssociatedWithDelegate?.Password, password)
                 ? adminUserAssociatedWithDelegate
                 : null;
+        }
+
+        public List<DelegateUser> GetVerifiedDelegateUsersAssociatedWithAdminUser(AdminUser adminUser, string password)
+        {
+            var delegateUsersAssociatedWithAdmin = new List<DelegateUser>();
+
+            if (string.IsNullOrWhiteSpace(adminUser.EmailAddress))
+            {
+                return delegateUsersAssociatedWithAdmin;
+            }
+
+            delegateUsersAssociatedWithAdmin = userService.GetDelegateUsersByUsername(adminUser.EmailAddress);
+
+            return delegateUsersAssociatedWithAdmin
+                .Where(du => cryptoService.VerifyHashedPassword(du.Password, password)).ToList();
         }
     }
 }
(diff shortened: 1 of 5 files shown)


DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -57,21 +57,30 @@ public IActionResult Index(LoginViewModel model)
                 return View("Index", model);
             }
 
+            if (verifiedAdminUser != null && verifiedDelegateUsers.Count == 0)
+            {
+                verifiedDelegateUsers =
+                    loginService.GetVerifiedDelegateUsersAssociatedWithAdminUser(verifiedAdminUser, model.Password);
```
stellake, line 63: Hmm.. I feel like we should get the associated delegates when we get the users in the first place, because:
* There is no reason to wait until we get the associated delegate accounts and if we get them early, the code above will deal with verifying the password. It's a bit different for finding an admin account associated with an unapproved delegate, because we need to handle that case differently from the UX perspective.
* It will keep the controller less crowded and make testing the code cleaner.

Also, we should only get the associated admin user if the delegate user is approved (moving that bit of logic up here means that if someone logs in with a delegate id that is for an unapproved account, we won't show them the "unapproved" page if they have an admin account associated with the email address of the delegate account).

DigitalLearningSolutions.Web/Views/Login/Index.cshtml
```diff
@@ -58,9 +58,9 @@
         <span class="nhsuk-u-visually-hidden">Information: </span>
         <p>
           By logging in, I consent to my details being stored and processed in line with your
-          <a href="https://www.hee.nhs.uk/about/privacy-notice">Privacy Policy</a>
+          <a target="_blank" href="https://www.hee.nhs.uk/about/privacy-notice">Privacy Policy (opens in new window)</a>
```
stellake, line 61: This seems a bit clunky - I wonder if we could do something proposed here: https://designnotes.blog.gov.uk/2016/11/28/removing-the-external-link-icon-from-gov-uk/

```
If your users do need to know in advance that a link goes to another website, it’s more effective to do this with words.

You can do this in the link text itself, or if you have a collection of links **you might want to do this in the immediately preceding text to avoid duplication.**
```

I don't have any good ideas though, so maybe we can send a screenshot of this into the external HEE Slack channel and ask them if they think this is ok.

DigitalLearningSolutions.Data/Services/LoginService.cs
```diff
@@ -9,8 +9,10 @@ public interface ILoginService
         public (AdminUser?, List<DelegateUser>) VerifyUsers(
```
stellake, line 9: Do you think it's worth adding a test for this bug? :)

DigitalLearningSolutions.Web/Controllers/LoginController.cs
DanBloxham-sw: That all makes sense, but I am a bit confused about the final bit with approval. I appreciate that if a user logs in with a delegate ID for an unapproved account, they should be shown the "Account Not Approved" page. But how will this work when we get the Choose A Centre stuff working? On there, I imagine if they log in with an unapproved delegate Id, they still want to see all of their roles on all centres, creating a different set of behaviours if they have multiple centres or not (I don't know if we really need to be concerned with it on this bug fix or not, but seeing as I am currently working on the Choose a Centre ticket I might as well ask)

For clarity:
- Imagine a situation where a user has 3 accounts: An unapproved delegate at Centre 1, an Admin account also at Centre 1, and an approved delegate at Centre 2 that will not be logged in to.
- As they have multiple centres, when they log in they would see "Delegate" at Centre 2, and "Delegate and Admin" at Centre 1. If they choose to log in to Centre 1, what should happen? I think they should get logged in with both delegate and admin accounts, thus avoiding the Account Not Approved page, despite the fact that the delegate account is not approved. They get around this because they know they have an Admin account.
- If they did not have that other account at Centre 2, they would only be detected as at Centre 1. Therefore, if they log in with the delegate Id, the admin account will not be picked up before the approval check, and thus will be shown the Account Not Approved page.

This would be different behaviours based when logging in, based on a separate account to the one they are trying to access. What are our thoughts on this?

stellake: The reason we show the unapproved page is to keep away users that get confused about the fact that they cannot log into their delegate account. Hence:
* If the user logs in with the unapproved delegate id (rather than their usual login that they're used to using with other accounts), we should always show unapproved page to communicate that the account is not approved. The main reason they'd be using this login is to access the new delegate account (this is different to the account they're used to using).
* If the user logs in with a different login (email/admin login/or approved delegate id), we should display the Choose a centre page with only approved accounts - so in your example it would be "Centre 1 - admin" and "Centre 2 - delegate".

DanBloxham-sw: Right, I think I get it now: If a user with an unapproved delegate account tries to log in with that delegate id, they will always be shown the unapproved page, regardless as to the number of centres available to their email address (which is tied to their account, but not used to log in in this case). And that should be the only way to "log in" to an unapproved account, as using the email address would return other centres and not show the unapproved ones.

stellake: Yes, great summary! 👍

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -48,6 +48,11 @@ public IActionResult Index(LoginViewModel model)
                 return View("Index", model);
             }
 
+            if (unverifiedAdminUser != null && unverifiedDelegateUsers.Count == 0)
+            {
+                unverifiedDelegateUsers = userService.GetDelegateUsersByUsername(unverifiedAdminUser.EmailAddress);
```
stellake, line 53: Could we just fetch that as part of the first call to the users service (we might need to change the name of the method). Then the user service could:
- Get admin by username
- If admin exists, get delegates by admin.emailAddress
- If not, get delegates by username
This would also reduce the amount of calls to db.

DanBloxham-sw: Yes, that makes a lot more sense.

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -64,14 +69,20 @@ public IActionResult Index(LoginViewModel model)
...
             }
 
-            LogIn(verifiedAdminUser, approvedDelegateUser, model.Password, model.RememberMe);
+            if (verifiedAdminUser == null && verifiedDelegateUsers.Count > 0)
+            {
+                verifiedAdminUser ??= loginService.GetVerifiedAdminUserAssociatedWithDelegateUser
+                (
+                    verifiedDelegateUsers.First(), model.Password
+                );
+            }
```
stellake, line 78: Could we simplify this with something like:
```
var admin = verifiedAdminUser ?? loginService.GetVerifiedAdminUserAssociatedWithDelegateUser(..)
LogIn(admin, ...)
```
Note that we don't care about verifiedDelegateUsers count being > 0, because if it is 0 and admin is null, the user will be redirected to account not approved page by default.

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
```diff
@@ -144,5 +144,44 @@ public void Unapproved_delegate_account_redirect_to_not_approved_page()
             // Then
             result.Should().BeViewResult().WithViewName("AccountNotApproved");
         }
+
+        [Test]
+        public void Log_in_with_admin_id_recovers_associated_delegate_users()
```
stellake, line 149: I think "recovers" is a little bit odd wording for this, maybe "fetches" associated delegate users?

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
```diff
@@ -144,5 +144,24 @@ public void Unapproved_delegate_account_redirect_to_not_approved_page()
             // Then
             result.Should().BeViewResult().WithViewName("AccountNotApproved");
         }
+
+        [Test]
+        public void Log_in_with_approved_delegate_id_fetches_associated_admin_user()
```
DanBloxham-sw, line 149: Previously, there was a test to ensure that getting an admin user would fetch the associated delegate users. Since the logic for getting the user is now in the user service, it is obscured from the controller and so can't be tested here. In order to test it in UserServiceTests, the UserService would have to be refactored  so it also has a DataService. I know that Alex has done that refactoring in 393, so when that is all merged I will have to remake that test.

stellake: Sounds good - do you mind coordinating with him and maybe just making a commit on his branch?

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -26,7 +26,9 @@ public UserService(IDbConnection connection)
         public (AdminUser?, List<DelegateUser>) GetUsersByUsername(string username)
         {
             var adminUser = GetAdminUserByUsername(username);
-            List<DelegateUser> delegateUsers = GetDelegateUsersByUsername(username);
+            string delegateUsername =
```
stellake, line 29: var :)

stellake (approved):
Just one comment about var - happy for this to be merged without rereview once that is fixed :)

PR #271, opened 2021-04-19
HEEDLS-393 Add profile picture and information from database

AlexJacksonDS:
Added the profile image to the My Details expander and converted the data shown on the my account page to all be retrieved from the database using only the Admin Id, Candidate Id and Centre Id from the Claims. The same default profile image is shown as the old site when there is no user profile image.

I've also refactored the UserService into a UserService and a UserDataService. The UserDataService contains all the methods that actually call the database, and the UserService contains logic which uses those. This allows more of the logic in the UserService to be unit tested easily.

Checked that the layout seems fine in Chrome and Chrome mobile devices emulation. Also ran all unit tests (and added some new ones)

image


DigitalLearningSolutions.Web.Tests/TestHelpers/MyAccountHelper.cs
```diff
@@ -0,0 +1,88 @@
+﻿namespace DigitalLearningSolutions.Web.Tests.TestHelpers
+{
+    using DigitalLearningSolutions.Data.Models.User;
+
+    internal class MyAccountHelper
+    {
+        internal static DelegateUser GetDefaultDelegateUser
```
DanBloxham-sw, line 7: Are these GetDefault...User not already present in the UserTestHelper in the data.tests project, or is there a benefit to having them in both test projects?

DigitalLearningSolutions.Web.Tests/ViewModels/MyAccount/MyAccountViewModelTests.cs
```diff
@@ -0,0 +1,69 @@
...
+            // When
+            var returnedModel = new MyAccountViewModel(adminUser, delegateUser, centreName);
+
+            // Then
+            returnedModel.FirstName.Should().BeEquivalentTo(adminUser.FirstName);
+            returnedModel.Centre.Should().BeEquivalentTo(centreName);
+            returnedModel.Surname.Should().BeEquivalentTo(adminUser.LastName);
+            returnedModel.ProfilePicture.Should().BeEquivalentTo(adminUser.ProfileImage);
+            returnedModel.DelegateNumber.Should().BeEquivalentTo(delegateUser.CandidateNumber);
+            returnedModel.User.Should().BeEquivalentTo(adminUser.EmailAddress);
```
DanBloxham-sw, line 28: There are a lot of assertions here to test each individual property. Do we think an Assert.Multiple() might be of benefit? https://docs.nunit.org/articles/nunit/writing-tests/assertions/multiple-asserts.html
The docs suggest it would allow multiple errors to be displayed at once, rather than terminating on the first, but I don't know if that's how the present code already works - or as they are testing what should be the same model if one property is wrong we can assume the rest will be wrong as well.

DigitalLearningSolutions.Data.Tests/Services/UserServiceTests.cs
```diff
@@ -1,66 +1,106 @@
...
             // Given
             var expectedAdminUser = UserTestHelper.GetDefaultAdminUser();
+            var expectedDelegateUsers = UserTestHelper.GetDefaultDelegateUser();
+            A.CallTo(() => userDataService.GetAdminUserByUsername(A<string>._)).Returns(UserTestHelper.GetDefaultAdminUser());
+            A.CallTo(() => userDataService.GetDelegateUsersByUsername(A<string>._)).Returns(new List<DelegateUser> { UserTestHelper.GetDefaultDelegateUser() });
 
             //When
-            var returnedAdminUser = userService.GetAdminUserByUsername("Username");
+            var (returnedAdminUser, _) = userService.GetUsersByUsername("Username");
+            var (_, returnedDelegateUsers) = userService.GetUsersByUsername("SV1234");
```
stellake, line 35: Given that we need to call this method twice, with different usernames, could we split it into two different tests?

DigitalLearningSolutions.Data.Tests/Services/UserServiceTests.cs
```diff
@@ -1,66 +1,106 @@
...
+            var expectedAdminUser = UserTestHelper.GetDefaultAdminUser();
+            var expectedDelegateUser = UserTestHelper.GetDefaultDelegateUser();
+            A.CallTo(() => userDataService.GetAdminUserById(A<int>._)).Returns(UserTestHelper.GetDefaultAdminUser());
+            A.CallTo(() => userDataService.GetDelegateUserById(A<int>._)).Returns(UserTestHelper.GetDefaultDelegateUser());
 
             //When
-            var returnedDelegateUsers = userService.GetDelegateUsersByUsername("SV1234");
-            returnedDelegateUsers.First().Id = expectedDelegateUsers.Id;
+            var (returnedAdminUser, _) = userService.GetUsersById("1", "2");
+            var (_, returnedDelegateUser) = userService.GetUsersById("1", "2");
```
stellake, line 55: Could we just do var (returnedAdminUser, returnedDelegateUser) and call userService once?

DigitalLearningSolutions.Data.Tests/Services/UserServiceTests.cs
AlexJacksonDS: Unsure, I didn't write the initial unit test, The only change I made to it was because I added the userDataService dependency for the methods it calls. I can look to change it

DigitalLearningSolutions.Data.Tests/Services/UserServiceTests.cs
AlexJacksonDS: Yeah probably, I just copied the pattern from the other existing tests I refactored slightly

DigitalLearningSolutions.Data.Tests/Services/UserServiceTests.cs
DanBloxham-sw: I think I initially wrote this test and it was just modified as part of the changes here. While it was originally written to test the scenario of a function returning both in one call, because of the way the test users are set up, they cannot both be returned at once. So I suppose splitting it into two might be best.

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -22,14 +24,15 @@ public IActionResult Index()
...
-            var firstName = User.GetCustomClaim(CustomClaimTypes.UserForename);
-            var surname = User.GetCustomClaim(CustomClaimTypes.UserSurname);
             var centreName = centresService.GetCentreName(centreId);
 
-            var model = new MyAccountViewModel(centreName, userEmail, delegateNumber, firstName, surname);
+            var userAdminId = User.GetCustomClaim(CustomClaimTypes.UserAdminId);
+            var userDelegateId = User.GetCustomClaim(CustomClaimTypes.LearnCandidateId);
+            var (adminUser, delegateUser) = userService.GetUsersById(userAdminId, userDelegateId);
+
+            var model = new MyAccountViewModel(adminUser, delegateUser, centreName);
```
stellake, line 34: I wonder if we even need the centre name anymore as it looks like we're joining with the centres table in the db queries when getting the user data..

AlexJacksonDS: Yeah we shouldn't need to, particularly as I need to get more information off the centres for the third subtask

DigitalLearningSolutions.Data/Models/User/User.cs
```diff
@@ -17,5 +17,7 @@ public abstract class User
         public string? Password { get; set; }
 
         public int? ResetPasswordId { get; set; }
+
+        public byte[] ProfileImage { get; set; }
```
stellake, line 21: Isn't this declared as nullable on the view model?

DigitalLearningSolutions.Data.Tests/Services/UserDataServiceTests.cs
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        [Test]
+        public void GetDelegateUserById_Returns_delegate_users()
+        {
+            // Given
+            var expectedDelegateUsers = UserTestHelper.GetDefaultDelegateUser();
+
+            //When
+            var returnedDelegateUser = userDataService.GetDelegateUserById(299639);
```
stellake, line 69: I'm a little bit nervous doing this by id because it will most likely break for some people locally (as the add test users script can be ran on the db backup at different times). At the same time, I don't want any password hashes being hardcoded in the codebase. 
Do you think we could potentially fetch the users based on an id on the db backup (e.g. 3, which will be the same for everyone) and then check that the fields except for the password hash are as expected? If doing it is a bit messy, we can just check that the fields we care about are as expected (the ones you'll be using on the My account page.

stellake: Or to be honest, we could just create a script that replaces one of the user password hashes that is in the db backup with "password" and then compare against that (which I think is even better than creating/using the new test users) 👀  ?

stellake: Any ideas/opinions? :)

AlexJacksonDS: Yeah didn't think about whether it would break on other people's environments. It would definitely be messy (I guess that is to be expected as these aren't really unit tests) but I could write tests that query against known users from the db backup. Would end up hardcoding in a load of expected user values with the details I need to check instead.

DanBloxham-sw: I don't believe any password hashes are currently stored in the codebase. The UserTestHelper defaults just have "password" as the password by default, and I don't think that is overwritten anywhere. The test user that that is based on also just has password as "password". If one of the records in the db backup (prior to migrations/scripts), can be safely modified to be the new test user, I don't see any problem with, and is almost certainly preferable to the new test user with the potentially varying Id.

stellake: Cool - so maybe we can replace the test user script with a different script that just changes the pw of a single candidate and single admin record that are both present in the db backup (and maybe modify some other fields we might need)? Then, we can make default users have these fields and use the same users for all of these data tests.

DigitalLearningSolutions.Web.Tests/ViewModels/MyAccount/MyAccountViewModelTests.cs
AlexJacksonDS: Looked into this, apparently FluentAssertions don't work with the NUnit Assert.Multiple, but it has a similar AssertionScope that I'll use instead.

DigitalLearningSolutions.Data.Tests/Services/UserServiceTests.cs
AlexJacksonDS: Since this test is no longer testing against the database, it doesn't actually matter what string we are inputting since the fake UserDataService is accepting A<string>._. Combined it into a single call in a single test. There are separate tests in the UserDataServiceTests that test against the actual data

DigitalLearningSolutions.Data.Tests/Services/UserDataServiceTests.cs
AlexJacksonDS: I've updated the PopulateTestUsers script to instead update the first record (other than the admin record ID 1) of each table with the properties we were setting on the created users. These are AdminID 7 and CandidateID 1.
I've also added an update statement for the CentreID 2 so we can set a few of the properties on there as needed.
We should now be able to test against these users/centre for all of the tests we need to, amending the update statements with further data as necessary (e.g. I'll be adding values to the AnswerX fields for 392)

AlexJacksonDS: This may mean we need to remove the created test users from any databases where they've been added

DigitalLearningSolutions.Data.Tests/Helpers/UserTestHelper.cs
```diff
@@ -79,5 +79,10 @@ public static AdminUser GetDefaultAdminUser
                 IsFrameworkDeveloper = isFrameworkDeveloper
             };
         }
+
+        public static string GetDefaultCentreName(string centreName = "North West Boroughs Healthcare NHS Foundation Trust")
```
stellake, line 83: Could this be a variable?

DigitalLearningSolutions.Data.Tests/Services/UserDataServiceTests.cs
```diff
@@ -0,0 +1,72 @@
...
+
+            // Then
+            returnedAdminUser.Should().BeEquivalentTo(expectedAdminUser);
+        }
+
+        [Test]
+        public void GetDelegateUsersByUsername_Returns_delegate_users()
+        {
+            // Given
+            var expectedDelegateUsers = UserTestHelper.GetDefaultDelegateUser();
```
stellake, line 37: Super minor, but should this be singular "User"

DigitalLearningSolutions.Data.Tests/Helpers/UserTestHelper.cs
AlexJacksonDS: Ah whoops I intended to delete that, it's not used any more

DigitalLearningSolutions.Data.Tests/Services/UserDataServiceTests.cs
```diff
@@ -0,0 +1,72 @@
...
+
+            // When
+            var returnedAdminUser = userDataService.GetAdminUserById(7);
+
+            // Then
+            returnedAdminUser.Should().BeEquivalentTo(expectedAdminUser);
+        }
+
+        [Test]
+        public void GetDelegateUserById_Returns_delegate_users()
```
stellake, line 60: ditto about the singular of "user"

DigitalLearningSolutions.Data.Tests/Services/UserServiceTests.cs
```diff
@@ -1,66 +1,117 @@
...
         }
 
         [Test]
-        public void GetAdminUserByUsername_Returns_admin_user()
+        public void GetUsersByUsername_Returns_admin_user_and_delegate_users()
         {
             // Given
             var expectedAdminUser = UserTestHelper.GetDefaultAdminUser();
+            var expectedDelegateUsers = UserTestHelper.GetDefaultDelegateUser();
+            A.CallTo(() => userDataService.GetAdminUserByUsername(A<string>._)).Returns(UserTestHelper.GetDefaultAdminUser());
```
stellake, line 30: Should these return "expected admin user" etc (across all the tests)?

DigitalLearningSolutions.Web/Views/MyAccount/Index.cshtml
```diff
@@ -82,14 +83,32 @@
...
+            </dt>
+            <dd class="nhsuk-summary-list__value">
+              <div class="profile-picture__image-box">
+                @if (Model.ProfilePicture != null)
+                {
+                  <img class="profile-picture__image" src="data:image;base64,@Convert.ToBase64String(Model.ProfilePicture)" alt="Profile Picture" />
+                }
+                else
+                {
+                  <img class="profile-picture__image" src="@Url.Content("~/images/avatar.png")" alt="Default Profile Picture" />
```
stellake, line 107: Just a thought and I'm not 100% sure, but could "Default profile picture" indicate that this will be the user's profile picture if they don't upload anything? Would it be worth naming it "No profile picture"?

stellake (approved):
A few really small comments - happy for you to merge once these are addressed. Don't forget to post in the hee-internal group for people to run the new migration after you've pressed the merge button.

Well done refactoring the userService into service & data service! 👍

DigitalLearningSolutions.Data.Tests/Services/UserServiceTests.cs
DanBloxham-sw: I think that they can be, as these should be testing what the UserService functions return rather than what is passed in to the UserService function, so it should be safe to reuse the expected variable. So I think that might be a better way forward.

DigitalLearningSolutions.Web/Views/MyAccount/Index.cshtml
AlexJacksonDS: Unsure, I guess it depends how the rest of the site and old behaves when a user doesn't have a set profile picture, do they always display with the default? Do we not display a profile picture? I went with default as it implied to me it was a non-user set profile picture.

PR #274, opened 2021-04-20
HEEDLS-346: show notifications in front end

JBsoftwire:
Final MR for HEEDLS-346, displaying the data in the front end.

image
image

image
image

### DIFF
--- a/DigitalLearningSolutions.Web/Styles/myAccount/notificationPreferences.scss
+++ b/DigitalLearningSolutions.Web/Styles/myAccount/notificationPreferences.scss
@@ -3,3 +3,25 @@
 .top-spaced-goback {
   margin-top: 32px;
 }
+
+.preference-list{
+  margin-bottom: 16px;
+}
+
+.preference-row {
+  border-bottom: 1px solid $color_nhsuk-grey-4;
+  margin-top: 24px;
+}
+
+.preference-row:first-child {
+  margin-top: 0;
+}
+
+.preference-button {
+  margin-top: 12px;
+  width: 10em;
+}
+
+.subscription-button--active {
+  background-color: $nhsuk-link-color;
+}
--- a/DigitalLearningSolutions.Web/Styles/myAccount/notificationPreferences.scss
+++ b/DigitalLearningSolutions.Web/Styles/myAccount/notificationPreferences.scss
@@ -17,11 +17,16 @@
   margin-top: 0;
 }
 
-.preference-button {
-  margin-top: 12px;
-  width: 10em;
+.preference-tag {
+  width: 100%;
+  max-width: 10em;
+  text-align: center;
+  margin-bottom: nhsuk-spacing(4);
 }
 
-.subscription-button--active {
-  background-color: $nhsuk-link-color;
+@media (min-width: 48.0625em) {
+  .preference-tag {
+    margin-top: nhsuk-spacing(4);
+    margin-bottom: 0;
+  }
 }
--- a/DigitalLearningSolutions.Web/Styles/myAccount/notificationPreferences.scss
+++ b/DigitalLearningSolutions.Web/Styles/myAccount/notificationPreferences.scss
@@ -22,10 +22,8 @@
   max-width: 10em;
   text-align: center;
   margin-bottom: nhsuk-spacing(4);
-}
 
-@media (min-width: 48.0625em) {
-  .preference-tag {
+  @include govuk-media-query($from: desktop) {
     margin-top: nhsuk-spacing(4);
     margin-bottom: 0;
   }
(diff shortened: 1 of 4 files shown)


DigitalLearningSolutions.Web/Views/NotificationPreferences/Index.cshtml
```diff
@@ -21,6 +21,24 @@
           </span>
         </summary>
         <div class="nhsuk-details__text">
+          <dl class="nhsuk-summary-list preference-list">
+            @foreach (var notification in Model.AdminNotifications) {
+              <div class="nhsuk-grid-row preference-row">
+                <div class="nhsuk-grid-column-three-quarters">
+                  <div class="nhsuk-heading-xs">@notification.NotificationName</div>
+                  <p>@Html.Raw(notification.Description)</p>
```
JBsoftwire, line 29: I'm always suspicious when I find a call to raw HTML, but this comes one-way out of the database so as long as what goes into that is properly controlled, we should be fine.

DigitalLearningSolutions.Web/Views/NotificationPreferences/Index.cshtml
```diff
@@ -34,6 +52,25 @@
...
+          <dl class="nhsuk-summary-list preference-list">
+            @foreach (var notification in Model.DelegateNotifications)
+            {
+              <div class="nhsuk-grid-row preference-row">
+                <div class="nhsuk-grid-column-three-quarters">
+                  <div class="nhsuk-heading-xs">@notification.NotificationName</div>
+                  <p>@Html.Raw(notification.Description)</p>
+                </div>
+                <div class="nhsuk-grid-column-one-quarter">
+                  <a class="preference-button nhsuk-button @(notification.Accepted ? "subscription-button--active" : "nhsuk-button--secondary")" role="button">
```
stellake, line 64: Could we make these tags as the user won't be able to interact with it - https://service-manual.nhs.uk/design-system/components/tag

Here are some colours: 
https://service-manual.nhs.uk/design-example/tag-colours?fullpage=undefined&blankpage=undefined

We could use grey for unsubscribed and either blue or green for subscribed? You could post screenshots to the external HEE channel and ask which version HEE prefers :)

JBsoftwire: Oup, I had been assuming the user would be clicking on these in some later ticket to subscribe and unsubscribe and using the Update Preferences button to send the request, but this makes more sense. I'll fix it.

DigitalLearningSolutions.Web/Views/NotificationPreferences/Index.cshtml
stellake: The users cannot change the notification data so I think that's ok. On another note - could we specify the font size on the paragraph too? (I think NHS has body font classes in addition to the heading ones)

DigitalLearningSolutions.Web/Views/NotificationPreferences/Index.cshtml
stellake: Yeah - unfortunately there's no good way of showing the tags in the wireframes, so it looks really similar to buttons 😢

DigitalLearningSolutions.Web/Views/NotificationPreferences/Index.cshtml
```diff
@@ -34,6 +52,25 @@
           </span>
         </summary>
         <div class="nhsuk-details__text">
+          <dl class="nhsuk-summary-list preference-list">
+            @foreach (var notification in Model.DelegateNotifications)
```
JBsoftwire, line 56: This code is duplicated and fixing both was a bit more fiddly than I'd like. Before I merge this, is there a good example of how to make it a component I can call in both places?

stellake: Yes, could you use partial views using a Partial Tag Helper - https://docs.microsoft.com/en-us/aspnet/core/mvc/views/partial?view=aspnetcore-3.1

There is an example how one is used in LearningPortal/Current index view - the cards are rendered as `<partial name="SelfAssessments/_SelfAssessmentCard" model="currentLearningItem" />`

DigitalLearningSolutions.Web/Styles/myAccount/notificationPreferences.scss
```diff
@@ -3,3 +3,30 @@
...
+}
+
+.preference-tag {
+  width: 100%;
+  max-width: 10em;
+  text-align: center;
+  margin-bottom: nhsuk-spacing(4);
+}
+
+@media (min-width: 48.0625em) {
```
stellake, line 27: Steve has used this syntax in his MR, which I think is nicer and ensures we use the NHS breakpoints (feel free to check out his MR):
@include govuk-media-query($from: tablet) {
    // Styles
  }

stellake (approved):
Looks good - just a comment about the media query. 👍 

No need for re-review so feel free to merge when that's implemented.

PR #277, opened 2021-04-21
HEEDLS-395 basic registration frontend

davidm-m:
Screenshots of every page on desktop:
image
image
image
image

And on mobile:
image
image
image
image
image


DigitalLearningSolutions.Data/Services/CentresService.cs
```diff
@@ -48,5 +50,16 @@ FROM Centres
 
             return name;
         }
+
+        public IEnumerable<(int, string)> GetActiveCentres()
```
stellake, line 54: Ah, we should probably add a test for the new data methods and as per my comment in the hee-internal channel, move them to a new "DataServices" folder (which will effectively be repositories)

DigitalLearningSolutions.Data/Services/CentresService.cs
```diff
@@ -48,5 +50,16 @@ FROM Centres
...
             return name;
         }
+
+        public IEnumerable<(int, string)> GetActiveCentres()
+        {
+            var centres = connection.Query<(int, string)>(
+                @"SELECT CentreID, CentreName
+                        FROM Centres
+                        WHERE Active = 1
+                        ORDER BY CentreName"
```
stellake, line 60: Should we make it clear in the method name that these will be alphabetically ordered? Alternatively, we could do the ordering in the code so this could be reused for other functionality too, but I'm struggling to think of other places that might need all of the active centres. 🤔

DigitalLearningSolutions.Data/Services/JobGroupsService.cs
```diff
@@ -0,0 +1,53 @@
...
+            var name = connection.QueryFirstOrDefault<string?>(
+                @"SELECT JobGroupName
+                        FROM JobGroups
+                        WHERE JobGroupID = @jobGroupId",
+                new { jobGroupId }
+            );
+            if (name == null)
+            {
+                logger.LogWarning(
+                    $"No centre found for job group id {jobGroupId}"
```
stellake, line 36: Centre? 👀

DigitalLearningSolutions.Data/Services/JobGroupsService.cs
```diff
@@ -0,0 +1,53 @@
...
+                );
+            }
+
+            return name;
```
stellake, line 40: Given that this returns the name even if it's null, should we make the return value nullable? Or are we always expecting the name to exist and are happy if it throws if that's not the case?

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -0,0 +1,169 @@
...
+        public IActionResult Index()
+        {
+            if (User.Identity.IsAuthenticated)
+            {
+                return RedirectToAction("Index", "Home");
+            }
+
+            var delegateRegistrationData = TempData.Get<DelegateRegistrationData>();
+
+            if (delegateRegistrationData == null || !Request.Cookies.ContainsKey(CookieName))
```
stellake, line 33: VS seems to think that this cannot be null - could we fix the return type in TempDataExtension?

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -0,0 +1,169 @@
...
+            {
+                var id = Guid.NewGuid();
+                delegateRegistrationData = new DelegateRegistrationData { Id = id };
```
stellake, line 36: Should we either:
1. Use the id created in the DelegateRegistrationData constructor 
2. Use the id generated in code, but add the id as a param in the DelegateRegistrationData constructor so we don't need to set the id and then override its value

DigitalLearningSolutions.Data/Services/JobGroupsService.cs
davidm-m: I just forgot to make this one nullable, nice catch

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -0,0 +1,169 @@
...
+
+                Response.Cookies.Append(
+                    CookieName,
+                    id.ToString(),
+                    new CookieOptions
+                    {
+                        Expires = DateTimeOffset.UtcNow.AddDays(30)
```
stellake, line 43: Any reason for the choice of 30? Also, should we pull the 30 into variable at the top of the file (e.g. registrationCookieExpiryTimeInDays) for extra emphasis?

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -0,0 +1,169 @@
...
+        [HttpPost]
+        public IActionResult Index(RegisterViewModel model)
+        {
+            if (!ModelState.IsValid)
+            {
+                return View(model);
+            }
+
+            var data = TempData.Get<DelegateRegistrationData>();
+            data.RegisterViewModel = model;
```
stellake, line 62: Could this be null?

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -0,0 +1,169 @@
...
+            TempData.Set(data);
+
+            return RedirectToAction("LearnerInformation");
+        }
+
+        [ServiceFilter(typeof(RedirectEmptySessionData<DelegateRegistrationData>))]
+        [Route("Register/learner-information")]
```
stellake, line 69: Is there a reason we're overriding the default routes here?

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
davidm-m: Rider has the same warning and I forgot to fix it 🙄
Making the return type nullable clears it all up

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
davidm-m: I'm not totally sure why we need a cookie anyway. [The example multi-page form](https://github.com/tomdoughty/nhsuk.base-application/blob/main/src/nhsuk.base-application/Controllers/ExampleFormController.cs) sets one for 30 days but it looks like TempData is stored in a cookie anyway, and we never use this id for anything, so I don't know if we need it.

DigitalLearningSolutions.Web/ViewModels/Register/LearnerInformationViewModel.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace DigitalLearningSolutions.Web.ViewModels.Register
+{
+    using System.ComponentModel.DataAnnotations;
+
+    public class LearnerInformationViewModel
+    {
+        public LearnerInformationViewModel()
+        {
+            Centre = 0;
```
stellake, line 9: If these are nullable fields, do we need to set them as "0" in the constructor?

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
davidm-m: The ServiceFilter on each of these methods guarantees that it won't be null or else they'll be redirected to the first page of the form with a new DelegateRegistrationData. I forgot to add it to this specific method though, I'll just stick that on (and assert not null for all the places we can).

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
davidm-m: Good question 🤔
I guess the answer's no, I'll take them out

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
davidm-m: Good idea, I'll just use the one from the constructor.

DigitalLearningSolutions.Web/Views/Register/Password.cshtml
```diff
@@ -0,0 +1,61 @@
...
+          </span>
+        }
+        <input class="nhsuk-input nhsuk-u-width-one-half @confirmPasswordInputErrorClass" id="ConfirmPassword" spellcheck="false" autocomplete="new-password" aria-describedby="confirm-password-error" asp-for="ConfirmPassword"/>
+      </div>
+
+      <button class="nhsuk-button" type="submit">Next</button>
+    </form>
+
+    <p>
+      <vc:back-link asp-controller="Register" asp-action="learner-information" />
```
stellake: Not sure about marking the back link as paragraph in HTML syntax - is this a weird accessibility thing? 😮

DigitalLearningSolutions.Web/Views/Register/Summary.cshtml
```diff
@@ -0,0 +1,107 @@
...
+          @if (termsError) {
+            <span class="nhsuk-error-message" id="terms-error">
+              <span class="nhsuk-u-visually-hidden">Error:</span> @ViewData.ModelState["Terms"].Errors[0].ErrorMessage
+            </span>
+          }
+          <div class="nhsuk-checkboxes">
+            <div class="nhsuk-checkboxes__item">
+              <input class="nhsuk-checkboxes__input" id="terms" name="terms" asp-for="Terms"/>
+              <label class="nhsuk-label nhsuk-checkboxes__label" for="terms">
+                I have read, understood and agree to abide by the Digital Learning Solutions Terms and Conditions.
```
stellake, line 100: We should make the "Terms and Conditions" a link that opens in a new tab, like on the current website. Note that the NHS guideline is to have "(opens in new window)" after a link that opens in new tab, but HEE have explicitly made a decision not to add this additional text!

DigitalLearningSolutions.Data/Services/CentresService.cs
davidm-m: Yeah, my logic was that ordering by id would be arbitrary and not very useful, especially for such a long list, whereas ordering alphabetically would be more intuitive.

DigitalLearningSolutions.Data/Services/CentresService.cs
davidm-m: I'm having trouble thinking of the sort of tests we can add for GetActiveCentres. I guess that it returns at least one centre, but we can't even check that they're all active without extending the method.

DigitalLearningSolutions.Web/ViewModels/Register/RegisterViewModel.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace DigitalLearningSolutions.Web.ViewModels.Register
+{
+    using System.ComponentModel.DataAnnotations;
+
+    public class RegisterViewModel
+    {
+        [Required(ErrorMessage = "Please enter a first name.")]
```
stellake, line 7: I'm wondering if we should use error messages in the NHS example format (I remember reading that Please shouldn't be used, maybe it was in govuk website):
image

stellake: Here it is: https://design-system.service.gov.uk/components/error-message/#be-clear-and-concise

(I don't think we follow it everywhere, but might be good to try and do it as much as possible)

stellake:
Styling wise looks good, just two minor comments:
- Could we make the centre and job group drop downs the same size on desktop?
- Could we move the "Alternatively" block on the first page slightly lower to show that it's separate form the form (right now the spacing between the title and form is much bigger than the spacing between the form and the "Alternatively" section)?

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
stellake: Ah, interesting! Do you think there'll be some issues storing the password data in a cookie (especially given the non-hashed form)?

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -0,0 +1,164 @@
...
+            }
+
+            // TODO HEEDLS-396 create new user and redirect to confirmation page
+            return View(model);
+        }
+
+        private SummaryViewModel MapToSummary(DelegateRegistrationData data)
+        {
+            var centre = centresService.GetCentreName((int)data.LearnerInformationViewModel.Centre!);
+            var jobGroup = jobGroupsService.GetJobGroupName((int)data.LearnerInformationViewModel.JobGroup!);
```
stellake, line 153: What happens in the UI if these ids don't exist (e.g. cookie has expired)?

DigitalLearningSolutions.Data/Services/CentresService.cs
stellake: Yeah, I agree with this so maybe just make it clear that we use ordering in the name of the method? 👍

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -0,0 +1,164 @@
...
+            return RedirectToAction("LearnerInformation");
+        }
+
+        [ServiceFilter(typeof(RedirectEmptySessionData<DelegateRegistrationData>))]
+        [HttpGet]
+        public IActionResult LearnerInformation()
+        {
+            var data = TempData.Get<DelegateRegistrationData>();
+            var viewModel = data!.LearnerInformationViewModel;
+            ViewBag.Centres = centresService.GetActiveCentres();
```
stellake, line 75: Ah, one thing I forgot to ask - is there a reason we're setting these on ViewBag rather than on a view model where they'd be typed?

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
davidm-m: Documentation is really sparse as to what TempData is and how it works but it _looks_ like the cookie is encrypted, so I don't think this will be a problem, especially if we make sure it's cleared once they actually register.

davidm-m:
>     * Could we make the centre and job group drop downs the same size on desktop?

Easiest way to match them up but not unnecessarily cut off the longer centre names is to make both full width. They're quite wide now on desktop but I think it's fine.
image

DigitalLearningSolutions.Web/Views/Register/Password.cshtml
```diff
@@ -0,0 +1,61 @@
...
+  var confirmPasswordError = ViewData.ModelState["ConfirmPassword"]?.Errors?.Count > 0;
+  var passwordInputErrorClass = passwordError ? "nhsuk-select--error" : "";
+  var passwordFormErrorClass = passwordError ? "nhsuk-form-group--error" : "";
+  var confirmPasswordInputErrorClass = confirmPasswordError ? "nhsuk-select--error" : "";
+  var confirmPasswordFormErrorClass = confirmPasswordError ? "nhsuk-form-group--error" : "";
+}
+
+<div class="nhsuk-grid-row">
+  <div class="nhsuk-grid-column-full">
+    <form method="post" asp-action="Password">
```
SteveJacksonSoft, line 17: Just had a thought - it's probably worth having anti-forgery tokens in these forms if you haven't already added them: https://docs.microsoft.com/en-us/aspnet/web-api/overview/security/preventing-cross-site-request-forgery-csrf-attacks

Otherwise someone who knows the POST url could get users to change their passwords involuntarily since we're using cookie auth.

davidm-m: We already do I'm pretty sure, visiting the website sets a `.AspNetCore.Antiforgery` cookie

DigitalLearningSolutions.Web/ViewModels/Common/PasswordViewModel.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace DigitalLearningSolutions.Web.ViewModels.Common
+{
+    using System.ComponentModel.DataAnnotations;
+
+    public class PasswordViewModel
+    {
+        [Required(ErrorMessage = "Enter a password.")]
+        [MinLength(8, ErrorMessage = "Password must be at least 8 characters.")]
```
stellake, line 8: Do we need any max length constraints (thinking about db constraints)?

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
davidm-m: I've finally found [some concrete docs](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/app-state?view=aspnetcore-3.1#tempdata-providers) on it, and TempData cookies are encrypted! I'll make sure it's only ever stored hashed as part of the backend work and then I think everything will be fine.

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
davidm-m: The main reason is that I didn't want to store them in TempData, as cookies have a 4096 byte limit and at the moment TempData is just a collection of all the ViewModels. We could get around this in a few different ways but it feels simplest to me to just chuck them into ViewBag when they're needed.

DigitalLearningSolutions.Web/ViewModels/Common/PasswordViewModel.cs
davidm-m: Since passwords are hashed and truncated we don't need a limit on them as far as the db is concerned, but other fields probably do. We should also enforce size limits on everything to make sure we don't go over 4096 bytes in TempData since that's the maximum size of a cookie.

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
davidm-m: If the cookie's expired then we will redirect back to the first page of the form before this function gets called. Theoretically though someone could visit the first page to get the cookie and then manually enter the summary URL which would cause a problem at the moment.
I've got 2 ideas for how to deal with that:
* Automatically redirect back to the appropriate page for the first missing or invalid field when loading the summary page
* Display error messages on the summary page for each missing or invalid field
Any preference on which one? I'm leaning towards the first because I don't know if there's an NHS standard way to display errors in tables

stellake: Yes, I think you're right - the first option sounds good. If it's too time consuming we might be able to leave it, because the probability of this happening should be fairly slim. :)

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
stellake: Oh yes, I wasn't implying to keep them in TempData, just in a typed model. If it's a lot more work to add them, happy to leave it! 👍

DigitalLearningSolutions.Web/ViewModels/Common/PasswordViewModel.cs
```diff
@@ -0,0 +1,21 @@
...
+        [MinLength(8, ErrorMessage = "Password must be at least 8 characters")]
+        [MaxLength(100, ErrorMessage = "Password must be at most 100 characters")]
+        [RegularExpression(@"(?=.*?[^\w\s])(?=.*?[0-9])(?=.*?[A-Za-z]).*",
+            ErrorMessage = "Password must contain at least 1 letter, 1 number and 1 symbol")]
+        [DataType(DataType.Password)]
+        public string? Password { get; set; }
+
+        [Required(ErrorMessage = "Confirm your password")]
+        [DataType(DataType.Password)]
+        [Compare("Password", ErrorMessage = "Password and confirm password do not match.")]
```
stellake, line 17: Cheeky full stop at the end of the message :)

DigitalLearningSolutions.Web/ViewModels/Register/SummaryViewModel.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace DigitalLearningSolutions.Web.ViewModels.Register
+{
+    using System.Collections.Generic;
+    using System.ComponentModel.DataAnnotations;
+
+    public class SummaryViewModel: IValidatableObject
+    {
+        public string? FirstName { get; set; }
```
stellake, line 8: Can these be non-optional? (I'd expect these to always exist in summary page and we could set values in the constructor on page load)

DigitalLearningSolutions.Web/Controllers/FrameworksController/AssessmentQuestions.cs
```diff
@@ -213,7 +213,7 @@ public IActionResult EditAssessmentQuestion(int frameworkId, int assessmentQuest
             {
                 return StatusCode(403);
             }
-            SessionAssessmentQuestion sessionAssessmentQuestion = TempData.Get<SessionAssessmentQuestion>();
+            SessionAssessmentQuestion sessionAssessmentQuestion = TempData.Peek<SessionAssessmentQuestion>();
```
stellake, line 216: So beautiful! 🤩

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
stellake: Could we add a TODO for hashing the password just in case? 🙏

DigitalLearningSolutions.Data.Tests/DataServices/CentresDataServiceTests.cs
```diff
@@ -1,29 +1,30 @@
-﻿namespace DigitalLearningSolutions.Data.Tests.Services
```
stellake, line 1: Could we add the new page to the Automated UI tests (don't mind if it's part of the next ticket as you can probably simulate the whole registration journey to go through all the pages - happy to have a chat about this)?

DigitalLearningSolutions.Web/Views/Register/Password.cshtml
davidm-m: [This is done automatically](https://docs.microsoft.com/en-us/aspnet/core/security/anti-request-forgery?view=aspnetcore-3.1#aspnet-core-antiforgery-configuration), thanks @DanBloxham-sw for the link 😀

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
davidm-m: Added a TODO for this

stellake (approved):
Looks good! :)

PR #278, opened 2021-04-22
HEEDLS-361 - Choose a Centre on log in - back end and front end

DanBloxham-sw:
Made changes to Log In such that if the user has accounts at different centres available (can vary based on which account log in details are used) they will be taken to a new page that allows them to choose which centre (and thus which accounts) they will log in as.

**Minor concern: spacing/size of the Admin/Delegate role tags**
In the screenshots below, we can see that the tags for Admin and the tags for Delegate roles are not the same size, which can lead to the page not looking as neat as it perhaps could be. The mobile view is also somewhat spaced out when there are both tags present. I'm not sure what the best way to sort this would be (or if it needs sorting at all) - I tried using the nhsuk-u-width classes, but these either compress the tag boundaries to be shorter than the words in some screen sizes, or cause the tags to be displayed one-on-top of the other with additional padding. Any thoughts?

Tested:
- Ran all unit tests
- Manually ran through different combinations of accounts (including approved and unapproved)
- Manually tested Chrome, IE11, Edge, Firefox
- Tested chrome's mobile view
- Tested No JS
- Tested with screen reader

Screen shots of new page:
Choose a centre

Mobile view:
Choose a centre mobile view


DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
```diff
@@ -163,5 +165,94 @@ public void Log_in_with_approved_delegate_id_fetches_associated_admin_user()
...
+            var result =
+                controller.Index(LoginTestHelper.GetDefaultLoginViewModel());
+
+            // Then
+            result.Should().BeRedirectToActionResult()
+                .WithControllerName("Home").WithActionName("Index");
+        }
+
+        [Test]
+        public void Multiple_accounts_with_different_approved_statuses_should_use_accounts_with_approved_statuses()
```
stellake, line 212: This is good, but the test name is a bit misleading - this test would pass if it only used accounts with unapproved statuses (same with the password test above).

DigitalLearningSolutions.Web/Views/Login/ChooseACentre.cshtml
```diff
@@ -0,0 +1,60 @@
...
+
+    <h1 id="page-heading" class="nhsuk-heading-xl">Choose a Centre</h1>
+    <p class="nhsuk-body-l">
+      You are registered at more than one centre. Please choose the centre you would like to sign in to.
+    </p>
+
+    <table role="table" class="nhsuk-table-responsive">
+      <thead role="rowgroup" class="nhsuk-table__head">
+      <tr role="row">
+        <th role="columnheader" class="" scope="col">
```
stellake, line 20: Are these empty classes required?

DigitalLearningSolutions.Web/Views/Login/ChooseACentre.cshtml
```diff
@@ -0,0 +1,60 @@
...
+              @if (centreRow.IsAdmin) {
+                <strong class="nhsuk-tag nhsuk-tag--blue centre-role-tag">Admin</strong>
+              }
+              @if (centreRow.IsDelegate) {
+                <strong class="nhsuk-tag nhsuk-tag--grey centre-role-tag">Delegate</strong>
+              }
+            </td>
+            <td role="cell"class="nhsuk-table__cell">
+              <input type="hidden" name="centreId" value="@centreRow.CentreId"/>
+              <button class="nhsuk-button nhsuk-u-margin-bottom-1 nhsuk-u-padding-bottom-0 nhsuk-u-padding-top-0" type="submit">
```
stellake, line 48: I wonder if instead of making this a form and adding this hidden input, we could navigate to a url where centre id is either embedded in the url or a query string (e.g. /Centre/1/Login or /Login?CentreId=1 or /ChooseCentre?CentreId=1) 🤔 Then this can be a simple anchor tag with a role button rather than a form submission, which is a bit odd after the user has already submitted a form (not 100% sure how this will work with a screenreader)

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -40,37 +41,84 @@ public IActionResult Index(LoginViewModel model)
...
+
+            if (availableCentres.Count == 1)
+            {
+                return LogIn(verifiedAdminUser, approvedDelegateUsers.FirstOrDefault(), model.RememberMe);
+            }
+
+            TempData.Clear();
+            TempData["RememberMe"] = model.RememberMe;
+            TempData.Set(adminUser);
+            TempData.Set(approvedDelegateUsers);
```
stellake, line 81: Could we extract setting TempData into a method?

Also, I think I'd have a minor preference to use spacing to group some of the logic together - e.g. remove L72 as then the var available centres is together with where it's used, in the "Check available centres" section.

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -40,37 +41,84 @@ public IActionResult Index(LoginViewModel model)
...
+
+            return View("ChooseACentre", availableCentres);
```
stellake, line 83: I'd have a minor preference of having ChooseACentreViewModel with a field userDetailsInCentre[] or something similar. (so instead of passing the array to the view, we pass a viewmodel that has a field for this array. Then, if we ever need to add any other properties, it'll be easy to do :)

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -40,37 +41,84 @@ public IActionResult Index(LoginViewModel model)
...
+        }
+
+        [HttpPost]
+        public IActionResult ChooseACentre(int centreId)
+        {
+            var rememberMe = (bool)TempData["RememberMe"];
+            var adminAccount = TempData.Get<AdminUser>();
```
stellake, line 90: Does this "Get" actually delete the data (it looks like it's "Peek" behind the schenes)?

DigitalLearningSolutions.Web/ViewModels/Login/ChooseACentreViewModel.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace DigitalLearningSolutions.Web.ViewModels.Login
+{
+    public class ChooseACentreViewModel
+    {
+        public ChooseACentreViewModel()
```
stellake, line 5: Is this used?

stellake:
I really like the desktop design. I know that the client was keen to see the version with links instead of buttons - do you think we could show both versions in demo tomorrow and see what they think? We can also see what they think about the tag colours. :)

Regarding the mobile, I think moving the tags next to each other might make sense. I also wonder if it might look nicer if we didn't have a border between the centre name and tags in the mobile version.
image

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -40,37 +41,84 @@ public IActionResult Index(LoginViewModel model)
...
+
+            return LogIn(adminAccountForChosenCentre, delegateAccountForChosenCentre, rememberMe);
         }
 
-        private void LogIn(AdminUser? adminUser, DelegateUser? delegateUser, bool rememberMe)
+        private List<ChooseACentreViewModel> GetAvailableCentres(AdminUser? adminUser, List<DelegateUser> delegateUsers)
+        {
+            var availableCentres = new List<ChooseACentreViewModel>();
+
+            foreach (var delegateUser in delegateUsers)
```
stellake, line 104: Could we just map the delegateUser to the available centres here instead to have this more concise (.Select)?

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
DanBloxham-sw: So should the test itself be kept the same, but rename it to something like "When Log In finds approved and unapproved accounts it should log in to approved accounts only"?
And the password test becomes "When Log In finds accounts with different passwords it should log in to accounts that match input password only"

DigitalLearningSolutions.Web/Views/Login/ChooseACentre.cshtml
DanBloxham-sw: The empty classes shouldn't be doing anything, but they are present in the examples on the NHS guidelines. https://service-manual.nhs.uk/design-system/components/table guessing we would prefer them to be removed?

DanBloxham-sw:
I've updating the styling among other things, so while the desktop view of the page is unaffected, the mobile/tablet view now has extra headers (that should have been included in the first place) and some extra styling so that the tags are close together, as well as removing lines between the cells. Screen shot below. Though I realise now that a user looking at this page on mobile won't see the warning at the bottom unless they scroll all the way there. Do we want to move it to the top of the page (underneath the other paragraph asking the user to pick a centre) or just leave it as is because it's not that important?

Choose a centre mobile view reformatted

stellake:
Re the new mobile version - I think it's ok to leave the warning at the bottom for now, but I'll make a note to ask about this in the user testing session (I expect the number of users that use the site on mobile to be quite low, and the number of users that have accounts in several centres to be minima).
Styling wise, I think we should change the "Choose" button margins such that there's more space between the button and the border than the button and the rest centre-specific data (just to emphasise that these should be grouped together). Otherwise, it looks just like I imagined! 🎉 🙏

stellake:
Ah, looks like you have some merge conflicts that need to be resolved 😢

DigitalLearningSolutions.Web/Views/Login/ChooseACentre.cshtml
stellake: Yep - no need for empty classes, they'll just make everything else harder to read (not sure why the example has empty classes 🤔 ).

DigitalLearningSolutions.Web/ViewModels/Login/ChooseACentreViewModel.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace DigitalLearningSolutions.Web.ViewModels.Login
+{
+    using System.Collections.Generic;
+
+    public class ChooseACentreViewModel
+    {
+        public ChooseACentreViewModel()
+        {
+            CentreUserDetails = new List<CentreUserDetails>();
```
stellake, line 9: Would it be better to set the userDetails in the constructor rather than declaring it as an empty list and then setting the field?

DigitalLearningSolutions.Web/Styles/login/chooseACentre.scss
```diff
@@ -0,0 +1,13 @@
+﻿@import "~nhsuk-frontend/packages/core/all";
+
+.centre-role-tag {
+    text-align: center;
+    margin-right: 5px;
+}
+
+.choose-centre-td {
+  @include mq($until: desktop) {
```
stellake, line 9: Is there a reason you're using "mq" over "govuk-media-query" syntax (not sure if they're any different)? Also, I think !important might cause some issues in the future (as nothing can override it), so we could make use of the CSS specificity and do something like Steve has done in https://github.com/TechnologyEnhancedLearning/DLSV2/pull/281/files#
(check out welcome.scss L16 - class in a class is more specific than just a class so CSS will use that style when deciding between the two)

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -81,6 +127,7 @@ private void LogIn(AdminUser? adminUser, DelegateUser? delegateUser, bool rememb
                 IssuedUtc = DateTime.UtcNow
             };
             HttpContext.SignInAsync("Identity.Application", new ClaimsPrincipal(claimsIdentity), authProperties);
+            return RedirectToAction("Index", "Home");
```
stellake, line 130: Is there a reason behind this change (now the login method logs the user in and redirects to home page)?

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -40,37 +41,82 @@ public IActionResult Index(LoginViewModel model)
...
+            TempData.Clear();
+
+            var adminAccountForChosenCentre = adminAccount?.CentreId == centreId ? adminAccount : null;
+            var delegateAccountForChosenCentre =
+                approvedDelegateAccounts?.FirstOrDefault(du => du.CentreId == centreId);
+
+            return LogIn(adminAccountForChosenCentre, delegateAccountForChosenCentre, rememberMe);
+        }
+
+        private List<CentreUserDetails> GetAvailableCentres(AdminUser? adminUser, List<DelegateUser> delegateUsers)
```
stellake, line 96: I wonder whether this would be better in a helper or a service to make unit testing easier.. 🤔 What do you think?

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
stellake: Maybe something like:
"When_user_has_multiple_accounts_with_different_passwords_redirects_home_rather_than_choose_a_centre" (or redirects home/does not redirect to choose a centre)
I think the main thing is that we don't test which user is logged in, but that both are not considered valid accounts for Choose a centre page. So "only_log_in_to_ones_matching_input_password" could not be true (e.g. we could only log in the account where password doesn't match), but the test would pass .

DigitalLearningSolutions.Web/Controllers/LoginController.cs
DanBloxham-sw: I just thought that because the LogIn method will be called from the single centre log in and also the choose a centre log in that both send the user to the same place, I might as well save a line of code and bung it in there. Would we prefer the extra specificity of before where each method returning specific actions rather than having to look through the other methods it calls?

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
DanBloxham-sw: When initially writing the test, I was trying to figure out the best way to check that accounts without password matches were not available, and could only arbitrarily do this by making the case where:
- If all accounts were seen as acceptable for logging in, then there would be too many centres and thus would be redirected to the Choose A Centre page
- If accounts without password matches were excluded, then the acceptable accounts only have 1 centre, and thus log in and get taken to the Home page

I think I see now what you mean, that the test itself is not specific enough. However, I'm not sure your suggested test names make the situation any better - that name would not work if the user had 5 different delegate accounts at different centres, and some of them had the correct password and some did not.

I think the test itself needs redoing, as the point was to make sure that accounts that don't match the password are not considered acceptable. Currently, I don't think there is an easy way to do this, but your other comment about checking available centres in a helper/service might be the way forward, so I'll refactor into using that.

DigitalLearningSolutions.Web/Controllers/LoginController.cs
stellake: I think it's ok actually as I don't think the redirects should ever be different between those pages.

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -55,5 +58,19 @@ public UserService(IUserDataService userDataService)
 
             return (adminUser, delegateUser);
         }
+
+        public List<CentreUserDetails> GetAvailableCentres(AdminUser? adminUser, List<DelegateUser> delegateUsers)
+        {
+            var availableCentres = delegateUsers.Select(du => new CentreUserDetails
```
stellake, line 64: Super minor, but could we reformat this such that .Select is on a new line and the .List is on a new line to make it easier to read?

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -1,13 +1,16 @@
...
+    using System.Linq;
     using DigitalLearningSolutions.Data.Models.User;
 
     public interface IUserService
     {
         public (AdminUser?, List<DelegateUser>) GetUsersByUsername(string username);
         public (List<AdminUser>, List<DelegateUser>) GetUsersByEmailAddress(string emailAddress);
         public (AdminUser?, DelegateUser?) GetUsersById(string? adminId, string? delegateId);
+
+        public List<CentreUserDetails> GetAvailableCentres(AdminUser adminUser, List<DelegateUser> delegateUsers);
```
stellake, line 13: I'm thinking that the name GetUserCentres might work better here, as "Available" feels more ambiguous, but this could be subjective.

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
```diff
@@ -163,5 +170,106 @@ public void Log_in_with_approved_delegate_id_fetches_associated_admin_user()
...
+            A.CallTo(() => loginService.VerifyUsers(A<string>._, A<AdminUser>._, A<List<DelegateUser>>._))
+                .Returns((expectedAdminUser, new List<DelegateUser>()));
+
+            // When
+            controller.Index(LoginTestHelper.GetDefaultLoginViewModel());
+
+            // Then
+            A.CallTo(() => userService.GetAvailableCentres(
+                    A<AdminUser>.That.Matches(au => au.IsSameOrEqualTo(expectedAdminUser)),
+                    A<List<DelegateUser>>.That.Matches(list => list.SequenceEqual(new List<DelegateUser>()))))
```
stellake, line 219: Could we use .That.IsEmpty() here?

stellake: And maybe IsEqualTo for admin user?

stellake (commented):
Looking good - just a few minor naming/formatting comments!

stellake (approved):
Looks good!

PR #279, opened 2021-04-22
HEEDLS-392 Add delegate details to My Details expander

AlexJacksonDS:
Added the extra delegate fields to the My Account page. This pull extra details from the Candidate table using the existing queries, and there is a new pair of services to get the custom prompts for a particular centre.

Added new unit tests for these. Checked that the layout worked in desktop and mobile size and checked that a screen reader reads out the values correctly.

image


DigitalLearningSolutions.Data/Services/UserDataService.cs
```diff
@@ -70,9 +70,17 @@ public UserDataService(IDbConnection connection)
...
                         cd.Approved,
-                        cd.ProfileImage
+                        cd.ProfileImage,
+                        cd.Answer1,
+                        cd.Answer2,
+                        cd.Answer3,
+                        cd.Answer4,
+                        cd.Answer5,
+                        cd.Answer6,
+                        jg.JobGroupName
```
AlexJacksonDS, line 80: I noticed that as part of the registration journey we now have an IJobGroupsService method string? GetJobGroupName(int jobGroupId);
Should I be using this and only querying the cd.JobGroupId here rather than just joining on the table?

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -28,7 +27,9 @@ public IActionResult Index()
             var userDelegateId = User.GetCustomClaim(CustomClaimTypes.LearnCandidateId);
             var (adminUser, delegateUser) = userService.GetUsersById(userAdminId, userDelegateId);
 
-            var model = new MyAccountViewModel(adminUser, delegateUser);
+            var customPrompts = customPromptsService.GetCustomPromptsForCentreByCentreId(delegateUser?.CentreId ?? adminUser?.CentreId);
```
DanBloxham-sw, line 30: It looks to me like the CustomPrompts will always be obtained as long as the user has either an admin or a delegate account. Is that the case? If so, is there a particular need to grab them when no delegate account is found, given we are only displaying them when there is a delegate account?

AlexJacksonDS: Yeah I think I have the logic here slightly wrong. Should just be trying to pass in delegateUser?.CentreId

DigitalLearningSolutions.Web/ViewModels/MyAccount/MyAccountViewModel.cs
```diff
@@ -1,19 +1,39 @@
...
+        }
+
+        private CustomFieldViewModel? PopulateCustomFieldViewModel(CustomPrompt? customPrompt, string? answer)
+        {
+            if (customPrompt == null)
+            {
+                return null;
+            }
+
+            return new CustomFieldViewModel(customPrompt.CustomPromptText, customPrompt.Mandatory, answer);
```
DanBloxham-sw, line 36: The ticket mentions that one of the things that should be displayed is the options - "Options is all potential options that should be displayed in a dropdown. If it’s null, we should display a text input." Are the options not being brought into the view model here for any particular reason?

AlexJacksonDS: No dropdown in display mode as it is read only.
"The "options" on the ticket is a bit misleading because you don't need to display them at this stage"
They'll be used for the Edit Details page.

DigitalLearningSolutions.Web/ViewModels/MyAccount/MyAccountViewModel.cs
```diff
@@ -1,19 +1,39 @@
...
             Centre = adminUser?.CentreName ?? delegateUser?.CentreName;
             DelegateNumber = delegateUser?.CandidateNumber;
+            JobGroup = delegateUser?.JobGroupName;
+
+            CustomField1 = PopulateCustomFieldViewModel(customPrompts?.CustomField1, delegateUser?.Answer1);
+            CustomField2 = PopulateCustomFieldViewModel(customPrompts?.CustomField2, delegateUser?.Answer2);
+            CustomField3 = PopulateCustomFieldViewModel(customPrompts?.CustomField3, delegateUser?.Answer3);
+            CustomField4 = PopulateCustomFieldViewModel(customPrompts?.CustomField4, delegateUser?.Answer4);
+            CustomField5 = PopulateCustomFieldViewModel(customPrompts?.CustomField5, delegateUser?.Answer5);
+            CustomField6 = PopulateCustomFieldViewModel(customPrompts?.CustomField6, delegateUser?.Answer6);
```
DanBloxham-sw, line 26: From what I can tell, this is the first point where the answers and prompts are linked together in a single object. Would it be worth handling these multiple CustomFields from this point onwards in a list or an array or something?
I think before this the field numbers have to be preserved as they come from different tables and the ordering matters, but a from here we might be able to cut down on nearly-identical code with looping. Or  would we prefer to keep it to a specific number order to keep it inline with the database?

DigitalLearningSolutions.Data/Services/UserDataService.cs
DanBloxham-sw: I want to say that it is fine to grab the name as part of this query, as it is only grabbing the name from an Id that is already stored on the record. If we wanted to grab more details from JobGroups, then we should almost certainly call the other service. It does mean that the GetJobGroupName method is slightly less useful than it could be, but I'm leaning toward having a more redundant method that is only used in specific circumstances and having less calls to the services/DB for something this simple. Though @stellake  might have other ideas and I am happy to be overruled.

DigitalLearningSolutions.Data.Tests/Services/UserDataServiceTests.cs
```diff
@@ -60,7 +60,7 @@ public void GetAdminUserById_Returns_admin_user()
         public void GetDelegateUserById_Returns_delegate_users()
         {
             // Given
-            var expectedDelegateUsers = UserTestHelper.GetDefaultDelegateUser();
+            var expectedDelegateUsers = UserTestHelper.GetDefaultDelegateUser(jobGroupName: "Nursing / midwifery");
```
DanBloxham-sw, line 63: If the default delegate user should have a job group name of "Nursing / midwifery" I think it should be put in the GetDefaultDelegateUser default options. 

I suspect that might mess up other tests - e.g. some data service tests will use queries that don't recover all the columns (as they won't get used further down the line), and then compare if the default model is equivalent to the recovered data (which will have nulls for those missed columns). So we might want a decision for how to move forward. I see three options:

1. Change all the queries that retrieve data to always retrieve ALL the columns, and change the GetDefault...() methods to match all the columns on the DB record they are for as the defaults.
2. Don't change the queries that retrieve data, but change the GetDefault...() as in number 1. This would mean each test would have to specify the null columns to overwrite the defaults in GetDefault...() so there would be a lot of going back and forth "correcting" tests when new columns are added.
3. Keep it as it is in this file, so queries only retrieve what is necessary, but then we have to specify in the tests what the values the expected one should have.

My preference is for option 1, though it may require a large amount of overhauling depending on how many data services don't get the full set of columns for their tables. It would probably entail us doing what we are doing for the data services split - let everyone know, and then when they work on a service they can change the rest of it to match, so we only change one service at a time. Thoughts?

DigitalLearningSolutions.Data.Tests/Services/CustomPromptsDataServiceTests.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        [SetUp]
+        public void Setup()
+        {
+            var connection = ServiceTestHelper.GetDatabaseConnection();
+            customPromptsDataService = new CustomPromptsDataService(connection);
+        }
+
+        [Test]
+        public void GetCentreCustomPromptsByCentreId()
```
DanBloxham-sw, line 20: Is it worth naming this test to be more "sentence-like" rather than just the name of the method (even though it is pretty basic)?

AlexJacksonDS: yes, forgot to rename it

DigitalLearningSolutions.Data.Tests/Services/UserDataServiceTests.cs
AlexJacksonDS: You are exactly right, it was because of a failing unit test I did it this way, one query not having all the properties that the other did. I do think all relevant queries should probably be populating all the properties on the. e.g. all AdminUser queries should return fully populated AdminUsers. Does bring us back around to the JobGroupService question then. Should we populate the ID and look it up, or add the Join to every query?

DanBloxham-sw: I think if we are deciding to make all queries populate all properties then we should use JobGroupService for the acquiring of the name (and thus, keeping with the same principal, do the same for the CentreName). 

Though I am not sure if we should then be adding CentreName/JobGroupName to the data models or keep them separate. If we don't add them, then the data model and data service will be clear matches to the database, but we will likely have to refactor other services/controllers to obtain the names properly. If we add them to the model (but have default values as null or empty), then we can easily refactor the service methods to make the calls to get the additional properties and leave most of the code the same, but then we don't have a model = database pattern.

DigitalLearningSolutions.Web/Views/MyAccount/Index.cshtml
```diff
@@ -92,6 +92,49 @@
...
+                Job group:
+              </dt>
+              <dd class="nhsuk-summary-list__value">
+                @Model.JobGroup
+              </dd>
+            </div>
+
+            @if (Model.CustomField1 != null)
+            {
+              <partial name="MyDetailsCustomField" model="@Model.CustomField1" />
```
stellake, line 108: I'm wondering if it would be nicer to just pass in an array of CustomFields/CustomPrompts and use foreach here rather than hardcoding CustomField1, CustomField2 etc (we can filter out the null values)?

DigitalLearningSolutions.Web/ViewModels/MyAccount/MyAccountViewModel.cs
stellake: Ah, just saw this comment - I think my mind went the same direction below. :)

DigitalLearningSolutions.Data/Services/CustomPromptsService.cs
```diff
@@ -0,0 +1,45 @@
...
+                CustomField1 = PopulateCustomPrompt(result.CustomField1PromptId, result.CustomField1Prompt, result.CustomField1Options, result.CustomField1Mandatory),
+                CustomField2 = PopulateCustomPrompt(result.CustomField2PromptId, result.CustomField2Prompt, result.CustomField2Options, result.CustomField2Mandatory),
+                CustomField3 = PopulateCustomPrompt(result.CustomField3PromptId, result.CustomField3Prompt, result.CustomField3Options, result.CustomField3Mandatory),
+                CustomField4 = PopulateCustomPrompt(result.CustomField4PromptId, result.CustomField4Prompt, result.CustomField4Options, result.CustomField4Mandatory),
+                CustomField5 = PopulateCustomPrompt(result.CustomField5PromptId, result.CustomField5Prompt, result.CustomField5Options, result.CustomField5Mandatory),
+                CustomField6 = PopulateCustomPrompt(result.CustomField6PromptId, result.CustomField6Prompt, result.CustomField6Options, result.CustomField6Mandatory)
+            };
+        }
+
+        private CustomPrompt? PopulateCustomPrompt(int promptId, string prompt, string? options, bool mandatory)
```
stellake, line 40: I think in theory the "prompt" can be null in the db (but there are currently no null values). I think this is a constraint issue that can hopefully be resolved in the future when we get to tidy up the db, but until then, could we just pass back null if the "prompt" is null (so just do an check of id != 0 && prompt != null).

DigitalLearningSolutions.Data/Models/CustomPrompts/CentreCustomPromptsResult.cs
```diff
@@ -0,0 +1,37 @@
+﻿namespace DigitalLearningSolutions.Data.Models.CustomPrompts
+{
+    public class CentreCustomPromptsResult
+    {
+        public int CentreId { get; set; }
+
+        public int CustomField1PromptId { get; set; }
```
stellake, line 7: Such a shame this is not nullable! 😢

DigitalLearningSolutions.Data/Services/UserDataService.cs
stellake: I think it makes sense to join here. 👍

DigitalLearningSolutions.Web.Tests/ViewModels/MyAccount/MyAccountViewModelTests.cs
```diff
@@ -27,6 +28,11 @@ public void MyAccountViewModel_AdminUser_and_DelegateUser_populates_expected_val
                 returnedModel.ProfilePicture.Should().BeEquivalentTo(adminUser.ProfileImage);
                 returnedModel.DelegateNumber.Should().BeEquivalentTo(delegateUser.CandidateNumber);
                 returnedModel.User.Should().BeEquivalentTo(adminUser.EmailAddress);
+                returnedModel.JobGroup.Should().BeEquivalentTo(delegateUser.JobGroupName);
```
stellake, line 31: I'm wondering if it's worth separating the customPrompts population into a separate test to make it easier to debug if anyone breaks it in the future?

DigitalLearningSolutions.Data.Tests/Helpers/CustomPromptsTestHelper.cs
```diff
@@ -0,0 +1,100 @@
+﻿namespace DigitalLearningSolutions.Data.Tests.Helpers
+{
+    using DigitalLearningSolutions.Data.Models.CustomPrompts;
+
+    public static class CustomPromptsTestHelper
+    {
+        public static CentreCustomPrompts GetDefaultCentreCustomPrompts
+        (
+            int centreId = 2,
```
stellake, line 9: Could we use centre 29 for the custom prompts data service tests as it has two custom prompts (one with options, one without) - this means that we can actually test some of these joins working correctly.

DigitalLearningSolutions.Data.Tests/Services/CustomPromptsDataServiceTests.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            // Given
+            var expectedCentreCustomPromptsResult = CustomPromptsTestHelper.GetDefaultCentreCustomPromptsResult();
+
+            //When
```
stellake, line 25: NIP: Super minor, but can we have a space before "When" like we do for other comments?

DigitalLearningSolutions.Data.Tests/Services/CustomPromptsServiceTests.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        [Test]
+        public void GetCustomPromptsForCentreByCentreId_Returns_Populated_CentreCustomPrompts()
+        {
+            // Given
+            var options = "options";
+            var expectedPrompt1 = CustomPromptsTestHelper.GetDefaultCustomPrompt(options: options);
+            var expectedCustomerPrompts = CustomPromptsTestHelper.GetDefaultCentreCustomPrompts(customPrompt1: expectedPrompt1);
+            A.CallTo(() => customPromptsDataService.GetCentreCustomPromptsByCentreId(A<int>._))
```
stellake, line 29: Could we specify the argument on the mock as the centre id below, just to make the mock more accurate? :) 

`A.CallTo(() => customPromptsDataService.GetCentreCustomPromptsByCentreId(2))...`

DigitalLearningSolutions.Data.Tests/Services/UserDataServiceTests.cs
stellake: I think the "getAccountByUsername" only cares about the fields that are required for logging in whereas "byId" cares about all the information. Hence, I think joining is still a good idea and we shouldn't change the code because of this single test (unless we truly believe that we'll need these extra fields later down the line). 
In general, I think it's ok to override this field in the "default user", because the default user is meant to be the simplest user and is there to avoid some code duplication (kind of like a builder).

DigitalLearningSolutions.Web/ViewModels/MyAccount/MyAccountViewModel.cs
AlexJacksonDS: I've had a look in the database, and we have a whole load of centres where the custom fields IDs are not populated in order, e.g. the IDs are 0,1,5 for prompts 1,2,3 respectively. This would give us null for Prompt1 currently, I'm not sure whether it would be better to keep it as the separate fields or add some kind of property to the view model to signify which field it is. I'm not sure which will be better when we are going in the other direction for Edit Details. Thoughts?

stellake: I think adding a property to the view model (or even a CustomPrompt that could be used here directly) to signify which field it is would be useful - it would work well with both display & edit pages and will keep the views simple too.

A bit separate point, but now that I think about it, do we even need the customPromptId as when it's 0, the custom field name should be null on join?

AlexJacksonDS: OK will refactor it that way instead, might end up needing to modify it a bit later for the Edit screen, but will deal with that when I get there

AlexJacksonDS: And yes, we shouldn't need the CustomPromptIds

DigitalLearningSolutions.Data/Models/CustomPrompts/CentreCustomPromptsResult.cs
```diff
@@ -0,0 +1,37 @@
...
+        public string? CustomField1Prompt { get; set; }
+        public string? CustomField1Options { get; set; }
+        public bool CustomField1Mandatory { get; set; }
+
+        public int CustomField2PromptId { get; set; }
```
stellake, line 12: Do we need these ids in the db level or is prompt name sufficient (as we only use the id for join)?

AlexJacksonDS: I'm using it to determine if it is a prompt or not in the CustomPromptService, if promptId != 0 && prompt != null then it is a prompt.

stellake: Would prompt != null be sufficient?

AlexJacksonDS: Having looked at the database, it probably should be. Will refactor it to remove the IDs

stellake (approved):
Looks good!

PR #280, opened 2021-04-22
Automated UI tests 

stellake:
Quick automated UI tests with Selenium + adding them to pipeline. Creating a MR now so people can start using these.

Left to do:
* Make error messages nicer and better readable. ==> Fixed.

Future work:
* Log in the user and test all the pages that user needs to be logged in for.

### DIFF
--- a/DigitalLearningSolutions.Data/Mappers/MapperHelper.cs
+++ b/DigitalLearningSolutions.Data/Mappers/MapperHelper.cs
@@ -6,6 +6,7 @@
     {
         public static void SetUpFluentMapper()
         {
+            FluentMapper.EntityMaps.Clear();
             FluentMapper.Initialize(fluentMapperConfig =>
             {
                 fluentMapperConfig.AddMap(new CurrentCourseMap());
--- a/DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests.cs
+++ b/DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests.cs
@@ -1,6 +1,7 @@
 namespace DigitalLearningSolutions.Web.AutomatedUiTests
 {
     using System;
+    using FluentAssertions;
     using OpenQA.Selenium;
     using OpenQA.Selenium.Chrome;
     using Selenium.Axe;
@@ -36,7 +37,7 @@ namespace DigitalLearningSolutions.Web.AutomatedUiTests
             var axeResult = new AxeBuilder(driver).Analyze();
 
             // then
-            Assert.Empty(axeResult.Violations);
+            axeResult.Violations.Should().BeEmpty();
         }
 
         private ChromeDriver CreateHeadlessChromeDriver()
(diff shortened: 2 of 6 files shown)


DigitalLearningSolutions.Data/Mappers/MapperHelper.cs
```diff
@@ -6,6 +6,7 @@ public static class MapperHelper
     {
         public static void SetUpFluentMapper()
         {
+            FluentMapper.EntityMaps.Clear();
```
stellake, line 9: Without this, we try and set the entity maps twice with a failure (because we need to start a fake test server). I can't think of any ways this would cause issues to the existing code, but it would be neater if we didn't have to do it. I might look into it later unless anyone has immediate ideas.

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        [Theory]
+        [InlineData("/Home")]
+        [InlineData("/Login")]
+        [InlineData("/ForgotPassword")]
+        [InlineData("/Register")]
+        public void Page_has_no_accessibility_errors(string url)
+        {
+            // when
+            driver.Navigate().GoToUrl(baseUrl + url);
```
DanBloxham-sw, line 36: As this goes directly to a url, does this mean it won't test pages that can't be reached through regular GET requests? e.g. the Choose A Centre page when logging in?

kevwhitt-hee:
Happy with this. @DanBloxham-sw makes a good point. Not sure if this scenario could also be catered for in future work or whether the effort to do so would stop it being useful? If so, we just need to remember to test those types of page manually, I guess.

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests.cs
DanBloxham-sw: It would be nice if the controller actions could be called instead of just heading to a url, but then I am very unfamiliar with this and have no idea if that's possible. Either way, I imagine it won't matter for the vast majority of pages, so it will just be a case of not relying on the tests when making those specific pages, which we've done up to this point anyway. So I'm happy with this.

stellake: The Selenium tests control the browser - they effectively open Chrome locally and interact with it. Just like you cannot select "POST" in normal Chrome browser when you enter a URL, you won't be able to do it when navigating in these Selenium tests (and it would defeat the point as we're trying to mimic user's experience). However (!), how the tests do work is that they allow you to interact with the browser, so you can fill in forms and press buttons (e.g. log in), which will allow us to test these pages. Given that the accessibility scanner is a bit slow, I would prefer to keep these tests limited (rather than accessibility testing every version of the page), but we can definitely discuss it if there are pages that would strongly benefit from several scans. 😄

stellake:
Regarding manual testing, I think these tests will never fully replace manual testing, but will hopefully help us identify some bugs early and also minimise the risk of some issues slipping through the testing phase. Will definitely create a second MR soon for the "logged in" pages though. 👍

DanBloxham-sw (approved):

PR #281, opened 2021-04-22
Heedls 338 landing page tweaks

SteveJacksonSoft:
@stellake Think I've fixed everything left from the unresolved threads in the old PR. I haven't resolved the threads, so if you want to go through and check you can. 😁
It's here: https://github.com/TechnologyEnhancedLearning/DLSV2/pull/245

This contains the change to the welcome page HEE wanted, and some CSS changes for mobile display of LearningContent section.


SteveJacksonSoft:
One thing I'd mention is that the mobile display of the welcome page looks slightly off to me - just the size of the of a gap below the auth buttons compared with the gap above them.
image

DigitalLearningSolutions.Web/Views/Home/Welcome.cshtml
```diff
@@ -9,9 +9,7 @@
 
 <div class="nhsuk-grid-row">
   <div class="nhsuk-grid-column-full">
-    <h1 class="nhsuk-heading-xl">A platform for digital learning delivery</h1>
-
-    <p class="nhsuk-u-font-size-32">Enabling health and care organisations to deliver digital learning to staff and students.</p>
+    <partial name="MiniHub/_MiniHubHeading" model="@Model.MiniHubNavigationModel"/>
```
stellake, line 12: 👍

stellake:
> One thing I'd mention is that the mobile display of the welcome page looks slightly off to me - just the size of the gap below the auth buttons compared with the gap above them. It might be fine though. What do you think?
> image

I agree with this - we could either increase the spacing below "Register" a tiny bit and reduce it around the "Log in" button, or put the buttons next to each other (if they fit on 320px screen) to amend this. :)

SteveJacksonSoft:
> 
> 
> > One thing I'd mention is that the mobile display of the welcome page looks slightly off to me - just the size of the gap below the auth buttons compared with the gap above them. It might be fine though. What do you think?
> > image
> 
> I agree with this - we could either increase the spacing below "Register" a tiny bit and reduce it around the "Log in" button, or put the buttons next to each other (if they fit on 320px screen) to amend this. :)

image
image

How about these? The narrower screenshot is 320px wide - don't think the buttons would fit side-by-side.

DigitalLearningSolutions.Web/Styles/home/welcome.scss
```diff
@@ -1,10 +1,32 @@
 @import "~nhsuk-frontend/packages/core/all";
 
+.welcome-page-button-container {
+  @include govuk-media-query($from: tablet) {
+    margin-bottom: 32px;
+  }
+
+  @include govuk-media-query($until: tablet) {
```
stellake, line 8: Do we need both media queries? I think the standard is to write a default style first and then if necessary override that style with a media query (rather than have media queries for until tablet and after tablet). e.g.
```
.welcome-page-button-container {
  margin-bottom: 26px;

  @include govuk-media-query($from: tablet) {
    margin-bottom: 32px;
  }
}
```

DigitalLearningSolutions.Web/Styles/home/welcome.scss
```diff
@@ -1,10 +1,32 @@
...
+  }
 }
 
 .auth-button--left {
-  margin-right: 32px;
+  @include govuk-media-query($from: tablet) {
+    margin-right: 32px;
+  }
+
+  @include govuk-media-query($until: tablet) {
```
stellake, line 28: ditto about the double queries

DigitalLearningSolutions.Web/Styles/home/welcome.scss
SteveJacksonSoft: Yeah, I don't know the benefit of doing that, but happy to make that change.

JBsoftwire: There was... some reason I did it this way but I can't remember what, so if it works fine without it that's probably better.

stellake (approved):
Looks good!

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests.cs
```diff
@@ -1,50 +1,52 @@
...
+        public void Dispose()
+        {
+            driver.Quit();
+            driver.Dispose();
+        }
+
+        [Theory]
+        [InlineData("/Home/Welcome")]
+        [InlineData("/Home/Products")]
+        [InlineData("/Home/LearningContent")]
```
SteveJacksonSoft, line 31: These are the lines I changed in this file. I've just noticed I've got AutoCRLF turned off. Do you want me to turn it on?

DigitalLearningSolutions.Web/Views/Shared/MiniHub/_MiniHubHeading.cshtml
```diff
@@ -2,7 +2,7 @@
 @model MiniHubNavigationModel
 
 <h1 class="nhsuk-heading-xl">
-  <span role="text">
```
SteveJacksonSoft, line 5: This isn't part of the official ARIA spec and is only supported on certain browsers (that broke the accessibility test). I tried without this line and it didn't change the screenreader behaviour, so I reckon best to remove it.

README.md
```diff
@@ -137,7 +137,7 @@ Open the test file, find the test you want to run, click the icon to the left of
 Open the file and click the icon to the left of the class name.
 
 ## Run all the tests
-Open the solution explorer. Right click the test project you want (DigitalLearningSolutions.Web.Tests or DigitalLearningSolutions.Data.Tests) and select "Run tests".
+Open the solution explorer. Right click the test project you want (DigitalLearningSolutions.Web.Tests, DigitalLearningSolutions.Data.Tests, etc.) and select "Run tests".
```
stellake, line 140: Nice! 🎉

PR #284, opened 2021-04-23
HEEDLS-366 - Log In now records admin session and session services re…

DanBloxham-sw:
…named to specify user type

Tested by:
- running all unit tests 
- manually logging in as an admin
- manually logging in as delegate to make sure no admin session is made.


DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -72,6 +74,11 @@ public IActionResult Index(LoginViewModel model)
 
         private void LogIn(AdminUser? adminUser, DelegateUser? delegateUser, bool rememberMe)
```
stellake, line 75: Could you change this method name to "LogInAndRecordAdminSession" to make sure it's not misused?

stellake: Or alternatively have a recordAdminSession function before "LogIn" is called 🤔 😄

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
```diff
@@ -163,5 +166,44 @@ public void Log_in_with_approved_delegate_id_fetches_associated_admin_user()
...
+            // Given
+            controller = LoginTestHelper.GetLoginControllerWithSignInFunctionality(loginService, userService, sessionService);
+            var expectedDelegates = new List<DelegateUser> { UserTestHelper.GetDefaultDelegateUser(approved: true) };
+            A.CallTo(() => userService.GetUsersByUsername(A<string>._))
+                .Returns((null, expectedDelegates));
+            A.CallTo(() => loginService.VerifyUsers(A<string>._, A<AdminUser>._, A<List<DelegateUser>>._))
+                .Returns((null, expectedDelegates));
+            A.CallTo(() => loginService.GetVerifiedAdminUserAssociatedWithDelegateUser(A<DelegateUser>._, A<string>._))
+                .Returns(null);
+            // When
```
stellake, line 201: Super minor, but could we add an empty line before "// When" ?

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
```diff
@@ -163,5 +166,44 @@ public void Log_in_with_approved_delegate_id_fetches_associated_admin_user()
             A.CallTo(() => loginService.GetVerifiedAdminUserAssociatedWithDelegateUser(testDelegate, A<string>._))
                 .MustHaveHappened();
         }
+
+        [Test]
+        public void Log_in_as_admin_records_admin_session()
```
stellake, line 171: Nice tests 👍

DigitalLearningSolutions.Data/Services/SessionDataService.cs
```diff
@@ -35,18 +37,28 @@ public int StartOrRestartSession(int candidateId, int customisationId)
...
                    WHERE [SessionID] = @sessionId AND Active = 1;",
                 new { sessionId }
             );
         }
+
+        public int StartAdminSession(int adminId)
+        {
+            return connection.QueryFirst<int>(
+                @"INSERT INTO AdminSessions (AdminID, LoginTime, Duration, Active)
+                  VALUES (@adminId, GetUTCDate(), 0, 0);
```
stellake, line 58: nice UTC date :)

DigitalLearningSolutions.Data.Tests/Services/SessionDataServiceTests.cs
```diff
@@ -139,13 +139,28 @@ public void UpdateSessionDuration_Should_Not_Update_Inactive_Session()
...
         }
+
+        [Test]
+        public void StartAdminSession_Should_Start_Users_New_Admin_Session()
+        {
+            using (new TransactionScope())
+            {
+                // When
+                var adminUser = UserTestHelper.GetDefaultAdminUser();
+                var sessionId = sessionDataService.StartAdminSession(adminUser.Id);
```
stellake, line 158: Do we need the "default admin user" rather than using an id of some sort directly?

DigitalLearningSolutions.Data.Tests/Services/SessionDataServiceTests.cs
```diff
@@ -139,13 +139,28 @@ public void UpdateSessionDuration_Should_Not_Update_Inactive_Session()
...
+
+                // Then
+
+                sessionId.Should().NotBe(0);
```
stellake, line 162: I think it might be worth doing something similar to the candidate tests where we fetch the session from the db (using the existing test helper) and check that admin id is associated with that particular session id. 
Also, super minor, but we can probably get rid of L161. 👀

stellake (commented):
A few minor comments, but looking good in general! :)

DigitalLearningSolutions.Data.Tests/Services/SessionDataServiceTests.cs
DanBloxham-sw: The AdminSessions table has the AdminID as a foreign key, so will only work if the ID used can be found in the database. My thinking here was that we already have an admin that we know is going to be in the database, so I might as well grab that and use it, rather than specifying an ID manually. Manually setting the ID might cause the test to break if the default user is changed about in the future on the database. Is there any harm to how the test is set up currently?

stellake: In every test we write, we should assume that all the data in the database backup is there, so any existing admin id will do. I feel like calling the adminUser builder here gives this test an extra layer of complication that can be avoided.

DigitalLearningSolutions.Data.Tests/DataServices/SessionDataServiceTests.cs
```diff
@@ -1,151 +1,167 @@
...
-
-                // Then
-                var updatedSessions = sessionTestHelper.GetCandidateSessions(candidateId);
-
-                updatedSessions.Should().BeEquivalentTo(startingSessions);
-            }
-        }
-    }
-}
+﻿namespace DigitalLearningSolutions.Data.Tests.DataServices
```
stellake, line 1: Yeeess! 🎉

DigitalLearningSolutions.Data.Tests/Services/SessionServiceTests.cs
```diff
@@ -9,131 +10,146 @@
...
 
             // Then
             httpContextSession.Keys.Should().BeEmpty();
         }
+
+        [Test]
+        public void StartAdminSession_should_call_data_service()
+        {
+            // Given
+            A.CallTo(() => sessionDataService.StartAdminSession(A<int>._)).Returns(1);
```
stellake, line 146: Could we make this test slightly more accurate by only returning "1" if the "sessionDataService.StartAdminSession" is called with the right param:
`A.CallTo(() => sessionDataService.StartAdminSession(1)).Returns(1);`

stellake: Oh, ignore this, we check it below, so it actually doesn't matter 😄

DigitalLearningSolutions.Data.Tests/DataServices/SessionDataServiceTests.cs
```diff
@@ -1,151 +1,167 @@
...
+            {
+                // When
+                var adminUser = UserTestHelper.GetDefaultAdminUser();
+                var sessionId = sessionDataService.StartAdminSession(adminUser.Id);
+
+                // Then
+                var newAdminSession = sessionTestHelper.GetAdminSession(sessionId);
+                newAdminSession.Should().NotBe(null);
+                newAdminSession.AdminId.Should().Be(adminUser.Id);
+                newAdminSession.LoginTime.Should().BeCloseTo(DateTime.UtcNow, twoMinutesInMilliseconds);
```
stellake, line 163: Really minor, but I think it should be (just the location of the helping comments):
// Given session id is X
// When GetAdminSession is called with that id
// Then newAdminSession is ... (assertions)

stellake (approved):
Looks good!

PR #286, opened 2021-04-23
Heedls 406 fix forgot password email url

AlexJacksonDS:
Added an extra slash in the forgot password url if it isn't already there before appending the ResetPassword part of the path.

The port it being added because we are just using string interpolation on the UriBuilders rather than on UriBuilder.Uri. I've removed it from the password url by changing the string interpolation values.

I've checked elsewhere in the solution and there are also similar instances of this being done in the same way (UriBuilder in string interpolation) in the NotificationService and FrameworkNotificationService. I've not seen emails generated from those, but we may want to also update those to be using .Uri as well.

Checked that the url was correct in debug locally, i.e. this doesn't remove the port from localhost.
Ran all unit tests.

### DIFF
--- a/DigitalLearningSolutions.Data/Services/PasswordResetService.cs
+++ b/DigitalLearningSolutions.Data/Services/PasswordResetService.cs
@@ -95,20 +95,20 @@
             }
             resetPasswordUrl.Path += "ResetPassword";
             resetPasswordUrl.Query = $"code={resetHash}&email={emailAddress}";
-
+            
             string emailSubject = "Digital Learning Solutions Tracking System Password Reset";
 
             BodyBuilder body = new BodyBuilder
             {
                 TextBody = $@"Dear {firstName},
                             A request has been made to reset the password for your Digital Learning Solutions account.
-                            To reset your password please follow this link: {resetPasswordUrl}
+                            To reset your password please follow this link: {resetPasswordUrl.Uri}
                             Note that this link can only be used once and it will expire in two hours.
                             Please don’t reply to this email as it has been automatically generated.",
                 HtmlBody = $@"<body style= 'font - family: Calibri; font - size: small;'>
                                     <p>Dear {firstName},</p>
                                     <p>A request has been made to reset the password for your Digital Learning Solutions account.</p>
-                                    <p>To reset your password please follow this link: <a href=""{resetPasswordUrl}"">{resetPasswordUrl}</a></p>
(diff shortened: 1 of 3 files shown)


DigitalLearningSolutions.Data/Services/PasswordResetService.cs
```diff
@@ -89,22 +89,26 @@ END CATCH
         private Email GeneratePasswordResetEmail(string emailAddress, string resetHash, string firstName, string baseUrl)
         {
             UriBuilder resetPasswordUrl = new UriBuilder(baseUrl);
+            if (!resetPasswordUrl.Path.EndsWith('/'))
```
stellake, line 92: Do we know why the path doesn't end with "/" sometimes, but does in other times (the emails work ok in staging)? Is the AppRootPath consistent across the environments?

AlexJacksonDS: It's because the UAT has some path already, i.e. at www.example.com/UAT rather than at test.example.com. When we create the UriBuilder with baseUrl with a url like this we get:

- Host: www.example.com
- Path: /UAT

rather than
- Host: test.example.com
- Path: /

DanBloxham-sw:
Regarding the NotificationService and FrameworkNotificationService - originally, there was just the NotificationService that created/sent emails, so I had copied that when creating the reset password email stuff. The FrameworkNotificationService was only added a few days ago and was probably also mostly copied from the NotificationService. Because they are all based originally around the NotificationService, and use exactly the same logic to generate those links, I think it is safe and probably for the best to change those as well (thought I also haven't tested the NotificationService and FrameworkNotificationService emails). Might be worth making a common method for these so we avoid this cycle of copying the old code?

DigitalLearningSolutions.Data/Services/PasswordResetService.cs
stellake: Ah, that makes sense! 👍 😅

DigitalLearningSolutions.Data/Services/PasswordResetService.cs
```diff
@@ -89,22 +89,26 @@ END CATCH
         private Email GeneratePasswordResetEmail(string emailAddress, string resetHash, string firstName, string baseUrl)
```
stellake, line 89: Can't reply to Dan's comment, but I agree that it makes sense to change it everywhere (unless it's too much work, in which case I can flag the Frameworks changes with Kevin)

stellake (approved):
Looks good! Before merging, could you add a list of emails where the URL should be tested to the JIRA ticket so we make sure nothing's broken? 🙏

PR #287, opened 2021-04-26
DLSV2-174 Fixes email service content encoding

kevwhitt-hee:
And fix to launchi9ng of certificates in legacy platform relating to DLSV2-173.


stellake (approved):
Nice! I would have personally chosen to separate creating the MultipartAlternative into a separate method and maybe add a comment about the ".toMessageBody" method of the BodyBuilder not working properly with NHS email addresses to avoid someone switching back to it in the future.

kevwhitt-hee:
I agree. Does this look better?

stellake:
Yep - looks great! :)

PR #288, opened 2021-04-26
HEEDLS-403 Basic edit details frontend changes

AlexJacksonDS:
WIP, will add further details/commits to this. Just seemed the easiest way to show the proposed changes


DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -32,5 +31,33 @@ public IActionResult Index()
 
             return View(model);
         }
+
+        [HttpGet]
+        public IActionResult EditDetails()
+        {
+            if (!User.Identity.IsAuthenticated)
```
stellake, line 38: Not sure how I missed it in the My account review, but we should be able to just use [Authorize] annotation instead of this. :)

DigitalLearningSolutions.Web/Views/MyAccount/EditDetails.cshtml
```diff
@@ -0,0 +1,60 @@
...
+
+  <div class="nhsuk-grid-row">
+    <div class="nhsuk-grid-column-full">
+      <vc:text-input asp-for="Email" label="Email Address" type="text" spell-check="false" autocomplete="email"></vc:text-input>
+    </div>
+  </div>
+
+  <div class="nhsuk-grid-row">
+    <div class="nhsuk-grid-column-full">
+      <vc:text-input asp-for="Password" label="Confirm password to save changes:" type="password" spell-check="false" autocomplete="email"></vc:text-input>
```
stellake, line 49: Had a really quick scan now, but think that this is definitely worth the refactor - these components are really easy to read as part of the HTML and will avoid a lot of the duplication. Will have a look at the details tomorrow 👍

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
AlexJacksonDS: OK will update all the existing ones while I'm doing this ticket

AlexJacksonDS:
Added the basic set of fields with a confirmation password field for the Edit details screen.

Checked that the layout works OK in desktop and mobile and ran unit tests

image

image

JBsoftwire (approved):
Looks good, just one question about some of the cshtml.

DigitalLearningSolutions.Web/Views/MyAccount/EditDetails.cshtml
```diff
@@ -0,0 +1,60 @@
...
+
+  <div class="nhsuk-grid-row">
+    <div class="nhsuk-grid-column-full">
+      <vc:text-input asp-for="LastName" label="Last Name" type="text" spell-check="false" autocomplete="family-name"></vc:text-input>
+    </div>
+  </div>
+
+  <div class="nhsuk-grid-row">
+    <div class="nhsuk-grid-column-full">
+      <vc:text-input asp-for="Email" label="Email Address" type="text" spell-check="false" autocomplete="email"></vc:text-input>
```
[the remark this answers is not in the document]
AlexJacksonDS: Yeah looks like you can, I've just tried it on them. The empty tag style is just the default that visual studio put.

DigitalLearningSolutions.Web/Views/Shared/Components/TextInput/Default.cshtml
```diff
@@ -0,0 +1,21 @@
...
+<div class="nhsuk-form-group @(Model.HasError?"nhsuk-form-group--error":"")">
+  <label class="nhsuk-label" for="@Model.Name">
+    @Model.Label
+  </label>
+  @if (Model.HasError) {
+    <span class="nhsuk-error-message" id="@Model.Name-error">
+      <span class="nhsuk-u-visually-hidden">Error:</span> @Model.ErrorMessage
+    </span>
+  }
+  <input class="nhsuk-input nhsuk-u-width-one-half @(Model.HasError?"nhsuk-input--error":"")"
```
stellake, line 13: Super minor, but having some spacing around the ? and : in ternary makes it a bit easier to read. Maybe worth running a formatter on the file?

DigitalLearningSolutions.Web/ViewModels/MyAccount/EditDetailsViewModel.cs
```diff
@@ -0,0 +1,33 @@
...
+        [Required(ErrorMessage = "Enter a last name.")]
+        public string? LastName { get; set; }
+
+        [Required(ErrorMessage = "Enter an email address.")]
+        [EmailAddress(ErrorMessage = "Enter a valid email address.")]
+        public string? Email { get; set; }
+
+        [Required(ErrorMessage = "Enter your password.")]
+        [DataType(DataType.Password)]
+        public string? Password { get; set; }
```
stellake, line 31: Could we add a maxLength of 100 constraint to all of these fields? 🙏  

[MaxLength(100, ErrorMessage = "First name must be at most 100 characters")]
[MaxLength(100, ErrorMessage = "Last name must be at most 100 characters")]
[MaxLength(100, ErrorMessage = "Email address must be at most 100 characters")]

Less concerned about PW as we're checking against existing hash, but feel free to add that:
[MaxLength(100, ErrorMessage = "Password must be at most 100 characters")]

DigitalLearningSolutions.Web/ViewModels/MyAccount/EditDetailsViewModel.cs
```diff
@@ -0,0 +1,33 @@
+﻿namespace DigitalLearningSolutions.Web.ViewModels.MyAccount
+{
+    using System.ComponentModel.DataAnnotations;
+    using DigitalLearningSolutions.Data.Models.User;
+
+    public class EditDetailsViewModel
+    {
+        public EditDetailsViewModel() { }
```
stellake, line 8: Do we need/use the empty constructor?

AlexJacksonDS: Yes, it's needed for some model binding thing. Without it we get:
System.InvalidOperationException: Could not create an instance of type 'DigitalLearningSolutions.Web.ViewModels.MyAccount.EditDetailsViewModel'. Model bound complex types must not be abstract or value types and must have a parameterless constructor. Alternatively, give the 'model' parameter a non-null default value.

DigitalLearningSolutions.Web/ViewComponents/ErrorSummaryViewComponent.cs
```diff
@@ -0,0 +1,28 @@
...
+            var orderedErrors = GetOrderedPropertyNames(errors, orderOfPropertyNames ?? new string[0]);
+
+            var errorSummaryViewModel = new ErrorSummaryViewModel(orderedErrors);
+            return View(errorSummaryViewModel);
+        }
+
+        private static List<ErrorSummaryListItem> GetOrderedPropertyNames(List<ErrorSummaryListItem> errors, string[] orderOfPropertyNames)
+        {
+            var orderedErrors = errors.OrderBy(e => orderOfPropertyNames.ToList().IndexOf(e.Key)).ToList();
+            return orderedErrors;
```
stellake, line 25: Two minor things, 
1. Can we return immediately rather than define var and immediately return it?
2. Should the method be named GetOrderedErrors?

DigitalLearningSolutions.Web/Startup.cs
```diff
@@ -139,13 +139,13 @@ public void Configure(IApplicationBuilder app, IMigrationRunner migrationRunner)
...
         {
-            context.HttpContext.Response.Redirect( $"{config["CurrentSystemBaseUrl"]}/home?action=login&app=lp");
+            context.HttpContext.Response.Redirect( "/Login");
             return Task.CompletedTask;
         }
 
         private Task RedirectToHome(RedirectContext<CookieAuthenticationOptions> context)
         {
-            context.HttpContext.Response.Redirect($"{config["CurrentSystemBaseUrl"]}/home");
+            context.HttpContext.Response.Redirect("/Home");
```
stellake, line 148: Nice! 👍

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -27,11 +43,44 @@ public IActionResult Index()
...
+        [HttpPost]
+        public IActionResult EditDetails(EditDetailsViewModel model)
+        {
+            if (!ModelState.IsValid)
+            {
+                return View(model);
+            }
+
+            var userAdminId = User.GetCustomClaim(CustomClaimTypes.UserAdminId);
+            var userDelegateId = User.GetCustomClaim(CustomClaimTypes.LearnCandidateId);
```
stellake, line 60: I think you should be able to use User.GetCandidateId() here (as per CustomClaims extension). I think Jon B added another extension method for the admin equivalent, which should be on the latest master (User.GetAdminId).

DigitalLearningSolutions.Data.Tests/Services/UserServiceTests.cs
```diff
@@ -137,5 +139,89 @@ public void GetAvailableCentres_returns_centres_correctly_ordered()
...
+                .Returns((adminUser, new List<DelegateUser>()));
+            A.CallTo(() => userDataService.UpdateAdminUser(adminUser)).DoesNothing();
+
+            // When
+            var result = userService.TryUpdateUsers(adminUser, null, password);
+
+            // Then
+            result.Should().BeTrue();
+            A.CallTo(() => userDataService.UpdateAdminUser(adminUser))
+                .MustHaveHappened();
```
stellake, line 159: The name of the test says "only updates admin", but there's no check for whether the delegate user is updated or not. Would it be worth adding a check that UpdateDelegateUser is never called?

DigitalLearningSolutions.Data.Tests/Services/UserServiceTests.cs
```diff
@@ -137,5 +139,89 @@ public void GetAvailableCentres_returns_centres_correctly_ordered()
...
+            // Then
+            result.Should().BeTrue();
+            A.CallTo(() => userDataService.UpdateDelegateUser(delegateUser))
+                .MustHaveHappened();
+            A.CallTo(() => userDataService.UpdateAdminUser(adminUser))
+                .MustHaveHappened();
+        }
+
+        
+
```
stellake, line 205: I think one empty line between tests is sufficient for readability :)

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -27,11 +43,44 @@ public IActionResult Index()
...
+                adminUser.LastName = model.LastName;
+                adminUser.EmailAddress = model.Email;
+            }
+            
+            if (delegateUser != null)
+            {
+                delegateUser.FirstName = model.FirstName;
+                delegateUser.LastName = model.LastName;
+                delegateUser.EmailAddress = model.Email;
+            }
```
stellake, line 75: Hmm.. There are a lot of issues around the accounts because of the db structure etc. I'm thinking that actually we might want to update these details based on a current email address and password combination such that accounts in other centres get updated too. 

So in this case we'd get the user email from the database based on ids, then do something like TryUpdateUserAccountDetails(email, password, newUserAccountDetails)

This method would:
- Find all valid users for the email-password combination (this way account details get updated for a different centres too)
- Make the updates based on these user ids.

I'm also not sure about overriding the existing user model properties, because the UPDATE method in the db is meant to update a very specific subset of the AdminUser/Candidate details. Having the data service method accept the user object could make it seem like every field will be updated.. 🤔

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -17,21 +19,52 @@ public MyAccountController(ICustomPromptsService customPromptsService, IUserServ
...
+
+        [HttpPost]
+        public IActionResult EditDetails(EditDetailsViewModel model)
+        {
+            if (!ModelState.IsValid)
+            {
+                return View(model);
+            }
+
+            var signedInEmail = User.GetEmail();
```
stellake, line 59: Edge case, but what if the user edits their email twice in one session (as we don't update the claims when updating the email address - or updating the email address from a different device)?

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -73,5 +76,30 @@ public List<CentreUserDetails> GetUserCentres(AdminUser? adminUser, List<Delegat
 
             return availableCentres.OrderByDescending(ac => ac.IsAdmin).ThenBy(ac => ac.CentreName).ToList();
         }
+
+        public bool TryUpdateUserAccountDetails(string password, string signedInEmail, string firstName, string surname, string email)
+        {
+            var (adminUsers, delegateUsers) = GetUsersByEmailAddress(signedInEmail);
```
stellake, line 82: We should never have more than one adminUser per email address as per db constraints. If it's easy to change the implementation of this method, would be good to get it in (I don't think it's used in many places).

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -73,5 +76,30 @@ public List<CentreUserDetails> GetUserCentres(AdminUser? adminUser, List<Delegat
...
+            }
+
+            if (verifiedAdminUser != null)
+            {
+                userDataService.UpdateAdminUser(firstName, surname, email, verifiedAdminUser.Id);
+            }
+
+            foreach (var verifiedDelegateUser in verifiedDelegateUsers)
+            {
+                userDataService.UpdateDelegateUser(firstName, surname, email, verifiedDelegateUser.Id);
```
stellake, line 99: Would it be worth having a single query based on a list of ids? Then rollbacks will be handled by default too if any update fails.

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
AlexJacksonDS: I guess it's best to partially revert back to the previous flow then? Get Users by ID from claims IDs, get users by email from those users, update all these users?

AlexJacksonDS: The concurrency issue with multiple devices is a bit more of a difficult issue to resolve, all I think we can do is just save whatever is latest since I don't think we have anything in place to handle that kind of issue

DigitalLearningSolutions.Data/Services/UserService.cs
AlexJacksonDS: Agreed, I added the SingleOrDefault call in my usage of the return from this. Will refactor to put that in the Data Service method instead

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
stellake: The proposal of id -> email -> update sounds good and given that the user gets the latest email from the db, we shouldn't have issues with email being outdated

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -73,5 +76,34 @@ public List<CentreUserDetails> GetUserCentres(AdminUser? adminUser, List<Delegat
...
+        public bool TryUpdateUserAccountDetails(int? adminId, int? delegateId, string password, string firstName, string surname, string email)
+        {
+            var (loggedInAdminUser, loggedInDelegateUser) = GetUsersById(adminId, delegateId);
+
+            var signedInEmail = loggedInAdminUser?.EmailAddress ?? loggedInDelegateUser?.EmailAddress;
+
+            var (adminUser, delegateUsers) = GetUsersByEmailAddress(signedInEmail);
+
+            var (verifiedAdminUser, verifiedDelegateUsers) =
+                loginService.VerifyUsers(password, adminUser, delegateUsers);
```
stellake, line 89: Could we pull this into its own private method - get all linked accounts or something similar? Also, I think we should handle this by returning no users:

> Possible null reference argument for parameter 'emailAddress' in '(AdminUser?, List<DelegateUser>) UserService.GetUsersByEmailAddress(string emailAddress)'.

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -73,5 +76,34 @@ public List<CentreUserDetails> GetUserCentres(AdminUser? adminUser, List<Delegat
...
+            }
+
+            if (verifiedAdminUser != null)
+            {
+                userDataService.UpdateAdminUser(firstName, surname, email, verifiedAdminUser.Id);
+            }
+
+            if (verifiedDelegateUsers.Count != 0)
+            {
+                userDataService.UpdateDelegateUsers(firstName, surname, email, verifiedDelegateUsers.Select(d => d.Id).ToArray());
```
stellake, line 103: This looks good, but could we pull the mapping to ids into a variable above this line for better readability?

stellake (approved):
Looks good! Two minor comments about readability, but happy for this to be merged without rereview when these are addressed. 👍

Also, good job with all the additional fixes!

DigitalLearningSolutions.Data/Services/UserService.cs
AlexJacksonDS: so new method GetAllAccounts(int? adminId, int? delegateId, string password) returning the output of the loginService call?

stellake: Yep, the new private method here which is effectively encapsulating these lines:

```
var (loggedInAdminUser, loggedInDelegateUser) = GetUsersById(adminId, delegateId);

            var signedInEmail = loggedInAdminUser?.EmailAddress ?? loggedInDelegateUser?.EmailAddress; // or return no users

            var (adminUser, delegateUsers) = GetUsersByEmailAddress(signedInEmail);

            return loginService.VerifyUsers(password, adminUser, delegateUsers);
```

PR #289, opened 2021-04-27
HEEDLS-393 Change left margin on button for mobile layout

AlexJacksonDS:
Small css change to remove the left margin from the button for the mobile layout so they line up.

image

### DIFF
--- a/DigitalLearningSolutions.Web/Styles/myAccount/myAccount.scss
+++ b/DigitalLearningSolutions.Web/Styles/myAccount/myAccount.scss
@@ -4,16 +4,16 @@
   padding: 16px 32px;
 }
 
-.account-summarylist_row:last-child {
+.account-summary-list__row:last-child {
   border: none;
   margin-bottom: 0;
 }
 
-.account-summarylist_row:last-child .nhsuk-summary-list__key {
+.account-summary-list__row:last-child .nhsuk-summary-list__key {
   border: none;
 }
 
-.account-summarylist_row:last-child .nhsuk-summary-list__value {
+.account-summary-list__row:last-child .nhsuk-summary-list__value {
   border: none;
   margin-bottom: 0px;
 }
@@ -24,7 +24,7 @@
   }
 }
 
-.account-detailslist_row:last-child {
+.account-details-list__row:last-child {
   margin-bottom: 0px;
 }
 
--- a/DigitalLearningSolutions.Web/Styles/myAccount/myAccount.scss
+++ b/DigitalLearningSolutions.Web/Styles/myAccount/myAccount.scss
@@ -18,8 +18,18 @@
   margin-bottom: 0px;
 }
 
-.profile-picture__label{
-    vertical-align: top;
+.account-details-list {
+  @include govuk-media-query($until: tablet) {
+    margin-bottom: 0px;
+  }
+}
+
+.account-detailslist_row:last-child {
+  margin-bottom: 0px;
+}
+
+.profile-picture__label {
+  vertical-align: top;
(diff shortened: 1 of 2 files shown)


DigitalLearningSolutions.Web/Views/MyAccount/Index.cshtml
```diff
@@ -114,7 +114,7 @@
         <a class="nhsuk-button" role="button">
           Edit details
         </a>
-        <a class="nhsuk-button nhsuk-button--secondary  nhsuk-u-margin-left-2" role="button">
+        <a class="nhsuk-button nhsuk-button--secondary" id="change-password-button" role="button">
```
SteveJacksonSoft, line 117: I think a simpler way of achieving this would be to put nhsuk-u-margin-right-2 on the `Edit Details` button instead - then you avoid the media query, which could cause issues for large mobiles where the buttons are still on the same line I guess?

SteveJacksonSoft:
Looking at the screenshot... I feel like the NHS button styles haven't been designed for mobiles, because the buttons always seem really far apart. I think they would look nicer closer together, and with equal amounts of white space at the top and bottom of the edit-details / change password area. What do you think?

In any case, @stellake has a better design mind than I, so I'll defer to her if she disagrees.

AlexJacksonDS:
I've reduced all the margins, and I do think it looks better
image

DigitalLearningSolutions.Web/Styles/myAccount/myAccount.scss
```diff
@@ -18,11 +18,27 @@
...
 
-.profile-picture__label{
-    vertical-align: top;
+.account-details-list {
+  @include govuk-media-query($until: tablet) {
+    margin-bottom: 0px;
+  }
+}
+
+.account-detailslist_row:last-child {
```
SteveJacksonSoft, line 27: Does this do anything?

SteveJacksonSoft:
Yeah, think that looks much better, nice!

DigitalLearningSolutions.Web/Styles/myAccount/myAccount.scss
AlexJacksonDS: It's to make sure these match, it has a bottom margin by default
image

SteveJacksonSoft: Oh, found the things it selects - must have missed them on the first pass.

Just on naming the classes in the MyAccount view - there are a few like account-detailslist_row and account-summarylist_row where the dashes don't line up with the class on the containing element (e.g. account-details-list). It might be a tad nicer if you could match those up.
Also, are you familiar with BEM naming? https://css-tricks.com/bem-101/
We seem to be roughly following that convention, though I don't believe we've taken a conscious decision to do so. According to that you'd have an extra underscore before "row", but I'll leave it to your discretion whether you want to follow that. :)

AlexJacksonDS: I just followed the naming convention started by the account-summary (I didn't do that part), should be easy enough to just change them all so that they are a bit more standard, will update them.

SteveJacksonSoft:
Cool, looks great to me!

stellake (approved):
Looks good to me! 

Just super NIP comment, which is that "0" doesn't need units in CSS (e.g. margin-bottom) and I can see a mixture of these in this file. My preference would be no units, but it's fairly subjective so don't mind as long as this is consistent.

No need for rereview though and looks good! 👍

PR #291, opened 2021-04-27
HEEDLS-394 - Removed the nav bar from Accessibility and Terms of Use …

DanBloxham-sw:
…pages and made sure they open in new tabs. Also made sure the "Menu" button that appears when the nav bar is collapsed no longer appears on these pages, and the visual separator is always present on these pages.

Test manually in chrome and ran all unit tests.

Screen shot of Terms of use page and Accessibility page:
Accessibility Page
Terms of Use page

### DIFF
--- a/DigitalLearningSolutions.Web/Views/LearningSolutions/AccessibilityHelp.cshtml
+++ b/DigitalLearningSolutions.Web/Views/LearningSolutions/AccessibilityHelp.cshtml
@@ -4,7 +4,6 @@
   ViewData["Title"] = "Digital Learning Solutions - Accessibility";
   ViewData["DoNotDisplayNavBar"] = true;
 }
-<link rel="stylesheet" href="@Url.Content("~/css/learningSolutions/accessibilityAndTermsOfUse.css")" asp-append-version="true">
 
 <div class="nhsuk-grid-row">
   <div class="nhsuk-grid-column-full">
--- a/DigitalLearningSolutions.Web/Views/LearningSolutions/AccessibilityHelp.cshtml
+++ b/DigitalLearningSolutions.Web/Views/LearningSolutions/AccessibilityHelp.cshtml
@@ -2,7 +2,9 @@
 @model AccessibilityHelpViewModel
 @{
   ViewData["Title"] = "Digital Learning Solutions - Accessibility";
+  ViewData["DoNotDisplayNavBar"] = true;
 }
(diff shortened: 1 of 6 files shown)


JBsoftwire (approved):
Looks good to me.

DigitalLearningSolutions.Web/Views/Shared/_Layout.cshtml
```diff
@@ -57,118 +57,122 @@
           </a>
         </div>
 
-        <div class="nhsuk-header__content" id="content-header">
-          <div class="nhsuk-header__menu">
-            <button class="nhsuk-header__menu-toggle" id="toggle-menu" aria-controls="header-navigation" aria-label="Open menu">Menu</button>
+        @if (ViewData["DoNotDisplayNavBar"] == null) {
```
stellake, line 60: It might be nicer to check != true to handle DoNotDisplayNavBar being false case :)

DigitalLearningSolutions.Web/Views/LearningSolutions/AccessibilityHelp.cshtml
```diff
@@ -2,7 +2,9 @@
 @model AccessibilityHelpViewModel
 @{
   ViewData["Title"] = "Digital Learning Solutions - Accessibility";
+  ViewData["DoNotDisplayNavBar"] = true;
 }
+<link rel="stylesheet" href="@Url.Content("~/css/learningSolutions/accessibilityAndTermsOfUse.css")" asp-append-version="true">
```
stellake, line 7: Hmm.. could we add conditional styling in the Layout component that depends on if the DoNotDisplayNavBar is true rather than create a new stylesheet and remember to import it everywhere we use the "DoNotDisplayNavBar" flag? (might be a bit less error prone)

DigitalLearningSolutions.Web/Views/Shared/_Layout.cshtml
```diff
@@ -57,119 +57,125 @@
...
+                    </svg>
+                  </a>
+                </li>
+              }
+            </ul>
+          </div>
+        </nav>
+      }
+
+      @{ var visualSeparatorStyle = (bool?)ViewData["DoNotDisplayNavBar"] == true ? "display : block;" : string.Empty; }
```
stellake, line 177: Could we add an extra class in the layout styles so the styling lives in the CSS files? :)

stellake (approved):
Looks good!

PR #292, opened 2021-04-27
Heedls 368 GET /ResetPassword logic

SteveJacksonSoft:
This involves validation of the GET request, redirection where appropriate, and the Error page.

I've manually tested hitting the /ResetPassword endpoint with:
* Missing query params
* Invalid query params (email wrong, incorrect hash or out-of-date hash)
* Valid query params
and everything works.

I've also tried testing the whole reset password journey so far and it seems to work (obvs ignoring the inability to send emails).


DigitalLearningSolutions.Data.Tests/Services/PasswordResetServiceTests.cs
```diff
@@ -1,67 +1,187 @@
...
+            var createTime = DateTime.UtcNow;
+
+            var resetPasswordModel = Builder<ResetPassword>.CreateNew()
+                .With(m => m.Id = 1)
+                .With(m => m.PasswordResetDateTime = createTime)
+                .With(m => m.ResetPasswordHash = "Old Hash Brown")
+                .Build();
+            A.CallTo(() => passwordResetDataService.FindAsync(1)).Returns(Task.FromResult(resetPasswordModel));
+
+            var candidate = Builder<DelegateUser>.CreateNew()
```
SteveJacksonSoft, line 108: @stellake This is the builder library I mentioned a couple of weeks back - I'm happy to replace these with plain class instances if you think it's better. I don't think the builder saves much code here, though it may be more useful in other tests. Happy to chat about it if it's worth it.

DigitalLearningSolutions.Web/Views/Shared/_ActionLink.cshtml
```diff
@@ -0,0 +1,12 @@
+@using DigitalLearningSolutions.Web.ViewModels.Common
+@model LinkViewModel
+
+<div class="nhsuk-action-link">
+  <a class="nhsuk-action-link__link" asp-controller="@Model.AspController" asp-action="@Model.AspAction">
+    <svg class="nhsuk-icon nhsuk-icon__arrow-right-circle" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" aria-hidden="true">
+      <path d="M0 0h24v24H0z" fill="none"></path>
+      <path d="M12 2a10 10 0 0 0-9.95 9h11.64L9.74 7.05a1 1 0 0 1 1.41-1.41l5.66 5.65a1 1 0 0 1 0 1.42l-5.66 5.65a1 1 0 0 1-1.41 0 1 1 0 0 1 0-1.41L13.69 13H2.05A10 10 0 1 0 12 2z"></path>
+    </svg>
+    <span class="nhsuk-action-link__text">Reset Password</span>
```
DanBloxham-sw, line 10: I think "Reset Password" should probably be supplied in a model here.

SteveJacksonSoft: D'oh

DigitalLearningSolutions.Web/Controllers/ResetPasswordController.cs
```diff
@@ -0,0 +1,39 @@
...
+    public class ResetPasswordController : Controller
+    {
+        private readonly IPasswordResetService passwordResetService;
+        public ResetPasswordController(IPasswordResetService passwordResetService)
+        {
+            this.passwordResetService = passwordResetService;
+        }
+
+        [HttpGet]
+        public async Task<IActionResult> Index(string email, string resetHash)
```
DanBloxham-sw, line 16: The query parameter "resetHash" here does not match the query parameter name in the email generated. When the email is created, it uses the parameter name "code" in the link instead. So when clicking the link in the reset password email, it won't be able to find "resetHash" and instead that will come up null. I think you can either change "resetHash" to "code" here, or you can change the link generation to use "resetHash" instead and keep this the same.

I tested renaming the "resetHash" to "code" and that allowed the reset password hash in the email link to be picked up.

DigitalLearningSolutions.Data/Services/PasswordResetService.cs
```diff
@@ -2,100 +2,111 @@
...
+            var resetPasswordIdsForEmail = adminUsersWithEmail.Select(user => user.ResetPasswordId)
+                .Concat(delegateUsersWithEmail.Select(user => user.ResetPasswordId))
+                .Where(id => id.HasValue)
+                .Select(id => id.Value);
+
+            var resetPasswordEntitiesInDb = await Task.WhenAll(
+                resetPasswordIdsForEmail.Select(id => passwordResetDataService.FindAsync(id)));
+
+            return resetPasswordEntitiesInDb.Any(
+                entity => entity != null
```
DanBloxham-sw, line 58: Can this be simplified to the following (using null conditional operator)?
entity => entity?.ResetPasswordHash == resetHash
                          && clockService.UtcNow - entity?.PasswordResetDateTime < TimeSpan.FromHours(2)

DigitalLearningSolutions.Data.Tests/Services/PasswordResetServiceTests.cs
```diff
@@ -1,67 +1,187 @@
...
+            // Then
+            isValid.Should().BeTrue();
+        }
+
+        [Test]
+        public async Task Reset_password_hash_is_invalid_if_no_user_matching_email()
+        {
+            // Given
+            var createTime = DateTime.UtcNow;
+            A.CallTo(() => clockService.UtcNow).Returns(createTime + TimeSpan.FromMinutes(2));
```
DanBloxham-sw, line 155: Very minor, could this use "GivenCurrentTimeIs"?

DigitalLearningSolutions.Data.Tests/Services/PasswordResetServiceTests.cs
```diff
@@ -1,67 +1,187 @@
...
+
+            var resetPasswordModel = Builder<ResetPassword>.CreateNew()
+                .With(m => m.Id = 1)
+                .With(m => m.PasswordResetDateTime = createTime)
+                .With(m => m.ResetPasswordHash = "New Hash Brown")
+                .Build();
+            GivenResetPasswordExists(resetPasswordModel);
+
+            A.CallTo(() => userService.GetUsersByEmailAddress("Albus.Dumbledore@Hogwarts.com"))
+                .Returns((new List<AdminUser>().ToList(), new List<DelegateUser>()));
```
DanBloxham-sw, line 165: Is the ".ToList()" necessary here?

DanBloxham-sw:
Looks mostly good to few, just a few minor questions about formatting, and a concern about the query parameter taken in that needs addressing, but should be a simple fix.

DigitalLearningSolutions.Web/Views/ResetPassword/Index.cshtml
```diff
@@ -0,0 +1 @@
+<h1>Reset Password</h1>
```
stellake, line 1: Super minor, but could we have this "Reset password" to make the capitalisation similar to other pages? Also, whilst we're here, might be worth adding the page title now, so it's not "- DLS" on the tab.

DigitalLearningSolutions.Web/Views/ResetPassword/Error.cshtml
```diff
@@ -0,0 +1,9 @@
+@using DigitalLearningSolutions.Web.ViewModels.Common
+<h1 class="nhsuk-heading-xl">Something went wrong...</h1>
+<p class="nhsuk-body-m">
+  There was a problem resetting your password.
+  The reset password link is only valid once, and expires after 2 hours.
+  Please try again.
+</p>
+
+<partial name="_ActionLink" model="@(new ActionLinkViewModel("ForgotPassword", "Index", "Reset password"))" />
```
stellake, line 9: Hmm.. I'm wondering whether it's worth making this VC so we can use it with a similar syntax as back button (I don't think we'd ever create the View model in the code so might make sense here) - what do you think?

DigitalLearningSolutions.Web/Views/ResetPassword/Error.cshtml
```diff
@@ -0,0 +1,9 @@
+@using DigitalLearningSolutions.Web.ViewModels.Common
+<h1 class="nhsuk-heading-xl">Something went wrong...</h1>
```
stellake, line 2: I'm not 100% sure about the "..." at the end of the title. I think it's worth taking a screenshot of the page and posting in the HEE external to see if they have any comments about the content of this page :)

DigitalLearningSolutions.Web/Controllers/ResetPasswordController.cs
```diff
@@ -0,0 +1,39 @@
...
+        private readonly IPasswordResetService passwordResetService;
+        public ResetPasswordController(IPasswordResetService passwordResetService)
+        {
+            this.passwordResetService = passwordResetService;
+        }
+
+        [HttpGet]
+        public async Task<IActionResult> Index(string email, string code)
+        {
+            if (string.IsNullOrEmpty(email) || string.IsNullOrEmpty(code))
```
stellake, line 18: I'm wondering whether we should also redirect the user if they are logged in  🤔

DigitalLearningSolutions.Web/Controllers/ResetPasswordController.cs
```diff
@@ -0,0 +1,39 @@
+namespace DigitalLearningSolutions.Web.Controllers
+{
+    using System.Threading.Tasks;
+    using DigitalLearningSolutions.Data.Services;
+    using Microsoft.AspNetCore.Mvc;
+
+    public class ResetPasswordController : Controller
+    {
+        private readonly IPasswordResetService passwordResetService;
```
stellake, line 9: Super NIP, but could we add an empty line before the constructor for better readability (I think the autoformatter should be doing it by default).

DigitalLearningSolutions.Data/Services/PasswordResetService.cs
```diff
@@ -1,120 +1,130 @@
...
+            this.userService = userService;
+            this.passwordResetDataService = passwordResetDataService;
+            this.logger = logger;
+            this.emailService = emailService;
+            this.clockService = clockService;
+        }
+
+        public async Task<bool> PasswordResetHashIsValidAsync(string emailAddress, string resetHash)
+        {
+            var (adminUsersWithEmail, delegateUsersWithEmail) = userService.GetUsersByEmailAddress(emailAddress);
```
stellake, line 47: Do you think it's worth doing a join with reset password table when fetching the users to simplify some of the logic below?

DigitalLearningSolutions.Data.Tests/Services/PasswordResetServiceTests.cs
```diff
@@ -1,67 +1,218 @@
...
+                .With(m => m.Id = 1)
+                .With(m => m.PasswordResetDateTime = createTime)
+                .With(m => m.ResetPasswordHash = "Old Hash Brown")
+                .Build();
+            A.CallTo(() => passwordResetDataService.FindAsync(1)).Returns(Task.FromResult(resetPasswordModel));
+
+            var candidate = Builder<DelegateUser>.CreateNew()
+                .With(user => user.ResetPasswordId = resetPasswordModel.Id)
+                .Build();
+            A.CallTo(() => userService.GetUsersByEmailAddress(A<string>._))
```
stellake, line 111: Super minor, but could we specify the email as an argument here (I think it could also benefit the tests above) to make sure the mock only returns if called with the correct email address?

DigitalLearningSolutions.Data.Tests/Services/PasswordResetServiceTests.cs
```diff
@@ -1,67 +1,218 @@
...
+            var createTime = DateTime.UtcNow;
+
+            var resetPasswordModel = Builder<ResetPassword>.CreateNew()
+                .With(m => m.Id = 1)
+                .With(m => m.PasswordResetDateTime = createTime)
+                .With(m => m.ResetPasswordHash = "New Hash Brown")
+                .Build();
+            GivenResetPasswordExists(resetPasswordModel);
+
+            var candidate = Builder<DelegateUser>.CreateNew().With(user => user.ResetPasswordId = resetPasswordModel.Id)
```
stellake, line 136: Minor, but could we put the "With" on a new line too for better readability (like above)

DigitalLearningSolutions.Data.Tests/Services/PasswordResetServiceTests.cs
```diff
@@ -1,67 +1,218 @@
...
+                .With(m => m.Id = 1)
+                .With(m => m.PasswordResetDateTime = createTime)
+                .With(m => m.ResetPasswordHash = "New Hash Brown")
+                .Build();
+            GivenResetPasswordExists(resetPasswordModel);
+
+            A.CallTo(() => userService.GetUsersByEmailAddress("Albus.Dumbledore@Hogwarts.com"))
+                .Returns((new List<AdminUser>(), new List<DelegateUser>()));
+
+            GivenCurrentTimeIs(createTime + TimeSpan.FromMinutes(2));
```
stellake, line 166: This reads so nicely 🙌

DigitalLearningSolutions.Web/Views/ResetPassword/Index.cshtml
SteveJacksonSoft: Oh, yeah, this page is going to be completely changed in HEEDLS-369 (This is just a placeholder view). So I'll make sure that's done then and there.

DigitalLearningSolutions.Web/Views/ResetPassword/Error.cshtml
SteveJacksonSoft: Sounds fine to me! I'll do that.

DigitalLearningSolutions.Data/Services/PasswordResetService.cs
SteveJacksonSoft: Oops, using EF too much has made me lazy! Yep, think that would make it speedier.

DigitalLearningSolutions.Web/Controllers/ResetPasswordController.cs
SteveJacksonSoft: Hmm, yeah, I'm not sure. I think we should be consistent with the ForgotPassword page, which does redirect to Login.

DigitalLearningSolutions.Web/Views/ResetPassword/Error.cshtml
SteveJacksonSoft: Done.

DigitalLearningSolutions.Web/Views/NotificationPreferences/_NotificationPreferenceExpander.cshtml
```diff
@@ -8,7 +9,7 @@
 <details class="nhsuk-details nhsuk-expander">
   <summary class="nhsuk-details__summary">
     <span class="nhsuk-details__summary-text">
-      My @(Model.UserType == UserTypes.Admin ? "admin" : "delegate") account notifications
+      My @(((UserType)Model.UserType).Equals(UserType.AdminUser) ? "admin" : "delegate") account notifications
```
stellake, line 12: I'm surprised the type doesn't get inferred automatically 👁️

DigitalLearningSolutions.Web/ViewModels/MyAccount/NotificationPreferencesViewModel.cs
```diff
@@ -1,6 +1,7 @@
 ﻿namespace DigitalLearningSolutions.Web.ViewModels.MyAccount
 {
```
stellake, line 2: @JBsoftwire FYI there'll be a minor merge conflict with the user types

DigitalLearningSolutions.Web/Controllers/ResetPasswordController.cs
```diff
@@ -0,0 +1,46 @@
...
+            {
+                return RedirectToAction("Index", "Home");
+            }
+
+            if (string.IsNullOrEmpty(email) || string.IsNullOrEmpty(code))
+            {
+                return RedirectToAction("Index", "Login");
+            }
+
+            var matchingUserRefs = await passwordResetService.GetValidMatchingUserReferencesAsync(email, code);
```
stellake, line 30: I wonder if this method should return a boolean and be something like ResetPasswordLinkIsValid(email, code) - if we need the user later down the line, maybe we can shuffle things around within the passwordResetService. What do you think? :)

DigitalLearningSolutions.Web.Tests/Controllers/ResetPasswordControllerTests.cs
```diff
@@ -0,0 +1,108 @@
...
+        public async Task Index_should_redirect_to_login_page_if_user_is_unauthenticated_and_code_is_missing()
+        {
+            // When
+            var result = await unauthenticatedController.Index("email", null);
+
+            // Then
+            result.Should().BeRedirectToActionResult().WithControllerName("Login").WithActionName("Index");
+        }
+
+        private ResetPasswordController GetControllerWithUser(ClaimsPrincipal user)
```
stellake, line 92: Really nice tests! ⭐ 

One thing, Daniel set up a ControllerContextHelper which should allow you to write something like this:
```
new Controller()
                .WithDefaultContext()
                .WithMockUser(false);
```

DigitalLearningSolutions.Data/Services/PasswordResetService.cs
```diff
@@ -1,119 +1,129 @@
...
+        }
+
+        public void GenerateAndSendPasswordResetLink(string emailAddress, string baseUrl)
+        {
+            (User? user, List<DelegateUser> delegateUsers) = userService.GetUsersByEmailAddress(emailAddress);
+            user ??= delegateUsers.FirstOrDefault() ??
+                     throw new UserAccountNotFoundException(
+                         "No user account could be found with the specified email address");
+            string resetPasswordHash = GenerateResetPasswordHash(user);
+            Email resetPasswordEmail = GeneratePasswordResetEmail(
```
stellake, line 70: We usually write "var" as types should be inferred.

DigitalLearningSolutions.Web/Controllers/ResetPasswordController.cs
stellake: In fact, when we reset a password (second part of the ticket), we don't care about whether the hash is correlated to the account entry or not - we should just change the pw for all the entries where the email matches the user email address as long as at least one of these entries matches the reset password hash (which is there to ensure that the user owns that email address).

DigitalLearningSolutions.Data.Tests/Services/NotificationPreferenceServiceTests.cs
```diff
@@ -0,0 +1,86 @@
+namespace DigitalLearningSolutions.Data.Tests.Services
+{
```
stellake, line 2: Nice! 👍

DigitalLearningSolutions.Web/Controllers/ResetPasswordController.cs
SteveJacksonSoft: Yeah, so we will need the user id and type to be tracked in TempData or something for actually resetting the password. This is the result of the shuffling things around that I did because I started thinking about that part of the ticket last week.

DigitalLearningSolutions.Data/Helpers/AuthHelpers.cs
```diff
@@ -0,0 +1,15 @@
+namespace DigitalLearningSolutions.Data.Helpers
+{
+    using System;
+    using DigitalLearningSolutions.Data.Models.DbModels;
+
+    public static class AuthHelpers
```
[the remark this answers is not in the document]
SteveJacksonSoft: Yes, I'm never sure how to name new helper classes. I could name it `PasswordHashValidationHelpers`, but then when people add new methods that possibly should go alongside this method they might just make a new class. What do you think the correct level of granularity is?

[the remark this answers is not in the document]
stellake: I think starting as PasswordHashValidationHelper would be ok, and in HLDs I can just point people towards this if there's something that'd go well with this. Then we can change the name to something less specific depending on what is added to the helper :)

DigitalLearningSolutions.Web/Views/NotificationPreferences/_NotificationPreferenceExpander.cshtml
stellake: Ah that's fair! I think swapping them around would be nicer, just to avoid the extra brackets, which aren't very easy to read. However, if you find it more difficult to read that way, then maybe the answer is to just pull this whole line into a variable 🤔

DigitalLearningSolutions.Data.Tests/DataServices/PasswordResetDataServiceTests.cs
```diff
@@ -15,107 +15,99 @@ namespace DigitalLearningSolutions.Data.Tests.DataServices
 
     public class PasswordResetDataServiceTests
```
SteveJacksonSoft, line 16: Sorry, I moved `Can_Create_And_Find_A_Password_Reset_For_User` up to the top - the only actual code change on this file is converting that from two separate tests to a theory.

DigitalLearningSolutions.Web/Views/NotificationPreferences/_NotificationPreferenceExpander.cshtml
SteveJacksonSoft: Ah, I tell a lie, you still need the explicit cast - `Enumeration.Equals()` takes an `object?`.

stellake: In that case, let's leave it! 👍

stellake (approved):
Looks great! 👍

PR #293, opened 2021-04-28
Heedls 399 user not signed in on old site after logging in on new site

DanBloxham-sw:
I have changed the initial cookie set up in Startup.cs to set the authorisation cookies' domain to a specific value which I have added into the config files. Previously, the domain was not explicitly set, so I presume the regular root path is used. In UAT, this becomes dls.nhs.uk/dev-my-learning-portal and so the authentication cookies would have included the path and thus not become useable on the old site. The specific domain name should fix this by not including that path (and manually setting the path to be empty).

I have tested that it does not break the cookies on development, so the locally hosted old site is logged in when already logged in on the new site. However, the ASP cookies can only have a domain based on the domain that is actually being hosted (e.g. localhost cookies cannot have a domain of "hee-dls-test". But I cannot guarantee that the bug on UAT is now fixed as I cannot test it there. I will try to look in to a better method of determining the base path so we don't have to look at the settings, but I am unsure if there is such a method.

I have ran all unit tests and they work fine.

### DIFF
--- a/DigitalLearningSolutions.Web/Startup.cs
+++ b/DigitalLearningSolutions.Web/Startup.cs
@@ -43,6 +43,7 @@ namespace DigitalLearningSolutions.Web
                 {
                     options.Cookie.Name = ".AspNet.SharedCookie";
                     options.Cookie.Domain = config.GetValue<string>("SharedCookieDomain");
+                    options.Cookie.Path = "/";
                     options.Events.OnRedirectToLogin = RedirectToLogin;
                     options.Events.OnRedirectToAccessDenied = RedirectToHome;
                 });
--- a/DigitalLearningSolutions.Web/Startup.cs
+++ b/DigitalLearningSolutions.Web/Startup.cs
@@ -1,5 +1,6 @@
 namespace DigitalLearningSolutions.Web
 {
+    using System.Configuration;
     using System.Data;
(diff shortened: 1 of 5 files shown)


DigitalLearningSolutions.Web/Startup.cs
```diff
@@ -43,18 +42,19 @@ public void ConfigureServices(IServiceCollection services)
                 .AddCookie("Identity.Application", options =>
                 {
                     options.Cookie.Name = ".AspNet.SharedCookie";
+                    options.Cookie.Domain = config.GetValue<string>("SharedCookieDomain");
```
stellake, line 45: This is interesting! When I inspect the cookie on UAT, the domain seems to be set properly, so I was wondering if this was required (e.g. would setting the path be sufficient?)
image

stellake: Oops, just realised this image doesn't have headers! Second try:
image

DanBloxham-sw: In which case, that hopefully means the changes can be reverted to just manually set the path to be empty. That should make the review easier when I get the change committed.

stellake (approved):
Looks good! 👍

PR #295, opened 2021-04-29
HEEDLS-348 - Implemented Log Out functionality

DanBloxham-sw:
Does exactly what is says on the tin.

Tested manually in chrome, and ran all unit tests. Also tested the old site to make sure the shared cookie/session is affected properly to log the user out on the old site.

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/ControllerHelpers/ControllerContextHelper.cs
+++ b/DigitalLearningSolutions.Web.Tests/ControllerHelpers/ControllerContextHelper.cs
@@ -34,13 +34,15 @@
             controller.TempData = tempData;
         }
 
-        public static void SetUpControllerWithSignInFunctionality<T>(ref T controller, string authenticationType = "")
+        public static void SetUpControllerWithServices<T>(ref T controller, string authenticationType = "")
             where T : Controller
         {
             SetUpControllerWithUser(ref controller, authenticationType);
 
             var authService = A.Fake<IAuthenticationService>();
             A.CallTo(() => authService.SignInAsync(A<HttpContext>._, A<string>._, A<ClaimsPrincipal>._,
+                A<AuthenticationProperties>._)).Returns(Task.CompletedTask);
+            A.CallTo(() => authService.SignOutAsync(A<HttpContext>._, A<string>._,
                 A<AuthenticationProperties>._)).Returns(Task.CompletedTask);
 
             var urlHelperFactory = A.Fake<IUrlHelperFactory>();
--- a/DigitalLearningSolutions.Web.Tests/ControllerHelpers/ControllerContextHelper.cs
+++ b/DigitalLearningSolutions.Web.Tests/ControllerHelpers/ControllerContextHelper.cs
@@ -3,7 +3,6 @@
     using System;
     using System.Security.Claims;
     using System.Threading.Tasks;
-    using DigitalLearningSolutions.Data.Tests.Helpers;
     using FakeItEasy;
     using Microsoft.AspNetCore.Authentication;
     using Microsoft.AspNetCore.Http;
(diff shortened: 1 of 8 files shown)


DigitalLearningSolutions.Web/Controllers/LogoutController.cs
```diff
@@ -0,0 +1,16 @@
...
+    using Microsoft.AspNetCore.Authentication;
+    using Microsoft.AspNetCore.Mvc;
+
+    public class LogoutController : Controller
+    {
+        [HttpGet]
+        public IActionResult Index()
+        {
+            HttpContext.SignOutAsync();
+            HttpContext.Response.Cookies.Delete("ASP.NET_SessionId");
```
DanBloxham-sw, line 12: The old site uses an additional cookie – ASP.NET_SessionId – that is not used on the new site, but is generated if the user logs in on the new site, and then goes to the old site in the same session. If it is not deleted when logging out, then the user will still be logged in on the old site until they close the browser (ending the session). Deleting this cookie means that they will be immediately logged out on the old site when they log out on the new site (though they will still have to refresh the page to see the effect).

DigitalLearningSolutions.Web.Tests/Controllers/Logout/LogoutControllerTests.cs
```diff
@@ -0,0 +1,60 @@
...
+                        User = user,
+                        Session = session,
+                        RequestServices = services
+                    }
+                }
+            };
+        }
+
+        [Test]
+        public void Logout_should_redirect_user_to_home_page()
```
JBsoftwire, line 51: Would it be feasible to also have a test that the intended logout has actually occurred? Right now it'd be possible to pass the tests without the logout controller method changing anything.

DanBloxham-sw: It doesn't look like there is an easy way to do this unfortunately. The SignOutAsync method is a static extension method that can't be intercepted, and due to the way that the authentication service is mocked, the User Identity details aren't accurately represented in tests. I tried having a look into incorporating something similar into the integration tests, but there seems to be similar issues with the authentication service.

stellake: Oh, I meant more if the authService that we mock when creating a controller gets called as a result of this? (I think it has SignOutAsync method on it that we mock explicitly)

stellake: Oh sorry, it's in ControllerContextHelper currently

DanBloxham-sw: Yeah the set up is all a bit confusing, but the mocked authService all gets bundled up together and put in to the HttpContext, which is where it is accessed from. 

Might be able to pull that out into the ControllerTests, I'll take a look.

DanBloxham-sw: You were right, I didn't even think of this approach but it works perfectly. I implemented a test that ensures that the SignOutAsync method is called when logging out, and did a similar test in the LoginControllerTests for SignInAsync.

DigitalLearningSolutions.Web.Tests/ControllerHelpers/ControllerContextHelper.cs
```diff
@@ -0,0 +1,56 @@
...
+    using FakeItEasy;
+    using Microsoft.AspNetCore.Authentication;
+    using Microsoft.AspNetCore.Http;
+    using Microsoft.AspNetCore.Mvc;
+    using Microsoft.AspNetCore.Mvc.Routing;
+    using Microsoft.AspNetCore.Mvc.ViewFeatures;
+
+    public static class ControllerContextHelper
+    {
+        public static void SetUpControllerWithUser<T>(ref T controller, string authenticationType = "")
```
stellake, line 16: A few questions about this class:
- What's the role of authentication type being a parameter rather than something hardcoded?
- Do you think that instead of passing the ref around and adding a controller as a param, we write extension methods similar to CustomClaimHelper that will allow us to use syntax like controller.SetUpMockUser()? This way we can also make the syntax a bit cleaner (I think you should be able to use something like SetUpControllerWithUser(this Controller controller, ...) { } + use controller.SetUpThis; controller.SetUpThat; in the tests)
- The "Set up controller with user" is actually setting up a mock user session and a fake temp data, which seems independent of the user session. Could we potentially separate this out into two methods that can be called where relevant?
- The "SetUpControllerWithServices" is actually setting up the user session too behind the scenes which could be confusing to those that call it - could we make these two independent. Maybe something like a following set of methods:
- SetUserSession
- SetUpFakeTempData (although this is so small and rare that we could just set it up in the test setup and actually have reference to the fake - then we can check whether particular temp data is attempted to be saved/cleared etc in the tests).
- SetUpFakeAuthenticationService

DanBloxham-sw: The authentication type determines what type of user is created:
- if it is an empty string then an unauthenticated user is set up (if this wasn't set up, then controller checks such as User.Identity.IsAuthenticated to redirect would fail with a null user)
- if it is anything other than an empty string, an authenticated user is set up.
I left it as a parameter so the same code can be run to set things up with just that one parameter deciding if a user should be authenticated, but I suppose a single bool UserIsAuthenticated would help make the method signature easier to understand.

The other remarks make sense, I'll look into them. Might have to be careful with the order the methods are called in as one will set up new ControllerContext and new DefaultHttpContext while the others will just modify them, but it should be feasible.

DanBloxham-sw: I've implemented those changes. The session setting has been removed as it wasn't being used anywhere. I also created the SetUpDefaultController for the initial set up of the default http context which will always need to be called before the others so I'm not sure if there is a better way around it.

stellake: I think that's fine! What I noticed is that you set these methods up more like a builder (returning itself), which is quite nice. What do you think about calling these methods "WithMockUser" etc so when you create the controller it reads like this:

```
var controller = new XXXController()
.WithDefaultContext()
.WithMockUser()
.WithMockServices();
```

I'm happy with either, but I think either way it would make sense to put the .SetUpXXX or .WithXXX onto new lines when defining the controller (for readability). Otherwise, looks great!

DanBloxham-sw: Yeah I put the return in so that it could all be done on a single line when the controller is first defined. Seeing as that is how it is usually used, I think the "WithMockX" make more sense. I'll make those changes.

stellake (approved):
Looks good! :)

PR #297, opened 2021-04-30
HEEDLS-349: backend wired up

JBsoftwire:


DigitalLearningSolutions.Data/Services/NotificationPreferencesDataService.cs
```diff
@@ -2,13 +2,16 @@
 {
```
davidm-m, line 2: We have a separate DataServices folder now this probably wants to move into

davidm-m (approved):
Looks good, only thing is NotificationPreferencesDataService should be moved into the DataServices folder.

DigitalLearningSolutions.Web/Controllers/NotificationPreferencesController.cs
```diff
@@ -65,10 +65,26 @@ private IEnumerable<NotificationPreference> GetNotificationPreferencesForUser(st
 
         [HttpPost]
         [Route("/NotificationPreferences/Edit/{userType}")]
-        public IActionResult SaveNotificationPreferences(IEnumerable<int> notifications)
+        public IActionResult SaveNotificationPreferences(string userType, IEnumerable<int> notifications)
```
stellake, line 68: Could we name this "notificationIds" to keep it consistent with the naming below? :)

DigitalLearningSolutions.Data/Services/NotificationPreferencesDataService.cs
```diff
@@ -72,5 +75,43 @@ public IEnumerable<NotificationPreference> GetNotificationPreferencesForDelegate
...
+                new { delegateId }
+            );
+
+            var notificationIdsWithDelegateId = notificationIds.Select(notificationId => new { delegateId, notificationId });
+
+            connection.Execute(
+                @"INSERT INTO NotificationUsers (NotificationId, CandidateId)
+                VALUES (@notificationId, @delegateId)",
+                notificationIdsWithDelegateId
+            );
```
stellake, line 114: If the second query fails, does the deletion get rolled back too?

DigitalLearningSolutions.Data.Tests/Services/NotificationPreferencesDataServiceTests.cs
```diff
@@ -92,5 +93,73 @@ public void Get_notification_preferences_for_delegate_should_return_empty_list_f
...
+
+        [Test]
+        public void Update_notification_preferences_for_admin_should_update_notification_preferences_for_admin()
+        {
+            using (var transaction = new TransactionScope())
+            {
+                try
+                {
+                    // when
+                    var notificationIds = new[] { 2, 3, 4 };
```
stellake, line 105: Do you think it might be worth structuring the test such that we filter the result by taking only accepted notification preferences and check that the ids are 2, 3 and 4? 
(no need to check the name as we're testing the GetPreferences method above)

DigitalLearningSolutions.Data.Tests/Services/NotificationPreferencesDataServiceTests.cs
```diff
@@ -1,6 +1,7 @@
 ﻿namespace DigitalLearningSolutions.Data.Tests.Services
```
stellake, line 1: Should this be in data services folder too? :P

DigitalLearningSolutions.Data/DataServices/NotificationPreferencesDataService.cs
```diff
@@ -72,5 +76,55 @@ public IEnumerable<NotificationPreference> GetNotificationPreferencesForDelegate
...
+                var notificationIdsWithAdminId =
+                    notificationIds.Select(notificationId => new { adminId, notificationId });
+
+                connection.Execute(
+                    @"INSERT INTO NotificationUsers (NotificationId, AdminUserId)
+                    VALUES (@notificationId, @adminId)",
+                    notificationIdsWithAdminId
+                );
+
+                transaction.Complete();
```
stellake, line 101: Does this work? Could you test if nothing gets committed when "Complete" is removed?
https://stackoverflow.com/questions/6860979/dapper-transactionscope

JBsoftwire: It does work, removing it from a function causes the associated tests to fail.

Reading around there seems to be some confusion because e.g. read-only statements can still 'succeed' (retrieve data) even if in an uncommitted transaction, but the Complete() is required to write. The more developed version would Dispose() in a catch or finally block, but I understand that to be what the using{} block converts to when compiled, so adding it myself would be redundant.

My understanding here is based on this question and answers: https://stackoverflow.com/questions/53657509/do-i-need-transactionscope-complete

stellake: Ah, nice! Thanks for checking! 🙏

stellake (approved):
Looks good - happy for you to merge without rereview once the tests are moved to dataServices too 👍

PR #298, opened 2021-05-05
HEEDLS-372 - Set up basic centre config page

DanBloxham-sw:
Couldn't really think of any unit tests for the new controller, so if any come to mind do let me know.

Tested:
- All unit tests
- Tested in Chrome, IE11, Edge, Firefox
- Tested screen reader
- Tested no JS
- Tested in Chrome mobile view

Screen shots of UI:

Centre Configuration

Mobile view:

Centre Configuration Mobile


DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfigurationController.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public IActionResult Index()
+        {
+            var centreId = User.GetCentreId();
+
+            var centreDetails = centresDataService.GetCentreDetailsById(centreId);
+
+            var model = new CentreConfigurationViewModel(centreDetails.CentreName, centreDetails.RegionName);
+
+            return View("~/Views/TrackingSystem/CentreConfiguration/Index.cshtml", model);
```
DanBloxham-sw, line 28: Note that the view specified here is not just "Index" as it is not picked up automatically. This is because the controller name is "CentreConfigurationController" rather than "TrackingSystemController". 

The ticket states we will have a lot of controllers in the Tracking System so it is best to have a new folder for them. Assuming this, I created the controller as its own CentreConfigurationController, rather than making a partial TrackingSystemController as is done elsewhere (e.g. LearningPortal). I did this because it will make the individual controllers a little bit easier to manage and unit test, but a consequence is the automatic pathing doesn't work right so the views have to be specified manually. If we want to switch to using a single TrackingSystemController that uses partial classes for all its "child controllers" then it will be easy enough to change back. Thoughts?

AlexJacksonDS: Had a quick google, and it might be possible to define custom 'search locations' in addition to the default ones. Might be something to consider if we can set up some standard way of naming all this stuff so that we can add some simple rule
https://docs.microsoft.com/en-us/aspnet/core/mvc/views/overview?view=aspnetcore-5.0#how-controllers-specify-views

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/Index.cshtml
```diff
@@ -0,0 +1,27 @@
...
+@{
+  ViewData["Title"] = "Centre Configuration";
+}
+
+<div class="nhsuk-grid-row">
+  <div class="nhsuk-grid-column-full">
+    <h1 id="page-heading" class="nhsuk-heading-xl">Centre Configuration</h1>
+    <div class="nhsuk-card">
+      <div class="nhsuk-card__content">
+        <table class="nhsuk-table">
```
AlexJacksonDS, line 12: I think we should be using a <dl class="nhsuk-summary-list"> here rather than a table, the same as the MyAccount Index

AlexJacksonDS (commented):
image
Also we should remove this line from the last row of the card, same as MyAccount

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/Index.cshtml
AlexJacksonDS: https://service-manual.nhs.uk/design-system/components/summary-list

DanBloxham-sw:
New screenshot of page using a summary list instead of a table below. I have made a new scss file to contain the new classes to remove the bottom border. These were basically just copied over from myAccount.scss. I've put them in a new file specific to the tracking system centre configuration, but it would be feasible to move the classes in myAccount.scss into a more general .scss file and rename them to be more generic so they can be reused. But for now I have kept the TrackingSystem styles separate. Do we think this is suitable?

Centre Configuration Restyled

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfigurationController.cs
DanBloxham-sw: Good call. I have looked into that and made some changes in StartUp.cs that follow along with this stackoverflow answer: https://stackoverflow.com/a/59218969

AlexJacksonDS (approved):
The custom search path was nicer than I expected.

I kind of agree on the css being common between this and the MyAccount page, I've just suggested moving some common stuff from there to a common file on another pull request so maybe just leave it for now and go back later?

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/Index.cshtml
```diff
@@ -0,0 +1,28 @@
...
+
+<link rel="stylesheet" href="@Url.Content("~/css/trackingSystem/centreConfiguration.css")" asp-append-version="true">
+
+@{
+  ViewData["Title"] = "Centre Configuration";
+}
+
+<div class="nhsuk-grid-row">
+  <div class="nhsuk-grid-column-full">
+    <h1 id="page-heading" class="nhsuk-heading-xl">Centre Configuration</h1>
```
stellake, line 12: Could we have h1 as "Centre configuration" to match capitalisation of other pages? 🙏

stellake (commented):
One comment about the h1 capitalisation, otherwise looks good! I think it might be worth commonising this CSS as part of this MR as otherwise it won't get done. 👀

DanBloxham-sw:
I've changed the h1/title to have correct capitalisation - not sure how I keep missing this, but I've made a note so hopefully it won't happen again.

I've also moved the stylings for the summary list classes into /Styles/index.scss as that is where most common classes seem to be placed. I couldn't think of a good name for the new general classes so I just called them "basic-..." as they are currently only used in very basic summary lists that don't show a whole lot of detail.

stellake (approved):
Looks good!

PR #301, opened 2021-05-05
HEEDLS-361 - Choose a centre bug fixes

DanBloxham-sw:
Details of bug fixes:

Choose button margins - manually set the upper margin to 0 using the NHS styling override classes

Roles section is narrow - set minimum width of the choose-centre-td so that the tags are never collapsed on top of one another - the centre name will instead flow over multiple lines.

The h1 Choose A Centre -> Choose a centre

Login failed on UAT - The TempData cookie was indeed causing the problem. This was due to the addition of the profile image to the GetAdminUserByUsername method, and that was passed in TempData. In the SetTempDataForChooseACentre method, I have ensured that the profile images are set to null for both admins and delegates, and this has fixed the problem. 

Additional bug I discovered while fixing the others: user could navigate using Back button to the ChooseACentre page after logging in, and could cause errors to occur by manually navigating to the URL for choosing a centre. I've refactored how the LoginController navigates to this page to prevent the bug from occurring. 

Table styling fix screenshots:

Table with styling fix mobile view

Table with styling fix


DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -98,16 +123,28 @@ public IActionResult ChooseACentre(int centreId)
...
         private void SetTempDataForChooseACentre(bool rememberMe, AdminUser? adminUser,
-            List<DelegateUser> approvedDelegateUsers)
+            List<DelegateUser> approvedDelegateUsers, ChooseACentreViewModel chooseACentreViewModel)
         {
+            if (adminUser != null)
+            {
+                adminUser.ProfileImage = null;
+            }
+
+            foreach (var user in approvedDelegateUsers.Where(du => du.ProfileImage != null))
```
DanBloxham-sw, line 138: Currently, no profile image is grabbed in GetDelegateUsersByUsername so approvedDelegateUsers will never have an entity where the profile image is not null. So this foreach loop doesn't really do anything, but acts as a preventative measure in case changes are made to GetDelegateUsersByUsername to grab the profile picture at some point in the future. Do we think we should keep it, or get rid of it and only implement it if it becomes necessary?

JBsoftwire (approved):

DigitalLearningSolutions.Web/Views/Login/ChooseACentre.cshtml
```diff
@@ -1,7 +1,7 @@
 ﻿@using DigitalLearningSolutions.Web.ViewModels.Login
 @model ChooseACentreViewModel
 @{
-  ViewData["Title"] = "Choose A Centre";
+  ViewData["Title"] = "Choose a centre";
```
stellake, line 4: This doesn't really matter, but I don't think there's an issue with the tab title's capitalisation - especially because (confusingly) every work is usually capitalised there :P

DigitalLearningSolutions.Web/Controllers/LoginController.cs
stellake: I think we should actually create a new model, which only has the fields that we need to save in temp data. Then whichever fields we add to the admin and/or delegate db queries won't be affecting Choose a centre behaviour, because there's only a set of fields we map to what's saved in the tempdata.

DigitalLearningSolutions.Web/Models/DelegateClaimDetails.cs
```diff
@@ -0,0 +1,28 @@
+﻿namespace DigitalLearningSolutions.Web.Models
+{
+    using DigitalLearningSolutions.Data.Models.User;
+
+    public class DelegateClaimDetails
```
stellake, line 5: Could we name these classes to DelegateLoginDetails and AdminLoginDetails as I think that's a bit more intuitive?

DanBloxham-sw: I'd hesitate on the "...Login..." part because to me that would imply it contains details for logging in (e.g. password), but then I can't think of a better name, so I'll go with your suggestions.

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -69,50 +71,85 @@ public IActionResult Index(LoginViewModel model)
                     model.Password);
 
             var availableCentres = userService.GetUserCentres(verifiedAdminUser, approvedDelegateUsers);
```
stellake, line 73: Ah, this complains in my IDE for adminUser potentially being null 👀

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -69,50 +71,85 @@ public IActionResult Index(LoginViewModel model)
...
+            (AdminClaimDetails? adminClaimDetails, List<DelegateClaimDetails> delegateClaimDetails) =
+                GetClaimsDetails(verifiedAdminUser, approvedDelegateUsers);
             if (availableCentres.Count == 1)
             {
-                if (verifiedAdminUser != null)
+                if (adminClaimDetails != null)
```
stellake, line 78: Maybe we could refactor this into StartAdminSession(adminLoginDetails), which checks for null - then we can use it in Choose a centre too?

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -69,50 +71,85 @@ public IActionResult Index(LoginViewModel model)
...
+            {
+                return RedirectToAction("Index", "Home");
+            }
+
+            var model = TempData.Peek<ChooseACentreViewModel>();
+
+            return View("ChooseACentre", model);
+        }
+
+        [ServiceFilter(typeof(RedirectEmptySessionData<List<DelegateUser>>))]
```
stellake, line 113: How come this is a list of delegate users? 👀

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -69,50 +71,85 @@ public IActionResult Index(LoginViewModel model)
...
-        private void SetTempDataForChooseACentre(bool rememberMe, AdminUser? adminUser,
-            List<DelegateUser> approvedDelegateUsers)
+        private (AdminClaimDetails?, List<DelegateClaimDetails>) GetClaimsDetails(AdminUser? adminUser, List<DelegateUser> delegateUsers)
+        {
+            var adminClaimDetails = adminUser != null ? new AdminClaimDetails(adminUser) : null;
+            var delegateClaimDetails = delegateUsers.Select(du => new DelegateClaimDetails(du)).ToList();
+            return (adminClaimDetails, delegateClaimDetails);
+        }
+
+        private void SetTempDataForChooseACentre(bool rememberMe, AdminClaimDetails? adminClaimDetails,
```
stellake, line 140: If we need to split the params on new lines, could we add each of them on a new line for better readability? 🙏  (I don't think the formatter does it by default)

DigitalLearningSolutions.Web/Controllers/LoginController.cs
DanBloxham-sw: I think I agree but I am not certain. If the null check was put inside StartAdminSession, then the code in here looks like we always start an admin session (though admittedly it would be with an adminDetails?.Id, so that might make it obvious that there are null checks inside)

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -105,51 +101,63 @@ public IActionResult ChooseACentre()
...
-            }
+                delegateLoginDetails?.FirstOrDefault(du => du.CentreId == centreId);
+
+            sessionService.StartAdminSession(adminAccountForChosenCentre?.Id);
 
             return LogIn(adminAccountForChosenCentre, delegateAccountForChosenCentre, rememberMe);
         }
 
-        private (AdminClaimDetails?, List<DelegateClaimDetails>) GetClaimsDetails(AdminUser? adminUser, List<DelegateUser> delegateUsers)
+        private (AdminLoginDetails?, List<DelegateLoginDetails>) GetClaimsDetails
```
stellake, line 127: Super NIP, but should this method be renamed too?

DigitalLearningSolutions.Web/Views/Login/ChooseACentre.cshtml
```diff
@@ -9,7 +9,7 @@
 <div class="nhsuk-grid-row">
   <div class="nhsuk-grid-column-full">
 
-    <h1 id="page-heading" class="nhsuk-heading-xl">Choose a Centre</h1>
+    <h1 id="page-heading" class="nhsuk-heading-xl">Choose a centre</h1>
```
stellake, line 12: This is beautiful! ⭐

PR #302, opened 2021-05-05
HEEDLS-346: move whitespace trimming from database to memory

JBsoftwire:
The 'TRIM()' function used here isn't present in the version of SQL Server used for the UAT DB, and its predecessors can only trim spaces not arbitrary characters.

### DIFF
--- a/DigitalLearningSolutions.Data/Services/NotificationPreferencesDataService.cs
+++ b/DigitalLearningSolutions.Data/Services/NotificationPreferencesDataService.cs
@@ -2,6 +2,7 @@
 {
     using System.Collections.Generic;
     using System.Data;
+    using System.Linq;
     using Dapper;
     using DigitalLearningSolutions.Data.Models;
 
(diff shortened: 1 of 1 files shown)


DanBloxham-sw (approved):

DigitalLearningSolutions.Data/Services/NotificationPreferencesDataService.cs
```diff
@@ -59,18 +65,23 @@ public IEnumerable<NotificationPreference> GetNotificationPreferencesForDelegate
...
                         ORDER BY n.NotificationID",
                 new { delegateId }
-            );
+            ).ToList();
+
+            foreach (var notificationPreference in preferences)
+            {
+                // the notification names in the DB have trailing whitespace
+                notificationPreference.NotificationName = notificationPreference.NotificationName.Trim();
+            }
```
stellake, line 82: Super minor, but could we extract this into a private RemoveTrailingSpace method to avoid duplication (especially because comments tend to get outdated pretty quickly)?

stellake (approved):
Looks good! 👍

PR #303, opened 2021-05-06
Heedls 407 single user no expandable

JBsoftwire:
When there is only one set of preferences, there is no expandable and no title.

image
image
image
image


DanBloxham-sw (approved):
Looks nice to me

DigitalLearningSolutions.Web/ViewModels/MyAccount/NotificationPreferencesViewModel.cs
```diff
@@ -10,25 +11,34 @@ public class NotificationPreferencesViewModel
...
-            AdminNotifications = new NotificationPreferenceExpanderViewModel(adminNotifications, UserTypes.Admin);
-            DelegateNotifications = new NotificationPreferenceExpanderViewModel(delegateNotifications, UserTypes.Delegate);
+            AdminNotifications = new NotificationPreferenceListViewModel(
+                adminNotifications,
+                UserTypes.Admin,
+                delegateNotifications.Any());
+            DelegateNotifications = new NotificationPreferenceListViewModel(
+                delegateNotifications,
+                UserTypes.Delegate,
+                adminNotifications.Any());
```
stellake, line 21: Could we write a small test for this functionality?

DigitalLearningSolutions.Web/Views/NotificationPreferences/_NotificationPreferenceListContainer.cshtml
```diff
@@ -0,0 +1,25 @@
+﻿@using DigitalLearningSolutions.Web.Helpers
```
stellake, line 1: Really minor thing, but if something is called container I'd expect it to take something to render inside it (e.g. like a layout does), whereas this seems to be a classic component. Maybe we could name it _UserNotificationPreferences.cshtml or something similar without a "Container" in it.

stellake (approved):
Looks good - just a minor comment about a missing empty line! No re-review needed though 👍

PR #304, opened 2021-05-06
HEEDLS-388 - Moved UserDataService to DataService folder and implemen…

DanBloxham-sw:
…ted inactive centre handling on log in

Tested:
- Ran all unit tests
- Manually tested in Chrome, IE11, Edge, Firefox
- Chrome mobile view
- No JS
- Screen reader

I'm wondering if the paragraph on the new CentreInactive page (screenshot below) should be split into two paragraphs (one for each sentence)? UPDATE: after the Show and Tell on Friday 7th of May, it was decided to keep the page as it currently is - so there is no need for any restyling or new content.

New CentreInactive error page:

Centre Inactive error page


davidm-m (approved):
All looks good to me!

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
```diff
@@ -342,5 +342,57 @@ public void Multiple_approved_accounts_at_same_centre_should_log_in()
...
+                .Returns((expectedAdminUser, expectedDelegateUsers));
+            var expectedApprovedDelegateUsers = expectedDelegateUsers.Where(du => du.CentreActive).ToList();
+
+            // When
+            controller.Index(LoginTestHelper.GetDefaultLoginViewModel());
+
+            // Then
+            A.CallTo(() => userService.GetUserCentres(
+                    null,
+                    A<List<DelegateUser>>.That.Matches(list => list.SequenceEqual(expectedApprovedDelegateUsers))))
```
stellake, line 369: We can use "IsSameSequenceAs" instead, which is a bit more concise and removes the need for the extra logic to get the expectedApprovedDelegateUsers (we'd have to pull userAtActiveCentre into a var though)
 
`A<List<DelegateUser>>.That.IsSameSequenceAs(userAtActiveCentre)))`

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
```diff
@@ -342,5 +342,57 @@ public void Multiple_approved_accounts_at_same_centre_should_log_in()
             result.Should().BeRedirectToActionResult()
                 .WithControllerName("Home").WithActionName("Index");
         }
+
+        [Test]
+        public void
+            When_user_has_accounts_with_inactive_centres_only_check_for_centres_that_are_active()
```
stellake, line 348: I think without context "check for centres" could mean anything in this title. I wonder if we could rename it such that it's more obvious what's happening.. Don't have any immediate good ideas though 😓

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -70,9 +70,19 @@ public IActionResult Index(LoginViewModel model)
                 loginService.GetVerifiedAdminUserAssociatedWithDelegateUser(verifiedDelegateUsers.First(),
                     model.Password);
 
-            var availableCentres = userService.GetUserCentres(verifiedAdminUser, approvedDelegateUsers);
+            var verifiedAdminUserWithActiveCentre = verifiedAdminUser?.CentreActive == true ? verifiedAdminUser : null;
+            var approvedDelegateUsersWithActiveCentre = approvedDelegateUsers.Where(du => du.CentreActive).ToList();
```
stellake, line 74: This feels like it should live in a service - maybe in userService?

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
```diff
@@ -342,5 +342,57 @@ public void Multiple_approved_accounts_at_same_centre_should_log_in()
...
+            A.CallTo(() => userService.GetUsersByUsername(A<string>._))
+                .Returns((expectedAdminUser, expectedDelegateUsers));
+            A.CallTo(() => loginService.VerifyUsers(A<string>._, A<AdminUser>._, A<List<DelegateUser>>._))
+                .Returns((expectedAdminUser, expectedDelegateUsers));
+            A.CallTo(() => userService.GetUserCentres(expectedAdminUser, expectedDelegateUsers))
+                .Returns(new List<CentreUserDetails>());
+
+            // When
+            var result =
+                controller.Index(LoginTestHelper.GetDefaultLoginViewModel());
```
stellake, line 392: Super minor, but should this be on a new line? (it doesn't seem like a particularly long line of code)

stellake (commented):
Looks good - I think I'd probably separate the active centre filtering into a service, especially because this controller is already doing quite a lot. It'll make testing that particular piece of functionality a bit neater too. :)

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
davidm-m: Something like `When_user_has_accounts_with_inactive_centres_can_only_login_to_active_centres`?

DigitalLearningSolutions.Data.Tests/Services/UserServiceTests.cs
```diff
@@ -119,7 +120,31 @@ public void GetUsersByUsername_with_admin_id_fetches_associated_delegate_users()
         }
 
         [Test]
-        public void GetAvailableCentres_returns_centres_correctly_ordered()
+        public void GetUsersWithActiveCentres_only_returns_users_with_active_centres()
```
stellake, line 123: I feel like we could take more advantage of moving this into a service as we could also test the case where the admin is returned when the account is for active centre (as currently all tests would pass even if adminUser that's returned is always null)

stellake (commented):
Looks good, just one test request. 🙏

DigitalLearningSolutions.Data.Tests/Services/UserServiceTests.cs
DanBloxham-sw: Nevermind on the improved readability question, I realised it was a lot easier than I anticipated for some reason. Made the change to individual parameters rather than a bool array.

stellake (approved):
Looks good! :)

PR #306, opened 2021-05-07
HEEDLS-396 delegate registration backend

davidm-m:
Added all of the backend functionality for delegate registration:
* Saving new users in the database using the stored procedure
* Adding a password hashing function and making sure passwords are only ever stored hashed
* Redirecting users to a confirmation page with their new candidate number after registration
* Sending an approval email to centre admins when a user registers for their centre

I've added tests to check
* Our hash password and verify password functions agree with each other
* An appropriate email is sent when a user registers
* A user's password is set correctly when they register

image
image


AlexJacksonDS (approved):
All looks fine, just got to sort out the merge conflict in Startup.cs

DigitalLearningSolutions.Web/Views/Register/Confirmation.cshtml
```diff
@@ -0,0 +1,19 @@
+﻿@using DigitalLearningSolutions.Web.ViewModels.Register
+@model ConfirmationViewModel
+
```
stellake, line 3: Do we need to set the title here so a tab has a correct title? Maybe "Registration Confirmation"?

DigitalLearningSolutions.Web/ViewModels/Register/SummaryViewModel.cs
```diff
@@ -5,20 +5,11 @@
 
     public class SummaryViewModel: IValidatableObject
     {
-        public SummaryViewModel(string firstName, string lastName, string email, string centre, string jobGroup)
```
stellake, line 8: Hmm.. Is there a reason we've removed the constructor and made all the fields optional?

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -138,29 +141,58 @@ public IActionResult Summary()
...
-                var data = TempData.Peek<DelegateRegistrationData>();
-                var viewModel = MapToSummary(data!);
+                var viewModel = MapToSummary(data);
                 viewModel.Terms = model.Terms;
                 return View(viewModel);
             }
 
-            // TODO HEEDLS-396 create new user and redirect to confirmation page
-            return View(model);
+            var baseUrl = ConfigHelper.GetAppConfig()["AppRootPath"];
```
stellake, line 152: I think we should use the "CurrentSystemBaseUrl" here to make sure the email links to the old system. We will change it once we rebuild the approval page :)

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests.cs
```diff
@@ -32,6 +32,10 @@ public void Dispose()
         [InlineData("/Login")]
         [InlineData("/ForgotPassword")]
         [InlineData("/Register")]
+        [InlineData("/Register/LearnerInformation")]
+        [InlineData("/Register/Password")]
+        [InlineData("/Register/Summary")]
+        [InlineData("/Register/Confirmation")]
```
stellake, line 38: Does this work without TempData? (it feels like this would just redirect the user)

I think we should simulate the registration process and then scan each page (a different test). I'm happy to have a call about this, but also, feel free to scan over the HEEDLS-NONE-auth-accessibility branch first commit to see how the login form is handled. Also if you comment out the chromeOptions.AddArgument("--headless");  line, you should be able to visualise what's happening in the browser when you run your tests with breakpoints.

DigitalLearningSolutions.Data/Models/Register/DelegateRegistrationModel.cs
```diff
@@ -0,0 +1,27 @@
+﻿namespace DigitalLearningSolutions.Data.Models.Register
+{
+    public class DelegateRegistrationModel
+    {
+        public DelegateRegistrationModel(string firstName, string lastName, string email, int centre, int jobGroup,
```
stellake, line 5: Ah, this is minor, but when the lines get so long that we need to split the params onto new lines, we should put each param onto a new line so it's easier to read.

DigitalLearningSolutions.Data/DataServices/RegistrationDataService.cs
```diff
@@ -0,0 +1,52 @@
...
+                Answer2 = "",
+                Answer3 = "",
+                Answer4 = "",
+                Answer5 = "",
+                Answer6 = "",
+                AliasID = "",
+                ExternalReg = 0,
+                SelfReg = 0,
+                NotifyDate = DateTime.Now,
+                Bulk = 1
```
stellake, line 43: Is this correct? I'd expect the "SelfReg" to be true when someone goes through the form. Also, do you know what the "Bulk" is (I can double check with Kevin if you're not sure)?

DigitalLearningSolutions.Data/DataServices/RegistrationDataService.cs
```diff
@@ -0,0 +1,52 @@
...
+                Answer1 = "",
+                Answer2 = "",
+                Answer3 = "",
+                Answer4 = "",
+                Answer5 = "",
+                Answer6 = "",
+                AliasID = "",
+                ExternalReg = 0,
+                SelfReg = 0,
+                NotifyDate = DateTime.Now,
```
stellake, line 42: Should this be UTC?

DigitalLearningSolutions.Data/DataServices/PasswordDataService.cs
```diff
@@ -0,0 +1,39 @@
...
+        public void SetPasswordByCandidateNumber(string candidateNumber, string passwordHash)
+        {
+            connection.Query(
+                @"UPDATE Candidates
+                        SET Password = @passwordHash
+                        WHERE CandidateNumber = @candidateNumber",
+                new { passwordHash, candidateNumber });
+        }
+
+        public void SetPasswordByCandidateId(int candidateId, string passwordHash)
```
stellake, line 30: Do we need both of these methods?

DigitalLearningSolutions.Web/ViewModels/Register/SummaryViewModel.cs
davidm-m: Models need to have a parameterless constructor, it's a limitation of MVC. When I tried to get around this (by providing a parameterless constructor as well as one with all the properties) there were still issues I would just have to overcomplicate things to fix so I decided it would be simpler to have a default constructor and use the object initialiser.

DigitalLearningSolutions.Data/DataServices/PasswordDataService.cs
davidm-m: I was making an assumption that when it comes to updating passwords elsewhere CandidateId would be a more useful identifier but that was probably too forward thinking. For this we only need the CandidateNumber one.

DigitalLearningSolutions.Data/DataServices/RegistrationDataService.cs
davidm-m: Yeah I actually meant to ask about those! SelfReg should be true, that makes sense, and yes I don't know what bulk is for.

stellake: I suspect "Bulk" is when the admins register delegates in bulk, but I'm not 100% sure. I will chat to Kevin about this tomorrow, but for now, could we keep SelfReg as 1 and Bulk as 0?

DigitalLearningSolutions.Web/ViewModels/Register/SummaryViewModel.cs
stellake: Hmm.. I'm pretty sure we have the parameterless + parameter constructors elsewhere (e.g. ForgotPasswordViewModel). What were the issues when you tried both?

DigitalLearningSolutions.Data.Tests/Services/CryptoServiceTests.cs
```diff
@@ -22,5 +22,17 @@ public void VerifyHashedPassword_Returns_false_if_hashed_password_is_empty()
             // Then
             Assert.IsFalse(passwordIsCorrect);
         }
+
+        [Test]
+        public void VerifyHashedPassword_Returns_true_for_hash_from_GetPasswordHash()
```
stellake, line 27: Nice test!

DigitalLearningSolutions.Data.Tests/Services/RegistrationServiceTests.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            // When
+            registrationService.RegisterDelegate(testRegistrationModel, "localhost");
+
+            // Then
+            A.CallTo(() =>
+                passwordDataService.SetPasswordByCandidateNumber(newCandidateNumber, passwordHash)
+            ).MustHaveHappened();
+        }
+    }
```
stellake, line 71: Could we also add a minor test for the RegisterDelegate method returning the new candidate number?

DigitalLearningSolutions.Web/ViewModels/Register/SummaryViewModel.cs
davidm-m: Ah, the errors I get are that it says all the other fields in the viewmodel are required even though the only we care about is if the terms and conditions box is ticked. Seeing that ForgotPasswordViewModel works makes me think it's trying to construct the viewmodel from what's being posted by the form. I could add hidden inputs for each of the other fields, though that still feels a bit hacky.

DigitalLearningSolutions.Data.Tests/DataServices/PasswordDataServiceTests.cs
```diff
@@ -0,0 +1,44 @@
...
+        public void Set_password_by_candidate_number_should_set_password_correctly()
+        {
+            using var transaction = new TransactionScope();
+            try
+            {
+                // When
+                var password = "hashedPassword";
+                var candidateNumber = "KW1";
+                passwordDataService.SetPasswordByCandidateNumber(candidateNumber, password);
+                var result = userDataService.GetDelegateUserById(1)?.Password;
```
davidm-m, line 33: Feels a bit cheeky relying on another service in a unit test... Any idea how to get around that?

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -138,29 +141,58 @@ public IActionResult Summary()
...
+            {
+                FirstName = data.RegisterViewModel.FirstName!,
+                LastName = data.RegisterViewModel.LastName!,
+                Email = data.RegisterViewModel.Email!,
+                Centre = centre!,
+                JobGroup = jobGroup!
+            };
+        }
+
+        private static DelegateRegistrationModel MapToDelegateRegistrationModel(DelegateRegistrationData data)
```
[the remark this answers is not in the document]
stellake: Ah, that's sad! Could we add a mapping helper for the registration journey and maybe do mapping to summary there too? I think these would be really nice to unit test outside the controller.

DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            this.registrationDataService = registrationDataService;
+            this.passwordDataService = passwordDataService;
+            this.emailService = emailService;
+            this.centresDataService = centresDataService;
+        }
+
+        public string RegisterDelegate(DelegateRegistrationModel delegateRegistrationModel, string baseUrl)
+        {
+            var candidateNumber = registrationDataService.RegisterDelegate(delegateRegistrationModel);
```
stellake, line 31: Are the following returns handled (from the stored procedure documentation)?

```
'-1' : unexpected error
'-2' : some parameters not supplied
'-3' : failed, AliasID not unique
'-4' : active candidate record for e-mail address already exists 
```

I guess -2 should never be the case, but I can see the rest potentially coming up.

DigitalLearningSolutions.Data/Models/Register/DelegateRegistrationModel.cs
```diff
@@ -0,0 +1,32 @@
...
+            string passwordHash)
+        {
+            FirstName = firstName;
+            LastName = lastName;
+            Email = email;
+            Centre = centre;
+            JobGroup = jobGroup;
+            PasswordHash = passwordHash;
+        }
+        public string FirstName { get; set; }
```
stellake, line 20: NIP, but missing an empty line above this

DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -0,0 +1,69 @@
...
+    using DigitalLearningSolutions.Data.DataServices;
+    using DigitalLearningSolutions.Data.Models.Email;
+    using DigitalLearningSolutions.Data.Models.Register;
+    using MimeKit;
+
+    public interface IRegistrationService
+    {
+        string RegisterDelegate(DelegateRegistrationModel delegateRegistrationModel, string baseUrl);
+    }
+    public class RegistrationService: IRegistrationService
```
stellake, line 13: NIP, but missing an empty line above this (the auto-formatter should get this, although not 100% about IntelliJ)

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests.cs
```diff
@@ -42,13 +42,55 @@ public void Page_has_no_accessibility_errors(string url)
...
+            var passwordForm = driver.FindElement(By.TagName("form"));
+            passwordForm.Submit();
+
+            var summaryResult = new AxeBuilder(driver).Analyze();
+
+            // then
+            registerResult.Violations.Should().BeEmpty();
+            learnerInformationResult.Violations.Should().BeEmpty();
+            passwordResult.Violations.Should().BeEmpty();
+            summaryResult.Violations.Should().BeEmpty();
```
davidm-m, line 86: This doesn't currently test the confirmation page as without having actually submitted this form trying to get to that page just redirects you.
With a function to call somewhere that runs the delete user stored procedure we could maybe test it but that feels outside the scope of this ticket - maybe I should add a todo here to test the confirmation page once we have a was to delete users.

stellake: I think it's ok as long as we run some accessibility checks on the confirmation page during manual testing 👍

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -138,30 +142,39 @@ public IActionResult Summary()
...
             }
 
-            // TODO HEEDLS-396 create new user and redirect to confirmation page
-            return View(model);
+            var baseUrl = ConfigHelper.GetAppConfig()["CurrentSystemBaseUrl"];
+            var candidateNumber = registrationService.RegisterDelegate(MappingHelper.MapToDelegateRegistrationModel(data), baseUrl);
+            if (candidateNumber == "-1")
+            {
+                Response.StatusCode = 500;
+                return View("../LearningSolutions/Error/UnknownError");
```
stellake, line 160: Could we redirect to the LearningSolutions/Error page instead of this (it has 500 error by default)

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -128,8 +130,10 @@ public IActionResult Password(PasswordViewModel model)
         [HttpGet]
         public IActionResult Summary()
         {
-            var data = TempData.Peek<DelegateRegistrationData>();
-            var viewModel = MapToSummary(data!);
+            var data = TempData.Peek<DelegateRegistrationData>()!;
+            var centre = centresDataService.GetCentreName((int)data.LearnerInformationViewModel.Centre!);
+            var jobGroup = jobGroupsDataService.GetJobGroupName((int)data.LearnerInformationViewModel.JobGroup!);
+            var viewModel = MappingHelper.MapToSummary(data, centre!, jobGroup!);
```
stellake, line 136: I wonder whether the MappingHelper should be renamed to RegistrationMappingHelper to be more specific..

davidm-m: It depends on whether we want to make a new Helper class if another controller needs to do similar things in the future or if we put them all in here. I sort of assumed the second (since it's a static class anyway) but happy to rename it if you want.

stellake: I'm worried that this class will get too big and hard to scan over if we just have one mapping helper per project (this has happened on previous projects I've worked on so keen to avoid it from the start)

stellake (approved):
Looks good - one comment about the mapper naming change + we still need some documentation that I'm happy to pick up if you're busy. No rereview required though! 👍

PR #307, opened 2021-05-07
Heedls 405 edit details profile picture

AlexJacksonDS:
Added the profile picture editor to the Edit Details screen. Added a FileInputViewComponent that is just a modified version of the TextInput so that I could have all the same error functionality. The rest of the image stuff is still in the main EditDetails view. I've added what I think are the allowed file types to the file input based on the old site, but I couldn't find anything for file size for the profile image. If we do need this we'll need to create another custom attribute to handle it.

Checked that the tab ordering makes sense, and that a screen reader reads it OK. Checked mobile layout in chrome devtools. Ran all unit tests

image

image

### DIFF
--- a/DigitalLearningSolutions.Data/Services/ImageResizeService.cs
+++ b/DigitalLearningSolutions.Data/Services/ImageResizeService.cs
@@ -1,0 +1,80 @@
+﻿namespace DigitalLearningSolutions.Data.Services
+{
+    using System;
+    using System.Drawing;
+    using System.Drawing.Drawing2D;
+    using System.Drawing.Imaging;
+    using System.IO;
+    using Microsoft.AspNetCore.Http;
+
+    public interface IImageResizeService
+    {
+        public byte[] ResizeProfilePicture(IFormFile formProfileImage);
+    }
+
+    public class ImageResizeService : IImageResizeService
+    {
+        public byte[] ResizeProfilePicture(IFormFile formProfileImage)
+        {
+            using var memoryStream = new MemoryStream();
+            formProfileImage.CopyTo(memoryStream);
+
+            return SquareImageFromMemoryStream(memoryStream, 300);
+        }
+
+        private byte[] SquareImageFromMemoryStream(MemoryStream memoryStream, int targetSideLength)
+        {
+            using var image = Image.FromStream(memoryStream);
+            var minSideLength = Math.Min(image.Height, image.Width);
+
+            using var squareImage = new Bitmap(minSideLength, minSideLength);
+            using var graphics = Graphics.FromImage(squareImage);
+            graphics.FillRectangle(new SolidBrush(Color.White), 0, 0, minSideLength, minSideLength);
+
+            // Calculate offsets as half the difference between the longer and shorter side
+            // This crops an equal amount from either side of the longer side of the source image
+            // so the resulting square image is centred in the source image
+            var topOffset = 0;
+            var leftOffset = 0;
+            if (image.Height > image.Width)
+            {
+                topOffset = (image.Height - image.Width) / 2;
+            }
+            else
+            {
+                leftOffset = (image.Width - image.Height) / 2;
+            }
+
+            graphics.DrawImage(image,
+                new Rectangle(0, 0, minSideLength, minSideLength),
+                new Rectangle(leftOffset, topOffset, image.Width - leftOffset * 2, image.Height - topOffset * 2),
+                GraphicsUnit.Pixel);
+
+            using var resizedImage = ResizeSquareImage(squareImage, targetSideLength);
+            using var result = new MemoryStream();
+            resizedImage.Save(result, ImageFormat.Jpeg);
+            return result.ToArray();
+        }
+
+        private Image ResizeSquareImage(Image image, int sideLength)
+        {
+            var destRect = new Rectangle(0, 0, sideLength, sideLength);
+            var destImage = new Bitmap(sideLength, sideLength);
+
+            destImage.SetResolution(image.HorizontalResolution, image.VerticalResolution);
+
+            using var graphics = Graphics.FromImage(destImage);
+            graphics.CompositingMode = CompositingMode.SourceCopy;
+            graphics.CompositingQuality = CompositingQuality.HighQuality;
+            graphics.InterpolationMode = InterpolationMode.HighQualityBicubic;
+            graphics.SmoothingMode = SmoothingMode.HighQuality;
+            graphics.PixelOffsetMode = PixelOffsetMode.HighQuality;
+
+            using var wrapMode = new ImageAttributes();
+            wrapMode.SetWrapMode(WrapMode.TileFlipXY);
+            graphics.DrawImage(image, destRect, 0, 0, image.Width, image.Height, GraphicsUnit.Pixel, wrapMode);
+
+            return destImage;
+        }
+    }
+}
--- a/DigitalLearningSolutions.Data/Services/ImageResizeService.cs
+++ b/DigitalLearningSolutions.Data/Services/ImageResizeService.cs
@@ -22,13 +22,25 @@
             return SquareImageFromMemoryStream(memoryStream, 300);
         }
 
-        private byte[] SquareImageFromMemoryStream(MemoryStream memoryStream, int targetSideLength)
+        private byte[] SquareImageFromMemoryStream(MemoryStream memoryStream, int targetSideLengthPx)
         {
             using var image = Image.FromStream(memoryStream);
+
+            using var squareImage = CropImageToCentredSquare(image);
+
+            using var resizedImage = ResizeSquareImage(squareImage, targetSideLengthPx);
+
+            using var result = new MemoryStream();
+            resizedImage.Save(result, ImageFormat.Jpeg);
+            return result.ToArray();
+        }
+
+        private Image CropImageToCentredSquare(Image image)
+        {
             var minSideLength = Math.Min(image.Height, image.Width);
 
-            using var squareImage = new Bitmap(minSideLength, minSideLength);
-            using var graphics = Graphics.FromImage(squareImage);
+            var returnSquareImage = new Bitmap(minSideLength, minSideLength);
+            using var graphics = Graphics.FromImage(returnSquareImage);
             graphics.FillRectangle(new SolidBrush(Color.White), 0, 0, minSideLength, minSideLength);
 
             // Calculate offsets as half the difference between the longer and shorter side
@@ -50,20 +62,17 @@
                 new Rectangle(leftOffset, topOffset, image.Width - leftOffset * 2, image.Height - topOffset * 2),
                 GraphicsUnit.Pixel);
 
-            using var resizedImage = ResizeSquareImage(squareImage, targetSideLength);
-            using var result = new MemoryStream();
-            resizedImage.Save(result, ImageFormat.Jpeg);
-            return result.ToArray();
+            return returnSquareImage;
         }
 
-        private Image ResizeSquareImage(Image image, int sideLength)
+        private Image ResizeSquareImage(Image image, int sideLengthPx)
         {
-            var destRect = new Rectangle(0, 0, sideLength, sideLength);
-            var destImage = new Bitmap(sideLength, sideLength);
+            var destRect = new Rectangle(0, 0, sideLengthPx, sideLengthPx);
+            var returnImage = new Bitmap(sideLengthPx, sideLengthPx);
 
-            destImage.SetResolution(image.HorizontalResolution, image.VerticalResolution);
+            returnImage.SetResolution(image.HorizontalResolution, image.VerticalResolution);
 
-            using var graphics = Graphics.FromImage(destImage);
+            using var graphics = Graphics.FromImage(returnImage);
             graphics.CompositingMode = CompositingMode.SourceCopy;
             graphics.CompositingQuality = CompositingQuality.HighQuality;
             graphics.InterpolationMode = InterpolationMode.HighQualityBicubic;
@@ -74,7 +83,7 @@
             wrapMode.SetWrapMode(WrapMode.TileFlipXY);
             graphics.DrawImage(image, destRect, 0, 0, image.Width, image.Height, GraphicsUnit.Pixel, wrapMode);
 
-            return destImage;
+            return returnImage;
         }
     }
 }
--- a/DigitalLearningSolutions.Web/Attributes/AllowedExtensionsAttribute.cs
+++ b/DigitalLearningSolutions.Web/Attributes/AllowedExtensionsAttribute.cs
@@ -1,0 +1,38 @@
+﻿namespace DigitalLearningSolutions.Web.Attributes
+{
+    using System.ComponentModel.DataAnnotations;
+    using System.IO;
+    using System.Linq;
+    using Microsoft.AspNetCore.Http;
+
+    public class AllowedExtensionsAttribute : ValidationAttribute
+    {
+        private readonly string[] extensions;
+        private readonly string? errorMessage;
+
+        public AllowedExtensionsAttribute(string[] allowedExtensions, string? errorMessage = null)
+        {
+            this.extensions = allowedExtensions;
+            this.errorMessage = errorMessage;
+        }
+
+        protected override ValidationResult IsValid(object value, ValidationContext validationContext)
+        {
+            if (value is IFormFile file)
+            {
+                var extension = Path.GetExtension(file.FileName);
+                if (!extensions.Contains(extension.ToLower()))
+                {
+                    return new ValidationResult(errorMessage ?? GetErrorMessage());
+                }
+            }
+            return ValidationResult.Success;
+        }
+
+        private string GetErrorMessage()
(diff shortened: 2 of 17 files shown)


DigitalLearningSolutions.Web/Attributes/AllowedExtensionsAttribute.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        public AllowedExtensionsAttribute(string[] allowedExtensions, string? errorMessage = null)
+        {
+            this.extensions = allowedExtensions;
+            this.errorMessage = errorMessage;
+        }
+
+        protected override ValidationResult IsValid(object value, ValidationContext validationContext)
+        {
+            if (value is IFormFile file)
```
davidm-m, line 21: I feel like if it isn't a file it should fail, but also that shouldn't ever come up.

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -49,23 +56,87 @@ public IActionResult EditDetails()
...
+            foreach (var key in ModelState.Keys.Where(k => k != "ProfilePicture"))
+            {
+                ModelState[key].Errors.Clear();
+                ModelState[key].ValidationState = ModelValidationState.Valid;
+            }
+
+            if (!ModelState.IsValid)
+            {
+                return View(model);
+            }
```
davidm-m, line 113: We've just ensured the ModelState is valid so why are we checking that here?

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -49,23 +56,87 @@ public IActionResult EditDetails()
...
+        private IActionResult EditDetailsPostSave(EditDetailsViewModel model)
         {
             if (!ModelState.IsValid)
             {
                 return View(model);
             }
 
+            if (model.ProfilePicture != null)
+            {
+                ModelState.AddModelError("ProfilePicture", "Preview your new profile picture before saving");
```
davidm-m, line 83: Would it be better to do this for them automatically (i.e. return `EditDetailsPostPreviewImage` with this error) so they don't have to click preview and submit the form again?

DigitalLearningSolutions.Web/Views/Shared/Components/TextInput/Default.cshtml
```diff
@@ -5,17 +5,23 @@
...
+    </div>
+  }
+  @if (Model.HasError)
+  {
     <span class="nhsuk-error-message" id="@Model.Name-error">
       <span class="nhsuk-u-visually-hidden">Error:</span> @Model.ErrorMessage
     </span>
   }
-  <input class="nhsuk-input nhsuk-u-width-one-half @(Model.HasError ? "nhsuk-input--error" : "")" 
+  <input class="nhsuk-input @Model.Class @(Model.HasError ? "nhsuk-input--error" : "")"
```
davidm-m, line 19: Have you updated any other places that use this ViewComponent so they're providing the class they expect?

davidm-m (commented):
A few small questions but overall looks good 😀
Any chance we could get a screenshot of what it looks like when there's an error (say if the file has the wrong extension)?

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
AlexJacksonDS: We've just cleared all errors except errors on the ProfilePicture property. This then throws it back out if the ProfilePicture is invalid (e.g. wrong file type)

AlexJacksonDS:
File format error message currently looks like this
image

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
davidm-m: Oh yeah, I just totally ignored the `k != "ProfilePicture"` bit 🤦‍♂️

DigitalLearningSolutions.Web/Views/Shared/Components/TextInput/Default.cshtml
AlexJacksonDS: Yes unless anyone else has started using them and currently has something in progress.

DigitalLearningSolutions.Web/Attributes/AllowedExtensionsAttribute.cs
AlexJacksonDS: I can't think of a way that would come up other than a developer putting this attribute on something that isn't a property for a file upload. I'll add some documentation to it to make it obvious.

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
stellake: Re UX, I think it makes sense not to auto-preview on submit, because otherwise the user will see the error message but also an image preview, which contradicts the error message. If I get a form error, I'd expect to take an action before submission based on the message. 🙏 

Also, HEE was happy with the behaviour, so I'm happy to leave it.

stellake:
UI wise, a few comments:
* The spacing around the border is a bit uneven around the Choose file input - also with error. (see image below)
* Minor but could we write “First name” over “First Name” etc for consistency with other areas of the system.
* Could we rename the page to “Edit details” rather than "My details" – this must have been missed in wireframes.

DigitalLearningSolutions.Data/Services/ImageResizeService.cs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        public byte[] ResizeProfilePicture(IFormFile formProfileImage)
+        {
+            using var memoryStream = new MemoryStream();
+            formProfileImage.CopyTo(memoryStream);
+
+            return SquareImageFromMemoryStream(memoryStream, 300);
+        }
+
+        private byte[] SquareImageFromMemoryStream(MemoryStream memoryStream, int targetSideLength)
```
stellake, line 25: Could we specify the units in this variable name - targetSideLengthInPx to make this a bit more concise. Then if someone wants to use this in the future, they don't need to dig into the implementation to find out :)

DigitalLearningSolutions.Data/Services/ImageResizeService.cs
```diff
@@ -0,0 +1,80 @@
...
+            using var resizedImage = ResizeSquareImage(squareImage, targetSideLength);
+            using var result = new MemoryStream();
+            resizedImage.Save(result, ImageFormat.Jpeg);
+            return result.ToArray();
+        }
+
+        private Image ResizeSquareImage(Image image, int sideLength)
+        {
+            var destRect = new Rectangle(0, 0, sideLength, sideLength);
+            var destImage = new Bitmap(sideLength, sideLength);
```
stellake, line 62: We're using the "using" keyword for the BitMap above, but not here - is this intentional?

AlexJacksonDS: Yes, it's the return value of the method, the call to the method has the using declaration. Otherwise it'll be disposed before it can be used.

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -49,23 +56,87 @@ public IActionResult EditDetails()
         }
 
         [HttpPost]
-        public IActionResult EditDetails(EditDetailsViewModel model)
+        public IActionResult EditDetails(EditDetailsViewModel model, string action)
+        {
+            switch (action)
```
stellake, line 61: My IDE recommends changing this to a switch expression, which I think reads much nicer :)

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -49,23 +56,83 @@ public IActionResult EditDetails()
...
+            foreach (var key in ModelState.Keys)
+            {
+                ModelState[key].Errors.Clear();
+                ModelState[key].ValidationState = ModelValidationState.Valid;
+            }
+            
+            ModelState.Remove("HasProfileImageBeenRemoved");
+            model.ProfilePicture = null;
+            model.NewProfileImage = null;
+            model.HasProfileImageBeenRemoved = true;
```
stellake, line 133: Do we need both "HasProfileImageBeenRemoved" and "ProfileImage" fields? Wouldn't ProfileImage be null if it's removed or do we have a different case to this?

DigitalLearningSolutions.Web/ViewModels/MyAccount/EditDetailsViewModel.cs
```diff
@@ -33,5 +36,14 @@ public class EditDetailsViewModel
         [DataType(DataType.Password)]
         [MaxLength(100, ErrorMessage = "Password must be at most 100 characters")]
         public string? Password { get; set; }
+
+        public byte[]? CurrentProfileImage { get; set; }
+
+        public byte[]? NewProfileImage { get; set; }
```
stellake, line 42: Do we need both the new and old profile image here?

DigitalLearningSolutions.Web/Views/MyAccount/EditDetails.cshtml
```diff
@@ -26,29 +28,81 @@
...
     <div class="nhsuk-grid-column-full">
-      <vc:text-input asp-for="Email" label="Email Address" type="text" spell-check="false" autocomplete="email" />
+      <div class="divider"></div>
     </div>
   </div>
 
   <div class="nhsuk-grid-row">
     <div class="nhsuk-grid-column-full">
-      <vc:text-input asp-for="Password" label="Confirm password to save changes:" type="password" spell-check="false" autocomplete="email" />
+      <vc:text-input asp-for="Password" label="Confirm password to save changes:" type="password" spell-check="false" autocomplete="" hint-text="" css-class="nhsuk-u-width-one-half" />
```
stellake, line 103: Ah, not a big fan of the empty hint-text etc fields.. Apparently it's only fixed in .NET 5 😢

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
AlexJacksonDS: Needed to distinguish between the 2 cases where NewProfileImage is null, when the user hasn't attempted to upload a new profile image before saving (they like their current one) and the case when the user has actively removed their profile picture.

All the logic here might be possible to refactor so I don't need the Current/New ProfileImages or this bool, but I'm not sure whether it will catch all cases easily

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -49,23 +56,83 @@ public IActionResult EditDetails()
...
+                ModelState[key].ValidationState = ModelValidationState.Valid;
+            }
+
+            if (!ModelState.IsValid)
+            {
+                return View(model);
+            }
+
+            // Remove the old new profile image value from the ModelState
+            ModelState.Remove("NewProfileImage");
```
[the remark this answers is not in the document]
AlexJacksonDS: Due to the way all the modelstate stuff works it needs to be removed before setting in the post otherwise it will keep its value if you go through this method a second time, e.g. if the crop was bad so you tried a different image, it'll still contain the old image and won't overwrite properly. I don't really understand why it behaves like this, but that's what I found. I can look to see if I can get one of these alternatives to work correctly.

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -49,23 +56,82 @@ public IActionResult EditDetails()
...
+            }
+
+            if (!ModelState.IsValid)
+            {
+                return View(model);
+            }
+
+            if (model.ProfilePicture != null)
+            {
+                ModelState.Remove("NewProfileImage");
```
stellake, line 113: Do we have the "NewProfileImage" field in the model anymore?

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
stellake: Oh, do you mean like a "Reset to old picture" logic? I suspect we want the user to always "Preview" if they've selected a new file, and if they haven't, they can click "Remove" or upload and preview a different file.
If we want the reset old picture functionality in the future, we can add it as future work, but it doesn't need to be handled now.

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
AlexJacksonDS: Ah whoops, problems with renaming and strings not being checked. Will update them all to be using nameof instead a string to prevent doing that again in the future. Might be something worth doing everywhere we are using these string property names (unless there is some reason you know not to use this)

DigitalLearningSolutions.Web/Views/MyAccount/EditDetails.cshtml
```diff
@@ -26,29 +28,64 @@
...
+        var hintText = @"To change your profile picture, select a new image and click the Preview button to preview it.
+                              If you upload a new profile image and do not preview it before saving, then your profile picture
+                              will not be updated.";
+      }
+      <input type="hidden" asp-for="ProfileImage" />
+      <vc:file-input asp-for="ProfilePicture" label="Profile picture (optional)" hint-text="@hintText" css-class="nhsuk-u-width-full" />
+    </div>
+    <div class="nhsuk-grid-column-one-half">
+      <input type="hidden" asp-for="HasProfileImageBeenRemoved" />
+      @if (!Model.HasProfileImageBeenRemoved)
```
stellake, line 65: Could we just move the Model.ProfileImage != null check here or is there a case where we need the empty profile-picture__image-box displayed? I actually wonder, whether having the default profile image as a placeholder like we do when we display the profile image would be useful for the users 🤔

DigitalLearningSolutions.Web/Views/Shared/Components/FileInput/Default.cshtml
```diff
@@ -0,0 +1,24 @@
...
+    <span class="nhsuk-error-message" id="@Model.Name-error">
+      <span class="nhsuk-u-visually-hidden">Error:</span> @Model.ErrorMessage
+    </span>
+  }
+  <input class="nhsuk-input @Model.Class @(Model.HasError ? "nhsuk-input--error" : "")" 
+         id="@Model.Id"
+         name="@Model.Name"
+         type="file"
+         aria-describedby="@Model.Name-error" />
+</div>
```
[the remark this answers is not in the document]
stellake: Fair enough!

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
AlexJacksonDS: Nope, the NewProfileImage (now renamed) byte[]? property is null until the user goes through the preview action when they want to submit a new image.
It is also null if they do nothing relating to images (i.e. keep old image)
And the third case, when the user clears the profile picture, it is also null.

We need to distinguish between the different cases when this property is null image removed vs image never set. i.e. remove profile picture and do nothing.

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -49,23 +56,82 @@ public IActionResult EditDetails()
...
+            {
+                ModelState.AddModelError("ProfilePicture", "Preview your new profile picture before saving");
+                return View(model);
+            }
+
             var userAdminId = User.GetAdminId();
             var userDelegateId = User.GetNullableCandidateId();
 
-            if (!userService.TryUpdateUserAccountDetails(userAdminId, userDelegateId, model.Password,  model.FirstName, model.LastName, model.Email))
+            var profileImageToSave = model.HasProfileImageBeenRemoved ? null : model.ProfileImage;
```
stellake, line 86: Do we not set the ProfileImage to null when we remove it? (so could the "profileImageToSave" be model.ProfileImage -- note that I could be missing something)

DigitalLearningSolutions.Web/Attributes/AllowedExtensionsAttribute.cs
stellake: I guess we could always only return Success if the allowed extensions contains the file extensions, but I don't mind too much given that we control where this is used. Will leave it up to you :)

stellake (approved):
Just a really minor comment about formatting, but otherwise looks great! Happy for you to merge without rereview 👍

DigitalLearningSolutions.Web/Attributes/AllowedExtensionsAttribute.cs
AlexJacksonDS: I think leaving it as it is is the best option. Currently if a developer puts it on the wrong property type, it just won't do anything, always returning Success because it isn't an IFormFile. If we change it to fail on non-IFormFile values and it gets put on the wrong type of property, it'll cause unexpected errors because the validation will always fail (which if it somehow made it through testing/release would cause issues).
I think the always success if incorrect property type is safer.

PR #310, opened 2021-05-07
HEEDLS-421: redirect to former page when logging in

JBsoftwire:
This uses the returnUrl implementation from https://github.com/TechnologyEnhancedLearning/DLSV2/commit/78d49feed4ec0ce443ec9b51c3ef811be1a16de7
and passes it through the login process, including login failures and choosing a centre.

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
@@ -14,6 +14,7 @@
     using FluentAssertions.AspNetCore.Mvc;
     using Microsoft.AspNetCore.Authentication;
     using Microsoft.AspNetCore.Http;
+    using Microsoft.Extensions.Logging;
     using NUnit.Framework;
 
     internal class LoginControllerTests
@@ -23,6 +24,7 @@
         private ILoginService loginService;
         private ISessionService sessionService;
         private IUserService userService;
+        private ILogger<LoginController> logger;
 
         [SetUp]
         public void SetUp()
@@ -30,8 +32,9 @@
             loginService = A.Fake<ILoginService>();
             userService = A.Fake<IUserService>();
             sessionService = A.Fake<ISessionService>();
+            logger = A.Fake<ILogger<LoginController>>();
 
-            controller = new LoginController(loginService, userService, sessionService)
+            controller = new LoginController(loginService, userService, sessionService, logger)
                 .WithDefaultContext()
                 .WithMockUser(false)
                 .WithMockTempData()
@@ -56,7 +59,7 @@
         public void Index_should_redirect_if_user_is_authenticated()
         {
             // Given
-            var controllerWithAuthenticatedUser = new LoginController(loginService, userService, sessionService)
+            var controllerWithAuthenticatedUser = new LoginController(loginService, userService, sessionService, logger)
                 .WithDefaultContext()
                 .WithMockUser(true);
 
@@ -108,9 +111,12 @@
                 .Returns(
                     new List<CentreUserDetails> { new CentreUserDetails(1, "Centre 1", true, true) });
 
+            var returnUrl = "/some/other/page";
+            var urlHelper = controller.Url;
+            A.CallTo(() => urlHelper.IsLocalUrl(returnUrl)).Returns(true);
+
             // When
             var loginViewModel = LoginTestHelper.GetDefaultLoginViewModel();
-            var returnUrl = "/some/other/page";
             loginViewModel.ReturnUrl = returnUrl;
             var result = controller.Index(loginViewModel);
 
@@ -134,9 +140,12 @@
                 .Returns(
                     new List<CentreUserDetails> { new CentreUserDetails(1, "Centre 1", true, true) });
 
+            var returnUrl = "www.suspicious.com";
+            var urlHelper = controller.Url;
+            A.CallTo(() => urlHelper.IsLocalUrl(returnUrl)).Returns(false);
+
             // When
             var loginViewModel = LoginTestHelper.GetDefaultLoginViewModel();
-            var returnUrl = "www.suspicious.com";
             loginViewModel.ReturnUrl = returnUrl;
             var result = controller.Index(loginViewModel);
 
--- a/DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
@@ -1,5 +1,6 @@
 ﻿namespace DigitalLearningSolutions.Web.Tests.Controllers.Login
 {
+    using System;
     using System.Collections.Generic;
     using System.Linq;
     using System.Security.Claims;
@@ -49,7 +50,7 @@
         public void Index_should_render_basic_form()
         {
             // When
-            var result = controller.Index("");
+            var result = controller.Index(string.Empty);
 
             // Then
             result.Should().BeViewResult().WithDefaultViewName();
@@ -64,7 +65,7 @@
                 .WithMockUser(true);
 
             // When
-            var result = controllerWithAuthenticatedUser.Index("");
+            var result = controllerWithAuthenticatedUser.Index(string.Empty);
 
             // Then
             result.Should().BeRedirectToActionResult()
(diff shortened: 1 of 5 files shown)


DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -176,6 +182,10 @@ bool rememberMe
                 IssuedUtc = DateTime.UtcNow
             };
             HttpContext.SignInAsync("Identity.Application", new ClaimsPrincipal(claimsIdentity), authProperties);
+            if (!string.IsNullOrEmpty(returnUrl))
+            {
+                return Redirect(returnUrl);
```
DanBloxham-sw, line 187: I think this might be susceptible to open-redirect attacks. Would it be possible to implement one of the methods found here: https://docs.microsoft.com/en-us/aspnet/core/security/preventing-open-redirects?view=aspnetcore-3.1#protecting-against-open-redirect-attacks  

I'm thinking that the IsLocalUrl method, with some additional logging as described in those docs would be useful, but am unsure if the logging is something we really need.

DanBloxham-sw: If we can get an alternative redirect in if it looks like an open-redirect attack, then it might be worth making a unit test for that situation.

JBsoftwire: I've added a fix for this, but it breaks the tests due to them needing a real UrlHelper. I haven't yet been able to work out how to initialise a new one so the ticket isn't complete.

DanBloxham-sw (approved):
I'm happy with this.

DigitalLearningSolutions.Web/ViewModels/Login/LoginViewModel.cs
```diff
@@ -9,6 +9,14 @@ public LoginViewModel()
             Username = string.Empty;
             Password = string.Empty;
             RememberMe = false;
+            ReturnUrl = string.Empty;
+        }
+        public LoginViewModel(string? returnUrl)
```
stellake, line 14: Super NIP, but missing an empty line above this (formatter should in theory catch this..).
Also, is there a reason we need to set the ReturnUrl in the constructor above given that it's nullable?

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -176,6 +185,14 @@ bool rememberMe
...
             HttpContext.SignInAsync("Identity.Application", new ClaimsPrincipal(claimsIdentity), authProperties);
+            if (!string.IsNullOrEmpty(returnUrl))
+            {
+                if (Url.IsLocalUrl(returnUrl))
+                {
+                    return Redirect(returnUrl);
+                }
+                logger.LogWarning($"Attempted login redirect to non-local returnUrl {returnUrl}");
+            }
             return RedirectToAction("Index", "Home");
```
stellake, line 196: Could we separate this return url handling into a private method that returns an optional redirect. Then the return statement in this method can be something like:
`return RedirectToReturnUrl(returnUrl) ?? RedirectToAction("Index", "Home");`

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -176,6 +185,14 @@ bool rememberMe
                 IssuedUtc = DateTime.UtcNow
             };
             HttpContext.SignInAsync("Identity.Application", new ClaimsPrincipal(claimsIdentity), authProperties);
+            if (!string.IsNullOrEmpty(returnUrl))
+            {
+                if (Url.IsLocalUrl(returnUrl))
+                {
+                    return Redirect(returnUrl);
+                }
+                logger.LogWarning($"Attempted login redirect to non-local returnUrl {returnUrl}");
```
stellake, line 194: Does this work ok with the old site (they're on the same baseUrl so should be ok)?

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
```diff
@@ -92,6 +95,65 @@ public void Successful_sign_in_should_render_home_page()
...
+            var result = controller.Index(loginViewModel);
+
+            // Then
+            result.Should().BeRedirectResult().WithUrl(returnUrl);
+        }
+
+        [Test]
+        public void Successful_sign_in_with_nonlocal_return_url_should_render_home_page()
+        {
+            //Given
```
stellake, line 130: Super NIP, but missing a space after "//"

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
```diff
@@ -92,6 +95,65 @@ public void Successful_sign_in_should_render_home_page()
...
+            A.CallTo(() => userService.GetUsersByUsername(A<string>._))
+                .Returns((expectedAdmin, expectedDelegates));
+            A.CallTo(() => loginService.VerifyUsers(A<string>._, A<AdminUser>._, A<List<DelegateUser>>._))
+                .Returns((expectedAdmin, expectedDelegates));
+            A.CallTo(() => userService.GetUsersWithActiveCentres(A<AdminUser>._, A<List<DelegateUser>>._))
+                .Returns((expectedAdmin, expectedDelegates));
+            A.CallTo(() => userService.GetUserCentres(A<AdminUser>._, A<List<DelegateUser>>._))
+                .Returns(
+                    new List<CentreUserDetails> { new CentreUserDetails(1, "Centre 1", true, true) });
+
```
stellake, line 113: I wonder if we could separate this into a private method `GivenSignInIsSuccessful` to make these two tests a bit easier to read.

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
```diff
@@ -56,20 +59,20 @@ public void Index_should_render_basic_form()
...
         {
             // Given
-            var controllerWithAuthenticatedUser = new LoginController(loginService, userService, sessionService)
+            var controllerWithAuthenticatedUser = new LoginController(loginService, userService, sessionService, logger)
                 .WithDefaultContext()
                 .WithMockUser(true);
 
             // When
-            var result = controllerWithAuthenticatedUser.Index();
+            var result = controllerWithAuthenticatedUser.Index("");
```
stellake, line 67: Is this required?

JBsoftwire: It complains when I don't do it, and a simple null is ambiguous between the GET and POST methods. I suppose `string.Empty` would look a little neater though.

DigitalLearningSolutions.Web/Controllers/LoginController.cs
JBsoftwire: I think so but we should test it on UAT as there may be some trouble depending on how the absolute URLs are constructed.

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
```diff
@@ -92,12 +88,68 @@ public void Successful_sign_in_should_render_home_page()
...
+            // When
+            controller.Index(LoginTestHelper.GetDefaultLoginViewModel());
+
+            // Then
+            A.CallTo(() => authenticationService.SignInAsync(
+                    A<HttpContext>._, A<string>._, A<ClaimsPrincipal>._, A<AuthenticationProperties>._))
+                .MustHaveHappened();
+        }
+
+        private void GivenSignInIsSuccessful(AdminUser expectedAdmin, List<DelegateUser> expectedDelegates)
```
stellake, line 151: Could we move the private method at the end of the file (first public methods, then private)? 🙏 

Also, do we need to keep the expectedAdmin & expectedDelegates in the test or can they be part of the GivenSignInIsSuccessful (I guess it would need a rename to admin/delegates)

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
stellake: You should be able to do this in the actual controller params, which will give the returnUrl a default value if not provided - (string? returnUrl = null)

Then we should be able to do Index() in these tests. :)

stellake (approved):
Looks good! 👍

PR #311, opened 2021-05-10
HEEDLS-373 - Centre Configuration Centre Details implemented

DanBloxham-sw:
NotifyEmail, BannerText, SignatureImage, and CentreLogo are now retrieved and displayed on the Centre Configuration Page.

Tested:
- Chrome, Firefox, Edge, IE 11
- Mobile view/zoom, No JS, Screen reader
- Ran all unit tests
- Tested on Chrome with a number of different images for different widths/heights/areas/aspect ratios to make sure they all look nice (see comment on ticket for IDs)

Few things of note:
There are no centres with a non-null SignatureImage. The CentreLogo is present in a lot of centres, but the dimensions vary wildly so I included some extra css to just display the image as is and downsize the image if necessary (the regular nhsuk-image__img class would always make the image fill the width, which means that some of the smaller images would just become big blurry messes)

The CentreLogo is sometimes not null, but is still an empty image - e.g. Test Centre with ID 101. This means it evades the current null checking and produces an image like in the final screenshot. Do we want a little extra logic in the GetCentreDetailsByID method to replace that with null when it is retrieved so nothing is displayed at all (to match the SignatureImage display above it)?

There is a problem with the NHS card component on mobile view - when the contents are a long string that cannot be wrapped onto multiple lines because of a lack of whitespace, then the string remains on a single line going off the side of the screen, extending the card with it (see screen shot below). This will be apparent in testing where some centre banner text is a long string of "xxxxxxxx" etc. It can also occur if the user has a long email address. As this behaviour can also happen on the NHS card component examples, I don't think we have to deal with it, but certainly worth bearing in mind. 

Screen shots below
Regular page:
Centre Config Centre Details

Mobile view:
Centre Config Centre Details Mobile view

Mobile view with large banner text:
Mobile view with massive banner text

Test Centre image:
Test Centre 101 non null image


AlexJacksonDS (approved):
Looks good to me. 

I do think it would be nice if the long string in the banner text didn't also break the layout of the other components. Although the likelihood of it happening with real text is probably fairly low.

I think displaying the broken image is better than hiding it at this point, it signifies something is wrong that could be fixed by uploading a new image (when we have the functionality), and doesn't disguise the fact that the broken image could be being displayed elsewhere.

stellake:
Re questions:
- Broken images -- I think this is perfect for asking in the external DLS channel to see what HEE thinks
- Text wrapping -- I think we should use https://www.w3schools.com/cssref/css3_pr_word-break.asp in these cards
- Centre signatures -- I will ask Kevin tomorrow morning if he has any examples that we can use for testing purposes.

UX wise looks great, just a minor thing - can we use "Centre name" rather than "Centre Name" etc regarding the capitalisation to be consistent with the rest of the stuff we've build already? 🙏

stellake (commented):
Looks good! Just a question about an override and a comment about the UI. 👍

DigitalLearningSolutions.Data.Tests/DataServices/CentresDataServiceTests.cs
```diff
@@ -101,11 +101,16 @@ public void GetCentreDetailsById_should_return_the_correct_values()
...
+                RegionName = "North West",
+                NotifyEmail = "notify@test.com",
+                BannerText = "xxxxxxxxxxxxxxxxxxxx",
+                SignatureImage = null,
+                CentreLogo = null
             };
 
             // When
             var result = centresDataService.GetCentreDetailsById(2);
+            result.CentreLogo = null;
```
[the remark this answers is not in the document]
DanBloxham-sw: (For context - the LogoServiceTests only check if a logo is null or not. So I could check that instead.)

DigitalLearningSolutions.Data/DataServices/CentresDataService.cs
```diff
@@ -64,15 +64,36 @@ FROM Centres
...
                         WHERE CentreID = @centreId",
                 new { centreId }
             );
+
+            if (centre?.CentreLogo?.Length < 10)
+            {
+                centre.CentreLogo = null;
+            }
+
+            if (centre?.CentreLogo == null)
```
stellake, line 89: Oh this was meant to be logged only if centre == null (if we can't find it per id) 🙏

DanBloxham-sw: Good point, think I was having one of those days yesterday. Fixed now.

stellake (approved):
Looks good!

PR #313, opened 2021-05-11
HEEDLS-426 Split the custom prompts option strings

AlexJacksonDS:
Changed the options returned from the CustomPromptsService from a nullable string to a List<string> and moved the DataService to the DataServices folder.

I've tried this against all centres in the database, and the only one that wasn't either successfully split or null (empty list) was centre ID 435 which has an F1Options with a single option in it. i.e. we shouldn't have any issues with unexpected characters not splitting

### DIFF
--- a/DigitalLearningSolutions.Data/Models/CustomPrompts/CustomPrompt.cs
+++ b/DigitalLearningSolutions.Data/Models/CustomPrompts/CustomPrompt.cs
@@ -1,16 +1,30 @@
 ﻿namespace DigitalLearningSolutions.Data.Models.CustomPrompts
 {
+    using System.Collections.Generic;
+
     public class CustomPrompt
     {
         public CustomPrompt(string? text, string? options, bool mandatory)
         {
             CustomPromptText = text;
-            Options = options;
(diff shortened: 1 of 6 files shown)


DigitalLearningSolutions.Data/Models/CustomPrompts/CustomPrompt.cs
```diff
@@ -1,16 +1,30 @@
...
+        private List<string> SplitOptionsString(string? options)
+        {
+            var optionsList = new List<string>();
+            if (options != null)
+            {
+                optionsList.AddRange(options.Split(new [] { '\r', '\n' },
+                    System.StringSplitOptions.RemoveEmptyEntries));
+            }
+
+            return optionsList;
```
stellake, line 23: I've noticed that the data test for this checks the default mapping, but I'm wondering if we want to write a few unit tests just for this model to make sure the mapping in the constructor works as expected? (e.g. options = null, one item in a list, and a case with several options)

stellake (approved):
Looks great! 👍

PR #316, opened 2021-05-12
HEEDLS-376 - Centre manager details added to centre config page

DanBloxham-sw:
Tested:
- Ran all unit tests
- Checked in Chrome, IE11, Firefox, Edge,
- Checked Mobile view/zoom/screenreader/no js

Screenshots of new expander on centre config page:

Centre Manager Details

Centre Manager Details Mobile view

### DIFF
--- a/DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/Index.cshtml
+++ b/DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/Index.cshtml
@@ -75,5 +75,52 @@
         </dl>
       </div>
     </details>
+
+    <details class="nhsuk-details nhsuk-expander bottom-margin-nhsuk-card">
+      <summary class="nhsuk-details__summary">
+        <span class="nhsuk-details__summary-text">
+          Centre manager details
+        </span>
+      </summary>
+      <div class="nhsuk-details__text">
+        <dl class="nhsuk-summary-list">
+          <div class="nhsuk-summary-list__row">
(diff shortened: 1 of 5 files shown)


AlexJacksonDS (approved):
All looks fine

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/Index.cshtml
```diff
@@ -75,5 +75,52 @@
         </dl>
       </div>
     </details>
+
+    <details class="nhsuk-details nhsuk-expander bottom-margin-nhsuk-card">
```
stellake, line 79: This is looking good! I'm wondering whether it's worth pulling the expandables into CentreConfiguration specific partials just to improve the readability of this file as it's getting quite long. So we can have a partial for the centre details and a partial for the centre manager details. What do you think? :)

DanBloxham-sw: I'm happy to make it use partials, but I think my personal preference is to leave it as it is. Reason being is that we won't get any benefits to functionality from a number of unique partials, and I think the readability is fine as it is easy enough to collapse the different details elements in visual studio (and probably Rider as well). But I'm happy either way with whatever your final decision is.

stellake: I'm happy with that 👍

stellake (approved):
Looks good! 👍

PR #317, opened 2021-05-12
HEEDLS-NONE Add comment suppressions to ReSharper errors

davidm-m:
There are a couple of places in code where we rely on actions with certain names existing on controllers in generic code (extensions, service filters etc.). ReSharper doesn't like us doing this because if those actions don't exist on a controller we might get a runtime error.
I've added comments suppressing those errors where they already exist in code because I got sick of seeing the squiggly lines in my IDE.

Tested by doing ordinary development with these changes for over a week, I'm just pushing it in case it's useful to anyone else.

### DIFF
--- a/DigitalLearningSolutions.Web/Views/Frameworks/Shared/_NavMenuItems.cshtml
+++ b/DigitalLearningSolutions.Web/Views/Frameworks/Shared/_NavMenuItems.cshtml
@@ -1,5 +1,6 @@
 ﻿@using DigitalLearningSolutions.Web.Extensions
 <li class="nhsuk-header__navigation-item @Html.IsSelected("FrameworksDashboard")">
+  @* ReSharper disable once Mvc.ActionNotResolved *@
   <a class="nhsuk-header__navigation-link" asp-action="MyFrameworks">
     My Frameworks
     <svg class="nhsuk-icon nhsuk-icon__chevron-right" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" aria-hidden="true">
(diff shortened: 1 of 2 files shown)


DanBloxham-sw (approved):
Looks fine to me.

DigitalLearningSolutions.Web/Views/Frameworks/Shared/_NavMenuItems.cshtml
```diff
@@ -1,5 +1,6 @@
 ﻿@using DigitalLearningSolutions.Web.Extensions
 <li class="nhsuk-header__navigation-item @Html.IsSelected("FrameworksDashboard")">
+  @* ReSharper disable once Mvc.ActionNotResolved *@
   <a class="nhsuk-header__navigation-link" asp-action="MyFrameworks">
```
stellake, line 4: Could we actually fix these to have no issues? I think asp-action="FrameworksDashboard" asp-controller="Frameworks" should work?

davidm-m: The point of the ServiceFilter is that it's generic and so can be used on any controller. That does mean it relies on there being an action called Index on any controller it's used with though

stellake: I see - can we keep the comment there but fix the frameworks in that case? :)

stellake (approved):
Looks good! :)

PR #318, opened 2021-05-12
Heedls 380 manage registration prompts

AlexJacksonDS:
Added a screen to display the currently set custom registration prompts and a new button link on the CentreConfiguration

Checked the layout works down to mobile. 

I've put a message in slack to confirm the text for the free text prompts, so will potentially have to update that.

image

image

image

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Helpers/CustomPromptsTestHelper.cs
+++ b/DigitalLearningSolutions.Data.Tests/Helpers/CustomPromptsTestHelper.cs
@@ -7,43 +7,11 @@
     {
         public static CentreCustomPrompts GetDefaultCentreCustomPrompts
         (
-            int centreId = 29,
-            CustomPrompt? customPrompt1 = null,
-            CustomPrompt? customPrompt2 = null,
-            CustomPrompt? customPrompt3 = null,
-            CustomPrompt? customPrompt4 = null,
-            CustomPrompt? customPrompt5 = null,
-            CustomPrompt? customPrompt6 = null
+            List<CustomPrompt> customPrompts,
+            int centreId = 29
         )
         {
-            var centreCustomPrompts = new CentreCustomPrompts(centreId, new List<CustomPrompt>());
-
-            if (customPrompt1 != null)
-            {
-                centreCustomPrompts.CustomPrompts.Add(customPrompt1);
-            }
-            if (customPrompt2 != null)
-            {
-                centreCustomPrompts.CustomPrompts.Add(customPrompt2);
-            }
-            if (customPrompt3 != null)
-            {
-                centreCustomPrompts.CustomPrompts.Add(customPrompt3);
-            }
-            if (customPrompt4 != null)
-            {
-                centreCustomPrompts.CustomPrompts.Add(customPrompt4);
-            }
-            if (customPrompt5 != null)
-            {
-                centreCustomPrompts.CustomPrompts.Add(customPrompt5);
-            }
-            if (customPrompt6 != null)
-            {
-                centreCustomPrompts.CustomPrompts.Add(customPrompt6);
-            }
-
-            return centreCustomPrompts;
+            return new CentreCustomPrompts(centreId, customPrompts);
         }
 
         public static CustomPrompt GetDefaultCustomPrompt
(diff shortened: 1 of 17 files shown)


DanBloxham-sw (approved):
I can't see any problems with this, but I am wondering if it is worth using a bullet point list for the options display - looking at the screenshot provided in the description I feel like it might look better, but it is still good as is.

stellake (commented):
Looks good in general - just a few comments! 

Re styling, I think we should change the spacing a little bit to group things nicely.
1. I think it would be nicer to have a bit more space after the main title given that the page has no description. Maybe worth checking what we do in the notification preferences page?

2. and 3. I feel like the answers should be grouped together so we should make sure that the spacing 3 is bigger than spacing 2 in the image. I also like Daniel's suggestion of using bullet points. I think we need it to be a list of items from accessibility perspective anyway. 😄 

image

DigitalLearningSolutions.Web/Styles/learningMenu/index.scss
```diff
@@ -18,17 +18,17 @@
...
+  }
+
+  &.complete {
+    @extend .nhsuk-tag--green;
+  }
+}
+
+hr.thick {
+  border-bottom-width: 6pt;
 }
```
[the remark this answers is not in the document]
AlexJacksonDS: I have no idea, line endings maybe?

AlexJacksonDS:
Refactored the CustomPromptsService to return lists and restyled.

image

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts.cshtml
```diff
@@ -0,0 +1,84 @@
...
+              </strong>
+            </div>
+          </div>
+        </summary>
+        <div class="nhsuk-details__text">
+          <div class="nhsuk-grid-row nhsuk-u-margin-bottom-3">
+            <div class="nhsuk-grid-column-full">
+              @if (customField.Options.Count != 0) {
+                <p class="nhsuk-u-font-weight-bold">Answers delegate can choose from:</p>
+                <ul class="">
```
stellake, line 35: Do we need the empty class? :)

DigitalLearningSolutions.Web/Styles/learningMenu/index.scss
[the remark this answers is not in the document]
stellake: Cool - looks like we've gone from LF to CRLF, which is correct! 👍

DigitalLearningSolutions.Data/Services/CustomPromptsService.cs
```diff
@@ -26,21 +31,138 @@ public CustomPromptsService(ICustomPromptsDataService customPromptsDataService)
...
+                result.CustomField6Mandatory);
+            if (prompt6 != null)
+            {
+                list.Add(prompt6);
+            }
+
+            return list;
+        }
+
+        private List<CustomPromptWithAnswer> PopulateCustomPromptWithAnswerListFromCentreCustomPromptsResult(CentreCustomPromptsResult? result, DelegateUser delegateUser)
```
stellake, line 104: Ah, shame there isn't a nicer way of doing all this! One minor thing - I think the populate methods can be made static :)

DigitalLearningSolutions.Data.Tests/Helpers/CustomPromptsTestHelper.cs
```diff
@@ -15,26 +16,98 @@ public static CentreCustomPrompts GetDefaultCentreCustomPrompts
             CustomPrompt? customPrompt6 = null
```
stellake, line 16: I feel like this should accept a list of CustomPrompt. Then when called we could do something like:
var customPrompts = new List<CustomPrompt>() { expectedPrompt1, expectedPrompt2 };
GetDefaultCentreCustomPrompts(centreId, customPropmts)

DigitalLearningSolutions.Data.Tests/Services/CustomPromptsServiceTests.cs
```diff
@@ -25,8 +24,8 @@ public void Setup()
         public void GetCustomPromptsForCentreByCentreId_Returns_Populated_CentreCustomPrompts()
         {
             // Given
-            var expectedPrompt1 = CustomPromptsTestHelper.GetDefaultCustomPrompt(options: null, mandatory: true);
-            var expectedPrompt2 = CustomPromptsTestHelper.GetDefaultCustomPrompt(text: "Department / team", options: null, mandatory: true);
+            var expectedPrompt1 = CustomPromptsTestHelper.GetDefaultCustomPrompt(1,options: null, mandatory: true);
+            var expectedPrompt2 = CustomPromptsTestHelper.GetDefaultCustomPrompt(2,text: "Department / team", options: null, mandatory: true);
```
stellake, line 28: NIP, but missing a space after "," - should be picked up by a formatter 👍

DigitalLearningSolutions.Data/Services/CustomPromptsService.cs
```diff
@@ -26,21 +31,138 @@ public CustomPromptsService(ICustomPromptsDataService customPromptsDataService)
 
             var result = customPromptsDataService.GetCentreCustomPromptsByCentreId(centreId.Value);
 
-            return new CentreCustomPrompts
+            return new CentreCustomPrompts(result.CentreId, PopulateCustomPromptListFromCentreCustomPromptsResult(result));
+        }
+
+        public CentreCustomPromptsWithAnswers? GetCentreCustomPromptsWithAnswersByCentreIdAndDelegateUser(int? centreId,
```
stellake, line 37: I wonder if we want to ever handle the optional cases - I'm wondering whether accepting int centreId and DelegateUser would be better as then we wouldn't have to return/handle nullables everywhere. If the centreId doesn't exist on the claims, then it's ok to redirect the user to unexpected error page. If we want to be extra sure, we could always check for non-null centre id in the centre admin claim policy..

stellake (approved):
Looks good! :)

PR #319, opened 2021-05-12
Heedls 369 reset password frontend

SteveJacksonSoft:
Credits to David for most of the work.


DigitalLearningSolutions.Web/Views/Register/Password.cshtml
```diff
@@ -1,4 +1,4 @@
-﻿@using DigitalLearningSolutions.Web.ViewModels.Common
+@using DigitalLearningSolutions.Web.ViewModels.Common
```
SteveJacksonSoft, line 1: Not sure what happened here. Byte Order Mark? It's not a line ending change anyway.

SteveJacksonSoft: Yeah, looks like the byte order mark's been removed.

DigitalLearningSolutions.Web/Views/ResetPassword/Index.cshtml
```diff
@@ -1 +1,60 @@
...
+
+@{
+  var errorHasOccurred = !ViewData.ModelState.IsValid;
+  ViewData["Title"] = errorHasOccurred ? "Error: Reset Your Password": "Reset Your Password";
+  var passwordError = ViewData.ModelState["Password"]?.Errors?.Count > 0;
+  var confirmPasswordError = ViewData.ModelState["ConfirmPassword"]?.Errors?.Count > 0;
+  var passwordInputErrorClass = passwordError ? "nhsuk-select--error" : "";
+  var passwordFormErrorClass = passwordError ? "nhsuk-form-group--error" : "";
+  var confirmPasswordInputErrorClass = confirmPasswordError ? "nhsuk-select--error" : "";
+  var confirmPasswordFormErrorClass = confirmPasswordError ? "nhsuk-form-group--error" : "";
```
davidm-m, line 12: You can use the new TextInputViewComponent to avoid needing all these variables and most of the boilerplate for the inputs themselves.

davidm-m (approved):
LGTM! 🥳

DigitalLearningSolutions.Web/Views/ResetPassword/Index.cshtml
SteveJacksonSoft: Funky, I'll look into it, thanks.

DigitalLearningSolutions.Web/Views/MyAccount/EditDetails.cshtml
```diff
@@ -1,86 +1,86 @@
...
+    </div>
+  </div>
+
+  <div class="hidden-submit">
+    <button name="action" class="nhsuk-button" value="save">Save</button>
+  </div>
+
+  <div class="nhsuk-grid-row">
+    <div class="nhsuk-grid-column-full">
+      <vc:text-input asp-for="FirstName" label="First name" value="@Model.FirstName" type="text" spell-check="false" autocomplete="given-name" hint-text="" css-class="nhsuk-u-width-one-half"/>
```
DanBloxham-sw, line 32: Not quite sure why the value attribute has been added here. Is it something to do with the merge conflicts?

DigitalLearningSolutions.Web/Views/ResetPassword/Index.cshtml
```diff
@@ -1 +1,45 @@
...
+  var errorHasOccurred = !ViewData.ModelState.IsValid;
+  ViewData["Title"] = errorHasOccurred ? "Error: Reset Your Password" : "Reset Your Password";
+}
+
+<div class="nhsuk-grid-row">
+  <div class="nhsuk-grid-column-full">
+    <form method="post" asp-action="Index">
+
+      @if (errorHasOccurred) {
+        <partial name="_ErrorSummary"/>
```
DanBloxham-sw, line 14: I think it would be worth replacing this with the new error summary view component. Would probably look something like this:
<vc:error-summary order-of-property-names="@(new[] { nameof(Model.Password), nameof(Model.ConfirmPassword) })"/>

DanBloxham-sw (commented):
Looks mostly good, just a comment about the error summary, and some merge conflicts need dealing with.

DigitalLearningSolutions.Web/Views/MyAccount/EditDetails.cshtml
AlexJacksonDS: Looking back at the commits, I think it is to stop values that shouldn't be getting returned after posting, e.g. the password field. 

I suspect the merge conflict currently is just the line ending, I have no idea why we keep getting these, but it must be either myself or Steve at this point. I have git configured as Steve said in slack (git config core.autocrlf = false), but since these are all files that I created, something is clearly going wrong.

That being said, there may be future merge conflicts that one of us will have to deal with. I am in the process of adding another field to this page that uses the same text input component.

SteveJacksonSoft: @AlexJacksonDS Spot on! It's because the value was being set explicitly in each text-input. The extra field allows you to choose whether the value gets set.
By the way Alex, is your IDE set to use CRLF line endings as well?

AlexJacksonDS: I believe so, new files I created in the last couple of days seem to be CRLF. I have no idea why some that I have previously created aren't. I'm using VS, which I think should be using windows default

SteveJacksonSoft: But I've decided it would improve it to use a flag instead and reduce data handling duplication.

DanBloxham-sw (approved):
I'm happy with this.

stellake (approved):
Looks good! 👍

PR #320, opened 2021-05-12
Heedls 359 application selector portal redirect

JBsoftwire:
This is a partial review for HEEDLS-359 that covers the redirection of delegate-only accounts to the Learning Portal, but not changes to the navigation bar.

### DIFF
--- a/DigitalLearningSolutions.Web/Attributes/DelegateOnlyInaccessibleAttribute.cs
+++ b/DigitalLearningSolutions.Web/Attributes/DelegateOnlyInaccessibleAttribute.cs
@@ -6,13 +6,13 @@
     using Microsoft.AspNetCore.Mvc.Filters;
 
     [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method)]
-    public class AdminOnlyAttribute : Attribute, IAuthorizationFilter
+    public class DelegateOnlyInaccessibleAttribute : Attribute, IAuthorizationFilter
     {
         public void OnAuthorization(AuthorizationFilterContext context)
         {
             var user = context.HttpContext.User;
 
-            if (!user.GetAdminId().HasValue)
+            if (user.Identity.IsAuthenticated && !user.GetAdminId().HasValue)
             {
                 context.Result = new RedirectToActionResult("Current", "LearningPortal", new {});
             }
--- a/DigitalLearningSolutions.Web/Controllers/ApplicationSelectorController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/ApplicationSelectorController.cs
@@ -1,5 +1,6 @@
 ﻿namespace DigitalLearningSolutions.Web.Controllers
 {
+    using DigitalLearningSolutions.Web.Attributes;
     using DigitalLearningSolutions.Web.Helpers;
     using DigitalLearningSolutions.Web.ViewModels.ApplicationSelector;
     using Microsoft.AspNetCore.Authorization;
@@ -8,6 +9,7 @@
     public class ApplicationSelectorController : Controller
     {
         [Authorize]
+        [DelegateOnlyInaccessible]
         public IActionResult Index()
         {
             var learningPortalAccess = User.GetCustomClaimAsBool(CustomClaimTypes.LearnUserAuthenticated) ?? false;
--- a/DigitalLearningSolutions.Web/Attributes/AdminOnlyAttribute.cs
+++ b/DigitalLearningSolutions.Web/Attributes/AdminOnlyAttribute.cs
@@ -1,0 +1,21 @@
+﻿namespace DigitalLearningSolutions.Web.Attributes
+{
+    using System;
+    using DigitalLearningSolutions.Web.Helpers;
+    using Microsoft.AspNetCore.Mvc;
+    using Microsoft.AspNetCore.Mvc.Filters;
+
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method)]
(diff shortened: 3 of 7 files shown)


DigitalLearningSolutions.Web/Attributes/DelegateOnlyInaccessibleAttribute.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace DigitalLearningSolutions.Web.Attributes
+{
+    using System;
+    using DigitalLearningSolutions.Web.Helpers;
+    using Microsoft.AspNetCore.Mvc;
+    using Microsoft.AspNetCore.Mvc.Filters;
+
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method)]
+    public class DelegateOnlyInaccessibleAttribute : Attribute, IAuthorizationFilter
```
DanBloxham-sw, line 9: We have custom policies for the Authorize attributes that redirect to the Log In page, which each policy requiring different claims to allow access to the page (e.g. UserCentreAdminOnly policy used on the TrackingSystem/CentreConfiguration controller). Is it possible to have this written as a custom policy so we have all that logic together? I suspect the policies do not allow specific redirects based on the policy so it might not be feasible - if that's the case then this is fine.

DigitalLearningSolutions.Web/Attributes/DelegateOnlyInaccessibleAttribute.cs
```diff
@@ -0,0 +1,21 @@
...
+    using Microsoft.AspNetCore.Mvc.Filters;
+
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method)]
+    public class DelegateOnlyInaccessibleAttribute : Attribute, IAuthorizationFilter
+    {
+        public void OnAuthorization(AuthorizationFilterContext context)
+        {
+            var user = context.HttpContext.User;
+
+            if (user.Identity.IsAuthenticated && !user.GetAdminId().HasValue)
```
DanBloxham-sw, line 15: This is not particularly important because I think this combination of IsAuthenticated and no AdminId is sufficient, but I'm wondering if we could check the claim LearnUserAuthenticated. This is a bool that is true if the user is logged into a delegate account. Because it is not mutually exclusive with the AdminId, we'd still need to check both. So functionally it wouldn't change but it feels better to me to check LearnUserAuthenticated and AdminId. But it shouldn't make a difference functionally, so I'm happy with either.

DigitalLearningSolutions.Web/Attributes/DelegateOnlyInaccessibleAttribute.cs
JBsoftwire: That the custom policy doesn't allow a different redirect is the conclusion Stella and I came to when we looked at this possibility before, but I'll take a specific look.

DigitalLearningSolutions.Web/Attributes/DelegateOnlyInaccessibleAttribute.cs
JBsoftwire: I think it's better to check it so that it's more robust to errors in the login/account system, otherwise we may risk e.g. an infinite redirect loop.

DigitalLearningSolutions.Web/Attributes/DelegateOnlyInaccessibleAttribute.cs
JBsoftwire: I've not found any indication this can be done simply with a policy.

DigitalLearningSolutions.Web/Attributes/DelegateOnlyInaccessibleAttribute.cs
JBsoftwire: I'm going to fix this in the review for the navigation bar changes, as it's low-priority and to do otherwise would require some annoying git surgery.

DanBloxham-sw (approved):
I'm happy with this provided the LearnUserAuthenticated check is implemented with the nav bar changes.

DigitalLearningSolutions.Web/Attributes/DelegateOnlyInaccessibleAttribute.cs
stellake: I think checking for the learnUserAuthenticated sounds good - just to make sure they don't get redirected back from the Learning Portal :)

DigitalLearningSolutions.Web/Controllers/ApplicationSelectorController.cs
```diff
@@ -8,6 +9,7 @@
     public class ApplicationSelectorController : Controller
     {
         [Authorize]
+        [DelegateOnlyInaccessible]
```
stellake, line 12: I'm now thinking whether a better name would be [RedirectDelegateOnlyToLearningPortal] - it's a bit clunkier, but I think makes it clearer what's happening without having to look at the attribute code. What do you think? 😄

JBsoftwire: I think that would be the most clear. Do you mind if I change this as part of the navigation bar review as with the learnUserAuthenticated check?

stellake: Don't mind as it's quite small and easy to remember - will approve this now :)

stellake (approved):

PR #322, opened 2021-05-12
HEEDLS-400 Added optional centreid parameter to Find Your Centre page

davidm-m:
Passes through a centreid parameter on the Find Your Centre page to the iframe, so we can load it already focussed on a particular centre if necessary. This isn't the whole ticket but other tickets rely on this working, so I'm putting it up and getting it merged early.

Tested in Firefox, Chrome, Edge and IE

### DIFF
--- a/DigitalLearningSolutions.Web/Controllers/FindYourCentreController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/FindYourCentreController.cs
@@ -1,12 +1,14 @@
 ﻿namespace DigitalLearningSolutions.Web.Controllers
 {
+    using DigitalLearningSolutions.Web.ViewModels.FindYourCentre;
     using Microsoft.AspNetCore.Mvc;
 
     public class FindYourCentreController : Controller
     {
-        public IActionResult Index()
+        public IActionResult Index(string centreId)
         {
-            return View();
+            var model = new FindYourCentreViewModel(centreId);
+            return View(model);
         }
     }
 }
--- a/DigitalLearningSolutions.Web/Controllers/FindYourCentreController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/FindYourCentreController.cs
@@ -5,9 +5,9 @@
 
     public class FindYourCentreController : Controller
     {
-        public IActionResult Index(string centreId)
+        public IActionResult Index(string? centreId)
         {
-            var model = new FindYourCentreViewModel(centreId);
+            var model = centreId == null ? new FindYourCentreViewModel() : new FindYourCentreViewModel(centreId);
             return View(model);
         }
     }
--- a/DigitalLearningSolutions.Web/ViewModels/FindYourCentre/FindYourCentreViewModel.cs
+++ b/DigitalLearningSolutions.Web/ViewModels/FindYourCentre/FindYourCentreViewModel.cs
@@ -1,14 +1,19 @@
 namespace DigitalLearningSolutions.Web.ViewModels.FindYourCentre
 {
+    using DigitalLearningSolutions.Web.Helpers;
+
     public class FindYourCentreViewModel
     {
-        public string CentreId { get; set; }
+        public string Url { get; set; }
 
-        public FindYourCentreViewModel() { }
+        public FindYourCentreViewModel()
+        {
+            Url = ConfigHelper.GetAppConfig()["CurrentSystemBaseUrl"] + "/findyourcentre?nonav=true";
+        }
 
         public FindYourCentreViewModel(string centreId)
         {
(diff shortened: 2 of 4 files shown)


DigitalLearningSolutions.Web/Views/FindYourCentre/Index.cshtml
```diff
@@ -1,7 +1,10 @@
 ﻿@using DigitalLearningSolutions.Web.Helpers
+@using DigitalLearningSolutions.Web.ViewModels.FindYourCentre
+@model FindYourCentreViewModel
 @{
   ViewData["Title"] = "Find Your Centre";
-  var url = ConfigHelper.GetAppConfig()["CurrentSystemBaseUrl"] + "/findyourcentre?nonav=true";
+  var url = "https://www.dls.nhs.uk/dev-prev" + "/findyourcentre?nonav=true";
```
DanBloxham-sw, line 6: Is there any particular reason that the base url has been changed from using a config value to specifically https://www.dls.nhs.uk/dev-prev?

davidm-m: Whoops! That wasn't mean to be committed

DanBloxham-sw (approved):
That's looking good now.

DigitalLearningSolutions.Web/Views/FindYourCentre/Index.cshtml
```diff
@@ -1,7 +1,10 @@
 ﻿@using DigitalLearningSolutions.Web.Helpers
+@using DigitalLearningSolutions.Web.ViewModels.FindYourCentre
+@model FindYourCentreViewModel
 @{
   ViewData["Title"] = "Find Your Centre";
   var url = ConfigHelper.GetAppConfig()["CurrentSystemBaseUrl"] + "/findyourcentre?nonav=true";
```
stellake, line 6: Super minor, but could we set this url in the viewmodel so we can easily unit test the url construction and move the logic away from the view? :)

DigitalLearningSolutions.Web/Controllers/FindYourCentreController.cs
```diff
@@ -1,12 +1,14 @@
 ﻿namespace DigitalLearningSolutions.Web.Controllers
 {
+    using DigitalLearningSolutions.Web.ViewModels.FindYourCentre;
     using Microsoft.AspNetCore.Mvc;
 
     public class FindYourCentreController : Controller
     {
-        public IActionResult Index()
+        public IActionResult Index(string centreId)
```
stellake, line 8: Could this be an optional query string?

DigitalLearningSolutions.Web/Views/FindYourCentre/Index.cshtml
```diff
@@ -1,9 +1,10 @@
-﻿@using DigitalLearningSolutions.Web.Helpers
```
stellake, line 1: Ah, super minor, but can you make sure you have all these changes: (copied from Steve's MR)

```
# Configuring text editor
To get useful git diffs and make code reviewing easier, it's helpful to have the same line separator and file encoding settings on our text editors.

Line separators should be CRLF (and git auto-CRLF should be turned off).
File encoding should be UTF-8 with BOM.

In JetBrains Rider these can be set in `Settings > Editor > File Encodings` (UTF-8, with BOM) and `Settings > Editor > Code Style` (CRLF).
```

DigitalLearningSolutions.Web/ViewModels/FindYourCentre/FindYourCentreViewModel.cs
```diff
@@ -0,0 +1,19 @@
...
+    public class FindYourCentreViewModel
+    {
+        public string Url { get; set; }
+
+        public FindYourCentreViewModel()
+        {
+            Url = ConfigHelper.GetAppConfig()["CurrentSystemBaseUrl"] + "/findyourcentre?nonav=true";
+        }
+
+        public FindYourCentreViewModel(string centreId)
```
stellake, line 14: I would have probably added a string? centreId constructor to avoid the null check in the controller, but I don't think it matters or makes anything less difficult to understand :)

stellake (approved):
Looks good + nice tests! 👍

DigitalLearningSolutions.Web/Views/FindYourCentre/Index.cshtml
davidm-m: I do, the reason the diffs are all weird here is that these files were all set to LF. They should be CRLF now 🙂

PR #324, opened 2021-05-13
HEEDLS-400 add banner to confirmation page

davidm-m:
Adds some extra info to the registration confirmation page if the user isn't automatically approved.

Tested in Firefox, Chrome, Edge and IE including at mobile widths for FF and Chrome. There's a slight issue that the anonymised banner text for many centres is one long string of x's without any spaces, which can mess up the formatting at low widths, but that shouldn't happen in production.

Screenshots:
Desktop
image

Mobile
image

Formatting going weird because of all the x's
image


DigitalLearningSolutions.Data/DataServices/RegistrationDataService.cs
```diff
@@ -29,7 +31,7 @@ public string RegisterDelegate(DelegateRegistrationModel delegateRegistrationMod
                 CentreID = delegateRegistrationModel.Centre,
                 JobGroupID = delegateRegistrationModel.JobGroup,
                 Active = 1,
-                Approved = 0,
+                Approved = approved ? 1 : 0,
```
AlexJacksonDS, line 34: Since this is always false at the moment I assume this is something that will actually do something in the future?

AlexJacksonDS (commented):
In the centre configuration details we are using word-break: break-word; to deal with the long banner text. I agree it's unlikely in production but we should probably be consistent.

Other than that and a question looks good.

DigitalLearningSolutions.Data/DataServices/RegistrationDataService.cs
davidm-m: Yeah, HEEDLS-402 covers adding cases where a new user is automatically approved.

DigitalLearningSolutions.Web/Views/Shared/Components/CentreInfo/Default.cshtml
```diff
@@ -0,0 +1,7 @@
+@using DigitalLearningSolutions.Web.ViewModels.Common
```
davidm-m, line 1: I'm worried I might be the cause of the LF issues we're having, since I made this file and it, CentreInfoViewComponent and CentreInfoViewModel apparently had LF line endings. They've all got CRLF now.
Interestingly the latest file I made, centreInfo.scss has the correct line endings without me doing anything.

AlexJacksonDS (approved):
Looks good, jjust need to sort out the merge conflicts

DigitalLearningSolutions.Web/Views/Shared/Components/CentreInfo/Default.cshtml
AlexJacksonDS: Having files as only LF shouldn't cause issues as long as we aren't changing them backwards and forwards each time. Sounds very similar to some of the issue I've been having, some of the files I've created are also LF and I have no idea why.

After looking at some of them that I was having issues with yesterday, it seems like merging was partially causing some of the issues. An LF file from master being merged into a local CRLF was resulting in mixed line endings. we're just going to keep an eye on them and try and make them consistent where possible (e.g. I'm going to go through and change a load of them before merging 404 after it has been reviewed so the review isn't confusing as to what has changed)

DigitalLearningSolutions.Data/DataServices/RegistrationDataService.cs
```diff
@@ -19,8 +19,10 @@ public RegistrationDataService(IDbConnection connection)
             this.connection = connection;
         }
 
-        public string RegisterDelegate(DelegateRegistrationModel delegateRegistrationModel)
+        public (string candidateNumber, bool approved) RegisterDelegate(DelegateRegistrationModel delegateRegistrationModel)
         {
+            var approved = false;
```
stellake, line 24: Could the approved flag be taken from the delegateRegistrationModel - I think Dan has added it in his MR, but yours will be merged first? (you could just set it to false in the constructor)

https://github.com/TechnologyEnhancedLearning/DLSV2/pull/339/files

+ then we don't need to return approved from the dataService (as that flag is specified in the service layer and is not dependent on creating a delegate entry in the db)

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -153,27 +153,43 @@ public IActionResult Summary(SummaryViewModel model)
...
             }
 
-            var viewModel = new ConfirmationViewModel(candidateNumber);
-            return View(viewModel);
+            var approved = (bool)approvedNullable;
+            var centreId = (int)centreIdNullable;
+
+            if (!approved)
+            {
+                var bannerText = centresDataService.GetBannerText(centreId);
```
stellake, line 184: Should we get the bannerText as part of the new VC for centre details (VCs can have DI, e.g. LogoViewComponent)?

DigitalLearningSolutions.Web/Views/Shared/Components/CentreInfo/Default.cshtml
```diff
@@ -0,0 +1,7 @@
+@using DigitalLearningSolutions.Web.ViewModels.Common
+@model CentreInfoViewModel
+
+<link rel="stylesheet" href="@Url.Content("~/css/shared/centreInfo.css")" asp-append-version="true">
+
+<p class="nhsuk-body-l">@Model.BannerText</p>
+<vc:action-link asp-controller="FindYourCentre" asp-action="Index"  asp-all-route-data="@Model.Routes" link-text="More information" />
```
stellake, line 7: Super minor, but could we name this "CentreContactInfo" in various places to be consistent with the "BannerText" rename in the admin UI, which will now be "Centre contact info". 🙏

stellake: Also, NIP, but the asp-all-route-data has two spaces in front of it 👀

DigitalLearningSolutions.Web/ViewModels/Common/CentreInfoViewModel.cs
```diff
@@ -0,0 +1,19 @@
...
+        {
+            Routes = new Dictionary<string, string>
+            {
+                { "centreId", centreId.ToString() }
+            };
+            BannerText = bannerText;
+        }
+
+        public Dictionary<string, string> Routes { get; set; }
+        public string BannerText { get; set; }
```
stellake, line 17: I think BannerText is optional in the DB at the moment, so we should only render the bannerText if it exists, otherwise just show the "More information" text. 
Also, can we post a pic of this in the external channel so HEE can come up with alternative wording (I think they requested it at the Show&Tell but I might misremember)

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
davidm-m: Ooh nice catch, I should have thought of that!

DigitalLearningSolutions.Web/ViewModels/Common/CentreInfoViewModel.cs
davidm-m: Nope you're right, they did ask - I'll do that.

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -153,27 +153,42 @@ public IActionResult Summary(SummaryViewModel model)
...
+            if (!approved)
+            {
+                var viewModel = new ConfirmationViewModel(candidateNumber, approved, centreId);
+                return View(viewModel);
+            }
+            else
+            {
+                var viewModel = new ConfirmationViewModel(candidateNumber, approved);
+                return View(viewModel);
+            }
```
stellake, line 191: I think this could probably be simplified to:
```
var centreIdForContactInformation = approved ? null : centreId;
var viewModel = new ConfirmationViewModel(candidateNumber, approved, centreIdForContactInformation);
return View(viewModel);
```

stellake (approved):
Looks good - just one suggestion on reducing duplication! No need for rereview though 👍

PR #326, opened 2021-05-13
HEEDLS-378 - Added Centre Content to centre config page with link to …

DanBloxham-sw:
…relevant FindYourCentre page

Note that the link to the FindYourCentre page won't do anything with the centreId until the changes for HEEDLS-400 are merged.

- Ran all unit tests
- Tested on Chrome, IE11, Firefox, Edge
- Tested with screenreader/zoom/mobile view/no JS

Screenshots of UI:
Centre Content

Centre Content Mobile view 1
Centre Content Mobile view 2


DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/_CentreStringArray.cshtml
```diff
@@ -0,0 +1,11 @@
+﻿@model string[]
+@if (Model.Length == 1) {
```
DanBloxham-sw, line 2: The reason that this partial determines behaviour based on the number of entries is this:
The OrganisationsCovered and TrainingVenues that use this partial to be displayed are stored in the database as a single string. They were populated on the old site by a textarea without any constraints on formatting, so users entered whatever they thought was sensible. 

This means you sometimes get multiple organisations/venues that are separated on new lines (which is dealt with when populating the view model) and sometimes you get multiple organisations/venues written as a sentence on a single line separated by commas ("and" or "&" sometimes separates values, there could be more that I haven't noticed). We can't separate the values apart based on these, as some single organisations/venues have commas etc. in their regular name or as part of an address.

As such, if we can separate the values based on new lines, then we shall display them on new lines in a list. If it is just a single value that we cannot separate, then we just display the text. If we displayed the single items in a list, then the screen reader will read "List with 1 item: [value]" which would be confusing if that value actually contained multiple organisations/venues separated by commas. So this logic makes sure no list is used if we can't separate the values.

Rather annoyingly, there are a few TrainingLocations that are given as an address, where each address line is separated onto a new line, and so will be displayed in a list. I think this is acceptable, as I think they are all single addresses.

AlexJacksonDS (approved):
Looks OK to me. The spacing on the mobile screen shot is a bit off compared to the similar buttons on the MyAccount page. The styling for this is in myAccount.scss, but I've moved it to index.scss in the pull request for 380.

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/CentreConfigurationViewModel.cs
```diff
@@ -28,5 +38,20 @@ public CentreConfigurationViewModel(Centre centre)
...
+        public string? CentreTelephone { get; set; }
+        public string? CentreEmail { get; set; }
+        public string? CentrePostcode { get; set; }
+        public string? OpeningHours { get; set; }
+        public string? CentreWebAddress { get; set; }
+        public string[] OrganisationsCovered { get; set; }
+        public string[] TrainingVenues { get; set; }
+        public string? OtherInformation { get; set; }
+
+        private string[] SplitMultiValuedStringIntoArray(string? multiValuedString)
```
stellake, line 50: Nice! 💯  Can we also write a few test cases for this?

stellake (commented):
Looks good, but agree with Alex on the buttons looking a bit odd on mobile! Hopefully we can just steal the styling from the My Account page 👀

DigitalLearningSolutions.Web.Tests/ViewModels/TrackingSystem/CentreConfigurationViewModelTests.cs
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        [Test]
+        public void CentreConfigurationViewModel_should_split_training_venues_on_new_lines()
+        {
+            // When
+            var centre = CentreTestHelper.GetDefaultCentre(
+                trainingVenues: "Address Line 1\r\nAddress Line 2\r\nAddress Line 3\r\nAddress Line 4");
+            var viewModel = new CentreConfigurationViewModel(centre);
+            var expectedTrainingCentres = new[] { "Address Line 1", "Address Line 2","Address Line 3","Address Line 4" };
```
stellake, line 54: Nice! Super NIP, but do you mind running the formatter on this file before merging 🙏

stellake (approved):
Looks good - just a comment about running the formatter :)

No rereview needed though. 👍

PR #327, opened 2021-05-14
HEEDLS-359: Extend changes to MyAccount child pages, add application …

JBsoftwire:
Some quick fixes to extend the nav bar fixes to other pages and add application names


stellake (approved):
Looks good! 👍

PR #328, opened 2021-05-14
HEEDLS-420 Accessibility scan for authenticated pages

stellake:
Tested on Jenkins with a broken page to check we get expected errors on failed scans. 👍

### DIFF
--- a/DigitalLearningSolutions.Web.AutomatedUiTests/SeleniumServerFactory.cs
+++ b/DigitalLearningSolutions.Web.AutomatedUiTests/SeleniumServerFactory.cs
@@ -1,18 +1,15 @@
-﻿namespace DigitalLearningSolutions.Web.AutomatedUiTests
+namespace DigitalLearningSolutions.Web.AutomatedUiTests
 {
     using System;
-    using System.Data;
+    using System.IO;
     using System.Linq;
-    using Dapper.FluentMap;
-    using DigitalLearningSolutions.Web.Helpers;
     using Microsoft.AspNetCore;
     using Microsoft.AspNetCore.Hosting;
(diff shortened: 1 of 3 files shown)


DigitalLearningSolutions.Web.AutomatedUiTests/SeleniumServerFactory.cs
```diff
@@ -38,11 +36,22 @@ protected sealed override TestServer CreateServer(IWebHostBuilder builder)
...
+                .UseSerilog()
+                .ConfigureAppConfiguration(configBuilder =>
+                {
+                    var jsonConfigSources = configBuilder.Sources
+                        .Where(source => source.GetType() == typeof(JsonConfigurationSource))
+                        .ToList();
+
+                    foreach (var jsonConfigSource in jsonConfigSources)
+                    {
+                        configBuilder.Sources.Remove(jsonConfigSource);
```
JBsoftwire, line 50: I take it this is removing config information to be replaced with settings from the SIT.json file. Does it not automatically override? Or have I misunderstood what's going on here?

This shouldn't block merging if its works, either way.

JBsoftwire (approved):

DigitalLearningSolutions.Web.AutomatedUiTests/SeleniumServerFactory.cs
stellake: Yep - this is what is happening. I created a new config for the app we're running in the server during SITs, where default db connection string is the test db connection string. This resolves the Jenkins issues, but also ensures that developers don't accidentally use the normal db connection for SITs. Hopefully having the SIT config separate makes it really clear how the SITs are configured in case we have issues in the future.

PR #329, opened 2021-05-17
HEEDLS-377 - edit centre manger details implemented and some detail l…

DanBloxham-sw:
…ist styling commonised

- Tested on chrome, IE11, Edge, Firefox
- Tested No JS, screen reader, mobile view/zoom
- Ran all unit tests.

Screen shots:
Regular page:
Edit page

With errors:
Edit page with errors

Mobile:
Edit page mobile view


DigitalLearningSolutions.Web/Styles/myAccount/myAccount.scss
```diff
@@ -1,15 +1,5 @@
 ﻿@import "~nhsuk-frontend/packages/core/all";
 
-.account-details-list {
```
DanBloxham-sw, line 3: Moved to index.scss and renamed to details-list-with-button

DigitalLearningSolutions.Web/Styles/myAccount/myAccount.scss
```diff
@@ -19,12 +9,6 @@
   height: 150px;
 }
 
-.left-button-mobile-margin-bottom {
```
DanBloxham-sw, line 22: This was already moved to index.scss in another ticket, but it was not removed from here, so I removed it now.

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/CentreManagerDetailsViewModel.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        public CentreManagerDetailsViewModel(Centre centre)
+        {
+            FirstName = centre.ContactForename;
+            LastName = centre.ContactSurname;
+            Email = centre.ContactEmail;
+            Telephone = centre.ContactTelephone;
+        }
+
+        [Required(ErrorMessage = "Please enter a first name.")]
```
AlexJacksonDS, line 18: It's not major as I think there will be another ticket to go though and check all of these, but I think we are going with the format 'Enter a first name' rather than 'please'

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/CentreManagerDetails.cshtml
```diff
@@ -0,0 +1,39 @@
+﻿@using DigitalLearningSolutions.Web.ViewModels.TrackingSystem
+@model CentreManagerDetailsViewModel
+
+@{
+  var errorHasOccurred = !ViewData.ModelState.IsValid;
+  ViewData["Title"] = errorHasOccurred ? "Error: Centre Manager Details" : "Centre Manager Details";
```
AlexJacksonDS, line 6: Do we want this capitalised like this?

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/CentreManagerDetails.cshtml
```diff
@@ -0,0 +1,39 @@
...
+      <vc:text-input asp-for="LastName" label="Last name" type="text" spell-check="false" autocomplete="family-name" hint-text="" css-class="nhsuk-u-width-one-half"/>
+
+      <vc:text-input asp-for="Email" label="Email" type="text" spell-check="false" autocomplete="email" hint-text="" css-class="nhsuk-u-width-one-half"/>
+
+      <vc:text-input asp-for="Telephone" label="Telephone (optional)" type="text" spell-check="false" autocomplete="" hint-text="" css-class="nhsuk-u-width-one-half"/>
+
+      <button class="nhsuk-button nhsuk-u-margin-top-4" type="submit" value="save">Save</button>
+    </form>
+
+    <div class="nhsuk-back-link">
```
AlexJacksonDS, line 29: There is a back link view component that you can use instead of this (unless for some reason this one is different)

DigitalLearningSolutions.Web/Styles/myAccount/myAccount.scss
AlexJacksonDS: Missed that, thanks

DigitalLearningSolutions.Data.Tests/DataServices/CentresDataServiceTests.cs
```diff
@@ -125,5 +126,33 @@ public void GetCentreName_should_return_null_when_the_centre_does_not_exist()
...
+                var lastName = "TestLastName";
+                var email = "test@email.com";
+                var telephone = "0123456789";
+
+                // When
+                centresDataService.UpdateCentreManagerDetails(2, firstName, lastName, email, telephone);
+                var updatedCentre = centresDataService.GetCentreDetailsById(2);
+
+                // Then
+                updatedCentre.ContactForename.Should().BeEquivalentTo(firstName);
```
AlexJacksonDS, line 147: Might be worth wrapping these in an AssertionScope so it checks them all on failure

AlexJacksonDS (commented):
A few small comments, looks mostly fine though. 

Can you check the line endings in CentreConfigurationController, I assume they are now consistent, but I've been having issues with them (it's another file I've touched that is showing up as all having changed)

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/CentreManagerDetails.cshtml
DanBloxham-sw: Looks like a vast majority of the ViewData titles have the first letter of each word capitalised. There are only a few titles where we only have the first first letter capitalised, and those tend to be in the recent additions of my own where I got confused between the capitalisation styles of the Title and the body. So I think we should prioritise consistency and keep the first letter of each word in titles capitalised. I'll probably have to go back in future to change the others at a later date.

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/CentreManagerDetails.cshtml
DanBloxham-sw: This one is mildly different as it uses "Cancel" as the link text rather than "Go Back". I was contemplating modifying the back link component to take in some link text in the model. Originally, I decided not to as:
- Putting the link text as an optional parameter wouldn't work as tag helpers don't seem to understand optional parameters ( https://github.com/dotnet/aspnetcore/issues/5011 ). So whenever the Go Back component was used, we'd have to specify the text.
- I think the vast majority of cases will probably just be "Go Back", so it will be more simple to not have to specify the text.

Though now I am working on other centre detail editing tickets, I see they will also want a Cancel back link, so I don't know if making a new Cancel Link view component (that would basically be identical to the back link, except for the text) would be preferred, or modifying the back link to take in link text would be easier.

DanBloxham-sw:
Just checked the line endings in CentreConfigurationController - they are all consistently CRLF.

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/CentreManagerDetails.cshtml
AlexJacksonDS: OK, the NHS back link is the styling around the text, so if you do find you want to make a component, I think changing the existing one to accept a parameter would probably be better. They require the parameters to be specified, but there is nothing stopping us defining it such that if you enter no text for the parameter (i.e. text="") then it would default to Go Back

AlexJacksonDS (approved):

DigitalLearningSolutions.Data/DataServices/CentresDataService.cs
```diff
@@ -1,122 +1,150 @@
...
+                    ContactForename = @firstName,
+                    ContactSurname = @lastName,
+                    ContactEmail = @email,
+                    ContactTelephone = @telephone
+                WHERE CentreId = @centreId",
+                new {firstName, lastName, email, telephone, centreId}
+            );
+        }
+        
+        public (string firstName, string lastName, string email) GetContactInfo(int centreId)
```
stellake, line 139: Not your change, but I wonder if we want to rename this method to "GetCentreManagerDetails" to be consistent with the update?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfigurationController.cs
```diff
@@ -1,46 +1,80 @@
...
+        [Route("CentreManagerDetails/Edit")]
+        public IActionResult EditCentreManagerDetails()
+        {
+            var centreId = User.GetCentreId();
+
+            var centreDetails = centresDataService.GetCentreDetailsById(centreId);
+
+            var model = new CentreManagerDetailsViewModel(centreDetails);
+
+            return View("CentreManagerDetails", model);
```
stellake, line 60: Minor, but should the view & viewmodel have "Edit" in the name?

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/CentreManagerDetails.cshtml
stellake: I think we should either create a new VC with "Cancel" text as it will be used in loads of places, or we can modify the back button to take text as a parameter (either Cancel or Go back - I think having a default might be confusing if text is set to "")

DigitalLearningSolutions.Data.Tests/DataServices/CentresDataServiceTests.cs
```diff
@@ -2,10 +2,12 @@
 {
```
stellake, line 2: Could we add the Centre configuration + Edit manager detail pages to the automated accessibility tests?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfigurationController.cs
DanBloxham-sw: I was conflicted on this. 

- On one hand, there is nothing stopping this being about any general centre manager details. We also have only one page (currently) that uses it, so its quite obvious what it links to, despite not matching the route name explicitly, and I doubt we will have any other pages about these details.
- One the other, it is currently not conforming to route name, and other edit pages use the Edit prefix.

Seeing how the route is /CentreManagerDetails/Edit, should the name change to CentreManagerDetailsEditView... or do we want EditCentreManagerDetailsView... ?

DigitalLearningSolutions.Data/DataServices/CentresDataService.cs
DanBloxham-sw: I think that would be a good change. But I could also remove this, and make the calls to this use GetCentreDetailsById instead, but that would retrieve more data from the DB that wouldn't get used. Thoughts?

stellake: I think GetCentreManagerDetails is good for now to avoid too big of a refactor. 👍

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfigurationController.cs
stellake: If it makes sense to change the route to /EditCentreManagerDetails to make the namings a bit nicer/more consistent, I'd go for that. 👍

If we ever need to  use more generic CentreManagerDetails components, we should put them into Shared folders, but this is not very likely to happen in the near future (based on all the future work I've seen so far) so probably YAGNI

DigitalLearningSolutions.Web/ViewComponents/BackLinkViewComponent.cs
```diff
@@ -5,9 +5,13 @@
...
     public class BackLinkViewComponent : ViewComponent
     {
-        public IViewComponentResult Invoke(string aspController, string aspAction)
+        public IViewComponentResult Invoke(string aspController, string aspAction, string linkText)
         {
-            return View(new LinkViewModel(aspController, aspAction));
+            if (string.IsNullOrWhiteSpace(linkText))
+            {
+                linkText = "Go back";
+            }
```
stellake, line 13: If the user specifies it as linkText="" then they've explicitly made the decision to have no text at the component, so overriding it can be really confusing behaviour. And the linkText will have to be specified in the HTML, because of how MVC works, so I'd be tempted to remove the default.

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/EditCentreManagerDetails.cshtml
```diff
@@ -0,0 +1,32 @@
...
+  ViewData["Title"] = errorHasOccurred ? "Error: Centre Manager Details" : "Centre Manager Details";
+}
+
+<div class="nhsuk-grid-row">
+  <div class="nhsuk-grid-column-full">
+    @if (errorHasOccurred) {
+      <vc:error-summary order-of-property-names="@(new[] { nameof(Model.FirstName), nameof(Model.LastName), nameof(Model.Email), nameof(Model.Telephone) })"/>
+    }
+
+    <h1 id="page-heading" class="nhsuk-heading-xl">Centre manager details</h1>
```
stellake, line 15: Super minor, but could we change the h1 + title to "Edit XXX"?

stellake (approved):
Just one comment about the title of the page, otherwise looks good! No need for rereview 👍

PR #330, opened 2021-05-17
Heedls 404 edit details delegate prompts

AlexJacksonDS:
Added the Job group dropdown and centre custom prompts to the edit details screen.

I suspect there may be some bit of logic, e.g. the validation of the custom fields that we might want to move into a common place for use in multiple controllers, but I'm not sure exactly what would need moving/how it would work, so it is in the place it is needed for now.

Checked the layout works OK in desktop and mobile and ran all unit tests.

image

image


AlexJacksonDS:
Also, again, I have a bunch of messing line endings showing as changes. They should all be CRLF now. I'm hoping this will be the last pull request where I have this issue.

DigitalLearningSolutions.Web/Views/MyAccount/EditDetails.cshtml
```diff
@@ -54,11 +54,33 @@
...
+    </div>
+
+    foreach (EditCustomFieldViewModel customField in ViewBag.CustomFields) {
+      <div class="nhsuk-grid-row @(ViewBag.CustomFields.IndexOf(customField) == ViewBag.CustomFields.Count - 1 ? "divider" : "")">
+        <div class="nhsuk-grid-column-full">
+          @if (customField.Options.Any())
+          {
+            <vc:select-list asp-for="@("Answer" + @customField.CustomFieldId)" label="@(customField.CustomPrompt + (customField.Mandatory ? "" : " (optional)"))" value="@customField.Answer" hint-text="" css-class="nhsuk-u-width-one-half" default-option="Select a @customField.CustomPrompt.ToLower()" select-list-options="@customField.Options"></vc:select-list>
+          } else {
+            <vc:text-input asp-for="@("Answer" + @customField.CustomFieldId)" label="@(customField.CustomPrompt + (customField.Mandatory ? "" : " (optional)"))" value="@customField.Answer" type="text" spell-check="false" autocomplete="" hint-text="" css-class="nhsuk-u-width-one-half" />
```
AlexJacksonDS, line 72: 'value' doesn't exist on the Text Input component yet, it's added as part of the front-end changes for 369. I've tested that this still works fine, it just doesn't do anything for now.

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -69,23 +84,45 @@ public IActionResult EditDetails(EditDetailsViewModel model, string action)
 
         private IActionResult EditDetailsPostSave(EditDetailsViewModel model)
         {
+            ValidateCustomPrompts(model);
+
             if (!ModelState.IsValid)
             {
+                ViewBag.JobGroupOptions = GetJobGroupItems(model.JobGroupId);
```
DanBloxham-sw, line 91: Feels like there is a lot of setting of ViewBag.JobGroupOptions and CustomFields as it is done each time a different problem occurs. Is there a way of commonising this, or perhaps using something similar to ViewBag so the values are retained? If it is too big an issue, don't worry about it.

DanBloxham-sw (commented):
Number of relatively minor comments/questions, but mostly looks good.

DanBloxham-sw (approved):
Looks good to me.

DigitalLearningSolutions.Web/Views/MyAccount/EditDetails.cshtml
stellake: This change should be merged to master now (I think Steve changed it from value to a flag though)

DigitalLearningSolutions.Data.Tests/Helpers/CustomPromptsTestHelper.cs
```diff
@@ -1,96 +1,96 @@
-﻿namespace DigitalLearningSolutions.Data.Tests.Helpers
-{
-    using System.Collections.Generic;
-    using DigitalLearningSolutions.Data.Models.CustomPrompts;
-
-    public static class CustomPromptsTestHelper
-    {
-        public static CentreCustomPrompts GetDefaultCentreCustomPrompts
+﻿namespace DigitalLearningSolutions.Data.Tests.Helpers
```
stellake, line 1: Could we change this back to CRLF + other affected files? Resharper's Cleanup Code should do the magic.. :)

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -83,7 +102,21 @@ private IActionResult EditDetailsPostSave(EditDetailsViewModel model)
...
+                model.Email!,
+                model.ProfileImage,
+                User.GetCentreId(),
+                model.JobGroupId,
+                model.Answer1,
+                model.Answer2,
+                model.Answer3,
+                model.Answer4,
+                model.Answer5,
+                model.Answer6))
```
stellake, line 119: This is getting pretty long and hard to read now. Could we create a class for holding this information that TryUpdateUserAccountDetails is called with?

DigitalLearningSolutions.Web/Helpers/CustomPromptHelper.cs
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        public List<EditCustomFieldViewModel> GetCustomFieldViewModelsForCentre(
+            int centreId,
+            string? answer1,
+            string? answer2,
+            string? answer3,
+            string? answer4,
+            string? answer5,
+            string? answer6)
```
stellake, line 24: Could we pass in delegateUser rather than all the answers here and then do the mapping to answers on L26?

DigitalLearningSolutions.Web/Helpers/SelectListHelper.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    using System.Collections.Generic;
+    using System.Linq;
+    using Microsoft.AspNetCore.Mvc.Rendering;
+
+    public static class SelectListHelper
+    {
+        public static IEnumerable<SelectListItem> GetOptionsWithSelectedText(List<(int id, string name)> options, string? selectedName)
+        {
+            var selectedOption = options.Where(o => o.name == selectedName).Select(o => o.id).SingleOrDefault();
```
stellake, line 11: Given that there's a few things happening here, could we format it such that every data transformation is on a new line to improve readability?
```
options.Where(o => o.name == selectedName)
.Select(o => o.id)
.SingleOrDefault();
```

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests.cs
```diff
@@ -5,6 +5,7 @@
     using DigitalLearningSolutions.Data.DataServices;
```
stellake, line 5: Could we add this page to the automated accessibility tests if it's not there already? 🙏

DigitalLearningSolutions.Data/DataServices/UserDataService.cs
```diff
@@ -210,5 +220,23 @@ public void UpdateDelegateUsers(string firstName, string surname, string email,
                 new { firstName, surname, email, profileImage, ids }
             );
         }
+
+        public void UpdateDelegateUserCentrePrompts(int id, int jobGroupId, string? answer1, string? answer2, string? answer3,
+            string? answer4, string? answer5, string? answer6)
```
stellake, line 225: Could we put every param on a new line if we need to break the line up to make it really easy to scan over all of them?

DigitalLearningSolutions.Data/DataServices/UserDataService.cs
```diff
@@ -210,5 +220,23 @@ public void UpdateDelegateUsers(string firstName, string surname, string email,
...
+                            Answer2 = @answer2,
+                            Answer3 = @answer3,
+                            Answer4 = @answer4,
+                            Answer5 = @answer5,
+                            Answer6 = @answer6
+                        WHERE CandidateID = @id",
+                new { jobGroupId, answer1, answer2, answer3, answer4, answer5, answer6, id }
+            );
+        }
     }
```
stellake, line 241: I think this file is LF 👀

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -133,6 +149,12 @@ public bool TryUpdateUserAccountDetails
             {
                 var delegateIds = verifiedDelegateUsers.Select(d => d.Id).ToArray();
                 userDataService.UpdateDelegateUsers(firstName, surname, email, profileImage, delegateIds);
+
+                if (verifiedDelegateUsers.Any(u => u.Id == delegateId))
+                {
+                    userDataService.UpdateDelegateUserCentrePrompts(delegateId!.Value, jobGroupId!.Value, answer1, answer2, answer3,
```
stellake, line 155: To avoid the jobGroupId!.Value, we could pass in an optional CentrePromptsAnswers? object, which will only be populated if the user is delegate, but will have the correct strong types for the fields.

DigitalLearningSolutions.Web.Tests/Helpers/SelectListHelperTests.cs
```diff
@@ -0,0 +1,93 @@
...
+    using System.Linq;
+    using DigitalLearningSolutions.Web.Helpers;
+    using FluentAssertions;
+    using FluentAssertions.Execution;
+    using NUnit.Framework;
+
+    public class SelectListHelperTests
+    {
+        private const string item1 = "item1";
+        private const string item2 = "item2";
```
stellake, line 13: I think the naming convention here is Item1 etc (capitalised)

DigitalLearningSolutions.Web/Helpers/CustomPromptHelper.cs
stellake: Ignore this - just noticed that it's used with model too! 👍

DigitalLearningSolutions.Web/Views/MyAccount/EditDetails.cshtml
```diff
@@ -1,95 +1,117 @@
...
+    </div>
+
+    foreach (EditCustomFieldViewModel customField in ViewBag.CustomFields) {
+      <div class="nhsuk-grid-row @(ViewBag.CustomFields.IndexOf(customField) == ViewBag.CustomFields.Count - 1 ? "divider" : "")">
+        <div class="nhsuk-grid-column-full">
+          @if (customField.Options.Any())
+          {
+            <vc:select-list asp-for="@("Answer" + @customField.CustomFieldId)" label="@(customField.CustomPrompt + (customField.Mandatory ? "" : " (optional)"))" value="@customField.Answer" hint-text="" css-class="nhsuk-u-width-one-half" default-option="Select a @customField.CustomPrompt.ToLower()" select-list-options="@customField.Options"></vc:select-list>
+          } else {
+            <vc:text-input asp-for="@("Answer" + @customField.CustomFieldId)" label="@(customField.CustomPrompt + (customField.Mandatory ? "" : " (optional)"))" populate-with-current-value="true" type="text" spell-check="false" autocomplete="" hint-text="" css-class="nhsuk-u-width-one-half" />
```
stellake, line 72: There are a lot of attribute values with some logic here so it might be worth separating this onto several lines like the text inputs on reset password page.
We should also be able to close the select-list VC with "/>" over </vc:select-list>

DigitalLearningSolutions.Web/Views/MyAccount/EditDetails.cshtml
```diff
@@ -1,95 +1,117 @@
...
+  @if (Model.IsDelegateUser) {
+    <input type="hidden" asp-for="IsDelegateUser"/>
+    <div class="nhsuk-grid-row">
+      <div class="nhsuk-grid-column-full">
+        <vc:select-list asp-for="JobGroupId" label="Job group" value="@Model.JobGroupId.ToString()" hint-text="" css-class="nhsuk-u-width-one-half" default-option="Select a job group" select-list-options="@ViewBag.JobGroupOptions"></vc:select-list>
+      </div>
+    </div>
+
+    foreach (EditCustomFieldViewModel customField in ViewBag.CustomFields) {
+      <div class="nhsuk-grid-row @(ViewBag.CustomFields.IndexOf(customField) == ViewBag.CustomFields.Count - 1 ? "divider" : "")">
```
stellake, line 66: Might be worth pulling the ViewBag.CustomFields.IndexOf(customField) == ViewBag.CustomFields.Count - 1 into a variable

DigitalLearningSolutions.Web/Helpers/SelectListHelper.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace DigitalLearningSolutions.Web.Helpers
+{
+    using System.Collections.Generic;
+    using System.Linq;
+    using Microsoft.AspNetCore.Mvc.Rendering;
+
+    public static class SelectListHelper
+    {
+        public static IEnumerable<SelectListItem> GetOptionsWithSelectedText(List<(int id, string name)> options, string? selectedName)
```
stellake, line 9: I wonder if this should be named "MapToSelectListItems" to make it clearer what this function does. 🤔 And `selectedName` could be `selectedOptionName ` for extra clarity (I wasn't sure what the "selected name" was initially)

DigitalLearningSolutions.Web/Helpers/SelectListHelper.cs
stellake, line 11: Should this be selectedOptionId? Also, could we replace o.id == selectedOption with o.name == selectedName or are we worried about duplicate names?

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -126,8 +159,45 @@ private IActionResult EditDetailsPostRemoveImage(EditDetailsViewModel model)
...
+
+        private void ValidateCustomPrompts(EditDetailsViewModel model)
+        {
+            var customFields = GetCustomFieldsWithModelAnswers(model);
+
+            foreach (var customField in customFields)
+            {
+                if (customField.Mandatory && customField.Answer == null)
+                {
+                    ModelState.AddModelError("Answer"+customField.CustomFieldId, $"{(customField.Options.Any() ? "Select" : "Enter" )} a {customField.CustomPrompt.ToLower()}");
```
stellake, line 193: Could we pull out the errorMessage into a local var here for readability?

DigitalLearningSolutions.Web.Tests/Helpers/SelectListHelperTests.cs
```diff
@@ -0,0 +1,93 @@
...
+            var result = SelectListHelper.GetOptionsWithSelectedText(inputList, item1);
+
+            using (new AssertionScope())
+            {
+                result.Should().NotBeNullOrEmpty();
+                result.Count().Should().Be(2);
+
+                result.SingleOrDefault(i => i.Value == 1.ToString()).Should().NotBeNull();
+                result.SingleOrDefault(i => i.Value == 1.ToString()).Selected.Should().BeTrue();
+
```
stellake, line 35: I'm quite keen on minimising logic in tests so I wonder if we could do something like this instead:
```
result.ElementAt(0).Value.Should().Be("1"); 
result.ElementAt(0).Text.Should().Be("item1"); 
etc
```

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -50,42 +62,73 @@ public IActionResult EditDetails()
...
+                : new CentreAnswersData(
+                    User.GetCentreId(),
+                    model.JobGroupId!.Value,
+                    model.Answer1,
+                    model.Answer2,
+                    model.Answer3,
+                    model.Answer4,
+                    model.Answer5,
+                    model.Answer6);
+
```
stellake, line 127: Might be worth pulling this into something like:
`private (AccountDetailsData, CentreAnswersData?) MapToUpdateAccountData(EditDetailsViewModel model, int? userAdminId, int? userDelegateId)`

DigitalLearningSolutions.Web/Views/MyAccount/EditDetails.cshtml
```diff
@@ -1,95 +1,169 @@
...
-  <div class="nhsuk-grid-column-full">
-    <vc:back-link asp-controller="MyAccount" asp-action="Index" link-text="Go back" />
-  </div>
-</div>
+                          Changes will not be made until the Save button below is clicked."; }
+      <input type="hidden" asp-for="ProfileImage"/>
+      <vc:file-input asp-for="ProfileImageFile" label="Profile picture (optional)" hint-text="@hintText" css-class="nhsuk-u-width-full"/>
+    </div>
+    <div class="nhsuk-grid-column-one-half">
+      <div class="profile-picture__image-box">
```
stellake, line 132: My IDE complains that profile-picture__image-box does not exist - do we need this div?

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests.cs
```diff
@@ -44,6 +44,7 @@ public void Page_has_no_accessibility_errors(string url)
 
         [Theory]
         [InlineData("/MyAccount")]
+        [InlineData("/MyAccount/EditDetails")]
```
stellake, line 47: 🤩

DigitalLearningSolutions.Web/Views/MyAccount/EditDetails.cshtml
AlexJacksonDS: Also removed it from the Index view that I copied it from

stellake (approved):
Looks good! 👍

PR #334, opened 2021-05-18
HEEDLS-405 Fix remove profile picture and hint text

AlexJacksonDS:
Fixes the removal of profile pictures and changed the hint text to be less confusing.

image


stellake (approved):
Looks good! 👍

PR #335, opened 2021-05-18
HEEDLS-380 Add role text to expander button contents

AlexJacksonDS:
Added role="text" to the child div of the summary so that it is all read out in one go rather than as 2 separate elements. Added as visually hidden span so that when reading it out it doesn't sound like clicking the button is mandatory.


stellake (approved):
Looks good! 👍

PR #339, opened 2021-05-19
HEEDLS-402 registration handle default approval

DanBloxham-sw:
If a user is registering for a delegate account from an IP address that matches to a prefix given in the centre's IPPrefix column, they should get automatically approved. They should also be automatically approved on local host. If they are automatically approved, no approval request email is sent.

- Ran all unit tests
- Manually tested by deploying to test environment and making new accounts.

I reckon this will have some merge conflicts with HEEDLS-400 when one gets merged. 

Side note: looks like CentresDataService was using LF instead of CRLF, so that has been cleaned up.


DigitalLearningSolutions.Data.Tests/DataServices/CentresDataServiceTests.cs
```diff
@@ -119,13 +119,24 @@ public void GetCentreDetailsById_should_return_null_centre_logo_if_logo_is_empty
         }
 
         [Test]
-        public void GetCentreName_should_return_null_when_the_centre_does_not_exist()
+        public void GetCentreIPPrefix_should_return_empty_array_when_the_centre_does_not_exist()
```
AlexJacksonDS, line 122: Was this a duplicate test or is github just being dumb?

AlexJacksonDS (approved):
Looks good, just one small question. Can you also check that CentresDataService is CRLF since it is showing as having a lot of changes.

DigitalLearningSolutions.Data.Tests/DataServices/CentresDataServiceTests.cs
DanBloxham-sw: This was a duplicate test (albeit the name was different) - see this test above:
Get_centre_name_should_return_null_when_the_centre_does_not_exist()

DanBloxham-sw:
I got Resharper to clean up code, and CentresDataService is now definitely CRLF. 

Also, when solving merge conflicts, it looked like RegistrationService and RegisterController both were set to LF after the merge when they were CRLF before. When I made the follow up commit to fix a mistake from the merge, I cleaned up the code, setting them back to CRLF (so the Files Changed section doesn't look like the entire file has been rewritten). Interestingly, RegistrationDataService (which also had merge conflicts) did not get set to LF and remained CRLF. So it looks like the line ending changes are from when merging conflicts, but occur unreliably.

DigitalLearningSolutions.Data.Tests/Services/RegistrationServiceTests.cs
```diff
@@ -1,111 +1,189 @@
-﻿namespace DigitalLearningSolutions.Data.Tests.Services
```
stellake, line 1: This seems to be changed into unix line endings on your branch 👀

DanBloxham-sw: You are correct - and I see what happened. My earlier comment mentioned RegistrationDataService not being changed - I must have completely forgotten that it wasn't RegistrationDataService that had a conflict, it was RegistrationServiceTests that had a conflict. As such, it did indeed seem to be the merge conflict resolution that caused the LF troubles. I'll quickly clean that up to fix it.

DigitalLearningSolutions.Data/DataServices/CentresDataService.cs
```diff
@@ -1,150 +1,169 @@
...
+                        FROM Centres
+                        WHERE CentreID = @centreId",
+                new { centreId }
+            );
+            return info;
+        }
+
+        public string[] GetCentreIPPrefix(int centreId)
+        {
+            var IPPrefix = connection.QueryFirstOrDefault<string?>(
```
stellake, line 156: I think the default naming convention for this + var below is ipPrefix in terms of capitalisation (the IDE should give warnings about this). I also wonder whether this field should be ipPrefixesString or ipPrefixesData or something (as ipPrefix could be misleading in terms of single vs multiple IPs.

DigitalLearningSolutions.Data/DataServices/CentresDataService.cs
```diff
@@ -1,150 +1,169 @@
...
+                @"SELECT IPPrefix
+                        FROM Centres
+                        WHERE CentreID = @centreId",
+                new { centreId }
+            );
+
+            var IPPrefixes = IPPrefix?.Split(',', StringSplitOptions.RemoveEmptyEntries);
+            IPPrefixes ??= new string[0];
+
+            return IPPrefixes;
```
stellake, line 166: Could we do return `IPPrefixes ?? new string[0];` instead of overriding the IPPrefixes value (it's a bit more concise)

DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -1,75 +1,105 @@
...
 
-        public RegistrationService(IRegistrationDataService registrationDataService,
-            IPasswordDataService passwordDataService, IEmailService emailService, ICentresDataService centresDataService)
+        public RegistrationService
+        (
+            IRegistrationDataService registrationDataService,
+            IPasswordDataService passwordDataService,
+            IEmailService emailService,
+            ICentresDataService centresDataService
+        )
```
stellake, line 29: Nice! 👍

DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -1,75 +1,105 @@
...
 
-        public (string candidateNumber, bool approved) RegisterDelegate(DelegateRegistrationModel delegateRegistrationModel, string baseUrl)
+        public (string candidateNumber, bool approved) RegisterDelegate
+        (
+            DelegateRegistrationModel delegateRegistrationModel,
+            string baseUrl,
+            string userIP
+        )
         {
+            var centreIPPrefixes = centresDataService.GetCentreIPPrefix(delegateRegistrationModel.Centre);
```
stellake, line 44: Should the function be GetCentreIPPrefixes? (plural)

DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -1,75 +1,105 @@
...
     using DigitalLearningSolutions.Data.DataServices;
     using DigitalLearningSolutions.Data.Models.Email;
     using DigitalLearningSolutions.Data.Models.Register;
     using MimeKit;
 
     public interface IRegistrationService
     {
-        (string candidateNumber, bool approved) RegisterDelegate(DelegateRegistrationModel delegateRegistrationModel, string baseUrl);
+        (string candidateNumber, bool approved) RegisterDelegate(DelegateRegistrationModel delegateRegistrationModel,
+            string baseUrl, string userIp);
```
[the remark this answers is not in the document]
DanBloxham-sw: Had a fiddle with resharper and got the below settings which seem to do what we want. 
Resharper settings

The _Max Formal Parameters_ is the default, but I've got line split set to 120 in other settings, so I think it gets effectively overwritten.

DigitalLearningSolutions.Data.Tests/Services/RegistrationServiceTests.cs
```diff
@@ -41,15 +60,59 @@ public void Setup()
...
+                emailService,
+                centresDataService
+            );
+        }
+
+        [Test]
+        public void Registering_delegate_with_approved_IP_registers_delegate_as_approved()
+        {
+            // When
+            registrationService.RegisterDelegate(testRegistrationModel, "localhost", "123.456.789.100");
```
stellake, line 76: It might be nicer to write something like this so the person that reads the test doesn't need to scroll all the way up to see whether the IP actually has a correct prefix :)

```
// Given 
var clientIP = approvedIPPrefix + ".100";

// When
etc
```

DigitalLearningSolutions.Data.Tests/Services/RegistrationServiceTests.cs
```diff
@@ -41,15 +60,59 @@ public void Setup()
...
+
+            // Then
+            A.CallTo(() =>
+                    registrationDataService.RegisterDelegate(
+                        A<DelegateRegistrationModel>.That.Matches(d => d.Approved)))
+                .MustHaveHappened();
```
stellake, line 82: Could we also check here that the "approved" which is returned by the registrationService.RegisterDelegate(testRegistrationModel, "localhost", "123.456.789.100"); 

is true?

stellake (approved):
Looks good! 👍

PR #341, opened 2021-05-19
HEEDLS-446 Check for existing candidates with email address

davidm-m:
Checks to see if any existing candidates have a given email address during registration. We do this check on get and on post so if we encounter the error at the final stage (e.g. because the stored procedure tells us that email is already in use) we can just redirect back to the right page and the error is displayed.
I've tested both trying to register with an email address already in the database, and getting halfway through the registration journey before completing a registration with the same email in another browser, then trying to finish the first to make sure I'm booted back to the start with an appropriate error.


DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -31,7 +31,7 @@ public string RegisterDelegate(DelegateRegistrationModel delegateRegistrationMod
         {
             var candidateNumber = registrationDataService.RegisterDelegate(delegateRegistrationModel);
             // TODO HEEDLS-446 Handle return string "-4" for duplicate emails
```
DanBloxham-sw, line 33: This TODO comment can probably be removed.

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -1,162 +1,162 @@
...
+﻿namespace DigitalLearningSolutions.Data.Services
+{
+    using System.Collections.Generic;
+    using System.Linq;
+    using DigitalLearningSolutions.Data.DataServices;
+    using DigitalLearningSolutions.Data.Models.User;
+
+    public interface IUserService
+    {
+        public (AdminUser? adminUser, List<DelegateUser> delegateUsers) GetUsersByUsername(string username);
```
DanBloxham-sw, line 10: Is there a particular reason why these outputs now have names in the interface definition? 

Also it looks like line endings have been changed on this file - can we check that the line endings are correct, as it looks like this UserService file already has CRLF on master.

DanBloxham-sw (commented):
Looks mostly good, couple of very minor comments/questions.

DigitalLearningSolutions.Data/Services/UserService.cs
davidm-m: They have names to make it easier to use them elsewhere - if you only want the adminUser you can do `GetUsersByUsername.adminUser` instead of `GetUsersByUsername.item1` which is much nicer. I had to do something like this so I updated all the interface definitions.

davidm-m: The line endings is weird, if I check out current master they're CRLF but I branched off from `32db337b9cdc5aa91a7f33fe9559050257bc805c` which has them as LF for me at least.

DanBloxham-sw (approved):
Looking good to me.

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -55,13 +57,24 @@ public IActionResult Index()
...
         }
 
         [ServiceFilter(typeof(RedirectEmptySessionData<DelegateRegistrationData>))]
         [HttpPost]
         public IActionResult Index(RegisterViewModel model)
         {
+            if (model.Email != null && userService.GetUsersByEmailAddress(model.Email).delegateUsers.Count != 0)
+            {
+                ModelState.AddModelError(nameof(RegisterViewModel.Email), "A user with this email address already exists");
+            }
```
stellake, line 76: Could we pull this into a private method ValidateEmailAddress or similar that can be called both here and above?

stellake (commented):
One minor request for refactor, otherwise looks great!

stellake (approved):
Looks good! 👍

PR #345, opened 2021-05-20
HEEDLS-NONE - Updated .editorconfig so parameters are placed multiple…

DanBloxham-sw:
… lines if too long.


stellake (approved):
Looks great - thank you! 👍

PR #353, opened 2021-05-24
HEEDLS-455 Change dl element with no dt children to a div

AlexJacksonDS:
Changed the dl element in the notification list that had no child dt/dd elements to a div. Style should still be the same as I've not changed the css classes.
Added Notification preferences pages to accessibility tests now they pass.

Checked the layout wasn't broken by this change and the expanders etc. still work.

### DIFF
--- a/DigitalLearningSolutions.Web/Views/NotificationPreferences/_NotificationPreferenceList.cshtml
+++ b/DigitalLearningSolutions.Web/Views/NotificationPreferences/_NotificationPreferenceList.cshtml
@@ -3,21 +3,21 @@
 
 <link rel="stylesheet" href="@Url.Content("~/css/myAccount/notificationPreferences.css")" asp-append-version="true">
 
-<div class="nhsuk-grid-column-full nhsuk-summary-list preference-list">
+<dl class="nhsuk-grid-column-full nhsuk-summary-list preference-list">
   @foreach (var notification in Model.Notifications) {
     <div class="nhsuk-grid-row preference-row">
-      <div class="nhsuk-grid-column-three-quarters preference-text-column">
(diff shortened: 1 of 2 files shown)


davidm-m (approved):
Looks good!

DigitalLearningSolutions.Web/Views/NotificationPreferences/_NotificationPreferenceList.cshtml
```diff
@@ -3,7 +3,7 @@
 
 <link rel="stylesheet" href="@Url.Content("~/css/myAccount/notificationPreferences.css")" asp-append-version="true">
 
-<dl class="nhsuk-grid-column-full nhsuk-summary-list preference-list">
+<div class="nhsuk-grid-column-full nhsuk-summary-list preference-list">
```
stellake, line 6: Re HTML markup for screen readers, should this be key-value (dt - dd) combination? The key would be the name of the notification with the description and the value is its status.. 🤔 

https://service-manual.nhs.uk/design-system/components/summary-list

AlexJacksonDS: Doing this does remove the whole list reading stuff, but it also isn't really an NHS summary list any more. If we want to keep the dl/dt/dd structure, will need to do a whole load of styling changes, simply changing the 2 child divs to a dt/dd pair causes all the styling to break

stellake: I think we should use the accessible markup where-ever possible, so we should probably give this change a go! But maybe timebox this (keep at 1h max)?

stellake (approved):
Looks good! 👍

PR #354, opened 2021-05-24
Heedls 402 handle default approvals bug fixing

DanBloxham-sw:
Changed the Ip grabbing to grab forwarded value if present.

Tried deploying to test, this time it does indeed grab the same IP as https://whatismyipaddress.com/ and does indeed automatically approve delegates when the Centre record is updated with that IP prefix.

### DIFF
--- a/DigitalLearningSolutions.Web/Controllers/RegisterController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/RegisterController.cs
@@ -161,6 +161,10 @@
 
             var baseUrl = ConfigHelper.GetAppConfig()["CurrentSystemBaseUrl"];
             var userIP = Request.HttpContext.Connection.RemoteIpAddress.ToString();
+            if (Request.Headers.ContainsKey("X-Forwarded-For"))
+            {
+                userIP = Request.Headers["X-Forwarded-For"].ToString();
+            }
             var (candidateNumber, approved) =
                 registrationService.RegisterDelegate(RegistrationMappingHelper.MapToDelegateRegistrationModel(data),
                     baseUrl, userIP);
--- a/DigitalLearningSolutions.Web/Controllers/RegisterController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/RegisterController.cs
@@ -160,14 +160,12 @@
             }
 
             var baseUrl = ConfigHelper.GetAppConfig()["CurrentSystemBaseUrl"];
-            var userIP = Request.HttpContext.Connection.RemoteIpAddress.ToString();
-            if (Request.Headers.ContainsKey("X-Forwarded-For"))
-            {
-                userIP = Request.Headers["X-Forwarded-For"].ToString();
-            }
+            var userIp = Request.Headers.ContainsKey("X-Forwarded-For") ?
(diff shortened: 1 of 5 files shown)


DigitalLearningSolutions.Web/Startup.cs
```diff
@@ -148,8 +148,7 @@ public void Configure(IApplicationBuilder app, IMigrationRunner migrationRunner)
         {
             app.UseForwardedHeaders(new ForwardedHeadersOptions
             {
-                ForwardedHeaders = ForwardedHeaders.XForwardedFor |
-                                   ForwardedHeaders.XForwardedProto
+                ForwardedHeaders = ForwardedHeaders.XForwardedFor | ForwardedHeaders.XForwardedProto
```
davidm-m, line 151: Just to satisfy my curiosity, what does this vertical bar do?

davidm-m (approved):
Looks good - I like the renaming of IP to Ip!

DigitalLearningSolutions.Web/Startup.cs
DanBloxham-sw: After doing a little digging (I initially just copied this from msdn docs to see if it worked) it looks to be an operator that creates a bitwise combination of its (enum) values for the ForwadedHeaders property (https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.httpoverrides.forwardedheaders?view=aspnetcore-3.1).

This is because the property itself is defined with Flags attribute, which seems to basically be a way to specify multiple potential enum combinations. The Hue example and the PhoneService in the docs seems to be a better explanation than I can muster. https://docs.microsoft.com/en-us/dotnet/api/system.flagsattribute?view=netcore-3.1

davidm-m: Ooh nice, thanks for the research 🙂

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -160,10 +160,12 @@ public IActionResult Summary(SummaryViewModel model)
             }
 
             var baseUrl = ConfigHelper.GetAppConfig()["CurrentSystemBaseUrl"];
-            var userIP = Request.HttpContext.Connection.RemoteIpAddress.ToString();
+            var userIp = Request.Headers.ContainsKey("X-Forwarded-For") ?
+                Request.Headers["X-Forwarded-For"].ToString() :
```
stellake, line 164: Re conversation we had before, did you find any documentation around how MVC handles several values for X-Forwarded-For? https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For

DanBloxham-sw: From what I can tell, there is no unique way that it is handled in MVC, which means that the X-Forwarded-For could be a comma separated list as specified in the docs you linked. In the cast that a comma separated list is retrieved, because it is all handled as a string where the client Ip is first, the code should still work as is.

stellake (approved):
Looks good

PR #355, opened 2021-05-24
HEEDLS-456 Check page title in accessibility tests

AlexJacksonDS:
Added checks for the page heading (h1 element text) to the accessibility tests

### DIFF
--- a/DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests.cs
+++ b/DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests.cs
@@ -37,15 +37,9 @@ namespace DigitalLearningSolutions.Web.AutomatedUiTests
         {
             // when
             driver.Navigate().GoToUrl(baseUrl + url);
-            var axeResult = new AxeBuilder(driver).Analyze();
-            var h1Element = driver.FindElement(By.TagName("h1"));
 
             // then
-            using (new AssertionScope())
-            {
-                axeResult.Violations.Should().BeEmpty();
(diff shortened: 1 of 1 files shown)


DanBloxham-sw (approved):
Looks good to me.

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests.cs
```diff
@@ -26,38 +27,48 @@ public void Dispose()
...
             var axeResult = new AxeBuilder(driver).Analyze();
+            var h1Element = driver.FindElement(By.TagName("h1"));
 
             // then
-            axeResult.Violations.Should().BeEmpty();
+            using (new AssertionScope())
+            {
+                axeResult.Violations.Should().BeEmpty();
+                h1Element.Text.Should().BeEquivalentTo(pageTitle);
+            }
```
stellake, line 48: I don't think we need to use the AssertionScope here, because the logic should be something like this:
```
// when 
navigate to url

// then
var h1Element = ...
h1Element.Text.Should().BeEquivalentTo(pageTitle);

var axeResult = ...
axeResult.Violations.Should().BeEmpty();
```

(No need to do the accessibility scans if we know that we're not on the correct page)

stellake: You could even pull these two assertions into private methods that both tests can call, but that's not too important :)

stellake (commented):
Looks good - just a comment about the assertions

stellake (approved):
Looks great! 👍

PR #356, opened 2021-05-24
HEEDLS-451 - New migration to add ShowOnMap cable to Centres table

DanBloxham-sw:
Added ShowOnMap to Centres table.

Tested manually migrating up and then migration down to make sure both work. Also checked the default values are all good. Ran all unit tests.

### DIFF
--- a/DigitalLearningSolutions.Web/Helpers/MigrationHelperMethods.cs
+++ b/DigitalLearningSolutions.Web/Helpers/MigrationHelperMethods.cs
@@ -6,52 +6,21 @@
 
     public static class MigrationHelperMethods
     {
-        public static IServiceCollection RegisterMigrationRunner(this IServiceCollection serviceCollection, string connectionString)
+        public static IServiceCollection RegisterMigrationRunner
+            (this IServiceCollection serviceCollection, string connectionString)
         {
             return serviceCollection.AddFluentMigratorCore()
-                .ConfigureRunner(rb => rb
-                    .AddSqlServer2016()
-                    .WithGlobalConnectionString(connectionString)
-                    .ScanIn(
-                        typeof(AddSelfAssessmentTables).Assembly,
-                        typeof(AddSelfAssessmentResultTimestamp).Assembly,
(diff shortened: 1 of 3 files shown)


DigitalLearningSolutions.Web/Helpers/MigrationHelperMethods.cs
```diff
@@ -47,7 +47,8 @@ public static IServiceCollection RegisterMigrationRunner(this IServiceCollection
                         typeof(FixCustomisationUpdateSPs).Assembly,
                         typeof(AddEmailToFrameworkCollaboratorTable).Assembly,
                         typeof(AddFrameworkReviewsTable).Assembly,
-                        typeof(AddArchivedDateToFrameworkReviews).Assembly
+                        typeof(AddArchivedDateToFrameworkReviews).Assembly,
+                        typeof(AddMapDisplayColumnToCentres).Assembly
```
SteveJacksonSoft, line 51: Do you know if you actually need to add this? I'd have thought all these classes would be in the same assembly, but I don't confess to understand assemblies or FluentMigrator particularly well.

SteveJacksonSoft (commented):
Looks good! Just one question, but I'll pass the review on to Stella.

DigitalLearningSolutions.Web/Helpers/MigrationHelperMethods.cs
DanBloxham-sw: You are absolutely right. I was blindly following the readme instructions for setting up new migrations, which mentions adding each one. I've tested my new migration without the addition in ScanIn, and it still works fine (some debug comparisons show that each of these assemblies are definitely exactly the same, so we only need one as they are all from the same assembly).

@stellake  if you think we should, I am happy to cutback the ScanIn procedure to a single assembly and update the readme file as part of this ticket.

stellake: Sounds good! 👍

stellake (approved):
Looks great! 👍

PR #358, opened 2021-05-25
HEEDLS-446 Move email to first page of registration form and validate it with centre id

davidm-m:
This moves the centre field of the registration form from the Learner Information page to the first page and fixes any related logic. I also took the time to switch all the inputs on these two pages to our handy view components to make them easier to read and more maintainable in the future.

Tested by registering a bunch of times with the same email, checking that the validation worked as expected everywhere and moving backwards and forwards didn't lose data.

The first page of the registration form now looks like the below since the centre select needs to be very wide to display the full name of some centres:
image


DigitalLearningSolutions.Data/DataServices/UserDataService.cs
```diff
@@ -181,14 +181,19 @@ public List<DelegateUser> GetDelegateUsersByEmailAddress(string emailAddress)
         {
             List<DelegateUser> users = connection.Query<DelegateUser>(
                 @"SELECT
-                        CandidateID,
-                        FirstName,
-                        LastName,
-                        EmailAddress,
-                        Password,
-                        ResetPasswordID
```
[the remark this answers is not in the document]
AlexJacksonDS: I'm making a change to the same query for a 345 bug fix. Will update it to the same thing as here, but include the ResetPasswordID. It is used in the PasswordResetService.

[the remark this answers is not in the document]
davidm-m: Sounds good, removing ResetPasswordID here was an accident 😅

DigitalLearningSolutions.Data/DataServices/UserDataService.cs
```diff
@@ -181,14 +181,19 @@ public List<DelegateUser> GetDelegateUsersByEmailAddress(string emailAddress)
...
+                        ct.CentreName,
+                        ct.Active AS CentreActive,
+                        cd.EmailAddress,
+                        cd.FirstName,
+                        cd.LastName,
+                        cd.Password,
+                        cd.Approved
+                    FROM Candidates AS cd
+                    INNER JOIN Centres AS ct ON ct.CentreID = cd.CentreID
+                    WHERE cd.EmailAddress = @emailAddress",
```
stellake, line 196: This should have data tests for on the latest master - do you mind double checking that everything gets asserted properly? 🙏

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -54,9 +55,15 @@ public IActionResult Index()
                     });
                 TempData.Set(delegateRegistrationData);
 
-                return View();
+                ViewBag.CentreOptions = SelectListHelper.MapOptionsToSelectListItemsWithSelectedValue
+                    (centresDataService.GetActiveCentresAlphabetical(), null);
+
+                return View(delegateRegistrationData.RegisterViewModel);
```
stellake, line 61: Do we need this twice? Could we remove L28-61 potentially? As delegateRegistrationData.RegisterViewModel.Centre will be null and email validation passes with email being null. Now that I think about it, I think the validation only needs to happen on POST.

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -203,9 +212,17 @@ public IActionResult Confirmation()
 
         private void ValidateEmailAddress(RegisterViewModel model)
         {
-            if (model.Email != null && userService.GetUsersByEmailAddress(model.Email).delegateUsers.Count != 0)
+            if (model.Email == null)
+            {
+                return;
+            }
+
+            var duplicateUsers = userService.GetUsersByEmailAddress(model.Email).delegateUsers.Where
```
stellake, line 220: Could we add the .Where(...) onto a new line for better readability?

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -54,9 +55,15 @@ public IActionResult Index()
                     });
                 TempData.Set(delegateRegistrationData);
```
stellake, line 56: Could we write some controller tests for the new logic?

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
davidm-m: The validation needs to happen on GET in the unlikely case that someone gets beyond this page of the form, then someone else finishes registering with the same email address, and the first person finishes registering and gets a duplicate email error so we boot them back to this page. But yeah Lines 58-61 can go for sure.

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -203,9 +212,17 @@ public IActionResult Confirmation()
...
+                return;
+            }
+
+            var duplicateUsers = userService.GetUsersByEmailAddress(model.Email).delegateUsers.Where
+                (u => u.CentreId == model.Centre);
+
+            if (duplicateUsers.Count() != 0)
             {
-                ModelState.AddModelError(nameof(RegisterViewModel.Email), "A user with this email address already exists");
+                ModelState.AddModelError(nameof(RegisterViewModel.Email), "A user at this centre with this email address already exists");
```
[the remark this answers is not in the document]
davidm-m: No I think you're right, my wording always felt awkward to me 😄

davidm-m:
How it looks now with the centre being the first thing on the form:
image

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
stellake: Wait, how does a user get to the next page of the journey without posting the form (I'm not sure if I understood your description correctly)? 🤔

DigitalLearningSolutions.Web.Tests/Controllers/Register/RegisterControllerTests.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            centresDataService = A.Fake<ICentresDataService>();
+            cryptoService = A.Fake<ICryptoService>();
+            jobGroupsDataService = A.Fake<IJobGroupsDataService>();
+            registrationService = A.Fake<IRegistrationService>();
+            userService = A.Fake<IUserService>();
+        }
+
+        [Test]
+        public void IndexPost_with_existing_user_for_centre_fails_validation()
```
stellake, line 33: Nice! Should we add one where there's a user registered in a different centre (which is ok)?

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
davidm-m: Yeah I don't think I was clear enough. We need to have the check on a GET in case the registration stored procedure returns an error saying the email address is in use. There's only one case that I can think of where that might happen though:

1. A user completes the first page of the registration form
2. A different user (or the same user on a different computer) completes the whole registration form with the same centre and email address
3. The first user now tries to submit the form but the stored procedure gives us an error saying the email is already in use
4. We redirect the user back to the first page of the form which is a GET request, so to display the correct error we need to make the duplicate email check on a GET

DigitalLearningSolutions.Web.Tests/Controllers/Register/RegisterControllerTests.cs
davidm-m: Good call!

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
stellake: Oh I see - that makes sense! Could we add a quick comment about this here so if anyone ever reads through this, they won't get confused? :)

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests.cs
```diff
@@ -89,12 +89,12 @@ public void Registration_journey_has_no_accessibility_errors()
             lastName.SendKeys("User");
             var email = driver.FindElement(By.Id("Email"));
             email.SendKeys("candidate@test.com");
+            var centre = new SelectElement(driver.FindElement(By.Id("Centre")));
+            centre.SelectByValue("2");
```
stellake, line 93: This test is now moved to a different file on master - should be an easy one to resolve.

stellake (approved):
Looks good! Just two minor comments about the merge conflict and adding a quick comment. No need for rereview though 👍

PR #365, opened 2021-05-26
HEEDLS-464 - changed login procedure to only log in as candidate if o…

DanBloxham-sw:
…nly candidate number provided

Added extra check in the log in process so that admin account is retrieved only if it is at the same centre as the delegate account when logging in using delegate specific username. A new unit test has been added to test this functionality. The files were also cleaned up with the newish .editorconfig formatting, so the files now use CRLF line endings, and the spacing is a bit different from where lines were too long. I've put comments at the actual new additions.

Ran all unit tests and tried manually logging in to different accounts.


DigitalLearningSolutions.Data.Tests/Services/LoginServiceTests.cs
```diff
@@ -1,208 +1,249 @@
...
+                "password"
+            );
+
+            // Then
+            Assert.IsNull(returnedAdminUser);
+        }
+
+        [Test]
+        public void
+            GetVerifiedAdminUserAssociatedWithDelegateUser_Returns_nothing_when_associated_admin_account_is_at_different_centre()
```
DanBloxham-sw, line 211: This is the new test, and the only new addition in this file.

AlexJacksonDS (approved):
Looks fine, just got to resolve the merge conflict

stellake (approved):
Looks good!

PR #366, opened 2021-05-26
HEEDLS-345 Fix incorrect boolean condition

AlexJacksonDS:
Fixed an incorrect condition causing email validation to fail when unchanged. Added a couple of unit tests to check it.

### DIFF
--- a/DigitalLearningSolutions.Data/DataServices/UserDataService.cs
+++ b/DigitalLearningSolutions.Data/DataServices/UserDataService.cs
@@ -165,10 +165,10 @@
         {
             return connection.Query<AdminUser>(
                 @"SELECT
-                        AdminID,
-                        Forename,
-                        Surname,
-                        Email,
+                        AdminID AS Id,
+                        Forename AS FirstName,
+                        Surname AS LastName,
+                        Email AS EmailAddress,
                         Password,
                         ResetPasswordID
                     FROM AdminUsers
--- a/DigitalLearningSolutions.Data/DataServices/UserDataService.cs
+++ b/DigitalLearningSolutions.Data/DataServices/UserDataService.cs
@@ -181,14 +181,20 @@
         {
             List<DelegateUser> users = connection.Query<DelegateUser>(
                 @"SELECT
-                        CandidateID,
-                        FirstName,
-                        LastName,
-                        EmailAddress,
-                        Password,
-                        ResetPasswordID
-                    FROM Candidates
-                    WHERE (EmailAddress = @emailAddress)",
+                        cd.CandidateID AS Id,
+                        cd.CandidateNumber,
+                        cd.CentreID,
+                        ct.CentreName,
+                        ct.Active AS CentreActive,
+                        cd.EmailAddress,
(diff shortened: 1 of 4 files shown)


DanBloxham-sw (changes requested):
This all looks good to me, but there is one thing that will need changing.

I double checked, and you can indeed change a delegate account email address to the email address of another delegate account at the same centre, but the condition in the validation method should clearly exclude that. But I've figured out why it gets bypassed - the GetDelegateUsersByEmailAddress method (called as part of the GetUsersByEmailAddress) does not return the CentreId, so the delegateUsers populated by it will have a default CentreId (probably 0). Thus the non-zero centre Id passed in will never be the same, and thus the email address change occurs.

So the GetDelegateUsersByEmailAddress in the UserDataService will need a quick change to make sure the centre id is recovered for the users so it can be used in the validation.

AlexJacksonDS:
Got to love queries that don't fully populate objects. Will get that changed.

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -144,7 +144,7 @@ public List<CentreUserDetails> GetUserCentres(AdminUser? adminUser, List<Delegat
         public bool NewEmailAddressIsValid(string emailAddress, int? adminUserId, int? delegateUserId, int centreId)
         {
             var (adminUser, delegateUser) = GetUsersById(adminUserId, delegateUserId);
-            if (adminUser?.EmailAddress == emailAddress && delegateUser?.EmailAddress == emailAddress)
+            if ((adminUser == null || adminUser.EmailAddress == emailAddress) && (delegateUser == null || delegateUser.EmailAddress == emailAddress))
```
stellake, line 147: This statement is getting quite long - maybe worth pulling out into a method that's called `userEmailHasChanged(user, email)` or similar and doing if `(!userEmailHasChanged(adminUser, emailAddress) && !userEmailHasChanged(delegateUser, emailAddress) `

DanBloxham-sw (approved):
Requested change made.

DigitalLearningSolutions.Data/DataServices/UserDataService.cs
```diff
@@ -181,14 +181,20 @@ public List<DelegateUser> GetDelegateUsersByEmailAddress(string emailAddress)
...
+                        cd.CandidateNumber,
+                        cd.CentreID,
+                        ct.CentreName,
+                        ct.Active AS CentreActive,
+                        cd.EmailAddress,
+                        cd.FirstName,
+                        cd.LastName,
+                        cd.Password,
+                        cd.Approved,
+                        cd.ResetPasswordID
```
stellake, line 194: Do we need to change any test to check that these all come through fine now?

AlexJacksonDS: There are actually no tests for the GetByEmail methods. Should I write some as part of this change?

stellake (approved):
Looks good! 👍

PR #369, opened 2021-05-27
Heedls 453 add prompt second page

AlexJacksonDS:
Adds the second page of the Add registration prompt journey, configuring answers. There is a lot of common code here with the EditRegistrationPrompt screen, so I've refactored that at the same time.

Did a quick check that Editing still works fine and that the mobile layout works  (it is basically the same as the edit screen).

Also fixed a bug with the Add new answer textbox remaining populated after successful submission.

image


DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
```diff
@@ -53,15 +52,15 @@ public IActionResult EditRegistrationPrompt(int promptNumber)
         [Route("{promptNumber}/Edit")]
         public IActionResult EditRegistrationPrompt(EditRegistrationPromptViewModel model, string action)
         {
-            if (action.StartsWith("delete") && TryGetAnswerIndexFromAction(action, out var index))
+            if (action.StartsWith("delete") && TryGetAnswerIndexFromDeleteAction(action, out var index))
```
JBsoftwire, line 55: Can we make these action strings an enum or set of constants?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
```diff
@@ -103,88 +108,137 @@ public IActionResult AddRegistrationPromptSelectPrompt(AddRegistrationPromptSele
...
+            model.OptionsString = optionsString;
+
+            ModelState.Remove(nameof(RegistrationPromptAnswersViewModel.Answer));
+            model.Answer = null;
+        }
+
+        private void AddTotalStringLengthRegistrationPromptAnswersViewModelError
+            (RegistrationPromptAnswersViewModel model)
+        {
+            var remainingLength = 4000 - (model.OptionsString?.Length - 2 ?? 0);
```
JBsoftwire, line 234: Is it possible for the options string to have a length of less than 2?

JBsoftwire (commented):
I have a couple of questions, but didn't see any glaring issues. Since this is a blocker I think we can move it up to tech lead review.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
AlexJacksonDS: Technically yes, but not at the point where this would be a problem. This code will only be reached if (optionsString + '\r\n' + newAnswerString) > 4000. Which given the newAnswerString has a max length of 100 isn't possible until the optionsString has length greater than 3900

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
JBsoftwire: You probably also want to refer to the constant length for the string operation on line 257 in the controller

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests.cs
```diff
@@ -132,8 +132,25 @@ public void AddRegistrationPrompt_journey_has_no_accessibility_errors()
             var selectPromptForm = driver.FindElement(By.TagName("form"));
             selectPromptForm.Submit();
```
stellake, line 134: Nice! I wonder whether it's time to separate the journeys into different files.. 🤔 
So we could have AccessibilityTests folder and then BasicAccessibilityTests.cs, RegistrationJourneyAccessibilityTests.cs, AddCentreCustomPromptsAccessibilityTests.cs?
We could also create some helpers for logging in, submitting the form, filling in a text input, dropdown etc to make the tests nicer to read :)

DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsControllerTests.cs
```diff
@@ -33,24 +33,27 @@ public void Setup()
             registrationPromptsController = new RegistrationPromptsController(customPromptsService)
                 .WithDefaultContext()
                 .WithMockUser(true);
+            registrationPromptsController.TempData = new TempDataDictionary
```
stellake, line 36: Could we just do .withMockTempData above (there should be an extension)

DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsControllerTests.cs
```diff
@@ -172,13 +184,162 @@ public void AddRegistrationPromptSelectPrompt_loads_existing_temp_data()
...
+
+        [Test]
+        public void AddRegistrationPromptConfigureAnswers_next_updates_temp_data()
+        {
+            // Given
+            var expectedPromptModel = new AddRegistrationPromptSelectPromptViewModel
+            {
+                CustomPromptId = 1,
+                Mandatory = true
+            };
```
stellake, line 226: Could this benefit from a constructor for more concise declaration? (+ elsewhere to make it easier to read)

DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsControllerTests.cs
```diff
@@ -172,13 +184,162 @@ public void AddRegistrationPromptSelectPrompt_loads_existing_temp_data()
...
+            var initialTempData = new AddRegistrationPromptData
+                { SelectPromptViewModel = expectedPromptModel, ConfigureAnswersViewModel = model };
+            registrationPromptsController.TempData.Set(initialTempData);
+            var expectedViewModel = new RegistrationPromptAnswersViewModel
+            {
+                OptionsString = "Test\r\nAnswer"
+            };
+            const string action = "addPrompt";
```
stellake, line 269: I think naming things "model/Viewmodel/PromptModel" is fairly generic and makes this test a bit difficult to follow. I'd probably group same viewmodels together (declare same types next to each other) and have more specific names, like configureAnswersModel/expectedConfigureAnswersModel/selectPromptModel

DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsControllerTests.cs
```diff
@@ -172,13 +184,162 @@ public void AddRegistrationPromptSelectPrompt_loads_existing_temp_data()
...
+
+            // When
+            var result = registrationPromptsController.AddRegistrationPromptConfigureAnswers(model, action);
+
+            // Then
+            using (new AssertionScope())
+            {
+                registrationPromptsController.TempData.Peek<AddRegistrationPromptData>().Should().NotBeNull();
```
stellake, line 277: Should not be null seems unnecessary as we move to the next line anyway due to the assertion scope

DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsControllerTests.cs
```diff
@@ -172,13 +184,162 @@ public void AddRegistrationPromptSelectPrompt_loads_existing_temp_data()
...
+                registrationPromptsController.TempData.Peek<AddRegistrationPromptData>()!.SelectPromptViewModel.Should()
+                    .BeEquivalentTo(expectedPromptModel);
+                registrationPromptsController.TempData.Peek<AddRegistrationPromptData>()!.ConfigureAnswersViewModel
+                    .Should().BeEquivalentTo(expectedViewModel);
+                result.Should().BeViewResult();
+                result.As<ViewResult>().Model.Should().BeOfType<RegistrationPromptAnswersViewModel>();
+                result.As<ViewResult>().Model.As<RegistrationPromptAnswersViewModel>().Options.Count.Should().Be(2);
```
stellake, line 284: We could probably benefit from pulling some of these into private methods that can be used in these tests to make things easier to read and follow, e.g. AssertNumberOfConfiguredAnswersOnView(numberOfConfiguredAnswers) or something similar
```
result.Should().BeViewResult();
result.As<ViewResult>().Model.Should().BeOfType<RegistrationPromptAnswersViewModel>();
result.As<ViewResult>().Model.As<RegistrationPromptAnswersViewModel>().Options.Count.Should().Be(2);
```

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
```diff
@@ -18,6 +17,10 @@
     [Route("/TrackingSystem/CentreConfiguration/RegistrationPrompts")]
     public class RegistrationPromptsController : Controller
     {
+        public const string DeleteString = "delete";
+        public const string AddPromptString = "addPrompt";
+        public const string NextString = "next";
+        public const string SaveString = "save";
```
stellake, line 23: Should these be called actions?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
```diff
@@ -70,23 +73,29 @@ public IActionResult EditRegistrationPrompt(EditRegistrationPromptViewModel mode
...
+                    }
+                );
                 TempData.Set(addRegistrationPromptData);
+
+                SetViewBagCustomPromptNameOptions();
+            }
+            else
+            {
+                SetViewBagCustomPromptNameOptions(addRegistrationPromptData.SelectPromptViewModel.CustomPromptId);
             }
```
stellake, line 99: Do we need to separate this from L94? Could we just do:
`SetViewBagCustomPromptNameOptions(addRegistrationPromptData.SelectPromptViewModel?.CustomPromptId);`

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
```diff
@@ -103,88 +112,137 @@ public IActionResult AddRegistrationPromptSelectPrompt(AddRegistrationPromptSele
...
-            // TODO: HEEDLS-453 - redirect to next page
-            return RedirectToAction("Index");
+            return View(viewModel);
+        }
+
+        [HttpPost]
+        [Route("Add/ConfigureAnswers")]
+        [ServiceFilter(typeof(RedirectEmptySessionData<AddRegistrationPromptData>))]
+        public IActionResult AddRegistrationPromptConfigureAnswers
+            (RegistrationPromptAnswersViewModel model, string action)
```
stellake, line 135: This is weird formatting - could we set the formatter to put both params onto new lines in this case?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
```diff
@@ -103,88 +112,137 @@ public IActionResult AddRegistrationPromptSelectPrompt(AddRegistrationPromptSele
...
             }
 
-            var (optionsString, options) =
+            var optionsString =
                 NewlineSeparatedStringListHelper.AddStringToNewlineSeparatedList(model.OptionsString, model.Answer!);
 
             if (optionsString.Length > 4000)
             {
-                AddTotalStringLengthModelError(model);
+                AddTotalStringLengthRegistrationPromptAnswersViewModelError(model);
```
stellake, line 178: Maybe name this to something like SetConfiguredAnswersTooLongError or something similar?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
```diff
@@ -18,6 +17,10 @@
     [Route("/TrackingSystem/CentreConfiguration/RegistrationPrompts")]
     public class RegistrationPromptsController : Controller
     {
+        public const string DeleteString = "delete";
+        public const string AddPromptString = "addPrompt";
+        public const string NextString = "next";
```
stellake, line 22: Do we need Next & Save? Wouldn't the action always be Save technically?

AlexJacksonDS: I changed it to distinguish that the button is not save, it is go to next page of the journey

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
AlexJacksonDS: This seems to have been caused by slightly conflicting rules. Changed one so default wrapping should now be
```
public void ReallyLongMethodNameWithParameters(
  string param1,
  string param2,
)

File: DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/AccessibilityCollection.cs [hunk 14]
```diff
@@ -0,0 +1,9 @@
+﻿namespace DigitalLearningSolutions.Web.AutomatedUiTests.AccessibilityTests
+{
+    using Xunit;
+
+    [CollectionDefinition("Accessibility collection")]
```
stellake, line 5: Could we name this "Selenium test collection" as we probably don't want to restart the server for other tests (if we extend things and use these tests for stuff other than accessibility)

DigitalLearningSolutions.Web.AutomatedUiTests/TestHelpers/InputHelper.cs
```diff
@@ -0,0 +1,32 @@
+﻿namespace DigitalLearningSolutions.Web.AutomatedUiTests.TestHelpers
+{
+    using OpenQA.Selenium;
+    using OpenQA.Selenium.Support.UI;
+
+    public static class InputHelper
```
stellake, line 6: Nice! Minor thing - could we call this FormHelper as it also includes the button actions/form submission? 👍

stellake: To be honest, these could also be part of the DriverHelper class, given that they're extensions of it

DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsControllerTests.cs
stellake: I think you missed L168 👁️

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
stellake: Perfect, thank you! 👍

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
```diff
@@ -103,88 +107,145 @@ public IActionResult AddRegistrationPromptSelectPrompt(AddRegistrationPromptSele
...
-            return int.TryParse(action.Remove(0, 6), out index);
+            IgnoreAddNewAnswerValidation();
+
+            if (!ModelState.IsValid)
+            {
+                return View(model);
+            }
+
+            customPromptsService.UpdateCustomPromptForCentre
+                (User.GetCentreId(), model.PromptNumber, model.Mandatory, model.OptionsString);
```
stellake, line 157: Ah, could we use the same formatting in this case as we do for the function param declaration?

DigitalLearningSolutions.Web/appsettings.Development.json
```diff
@@ -1,6 +1,6 @@
 {
   "ConnectionStrings": {
-    "DefaultConnection": "Data Source=localhost;Initial Catalog=mbdbx101;Integrated Security=True;"
+    "DefaultConnection": "Data Source=localhost;Initial Catalog=mbdbx101_test;Integrated Security=True;"
```
stellake, line 3: Hmm.. Is this intended?

AlexJacksonDS: Nope was checking something, will revert

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
stellake: You could add a if optionsString < 4000 return statement to the top of this method to make this extra safe :)

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/AccessibilityCollection.cs
stellake: Super minor, but should we change the class name too?

stellake (approved):
Looks good - minor comment about a class name + suggestion to make the model error stuff in controller more safe. No need for rereview though 👍

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
AlexJacksonDS: I don't think that is the condition we want. Have added a check to return if the options string is null or has length less than 2. model.OptionsString < 4000 is always true. and optionsString (the updated one) it is always false at this point.

PR #370, opened 2021-05-27
HEEDLS-449 Add navbar to new page and fix claims checks

AlexJacksonDS:
Added navbar to the Edit Registration Prompt page and fixed the claims checking to actually check the value of the custom claims rather than just whether they have them.


stellake (approved):
Looks good!

PR #371, opened 2021-05-27
Heedls 423 approve delegates

JBsoftwire:


DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateApprovals/Index.cshtml
```diff
@@ -27,9 +28,12 @@
       }
       <div class="nhsuk-grid-row nhsuk-u-margin-top-7">
         <div class="nhsuk-grid-column-full">
-          <a class="nhsuk-button" href="" role="button">
-            Approve all
-          </a>
+          <form asp-controller="DelegateApprovals" asp-action="ApproveDelegatesForCentre" method="post">
```
JBsoftwire, line 31: Is there a more sensible way to get these post requests out without any JS? This works smoothly but feels like a hack.

DigitalLearningSolutions.Data/DataServices/UserDataService.cs
```diff
@@ -3,6 +3,7 @@
     using System.Collections.Generic;
     using System.Data;
     using System.Linq;
+    using System.Threading.Tasks;
```
AlexJacksonDS, line 6: Doesn't look like anything you've added to this file uses this using statement

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateApprovals/Index.cshtml
AlexJacksonDS: Do you mean in terms of having an individual form per button? The form with an asp-action is the standard way to do it.

The only other sensible way to do multiple posts per page with no JS is with the stuff I have been doing for the My Account Edit Details and Registration Prompt answer editing, but the workaround found for those was only needed due to the requirement to have forms within forms. I can give you a quick run through of that approach and why it was needed if you want, but I think the current approach here with multiple form elements is better.

AlexJacksonDS (approved):
Looks fine, just a tiny comment about an unused using statement

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests.cs
```diff
@@ -240,5 +240,38 @@ public void UpdateDelegateUserCentrePrompts_updates_user()
                 }
             }
         }
+
+        [Test]
+        public void ApproveDelegateUsers_approves_delegate_users()
+        {
+            using (var transaction = new TransactionScope())
```
stellake, line 247: Formatting wise, you should be able to do:
```
using var transaction = new TransactionScope(); 
try ... 
```

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests.cs
```diff
@@ -240,5 +240,38 @@ public void UpdateDelegateUserCentrePrompts_updates_user()
...
+
+                    delegate1.Approved.Should().BeFalse();
+                    delegate2.Approved.Should().BeFalse();
+
+                    // When
+
+                    userDataService.ApproveDelegateUsers(ids);
+
+                    delegate1 = userDataService.GetDelegateUserById(16);
+                    delegate2 = userDataService.GetDelegateUserById(28);
```
stellake, line 264: Instead of overriding delegate1 & delegate2, it might be worth declaring new vars, e.g. delegate1AfterUpdate or something similar (overriding variables can cause confusion at times)

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests.cs
```diff
@@ -240,5 +240,38 @@ public void UpdateDelegateUserCentrePrompts_updates_user()
...
+                    // Given
+                    var ids = new[] { 16, 28 };
+                    var delegate1 = userDataService.GetDelegateUserById(16);
+                    var delegate2 = userDataService.GetDelegateUserById(28);
+
+                    delegate1.Approved.Should().BeFalse();
+                    delegate2.Approved.Should().BeFalse();
+
+                    // When
+
```
stellake, line 260: NIP, but no need for empty line here

DigitalLearningSolutions.Data/Services/DelegateApprovalsService.cs
```diff
@@ -1,25 +1,39 @@
...
 
     public class DelegateApprovalsService : IDelegateApprovalsService
     {
         private readonly IUserDataService userDataService;
         private readonly ICustomPromptsService customPromptsService;
+        private readonly IEmailService emailService;
+        private readonly ILogger<DelegateApprovalsService> logger;
 
-        public DelegateApprovalsService(IUserDataService userDataService, ICustomPromptsService customPromptsService)
+        public DelegateApprovalsService(IUserDataService userDataService, ICustomPromptsService customPromptsService,
```
stellake, line 30: Could we run the formatter here?

DigitalLearningSolutions.Data/Services/DelegateApprovalsService.cs
```diff
@@ -30,5 +44,69 @@ public List<(DelegateUser delegateUser, List<CustomPromptWithAnswer> prompts)> G
...
+
+        public void ApproveAllUnapprovedDelegatesForCentre(int centreId)
+        {
+            var delegateUsers = userDataService.GetUnapprovedDelegateUsersByCentreId(centreId);
+
+            userDataService.ApproveDelegateUsers(delegateUsers.Select(d => d.Id));
+            foreach (var delegateUser in delegateUsers)
+            {
+                SendApprovalEmail(delegateUser);
+            }
```
stellake, line 75: I wonder if we could have SendEmails method in the EmailService so we don't need to connect-disconnect many times (as I suspect that's the bottleneck in terms of performance)

DigitalLearningSolutions.Data/Services/DelegateApprovalsService.cs
```diff
@@ -30,5 +44,69 @@ public List<(DelegateUser delegateUser, List<CustomPromptWithAnswer> prompts)> G
...
+            string candidateNumber,
+            string emailAddress)
+        {
+            string emailSubject = "Digital Learning Solutions Registration Approved";
+
+            var body = new BodyBuilder
+                {
+                    TextBody = $@"Your Digital Learning Solutions registration has been approved by your centre administrator.
+                            You can now login to the Digital Learning Solutions learning materials using your e-mail address or your Delegate ID number <b>""{candidateNumber}""</b> and the password you chose during registration.
+                            For more assistance in accessing the materials, please contact your Digital Learning Solutions centre.",
```
stellake, line 101: Might be worth adding a link to the sign in page here - could you send a screenshot of the email to the HEE external Slack and ask if it's something they'd like? :)

DigitalLearningSolutions.Data/Services/DelegateApprovalsService.cs
```diff
@@ -30,5 +44,69 @@ public List<(DelegateUser delegateUser, List<CustomPromptWithAnswer> prompts)> G
...
+                var delegateApprovalEmail = GenerateDelegateApprovalEmail(delegateUser.CandidateNumber, delegateUser.EmailAddress);
+                emailService.SendEmail(delegateApprovalEmail);
+            }
+        }
+
+        private static Email GenerateDelegateApprovalEmail(
+            string candidateNumber,
+            string emailAddress)
+        {
+            string emailSubject = "Digital Learning Solutions Registration Approved";
```
stellake, line 95: Can be a const :)

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/DelegateApprovalsController.cs
```diff
@@ -27,8 +27,24 @@ public IActionResult Index()
...
+            return RedirectToAction("Index", "DelegateApprovals");
+        }
+
+        [HttpPost]
+        [Route("/TrackingSystem/Delegates/Approve/ApproveForCentre")]
+        public IActionResult ApproveDelegatesForCentre(int centreId)
+        {
+            delegateApprovalsService.ApproveAllUnapprovedDelegatesForCentre(centreId);
+            return RedirectToAction("Index", "DelegateApprovals");
+        }
```
stellake, line 48: I don't think the user should ever be able to approve delegates in a centre that's other than what they've logged into, so might make sense to make this a GET endpoint /TrackingSystem/Delegates/Approve/All 

In fact, you could also avoid the POST in single delegate approve by adding the ID as a query string. Then the buttons can just be links to the relevant URLs.

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests.cs
```diff
@@ -240,5 +240,38 @@ public void UpdateDelegateUserCentrePrompts_updates_user()
                 }
```
stellake, line 240: Nice test! Could we also add some service and controller tests?

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests.cs
JBsoftwire: I prefer the explicit bracketing to make it clear where the using block is, do we prefer the alternative?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/DelegateApprovalsController.cs
JBsoftwire: I wanted to use POST since that's the type of operation being performed, but I'm not wedded to it. However, I can just get the centreId from the user claims in this endpoint either way.

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests.cs
stellake: I think the bracketing can be useful in a place where we need to distinguish or emphasise where it's used, whereas in these tests it's only causing more noise that makes it harder to follow. I'm not feeling super strongly about it so if you think for you it makes harder to follow the other way around, I'm happy to leave it (as long as there are no IDE warnings)

DigitalLearningSolutions.Data.Tests/Services/DelegateApprovalsServiceTests.cs
```diff
@@ -0,0 +1,129 @@
...
+
+            A.CallTo(() => userDataService.GetDelegateUserByIdFromCentre(2, 2)).Returns(expectedDelegateUser);
+            A.CallTo(() => userDataService.ApproveDelegateUsers(A<IEnumerable<int>>.That.IsSameSequenceAs(new[] { 2 }))).DoesNothing();
+            A.CallTo(() => emailService.SendEmail(A<Email>._)).DoesNothing();
+
+            // When
+            delegateApprovalsService.ApproveDelegate(2, 2);
+
+            // Then
+            A.CallTo(() => userDataService.ApproveDelegateUsers(A<IEnumerable<int>>.That.IsSameSequenceAs(new[] { 2 }))).MustHaveHappened();
```
stellake, line 71: Really minor, but I think IsSameSequenceAs can be just used with values, e.g.
`.IsSameSequenceAs(2) or IsSameSequenceAs(2, 3) `

This removes the extra syntax around creating an array :)

DigitalLearningSolutions.Data/DataServices/UserDataService.cs
```diff
@@ -105,6 +108,13 @@ public UserDataService(IDbConnection connection)
             return user;
         }
 
+        public DelegateUser? GetDelegateUserByIdFromCentre(int delegateId, int centreId)
+        {
+            var user = GetDelegateUserById(delegateId);
+
+            return user?.CentreId == centreId ? user : null;
+        }
```
stellake, line 116: This feels like the logic that should live in a service rather than a data service/repo

DigitalLearningSolutions.Data/Services/DelegateApprovalsService.cs
```diff
@@ -1,34 +1,135 @@
...
+                    LogNoEmailWarning(delegateUser.Id);
+                }
+                else
+                {
+                    var delegateApprovalEmail = GenerateDelegateApprovalEmail
+                        (delegateUser.CandidateNumber, delegateUser.EmailAddress);
+                    approvalEmails.Add(delegateApprovalEmail);
+                }
+            }
+            emailService.SendEmails(approvalEmails);
```
stellake, line 103: Might be worth pulling this into a private method "SendDelegateApprovalEmails" (+ in the single approval case). Then, the high level purpose of this method gets clearer to read :)

DigitalLearningSolutions.Data/Services/EmailService.cs
```diff
@@ -29,6 +31,55 @@ public class EmailService : IEmailService
...
+        }
+
+        public void SendEmails(IEnumerable<Email> emails)
+        {
+            var mailConfig = GetMailConfig();
+
+            using var client = smtpClientFactory.GetSmtpClient();
+            client.Timeout = 10000;
+            client.Connect(mailConfig.MailServerAddress, mailConfig.MailServerPort);
+            client.Authenticate(mailConfig.MailServerUsername, mailConfig.MailServerPassword);
```
stellake, line 63: What happens if connection to email client fails (is removing the try/catch intentional)?

DigitalLearningSolutions.Data/Services/EmailService.cs
```diff
@@ -29,6 +31,55 @@ public class EmailService : IEmailService
...
+                catch (Exception error)
+                {
+                    logger.LogError(error, "Sending an email has failed");
+                }
+            }
+
+            client.Disconnect(true);
+        }
+
+        private (string MailServerUsername, string MailServerPassword, string MailServerAddress, int MailServerPort,
```
stellake, line 81: Could we run the formatter here? 🙏

DigitalLearningSolutions.Data/Services/EmailService.cs
JBsoftwire: I moved the try-catch inside so it could deal with emails individually, but there also needs to be one outside for this case, whoops.

DigitalLearningSolutions.Data/Services/DelegateApprovalsService.cs
```diff
@@ -1,34 +1,133 @@
...
+            }
+
+            if (delegateUser.Approved)
+            {
+                logger.LogWarning($"Delegate user id {delegateId} already approved.");
+            }
+            else
+            {
+                var delegateList = new List<DelegateUser>{ delegateUser };
+                userDataService.ApproveDelegateUsers(delegateList.Select(d => d.Id));
```
stellake, line 69: Could this just use the delegateId (or delegateUser.Id) as a parameter?

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/Delegates/DelegateApprovalsViewModel.cs
```diff
@@ -9,11 +9,13 @@
...
     public class DelegateApprovalsViewModel
     {
-        public DelegateApprovalsViewModel(IEnumerable<UnapprovedDelegate> delegates)
+        public DelegateApprovalsViewModel(int centreId, IEnumerable<UnapprovedDelegate> delegates)
         {
+            CentreId = centreId;
             Delegates = delegates;
         }
 
+        public int CentreId { get; set; }
```
stellake, line 18: Do we need the CentreId anymore?

DigitalLearningSolutions.Data/Services/DelegateApprovalsService.cs
JBsoftwire: I think you're right - I can just use `params` myself

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/DelegateApprovalsController.cs
```diff
@@ -27,8 +27,26 @@ public IActionResult Index()
...
+        [Route("/TrackingSystem/Delegates/Approve")]
+        public IActionResult ApproveDelegate(int delegateId)
+        {
+            var centreId = User.GetCentreId();
+            delegateApprovalsService.ApproveDelegate(delegateId, centreId);
+            return RedirectToAction("Index", "DelegateApprovals");
+        }
+
+        [HttpPost]
+        [Route("/TrackingSystem/Delegates/Approve/ApproveForCentre")]
```
stellake, line 44: Could this route be `../Delegates/Approve/All` as everything on the whole site happens per centre (so we don't need for centre everywhere)

DigitalLearningSolutions.Data/Services/EmailService.cs
```diff
@@ -1,115 +1,183 @@
...
+            {
+                logger.LogError(error, "Sending emails has failed");
+            }
+        }
+
+        private void SendSingleEmailFromClient
+        (
+            Email email,
+            (string MailServerUsername, string MailServerPassword, string MailServerAddress, int MailServerPort, string
+                MailSenderAddress) mailConfig,
```
stellake, line 88: Is the whole mailConfig required here as a param?

stellake (approved):
Looks good - just a comment about the mail config as a param + there are some merge conflicts! 👍 
No need for rereview though unless the merge conflicts are bad (if so, let me know) :)

PR #372, opened 2021-05-28
HEEDLS-379 - Changed centre website details display to better method …

DanBloxham-sw:
…for preserving whitespace.

Instead of doing some ugly conversion to an array, I realised I could just use css instead. I've replaced the old complicated way with the nice new way.

Ran all unit tests, tested on chrome, and with screen reader.


DigitalLearningSolutions.Web/Styles/trackingSystem/centreConfiguration.scss
```diff
@@ -1,24 +1,28 @@
...
+    max-height: 250px;
+  }
+
+  @include govuk-media-query($until: tablet) {
+    max-width: 250px;
+    max-height: 200px;
+  }
+}
+
+.preserve-whitespace{
```
DanBloxham-sw, line 26: This new class is the only actual change. I double checked the file and it is now CRLF.

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/Index.cshtml
```diff
@@ -216,7 +216,7 @@
           Training venues
         </dt>
         <dd class="nhsuk-summary-list__value">
-          <partial name="_CentreStringArray" model="Model.TrainingVenues"/>
+          <span class="preserve-whitespace">@Model.TrainingVenues</span>
```
stellake, line 219: Is the span needed/could we add the preserve-whitespace class on the dd element?

DanBloxham-sw: I tried that first. The autoformatting will indent the value in the content on the dd on a new line, the indent will add extra space that will then be preserved in the HTML element, so that span does seem necessary.

stellake (approved):
Looks good! 👍

PR #373, opened 2021-05-28
HEEDLS-447 - Remove Preview On Website button when ShopOnMap is false.

DanBloxham-sw:
Ran all unit tests.
Manually tested on chrome.


stellake (approved):
Looks good! 👍

PR #376, opened 2021-06-02
Heedls 454 add prompt summary page

AlexJacksonDS:
Adds the final summary page and methods to save the new prompt to the centre

Checked that the layout looks OK for mobile and ran all unit tests.

image

image


DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
```diff
@@ -143,6 +143,37 @@ string action
             };
         }
 
+        [HttpGet]
+        [Route("Add/Summary")]
+        [ServiceFilter(typeof(RedirectEmptySessionData<AddRegistrationPromptData>))]
+        public IActionResult AddRegistrationPromptSummary()
+        {
+            var data = TempData.Peek<AddRegistrationPromptData>()!;
+            var promptName = customPromptsService.GetCustomPromptsAlphabeticalList().
```
DanBloxham-sw, line 152: I'm not sure how often this GetCustomPromptsAlphabeticalList is called only to recover a single value - if this occurs multiple times, do you think we should make a different method just to get a specific prompt by Id?

DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsControllerTests.cs
```diff
@@ -276,6 +276,31 @@ public void AddRegistrationPromptConfigureAnswers_returns_error_with_unexpected_
             result.Should().BeRedirectToActionResult().WithControllerName("LearningSolutions").WithActionName("Error");
         }
 
+        [Test]
+        public void AddRegistrationPromptSummary_calls_custom_prompt_service_and_redirects_to_index()
```
DanBloxham-sw, line 280: The title of this test implies to me that we test both that the service is called and the redirect occurs, while the test only appears to test the redirect. I think either the title should be changed to match the test, or a new assertion should be added to ensure that the service is called as expected.

DanBloxham-sw (commented):
Few minor questions.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
AlexJacksonDS: This is the only place where that is happening.

DanBloxham-sw (approved):
Looks good to me.

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/AddRegistrationPromptSummary.cshtml
```diff
@@ -0,0 +1,78 @@
...
+    <dl class="nhsuk-summary-list">
+      <div class="nhsuk-summary-list__row">
+        <dt class="nhsuk-summary-list__key">
+          Prompt
+        </dt>
+        <dd class="nhsuk-summary-list__value">
+          @Model.PromptName
+        </dd>
+        <dd class="nhsuk-summary-list__actions">
+          <a class="nhsuk-link" asp-action="AddRegistrationPromptSelectPrompt">
```
stellake, line 32: Ah, looks like you also have the nhsuk-links that are not in official examples :)

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/AddRegistrationPromptSummary.cshtml
```diff
@@ -0,0 +1,78 @@
...
+        <dd class="nhsuk-summary-list__value">
+          @if (Model.Answers.Any()) {
+            <ul>
+              @foreach (var answer in Model.Answers)
+              {
+                <li class="word-break">@answer</li>
+              }
+            </ul>
+          } else {
+            <partial name="_NoConfiguredAnswers"/>
```
stellake, line 63: Should we just leave it empty in the summary or maybe write something fairly short here given that it's a summary? Do you have any screenshots of the current version?

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/CentreConfiguration/AddRegistrationPromptSummaryViewModel.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace DigitalLearningSolutions.Web.ViewModels.TrackingSystem.CentreConfiguration
+{
+    using System.Collections.Generic;
+    using DigitalLearningSolutions.Web.Helpers;
+    using DigitalLearningSolutions.Web.Models;
+
+    public class AddRegistrationPromptSummaryViewModel
+    {
+        public AddRegistrationPromptSummaryViewModel(AddRegistrationPromptData data, string promptName
```
stellake, line 9: Ah - we might want to run a formatter here :)

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/CentreConfiguration/AddRegistrationPromptSummaryViewModel.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace DigitalLearningSolutions.Web.ViewModels.TrackingSystem.CentreConfiguration
```
stellake, line 1: Re folder structuring - should we add a /RegistrationPrompts folder under centre configuration like we do for views to keep things consistent?

AlexJacksonDS: I was going to sort that in the next ticket (381) that I haven't put for review yet as it needs methods from this one. Should I also move them to /Centre/Reg.... rather than /CentreConfiguration/Reg... at the same time?

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/AddCentreCustomPromptsAccessibilityTests.cs
```diff
@@ -18,20 +18,30 @@ public void AddRegistrationPrompt_journey_has_no_accessibility_errors()
...
             AddAnswer("Answer 1");
             AddAnswer("Answer 2");
+            AnalyzePageHeading("Configure answers");
             var configureAnswerWithAnswersResult = new AxeBuilder(Driver).Analyze();
+
             Driver.ClickButtonByText("Next");
 
+            AnalyzePageHeading("Summary");
+            var summaryResult = new AxeBuilder(Driver).Analyze();
+            Driver.SubmitForm();
```
stellake, line 38: Hmm.. Could we skip the "SubmitForm" step so we don't keep on adding the new prompts into the test db?

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/AccessibilityTestsBase.cs
```diff
@@ -22,5 +24,20 @@ public void Dispose()
...
+        public void AnalyzePageHeadingAndAccessibility(string pageTitle)
+        {
+            AnalyzePageHeading(pageTitle);
+
+            // then
+            var axeResult = new AxeBuilder(Driver).Analyze();
+            axeResult.Violations.Should().BeEmpty();
+        }
+
+        public void AnalyzePageHeading(string pageTitle)
```
stellake, line 37: Maybe ValidatePageHeading?

DigitalLearningSolutions.Data/Services/CustomPromptsService.cs
```diff
@@ -72,6 +74,23 @@ public List<(int id, string value)> GetCustomPromptsAlphabeticalList()
...
+                .OrderBy(n => n);
+
+            var promptNumbers = new List<int> { 1, 2, 3, 4, 5, 6 };
+            promptNumbers.RemoveAll(x => existingPromptNumbers.Contains(x));
+            var promptNumber = promptNumbers.Any() ? promptNumbers.Min() : (int?)null;
+
+            if (promptNumber != null)
+            {
+                customPromptsDataService.AddCustomPromptToCentre(centreId, promptNumber.Value, promptId, mandatory, options);
+            }
```
stellake, line 91: We might want to handle a failure with a warning log? And maybe redirect the user to the "Unexpected error" page?

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/CentreConfiguration/AddRegistrationPromptSummaryViewModel.cs
stellake: Yes please! But as part of a new ticket sounds good! 👍

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/AddRegistrationPromptSummary.cshtml
AlexJacksonDS: image

DigitalLearningSolutions.Data/Services/CustomPromptsService.cs
AlexJacksonDS: Agreed on the logging. The situations that this would occur seems unlikely but I guess it is better to handle it.

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/AddRegistrationPromptSummary.cshtml
stellake: Oh actually it doesn't look bad so let's leave it as is! 👍  Thanks for the screenshot! :)

DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsControllerTests.cs
```diff
@@ -276,6 +276,39 @@ public void AddRegistrationPromptConfigureAnswers_returns_error_with_unexpected_
             result.Should().BeRedirectToActionResult().WithControllerName("LearningSolutions").WithActionName("Error");
         }
 
+        [Test]
```
stellake, line 279: I'd probably add a quick check that the TempData is cleared into this test.

Maybe also worth adding a quick test for the error page redirect that we've added?

DigitalLearningSolutions.Data/Services/CustomPromptsService.cs
```diff
@@ -72,7 +96,36 @@ public List<(int id, string value)> GetCustomPromptsAlphabeticalList()
...
+        public bool AddCustomPromptToCentre(int centreId, int promptId, bool mandatory, string? options)
+        {
+            var centreCustomPrompts = GetCustomPromptsForCentreByCentreId(centreId);
+            var existingPromptNumbers = centreCustomPrompts.CustomPrompts
+                .Select(c => c.CustomPromptNumber)
+                .OrderBy(n => n);
+
+            var promptNumbers = new List<int> { 1, 2, 3, 4, 5, 6 };
+            promptNumbers.RemoveAll(x => existingPromptNumbers.Contains(x));
+            var promptNumber = promptNumbers.Any() ? promptNumbers.Min() : (int?)null;
```
stellake, line 108: I'm wondering if we could use "Except" syntax here: https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.except?view=netcore-3.1

Also, do we need to order the "existingPromptNumbers if we do .Min() later to get the promptNumber?

DigitalLearningSolutions.Data.Tests/Services/CustomPromptsServiceTests.cs
```diff
@@ -1,174 +1,288 @@
...
+            // Then
+            A.CallTo
+            (
+                () => customPromptsDataService.AddCustomPromptToCentre(1, 2, 1, true, null)
+            ).MustHaveHappened();
+            result.Should().BeTrue();
+        }
+
+        [Test]
+        public void AddCustomPromptToCentre_add_prompt_with_all_prompts_defined()
```
stellake, line 257: These tests are great, but I think we could improve the title namings a bit to make it clearer why the test fails/what the expected result is. For example:
AddCustomPromptToCentre_does_not_add_prompt_if_centre_has_all_prompts_defined
AddCustomPromptToCentre_add_prompt_at_lowest_possible_prompt_number_with_gaps_in_prompt_number

etc

stellake (approved):
Looks good! 👍

PR #378, opened 2021-06-02
Heedls 424 reject delegates

JBsoftwire:


DigitalLearningSolutions.Data/DataServices/UserDataService.cs
```diff
@@ -294,5 +297,61 @@ public void ApproveDelegateUsers(params int[] ids)
...
+	                    WHERE ProgressID IN (SELECT ProgressID FROM Progress WHERE CandidateID = @delegateId)
+
+                    DELETE FROM ProgressContributors
+	                    WHERE CandidateID = @delegateId
+
+                    DELETE FROM Progress
+                        WHERE CandidateID = @delegateId
+
+                    DELETE FROM Candidates
+                        WHERE CandidateID = @delegateId",
```
AlexJacksonDS, line 352: Do we want to wrap this in a transaction just in case something somehow goes wrong. (I don't really see how it would)

DigitalLearningSolutions.Data/Services/DelegateApprovalsService.cs
```diff
@@ -128,5 +160,34 @@ string emailAddress
...
+        private static Email GenerateDelegateRejectionEmail(
+            string? delegateName,
+            string emailAddress)
+        {
+            string emailSubject = "Digital Learning Solutions registration rejected"; // TODO AIR-424 capitalisation inconsistent with approvals
+
+            var body = new BodyBuilder
+            {
+                TextBody = $@"Dear {delegateName},
+                        Your Digital Learning Solutions (DLS) registration at the centre [Centre Name] has been rejected by an administrator.There are several reasons that this may have happened including:
```
AlexJacksonDS, line 173: Is [Centre Name] supposed to be an interpolated value? Similarly in the Html version

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateApprovals/DelegateRejectionPage.cshtml
```diff
@@ -0,0 +1,60 @@
...
+@using Microsoft.Extensions.Configuration
+@model RejectDelegateViewModel
+
+<link rel="stylesheet" href="@Url.Content("~/css/shared/cardWithButtons.css")" asp-append-version="true">
+<link rel="stylesheet" href="@Url.Content("~/css/trackingSystem/delegateApprovals.css")" asp-append-version="true">
+
+@{
+  ViewData["Title"] = "Reject delegate registration";
+  ViewData["Application"] = "Tracking System";
+  ViewData["HeaderPath"] = $"{Configuration["CurrentSystemBaseUrl"]}/tracking/dashboard";
```
AlexJacksonDS, line 13: This also needs a ViewData["HeaderPathName"] = "Tracking System";

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateApprovals/DelegateRejectionPage.cshtml
```diff
@@ -0,0 +1,60 @@
...
+}
+
+@section NavMenuItems {
+  <partial name="~/Views/TrackingSystem/Shared/_NavMenuItems.cshtml" />
+}
+
+<div class="nhsuk-grid-row">
+  <div class="nhsuk-grid-column-full">
+    <h1 class="nhsuk-heading-xl">Reject delegate registration</h1>
+    <h1 class="nhsuk-body-l">Are you sure you want to reject this user's registration?</h1>
```
AlexJacksonDS, line 23: I don't think this should be in a separate h1 element, but I'm not sure. Is there a reason why it is rather than being in a \<p> or \<h2> etc.?

AlexJacksonDS (commented):
Mostly fine, a couple of missing things

DigitalLearningSolutions.Data/DataServices/UserDataService.cs
JBsoftwire: No real reason not to I think, and I always favour paranoia around deletes.

JBsoftwire:
image
image
image

AlexJacksonDS (approved):
Looks good to me know, just need to sort out the merge conflicts as well at some point

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateApprovals/DelegateRejectionPage.cshtml
```diff
@@ -0,0 +1,61 @@
...
+@using Microsoft.Extensions.Configuration
+@model RejectDelegateViewModel
+
+<link rel="stylesheet" href="@Url.Content("~/css/shared/cardWithButtons.css")" asp-append-version="true">
+<link rel="stylesheet" href="@Url.Content("~/css/trackingSystem/delegateApprovals.css")" asp-append-version="true">
+
+@{
+    ViewData["Title"] = "Reject delegate registration";
+    ViewData["Application"] = "Tracking System";
+    ViewData["HeaderPath"] = $"{Configuration["CurrentSystemBaseUrl"]}/tracking/dashboard";
```
stellake, line 13: Ah - the new dashboard page should be on master now :)

DigitalLearningSolutions.Web/Styles/index.scss
```diff
@@ -29,6 +30,20 @@ a.delete-button:active {
...
+  margin-bottom: 16px !important;
+}
+
+button.delete-button:hover {
+  background-color: #aa2016;
+}
+
+button.delete-button:active {
+  background-color: #6b140e;
+}
```
stellake, line 45: Could we potentially commonise these styles with the a.delete-button? Also, I think the #d5281b colour is the nhsuk-error-colour:
https://service-manual.nhs.uk/design-system/styles/colour

DigitalLearningSolutions.Data/Services/DelegateApprovalsService.cs
```diff
@@ -128,5 +160,35 @@ string emailAddress
 
             return new Email(emailSubject, body, emailAddress);
         }
+
+        private static Email GenerateDelegateRejectionEmail(
+            string? delegateName,
+            string centreName,
+            string emailAddress)
+        {
+            string emailSubject = "Digital Learning Solutions registration rejected"; // TODO AIR-424 capitalisation inconsistent with approvals
```
stellake, line 169: AIR-424? 👁️ 

But also, lets use the approvals capitalisation here 👍

DigitalLearningSolutions.Data/Services/DelegateApprovalsService.cs
```diff
@@ -128,5 +160,35 @@ string emailAddress
...
+        private static Email GenerateDelegateRejectionEmail(
+            string? delegateName,
+            string centreName,
+            string emailAddress)
+        {
+            string emailSubject = "Digital Learning Solutions registration rejected"; // TODO AIR-424 capitalisation inconsistent with approvals
+
+            var body = new BodyBuilder
+            {
+                TextBody = $@"Dear {delegateName},
```
stellake, line 173: Both bodys need a space before "There are several reasons..."

DigitalLearningSolutions.Data/Services/DelegateApprovalsService.cs
```diff
@@ -128,5 +160,34 @@ string emailAddress
...
+                        Your Digital Learning Solutions (DLS) registration at the centre [Centre Name] has been rejected by an administrator.There are several reasons that this may have happened including:
+                        •You registered with a non-work email address which was not recognised by the administrator
+                        •Your DLS centre chooses to manage delegate registration internally
+                        •You have accidentally chosen the wrong centre during the registration process.
```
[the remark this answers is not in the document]
stellake: This seems to suggest that dashes are used instead:
https://www.litmus.com/blog/best-practices-for-plain-text-emails-a-look-at-why-theyre-important/

DigitalLearningSolutions.Data/Services/DelegateApprovalsService.cs
```diff
@@ -128,5 +160,35 @@ string emailAddress
...
+            string emailSubject = "Digital Learning Solutions registration rejected"; // TODO AIR-424 capitalisation inconsistent with approvals
+
+            var body = new BodyBuilder
+            {
+                TextBody = $@"Dear {delegateName},
+                        Your Digital Learning Solutions (DLS) registration at the centre {centreName} has been rejected by an administrator.There are several reasons that this may have happened including:
+                        •You registered with a non-work email address which was not recognised by the administrator
+                        •Your DLS centre chooses to manage delegate registration internally
+                        •You have accidentally chosen the wrong centre during the registration process.
+                        If you need access to the DLS platform, please use the <a>Find Your Centre</a> page to locate your local DLS centre and use the contact details provided to ask for help with registration.",
```
stellake, line 178: Need to add the Find your centre url at the end here and specify href below. :)

stellake:
UI wise, the new page looks good, but the "Cancel" button looks too close to the "Reject" button. Could we space it out a bit, such that the spacing between the user summary and Reject button is not bigger than the spacing between the "Reject" and "Cancel"?

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateApprovals/DelegateRejectionPage.cshtml
```diff
@@ -0,0 +1,61 @@
+﻿@inject IConfiguration Configuration
+
+@using DigitalLearningSolutions.Web.ViewModels.TrackingSystem.Delegates
+@using Microsoft.Extensions.Configuration
+@model RejectDelegateViewModel
+
+<link rel="stylesheet" href="@Url.Content("~/css/shared/cardWithButtons.css")" asp-append-version="true">
```
stellake, line 7: Are we using card with buttons on this page? 🤔

DigitalLearningSolutions.Data/Services/DelegateApprovalsService.cs
```diff
@@ -99,6 +100,37 @@ private void SendDelegateApprovalEmails(params DelegateUser[] delegateUsers)
 
             emailService.SendEmails(approvalEmails);
         }
+        public void RejectDelegate(int delegateId, int centreId)
```
stellake, line 103: NIP, but missing an empty line above

DigitalLearningSolutions.Data/Services/DelegateApprovalsService.cs
```diff
@@ -99,6 +100,37 @@ private void SendDelegateApprovalEmails(params DelegateUser[] delegateUsers)
...
+        {
+            var delegateUser = userDataService.GetDelegateUserById(delegateId);
+
+            if (delegateUser == null || delegateUser.CentreId != centreId)
+            {
+                throw new UserAccountNotFoundException($"Delegate user id {delegateId} not found at centre id {centreId}.");
+            }
+            if (delegateUser.Approved)
+            {
+                logger.LogWarning($"Delegate user id {delegateId} cannot be rejected as they are already approved.");
```
stellake, line 113: I wonder if in addition to logging we want to take the user to an "Unexpected error" page in this case, just to demonstrate that their Rejection action has not been successful..

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateApprovals/_unapprovedDelegateExpandable.cshtml
```diff
@@ -52,7 +52,11 @@
         Approve
       </button>
     </form>
-    <a class="nhsuk-button delete-button nhsuk-u-margin-bottom-0" href="" role="button">
+    <a class="nhsuk-button delete-button nhsuk-u-margin-bottom-0"
+       asp-controller="DelegateApprovals"
+       asp-action="DelegateRejectionPage"
+       asp-route-delegateId="@Model.Id"
+       role="button">
```
davidm-m, line 59: Came here because I stumbled upon this being an `<a>` while testing something else. Is there a reason this can't be a button? If we're using an `<a>` as a button it's not enough to just add `role="button"` to it, in particular it will still be missing keyboard controls e.g. spacebar to submit. See the [MDN docs on the button role here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role).
We can replace that functionality using js but that's extra complexity and breaks if someone has js disabled; ideally we would just always use a button.

stellake: Ah, great question! 

If we used a button to navigate the site, it would have to either have an onclick event of navigating the site using JS (setting location href, but this approach is JS dependent) or a form that needs to be submitted (which is clunky and not following MVC standard way of navigating the site). If the links fail AA accessibility standards we should either:
* Fix it with more spec than role="button" (note that role="button" seems to be used on Register to vote which is either AA or AAA compliant - can't remember exactly).
* Use links instead, which is something that HEE didn't want in terms of the design choice.

davidm-m: RtV is AAA rated I believe. Surprising that they can get away with that (but I agree they definitely do this!). I'm struggling to find the actual requirement to see whether it falls under AA or AAA or whatever but in all the w3c docs they say you must add support for enter and spacebar, and [here](https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20160317/examples/button/button.html) they specifically call an `<a>` without keypress handlers "non-compliant".

davidm-m: Also I'm happy to concede that I don't understand enough of the context here but since the approve button is using a form I don't see why we can't make it a multi-button form with an approve action and reject action.

stellake: Ah - I think we can change the endpoint to POST & use the form submit for consistency with Reject + compliance with the rules you found, but I don't think we'll be changing e.g. Log in and Register buttons into forms, if they're not truly forms and just for navigation 👍 :)

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateApprovals/_unapprovedDelegateExpandable.cshtml
```diff
@@ -46,14 +46,14 @@
...
   </details>
   <div class="flex-row">
-    <form class="delegate-approvals-button-wrapper" asp-controller="DelegateApprovals" asp-action="ApproveDelegate" method="post">
+    <form class="delegate-approvals-button-wrapper" method="post">
       <input type="hidden" name="delegateId" value="@Model.Id"/>
-      <button class="nhsuk-button nhsuk-u-margin-bottom-0 nhsuk-u-margin-right-2"  type="submit">
+      <button class="nhsuk-button nhsuk-u-margin-bottom-0 nhsuk-u-margin-right-2" asp-controller="DelegateApprovals" asp-action="ApproveDelegate">
         Approve
       </button>
+      <button class="nhsuk-button delete-button nhsuk-u-margin-bottom-0" asp-controller="DelegateApprovals" asp-action="DelegateRejectionPage">
```
stellake, line 54: Should the asp-action name be consistent with the approval (ApproveDelegate & RejectDelegate)?

stellake: Actually ignore this!

DigitalLearningSolutions.Data/DataServices/UserDataService.cs
```diff
@@ -340,6 +344,31 @@ public void ApproveDelegateUsers(params int[] ids)
             );
         }
 
+        public void RemoveDelegateUser(int delegateId)
+        {
+            using var transaction = new TransactionScope();
```
stellake, line 349: Just double checking that the register stored procedure doesn't create Progress and aspProgress entry (Kevin seemed to think it might, but might have misremembered)?

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests.cs
```diff
@@ -266,6 +278,30 @@ public void ApproveDelegateUsers_approves_delegate_users()
             }
         }
 
+        [Test]
+        public void RemoveDelegateUser_deletes_delegate_user()
```
stellake, line 282: Should we also write a test for not removing the delegate user if they've got progress?

stellake (commented):
Looks good to me assuming there are no issues with the "Progress/aspProgress" tables - just a question of whether there's value adding a test for the case where user has logged in before rejection. 👍

DigitalLearningSolutions.Data/DataServices/UserDataService.cs
JBsoftwire: I didn't think so but I'll take another look, best to check carefully

JBsoftwire: No hits for 'progress' in https://github.com/TechnologyEnhancedLearning/DLS-LEGACY/blob/master/database/dbo.uspSaveNewCandidate_V10.StoredProcedure.sql

stellake: As discussed - no issues with this.

stellake (approved):

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests.cs
JBsoftwire: I think we should, better safe than errors in the live DB

DigitalLearningSolutions.Data/DataServices/UserDataService.cs
```diff
@@ -340,6 +346,44 @@ public void ApproveDelegateUsers(params int[] ids)
...
+                    @"SELECT SessionID FROM Sessions WHERE CandidateID = @delegateId",
+                    new { delegateId }
+                );
+
+                if (existingSessions.Any())
+                {
+                    throw new UserAccountInvalidStateException(
+                        $"Delegate user id {delegateId} cannot be removed as they have already started a session."
+                        );
+                }
```
stellake, line 364: Oh sorry - I thought we'd change this at a db level, but at the same time, I think it might be worth getting the current state merged and then sorting it out in a different MR as Alex's ticket uses a few bits from here. (changing the DELETE_RULE for the Sessions table in the db - can we do it via a migration?)

stellake (approved):
Happy with this - let's do the db level delete_rule change as part of a differnet MR 👍

PR #381, opened 2021-06-03
HEEDLS-457 added basic reports page

JBsoftwire:


DigitalLearningSolutions.Web/Startup.cs
```diff
@@ -79,6 +79,7 @@ public void ConfigureServices(IServiceCollection services)
                 .AddRazorOptions(options =>
                 {
                     options.ViewLocationFormats.Add("/Views/TrackingSystem/{1}/{0}.cshtml");
```
stellake, line 81: Can we remove this option, as you mentioned earlier? :)

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Reports/Index.cshtml
```diff
@@ -0,0 +1,14 @@
+﻿@using DigitalLearningSolutions.Web.ViewModels.TrackingSystem.Centre.Reports
+
+@model ReportsViewModel
+
```
stellake, line 4: I think we also need to add page title, application name etc?

DigitalLearningSolutions.Web/Helpers/CustomPolicies.cs
```diff
@@ -5,7 +5,7 @@
     public class CustomPolicies
     {
         public const string UserOnly = "UserOnly";
-        public const string UserCentreAdminOnly = "UserCentreAdminOnly";
+        public const string UserCentreAdmin = "UserCentreAdminOnly";
```
stellake, line 8: Do we also want to change the string value?

JBsoftwire:
screenshots:
image
image

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Centre/Reports/ReportsController.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace DigitalLearningSolutions.Web.Controllers.TrackingSystem.Centre.Reports
```
stellake, line 1: I don't mind too much, but I suspect we won't need a "Reports" folder/can put ReportsController directly into the Centre folder as we don't have too many report related pages.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Centre/Reports/ReportsController.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace DigitalLearningSolutions.Web.Controllers.TrackingSystem.Centre.Reports
+{
+    using DigitalLearningSolutions.Web.Helpers;
+    using DigitalLearningSolutions.Web.ViewModels.TrackingSystem.Centre.Reports;
+    using Microsoft.AspNetCore.Authorization;
+    using Microsoft.AspNetCore.Mvc;
+
+    [Authorize(Policy = CustomPolicies.UserCentreAdmin)]
+    [Route("/TrackingSystem/Centre/Reports")]
+    public class ReportsController : Controller
```
stellake, line 10: Could we add this page to automated UI tests?

DigitalLearningSolutions.Web/Startup.cs
JBsoftwire: We can't - it breaks one of the other controllers. I assumed that was being changed in the other branch.

DigitalLearningSolutions.Web/Helpers/CustomPolicies.cs
JBsoftwire: I decided not to because I didn't know if it interacted with anything besides this, but after checking I don't think it does.

DigitalLearningSolutions.Web/Startup.cs
stellake: Ah, fair enough! Thanks :)

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Centre/Reports/ReportsController.cs
JBsoftwire: I'm inclined to leave it there right now just to avoid any issues with renaming index methods

stellake (approved):
Looks good! 👍

PR #383, opened 2021-06-03
HEEDLS-422 break words to avoid issues on mobile view

JBsoftwire:
A small fix for issues with card text with long words on small screens.


stellake (approved):
Looks good!

PR #384, opened 2021-06-03
HEEDLS-471 Add placeholder images to centre configuration

AlexJacksonDS:
Added a new image placeholder for the centre configuration and a new set of sizes in css for these.

Checked they work OK in chrome edge and IE.

image

image


DigitalLearningSolutions.Web/Styles/trackingSystem/centreConfiguration.scss
```diff
@@ -23,6 +23,27 @@
   }
 }
 
+.centre-detail--placeholder-image__downsize {
+  display: block;
+  width: 400px;
+  height: 200px;
```
davidm-m, line 29: I feel like we must just be able to set the width and some other property so the height scales with it.

davidm-m (approved):
Looks good, one comment on CSS but I don't know if it's worth doing.

DigitalLearningSolutions.Web/Styles/trackingSystem/centreConfiguration.scss
AlexJacksonDS: Looks like I might be able to do something like
```
.class {
    $width: 400px;
    width: $width;
    height: 0.5 * width
}
```
Not sure if it is worth it though

davidm-m: I think `height: auto` should just work

AlexJacksonDS: That produces square 400x400 images by default, which is what we were trying to avoid by setting the height to less

davidm-m: Ah sure, ignore me then

stellake (approved):
Looks good! :)

PR #385, opened 2021-06-03
HEEDLS-474 - Added side nav menu to delegate approval page

DanBloxham-sw:
The partial's logic to decide which link is "selected" will likely need to be changed when the new pages are introduced. 

Ran all unit tests
Tested in Chrome, IE, Edge, Firefox
Checked with screen reader, mobile view, no JS.

Screen shot of mobile view (note that the screen looks as it would if no new nav bar was present, as this mimics the behaviour found on the NHS site):

Delegates side menu mobile view

Screen shot of desktop view:
Delegates side menu

### DIFF
--- a/DigitalLearningSolutions.Web/Styles/trackingSystem/sideNavMenus.scss
+++ b/DigitalLearningSolutions.Web/Styles/trackingSystem/sideNavMenus.scss
@@ -1,0 +1,49 @@
+﻿@import "~nhsuk-frontend/packages/core/all";
+
+.side-nav-menu {
+  @include govuk-media-query($until: desktop) {
+      display: none;
+  }
+}
+
+.side-nav__heading {
+  font-weight: 600;
+  color: #4c6272;
+  margin-bottom: 12px;
+  padding-top: 4px;
+  font-size: 1.1875rem;
+  line-height: 1.25;
+}
+
+.side-nav__link {
+  text-decoration: none;
+  color: #005eb8;
+
+  &:visited {
+    color: #005eb8;
+  }
+
+  &:hover {
+    color: #7C2855;
+    text-decoration: underline;
+  }
+}
+
+.side-nav__list {
+  font-weight: 400;
+  margin-bottom: 0;
+  font-size: 1rem;
+  line-height: 1.3;
+}
+
+.side-nav__item {
+  padding-bottom: 4px;
+  padding-top: 4px;
+}
+
+.side-nav__item--selected {
+  border-left: 4px solid #005eb8;
+  margin-left: -12px;
+  padding-left: 8px;
+  font-weight: bold;
+}
--- a/DigitalLearningSolutions.Web/Views/TrackingSystem/Shared/_DelegatesSideNavMenu.cshtml
+++ b/DigitalLearningSolutions.Web/Views/TrackingSystem/Shared/_DelegatesSideNavMenu.cshtml
@@ -1,0 +1,32 @@
+﻿@using Microsoft.Extensions.Configuration
+@inject IConfiguration Configuration
+
+@{
+  var title = ViewData["Title"].ToString();
+  var selectedClass = "side-nav__item--selected";
+
+  var allDelegatesItemClass = title == "All delegates" ? selectedClass : string.Empty;
+  var delegateGroupsItemClass = title == "Delegate groups" ? selectedClass : string.Empty;
+  var courseDelegateItemClass = title == "Course delegates" ? selectedClass : string.Empty;
+  var approvedRegistrationItemClass = title == "Approve delegate registrations" ? selectedClass : string.Empty;
+}
+
+<link rel="stylesheet" href="@Url.Content("~/css/trackingSystem/sideNavMenus.css")" asp-append-version="true">
+
+<nav class="side-nav-menu">
+  <h2 class="side-nav__heading">Delegates</h2>
+  <ul class="nhsuk-list side-nav__list">
+    <li class="side-nav__item @allDelegatesItemClass">
+      <a class="side-nav__link" href="@Configuration["CurrentSystemBaseUrl"]/tracking/delegates">All delegates</a>
+    </li>
+    <li class="side-nav__item @delegateGroupsItemClass">
+      <a class="side-nav__link" href="@Configuration["CurrentSystemBaseUrl"]/tracking/delegategroups">Delegate groups</a>
+    </li>
+    <li class="side-nav__item @courseDelegateItemClass">
+      <a class="side-nav__link" href="@Configuration["CurrentSystemBaseUrl"]/tracking/coursedelegates">Course delegates</a>
+    </li>
+    <li class="side-nav__item @approvedRegistrationItemClass">
+      <a class="side-nav__link" asp-action="Index" asp-controller="DelegateApprovals">Approve registrations</a>
+    </li>
+  </ul>
+</nav>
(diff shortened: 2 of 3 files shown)


DigitalLearningSolutions.Web/Styles/trackingSystem/sideNavMenus.scss
```diff
@@ -0,0 +1,49 @@
...
+
+.side-nav-menu {
+  @include govuk-media-query($until: desktop) {
+      display: none;
+  }
+}
+
+.side-nav__heading {
+  font-weight: 600;
+  color: #4c6272;
```
AlexJacksonDS, line 11: We should probably be using the nhs defined colours for these rather than the explicit codes for them e.g. this one is $color_nhsuk-grey-1 (or various other specific names that can be found [here](https://service-manual.nhs.uk/design-system/styles/colour)

DigitalLearningSolutions.Web/Styles/trackingSystem/sideNavMenus.scss
```diff
@@ -0,0 +1,49 @@
...
+  margin-bottom: 12px;
```
AlexJacksonDS, line 12: We should also probably consider using the nhs defined spacing too rather than explicit px sizes

AlexJacksonDS (commented):
Some changes for the sass to use the nhs defined mixins for various things rather than explicit values.

It also doesn't behave exactly as per the design system site's side bar, which collapses below the main content when in tablet/mobile layout rather than being hidden. @stellake Not sure what behaviour we actually want here.

image

stellake:
@AlexJacksonDS @DanBloxham-sw 

I think we should move the side nav below the main content in mobile (like it is on NHS system), to make sure the user sees the main content of the page as a priority. 👍

DanBloxham-sw:
@AlexJacksonDS  @stellake  I've added the nav bar to the bottom, mimicking the logic on the NHS site. Screen shot is below, but due to the inconsistencies on the NHS site, I'm not sure what styling we want.

The side nav bar only collapses to the bottom of the page on "Contents" pages like this: https://service-manual.nhs.uk/design-system/components - note that the actual "Components" page you is not available on the left hand navigation bar. When collapsed to the bottom, it also has slightly different styling.

Non-Contents pages do not have the nav bar collapse to the bottom, e.g. https://service-manual.nhs.uk/design-system/components/breadcrumbs - these instead seem to require the user to navigate on small screens using the breadcrumbs to return to the Contents page. 

For now, I have basically shifted the side nav bar to the bottom, so that we retain the border/styling to highlight the selected page. 

If we want to keep the bottom nav bar for these pages, do we want to keep it as it is (with current page highlighted) or fully copy the style of the Contents page?

UPDATED Delegates side menu mobile view

stellake:
@DanBloxham-sw that looks great! Can we make the nav have the same left margin as the rest of the content and change the colour of the heading like the NHS does (but not worry about the breadcrumb style because we don't really have a parent page):
image

DanBloxham-sw:
New padding and heading colour on small screens:
UPDATED PADDING Delegates side menu mobile view

AlexJacksonDS (approved):
Looks good to me

DigitalLearningSolutions.Web/Styles/index.scss
```diff
@@ -194,3 +194,72 @@ ul.no-bullets {
...
+  }
+}
+
+.side-nav__heading {
+  font-weight: 600;
+  color: $color_nhsuk-grey-1;
+  @include nhsuk-responsive-margin(3, "bottom");
+  @include nhsuk-responsive-padding(1, "top");
+  font-size: 1.1875rem;
+  line-height: 1.25;
```
stellake, line 218: Re font-styles -- should we use any of the Typography classes for that rather than setting the font-size, font-weight and line-height explicitly?
https://service-manual.nhs.uk/design-system/styles/typography

DigitalLearningSolutions.Web/Styles/index.scss
```diff
@@ -194,3 +194,72 @@ ul.no-bullets {
...
+    @include govuk-media-query($from: tablet) {
+      @include govuk-media-query($until: desktop) {
+        font-size: 1.1875rem;
+      }
+    }
+  }
+
+  .side-nav__item--selected {
+    border-left: 4px solid $nhsuk-link-color;
+    margin-left: -0px;
```
stellake, line 262: 0 can be unitless and I don't think we need the "-" sign

DigitalLearningSolutions.Web/Views/TrackingSystem/Shared/_DelegatesSideNavMenu.cshtml
```diff
@@ -0,0 +1,30 @@
+﻿@using Microsoft.Extensions.Configuration
+@inject IConfiguration Configuration
+
+@{
+  var title = ViewData["Title"].ToString();
```
stellake, line 5: Not 100% sure about making this so decoupled with the ViewData title (which should be a purely UI value) - maybe we can pass in a value for the "selectedPage" which is from a list enums?

DigitalLearningSolutions.Web/Views/TrackingSystem/Shared/_DelegatesSideNavMenu.cshtml
```diff
@@ -0,0 +1,30 @@
+﻿@using Microsoft.Extensions.Configuration
```
stellake, line 1: Should this be under Delegates/Shared folder?

DigitalLearningSolutions.Web/Views/TrackingSystem/Shared/_DelegatesSideNavMenu.cshtml
```diff
@@ -0,0 +1,30 @@
+﻿@using Microsoft.Extensions.Configuration
+@inject IConfiguration Configuration
+
+@{
+  var title = ViewData["Title"].ToString();
+  var selectedClass = "side-nav__item--selected";
```
stellake, line 6: Can this be const? :)

DigitalLearningSolutions.Web/Styles/index.scss
DanBloxham-sw: Whoops, not sure how that happened, this should be -12px.

DigitalLearningSolutions.Web/Views/TrackingSystem/Shared/_DelegatesSideNavMenu.cshtml
DanBloxham-sw: As no Delegate/Shared folder currently exists, I just put it in TrackingSystem/Shared. Would probably be better off in specifically Delegates/Shared, so sure, I'll make a folder.

DanBloxham-sw:
Among other changes, I corrected the margin on the Selected link so that the border is more to the left of the text so it matches the NHS side bar:

UPDATING MARGIN on selected

DigitalLearningSolutions.Web/Helpers/TrackingSystem/SelectedPage.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace DigitalLearningSolutions.Web.Helpers.TrackingSystem
+{
+    public enum SelectedPage
```
stellake, line 3: Should this be called "DelegatePage"?

stellake (approved):
Looks good - just a minor naming change request, but no need for rereview :)

DigitalLearningSolutions.Web/Helpers/TrackingSystem/SelectedPage.cs
DanBloxham-sw: I couldn't decide between having a specific enum for each sub section of the tracking system (e.g. DelegatePage, CentrePage, etc.) or just a single enum that contains all pages in one. Do we have a preference?

stellake: Maybe use delegate specific for now so that's the only thing that can be passed into the delegate side menu partial. And if we decide to commonise some of this later, we can change it 🙏

PR #387, opened 2021-06-04
HEEDLS-479 Change mapping for answers so it picks correct values

AlexJacksonDS:
Fixes a bug with non-consecutive custom prompts being validated with the wrong answers.

Added a new unit test to check for this case.


DigitalLearningSolutions.Web/Helpers/CustomPromptHelper.cs
```diff
@@ -26,8 +26,15 @@ public CustomPromptHelper(ICustomPromptsService customPromptsService)
             var answers = new List<string?> { answer1, answer2, answer3, answer4, answer5, answer6 };
             var customPrompts = customPromptsService.GetCustomPromptsForCentreByCentreId(centreId);
 
-            return customPrompts.CustomPrompts.Select((cp, i) => new EditCustomFieldViewModel(cp.CustomPromptNumber,
-                cp.CustomPromptText, cp.Mandatory, cp.Options, answers[i])).ToList();
+            return customPrompts.CustomPrompts.Select(
+                (cp, i) => new EditCustomFieldViewModel(
```
DanBloxham-sw, line 30: Is this "i" still necessary?

DanBloxham-sw (approved):
Think this is good, minor change to remove redundant "i", but happy for it to move on without re-review once complete.

DigitalLearningSolutions.Web.Tests/Helpers/CustomPromptHelperTests.cs
```diff
@@ -54,5 +54,39 @@ public void GetCustomFieldViewModelsForCentre_returns_populated_list()
...
             }
         }
+
+        [Test]
+        public void GetCustomFieldViewModelsForCentre_returns_correctly_mapped_answers_with_gap_in_prompts()
+        {
+            // Given
+            var answer1 = "Answer1";
+            var answer2 = "Answer2";
+            var answer3 = "Answer3";
```
stellake, line 64: I think these can be const (IDE warnings)

Whilst we're there, could we also fix the other warnings in this test file? 🙏

stellake (approved):
Looks good! 👍

PR #390, opened 2021-06-07
HEEDLS-367 Changed centre id of RegistrationDataService test

davidm-m:
I had originally misdiagnosed the reason the test was failing: it wasn't because it couldn't delete the test user it was making; it was because a user with the same email and centre ID already existed! For some reason this made it throw an exception before the stored procedure returned - which would have given this away since it always returns -4 when a user already exists.
I've changed the centre ID used in the test and everything's dandy now.


DanBloxham-sw (approved):

stellake (approved):
Looks good! 👍

PR #391, opened 2021-06-07
Heedls 381 delete registration prompts

AlexJacksonDS:
Added new page for user confirmation of deletion of a registration prompt when it has answers and the backend functionality to delete it.

image

image


DanBloxham-sw (commented):
Few questions that might need some refactoring.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
```diff
@@ -178,6 +184,52 @@ public IActionResult AddRegistrationPromptSummaryPost()
...
+            {
+                ModelState.AddModelError(
+                    nameof(RemoveRegistrationPromptViewModel.Confirm),
+                    "You must confirm before deleting this prompt"
+                );
+                return View(model);
+            }
+
+            using var transaction = new TransactionScope();
+            try
```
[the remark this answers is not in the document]
AlexJacksonDS: Agreed that it should probably go in the data project. I think we should be fine if we just do a try-finally. This should allow the error to just propagate up to be handled in the controller

DigitalLearningSolutions.Web/Styles/index.scss
```diff
@@ -29,6 +29,20 @@ a.delete-button:active {
   background-color: #6b140e;
 }
 
+button.delete-button {
+  background-color: #d5281b;
+  box-shadow: 0 4px 0 #6b140e;
+  margin-bottom: 16px !important;
```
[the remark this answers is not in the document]
AlexJacksonDS: These styles are copied from the above a.delete-button styles. I'd be against changing the behaviour of them both for now. I suspect the reason it is there is because I've found a lot of the nhs styles use !important, and the only real way to override that is to use more of them

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
```diff
@@ -178,6 +184,52 @@ public IActionResult AddRegistrationPromptSummaryPost()
...
+        {
+            var delegateWithAnswerCount =
+                userDataService.GetDelegateCountWithAnswerForPrompt(User.GetCentreId(), promptNumber);
+            var customPrompts = customPromptsService.GetCustomPromptsForCentreByCentreId(User.GetCentreId());
+            var promptName = customPrompts.CustomPrompts.Single(c => c.CustomPromptNumber == promptNumber)
+                .CustomPromptText;
+
+            var model = new RemoveRegistrationPromptViewModel(promptName, delegateWithAnswerCount);
+
+            return View(model);
```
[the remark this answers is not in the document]
AlexJacksonDS: Ah no I messed up the logic here, it should be auto deleting if 0 delegates. Problems with working on a ticket a week apart due to dependencies

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
```diff
@@ -261,6 +304,19 @@ private IActionResult AddRegistrationPromptConfigureAnswersPostNext(Registration
...
+        private IActionResult TryRemoveRegistrationPrompt(int promptNumber)
+        {
+            try
+            {
+                customPromptsService.RemoveCustomPromptFromCentre(User.GetCentreId(), promptNumber);
+                return RedirectToAction("Index");
+            }
+            catch
+            {
+                return RedirectToAction("Error", "LearningSolutions");
```
DanBloxham-sw, line 316: I think the try catch -> redirect to error is redundant here as that should be the automatic behaviour.

DanBloxham-sw (approved):
Looks good to me. I think the try catch in the TryRemoveRegistrationPrompt method of the RegistrationPromptController is technically redundant, but I don't think it needs to be removed if you don't want it to be, it might just cut ahead of and replicate the automatic error handling.

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/RemoveRegistrationPrompt.cshtml
```diff
@@ -0,0 +1,83 @@
...
+  </div>
+</div>
+
+<div class="nhsuk-grid-row">
+  <div class="nhsuk-grid-column-one-quarter nhsuk-heading-l">
+    <div class="nhsuk-u-font-weight-bold">
+      Prompt:
+    </div>
+  </div>
+  <div class="nhsuk-grid-column-one-quarter nhsuk-heading-l nhsuk-u-font-weight-normal">
```
stellake, line 40: Should we change this to 3 quarters too?

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/RemoveRegistrationPrompt.cshtml
```diff
@@ -0,0 +1,83 @@
...
+    <input type="hidden" asp-for="PromptName" />
+    @Model.PromptName
+  </div>
+</div>
+
+<div class="nhsuk-grid-row">
+  <div class="nhsuk-grid-column-full nhsuk-body-m">
+    Some users have answered this prompt. By deleting this prompt, all the answers that the delegates have given will be deleted too.
```
stellake, line 48: Should this be in a <p> element?

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/RemoveRegistrationPrompt.cshtml
```diff
@@ -0,0 +1,83 @@
...
+  </div>
+</div>
+
+<form method="post" asp-action="RemoveRegistrationPrompt">
+  <div class="nhsuk-grid-row">
+    <div class="nhsuk-grid-column-full">
+      <div class="nhsuk-form-group @confirmFormErrorClass">
+        <fieldset class="nhsuk-fieldset" aria-describedby="confirm-error">
+          @if (confirmError)
```
stellake, line 57: Do you think it's worth creating a common component for checkboxes too?

DigitalLearningSolutions.Web/Styles/index.scss
```diff
@@ -15,18 +15,33 @@ ul > li > ul > li {
 .js-enabled .non-js-only {
   display: none;
 }
+
```
stellake, line 18: Ah - this is done in https://github.com/TechnologyEnhancedLearning/DLSV2/pull/378/files too, so worth coordinating with @JBsoftwire

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
stellake: The default behaviour logs an error too, which we should be doing! I think either logging a sensible error here that we can use for debugging in addition to the redirect, OR letting it bubble are both something we can do :)

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
```diff
@@ -178,6 +184,43 @@ public IActionResult AddRegistrationPromptSummaryPost()
...
+                userDataService.GetDelegateCountWithAnswerForPrompt(User.GetCentreId(), promptNumber);
+
+            if (delegateWithAnswerCount == 0)
+            {
+                return TryRemoveRegistrationPrompt(promptNumber);
+            }
+
+            var customPrompts = customPromptsService.GetCustomPromptsForCentreByCentreId(User.GetCentreId());
+            var promptName = customPrompts.CustomPrompts.Single(c => c.CustomPromptNumber == promptNumber)
+                .CustomPromptText;
```
stellake, line 201: I think this is used in a few place now - do you think it's worth creating a method to get prompt based on prompt number/id for centre?

DigitalLearningSolutions.Data/Services/CustomPromptsService.cs
```diff
@@ -108,7 +114,7 @@ public bool AddCustomPromptToCentre(int centreId, int promptId, bool mandatory,
 
             if (promptNumber != null)
             {
-                customPromptsDataService.AddCustomPromptToCentre(
+                customPromptsDataService.UpdateCustomPromptForCentre(
```
stellake, line 117: What's this change?

stellake: Nvm - I'm doing the review bottom-up! :)

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/RemoveRegistrationPrompt.cshtml
AlexJacksonDS: Not sure what you mean here, you need to escape the \<p> or whatever you've put with a \

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/RemoveRegistrationPrompt.cshtml
AlexJacksonDS: It's a bit harder than with the other components due to the need to handle multiple checkboxes in a group (and that they would usually not have any errors associated with them). I think we only have the option of either creating a Confirm checkbox component that covers all the stuff here but is not really reusable, or a component for the checkbox item nhsuk-checkboxes__item which feels a bit pointless as it is small.

DigitalLearningSolutions.Web/Styles/index.scss
AlexJacksonDS: Replace the version I have here with the version from that file.

stellake: Ah - can you pull the latest master for this, should be merged now? 🙏

DigitalLearningSolutions.Data.Tests/DataServices/CustomPromptsDataServiceTests.cs
```diff
@@ -97,5 +97,14 @@ public void AddCustomPromptToCentre_correctly_adds_custom_prompt()
             }
         }
 
+        [Test]
+        public void Get()
```
stellake, line 101: Is the test name intentional?

AlexJacksonDS: No, will change

stellake (approved):
Looks good - just need to rename a test, but no need for rereview 👍

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
AlexJacksonDS: Note here, this was the only place where it was being done like this. There is another place where a similar thing was being done, but with different input data. Do you want me to write another separate method to handle that?

PR #395, opened 2021-06-08
HEEDLS-355: Show multiple validation errors in text input VC

SteveJacksonSoft:
@stellake The registration create-password page doesn't use the VC, so it still only shows one validation error per input. Do you want me to change that so it uses the VC as well?

### DIFF
--- a/DigitalLearningSolutions.Web/Styles/viewComponents/textInput.scss
+++ b/DigitalLearningSolutions.Web/Styles/viewComponents/textInput.scss
@@ -1,8 +1,6 @@
-.nhsuk-error-message.validation-errors-container__message {
-  margin-bottom: 4px;
-}
+@import "~nhsuk-frontend/packages/core/all";
 
-.validation-errors-container {
-  padding-top: 4px;
-  padding-bottom: 16px;
+.nhsuk-error-message.error-message {
+  // Below class not set on elements directly because margin setting would get superseded by nhsuk-error-message rules.
+  @extend .nhsuk-u-margin-bottom-1;
 }
--- a/DigitalLearningSolutions.Web/Styles/viewComponents/textInput.scss
+++ b/DigitalLearningSolutions.Web/Styles/viewComponents/textInput.scss
@@ -1,0 +1,8 @@
+.nhsuk-error-message.validation-errors-container__message {
+  margin-bottom: 4px;
+}
+
+.validation-errors-container {
+  padding-top: 4px;
+  padding-bottom: 16px;
+}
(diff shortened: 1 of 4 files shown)


SteveJacksonSoft:
@stellake How does this look?
image

DigitalLearningSolutions.Web/ViewModels/Common/TextInputViewModel.cs
```diff
@@ -26,8 +29,8 @@ public TextInputViewModel
             SpellCheck = spellCheck;
             Autocomplete = autocomplete;
             HintText = hintText;
-            ErrorMessage = errorMessage;
-            HasError = hasError;
+            ErrorMessages = errorMessageList;
+            HasError = errorMessageList.Any(message => !string.IsNullOrWhiteSpace(message));
```
stellake, line 33: Just out of curiosity - is this consistent with the ErrorSummary component?

stellake:
> @stellake How does this look?
> image

This looks perfect! 👍 

Also, as discussed on standup, David will handle the registration page (just so it doesn't look like I'm ignoring you hehe). :)

DigitalLearningSolutions.Web/ViewModels/Common/TextInputViewModel.cs
SteveJacksonSoft: Ah, it's not actually.

SteveJacksonSoft: Do you want me to remove the predicate from the Any(), or add it to ErrorSummary?

stellake: I think it's almost impossible that we'll add a whitespace or null error in the code, so either is fine as long as we make sure that it's consistent across the two components :)

DigitalLearningSolutions.Web/Styles/viewComponents/textInput.scss
```diff
@@ -0,0 +1,8 @@
+.nhsuk-error-message.validation-errors-container__message {
+  margin-bottom: 4px;
+}
+
+.validation-errors-container {
+  padding-top: 4px;
+  padding-bottom: 16px;
+}
```
stellake, line 8: Do you think any of the nhsuk-spacing values would work here? https://service-manual.nhs.uk/design-system/styles/spacing

e.g.
margin-bottom: nhsuk-spacing(1);

stellake (approved):
Looks good - just a question whether we can use nhsuk-spacing values in CSS, but no need to rereview. 👍

PR #397, opened 2021-06-09
HEEDLS-450 - added side nav menu to centre pages and breadcrumbs to m…

DanBloxham-sw:
…anage delegate registration prompts page.

Note that the Delegate Side menu was dealt with in 474 (now merged). MR for reference: https://github.com/TechnologyEnhancedLearning/DLSV2/pull/385 

Ran all unit tests,
Tested with screen reader + mobile view.
Tested in Chrome, Firefox, IE, Edge

Screen shots:
Regular nav bar:

Centre Side nav menu

Breadcrumbs:

Manage delegate registration prompts breadcrumbs

Mobile view of nav bar: 

Centre Side bottom nav menu


DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Dashboard/Index.cshtml
```diff
@@ -17,214 +18,222 @@
...
-            <h2 class="nhsuk-card__heading nhsuk-heading-m">
-              <a class="nhsuk-card__link" href="#">Centre ranking</a>
-            </h2>
-            <p class="nhsuk-card__description">See centre rankings by week, month, or year</p>
-          </div>
-        </div>
-
-      </li>
-    </ul>
+<div class="nhsuk-grid-column-one-quarter">
```
DanBloxham-sw, line 21: Lots of line ending and indentation changes on this file. Only new addition is the changes to this one-quarter div, and another highlighted change near the bottom of the page.

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Dashboard/Index.cshtml
```diff
@@ -17,214 +18,222 @@
...
+          <a class="nhsuk-card__link" href="#">Centre ranking</a>
+        </h2>
+        <p class="nhsuk-card__description">See centre rankings by week, month, or year</p>
+      </div>
+    </div>
+
+  </li>
+</ul>
+
+<nav class="side-nav-menu-bottom" aria-label="Bottom navigation bar">
```
DanBloxham-sw, line 235: Lots of line ending and indentation changes on this file. Only new addition is this new nav and another highlighted change near the top of the page.

DigitalLearningSolutions.Web/Helpers/TrackingSystem/CentrePage.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace DigitalLearningSolutions.Web.Helpers.TrackingSystem
+{
+    public enum CentrePage
+    {
+        CentreDashboard,
+        CentreConfiguration,
+        CentreAdministrators,
+        Reports
```
AlexJacksonDS, line 8: Should these perhaps be named Dashboard/Configuration/Administrators/Reports i.e. drop the Centre from all of them since that is clear from the enum name (and all our urls will ultimately be /centre/pagename)

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Dashboard/Index.cshtml
AlexJacksonDS: Something has broken here to cause the incorrect indentation. Will mark a set of subsequent comments of issues I see.

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Dashboard/Index.cshtml
```diff
@@ -17,214 +18,222 @@
...
-      </li>
-    </ul>
+<div class="nhsuk-grid-column-one-quarter">
+  <nav class="side-nav-menu" aria-label="Side navigation bar">
+    <partial name="~/Views/TrackingSystem/Centre/Shared/_CentreSideNavMenu.cshtml" model="CentrePage.CentreDashboard" />
+  </nav>
+</div>
+<div class="nhsuk-grid-column-three-quarters">
+<h1 class="nhsuk-heading-xl" id="app-page-heading">Centre dashboard</h1>
+<p class="nhsuk-body-m">Welcome <span class="nhsuk-u-font-weight-bold">@Model.FirstName</span>. You are managing <span class="nhsuk-u-font-weight-bold">@Model.CourseCategory</span> courses.</p>
```
AlexJacksonDS, line 28: Incorrect indentation

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Dashboard/Index.cshtml
```diff
@@ -17,214 +18,222 @@
...
+        </dd>
+      </div>
+
+    </dl>
+
+    <a class="nhsuk-button" role="button">
+      Edit centre details
+    </a>
   </div>
+</details>
```
AlexJacksonDS, line 182: I think it is somewhere around here that it breaks

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Dashboard/Index.cshtml
```diff
@@ -17,214 +18,222 @@
...
+    </div>
+
+  </li>
+</ul>
+
+<nav class="side-nav-menu-bottom" aria-label="Bottom navigation bar">
+  <partial name="~/Views/TrackingSystem/Centre/Shared/_CentreSideNavMenu.cshtml" model="CentrePage.CentreDashboard" />
+</nav>
+</div>
 </div>
```
AlexJacksonDS, line 239: Incorrect indentation and Visual studio complains that this final div end tag is missing a start tag

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/Index.cshtml
```diff
@@ -18,7 +19,13 @@
...
 
 <div class="nhsuk-grid-row">
-<div class="nhsuk-grid-column-full">
+<div class="nhsuk-grid-column-one-quarter">
+  <nav class="side-nav-menu" aria-label="Side navigation bar">
+    <partial name="~/Views/TrackingSystem/Centre/Shared/_CentreSideNavMenu.cshtml" model="CentrePage.CentreConfiguration" />
+  </nav>
+</div>
+
+<div class="nhsuk-grid-column-three-quarters">
```
AlexJacksonDS, line 28: Something similar to the dashboard index file has happened here too, there are a lot of miss-aligned elements and elements that visual studio complains have no corresponding start/end tag

AlexJacksonDS (changes requested):

DigitalLearningSolutions.Web/Helpers/TrackingSystem/CentrePage.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace DigitalLearningSolutions.Web.Helpers.TrackingSystem
```
stellake, line 1: Given that the page enums are relevant to the Tracking System subfolders, can we put them into the relevant folders?

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/Index.cshtml
```diff
@@ -14,6 +14,17 @@
   <partial name="~/Views/TrackingSystem/Shared/_NavMenuItems.cshtml"/>
 }
 
+@section NavBreadcrumbs {
```
stellake, line 17: The more I think about it the more I wonder whether we should add the side menu to the Layout and define it similarly to the NavBreadcrumbs and the extra nav bar tabs. Then, the "double click for main content" will skip the side bar, which I think is expected + side nav vs bottom nav can be dealt with within the layout. What do you think?

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/Index.cshtml
```diff
@@ -14,6 +14,17 @@
   <partial name="~/Views/TrackingSystem/Shared/_NavMenuItems.cshtml"/>
 }
 
+@section NavBreadcrumbs {
+  <nav class="nhsuk-breadcrumb" aria-label="Breadcrumb">
+    <div class="nhsuk-width-container">
+      <ol class="nhsuk-breadcrumb__list">
+        <li class="nhsuk-breadcrumb__item"><a class="nhsuk-breadcrumb__link" asp-controller="CentreConfiguration" asp-action="Index">Centre Configuration</a></li>
```
stellake, line 21: Can this be "Centre configuration" (re capitalisation)

stellake:
Could we make sure that there's enough padding on the lhs of the selected page? 🙏 
image

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/Index.cshtml
DanBloxham-sw: If you mean that we just define a new section and render that at the top and bottom of the layout then I don't think that will work. RenderSection in the Layout seems to be restricted to rendering unique sections - calling RenderSection twice for the same section causes the page to break. An alternative to sections might be creating something like ViewData["SideNavMenu"] = "Centre" and ViewData["CurrentPage"] = CentrePage.Administrators and performing some logic on that, which would clean up having to define both in each file. Might become fiddly trying to set up a one-quarter grid column in the layout to work with all pages, given that not all pages are going to have this side menu.

DanBloxham-sw:
Note to self: Add an aria-current="page" to the selected item so it is read out by screen readers (similar to https://service-manual.nhs.uk/design-system/patterns/mini-hub)

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/Index.cshtml
stellake: Ah, that's fair! I think specifying it in ViewData will be too fiddly in the long run as the Layout already has quite a lot logic in it. The benefit of at least moving a side menu there is that the "Skip to main content" option will skip the side navigation bar and goes straight to content. Maybe we can just specify the side nav and bottom nav as separate sections?

DanBloxham-sw: Looks like the "Skip to main content" button already goes straight to the h1 in the main content (think we noticed this before when we were working with error summaries). Do we still need separate sections?

DanBloxham-sw:
Changed margin on selected item so that on smaller screens the item gets pushed to the right to stay in line with the other list items, but it remains consistent with the NHS site on bigger screens:

UPDATING MARGIN on selected MOBILE

DigitalLearningSolutions.Web/Helpers/TrackingSystem/Centre/CentrePage.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace DigitalLearningSolutions.Web.Helpers.TrackingSystem.Centre
```
stellake, line 1: Ah, sorry I just realised that this is under Helpers, but we should really add it to Models (and maybe even create enums folder)

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/Index.cshtml
stellake: Ah that's actually great news! In that case, I think it's ok! 👍

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Shared/_CentreSideNavMenu.cshtml
```diff
@@ -0,0 +1,36 @@
...
+  var administratorsItemClass = Model == CentrePage.Administrators ? selectedClass : string.Empty;
+  var reportsItemClass = Model == CentrePage.Reports ? selectedClass : string.Empty;
+}
+
+
+<h2 class="side-nav__heading">Centre</h2>
+<ul class="nhsuk-list side-nav__list">
+  <li class="side-nav__item @dashboardItemClass" aria-current="@(Model == CentrePage.Dashboard ? "page" : null)">
+    <a class="side-nav__link" asp-action="Index" asp-controller="Dashboard">Centre dashboard</a>
+  </li>
```
stellake, line 21: Do you think it's worth creating a VC for this like we've done with the text inputs to avoid the XXXItemClass and aria-current logic duplication, but still keep text in HTML?

DigitalLearningSolutions.Web/Helpers/TrackingSystem/Centre/CentrePage.cs
DanBloxham-sw: So for clarity, it should be under Web/Enums/TrackingSystem/Centre etc.?

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Shared/_CentreSideNavMenu.cshtml
```diff
@@ -0,0 +1,36 @@
...
+  <li class="side-nav__item @configurationItemClass" aria-current="@(Model == CentrePage.Configuration ? "page" : null)">
+    <a class="side-nav__link" asp-action="Index" asp-controller="CentreConfiguration">Centre configuration</a>
+  </li>
+
+  @if(User.GetCustomClaimAsBool(CustomClaimTypes.UserCentreManager) == true || User.GetCustomClaimAsBool(CustomClaimTypes.UserUserAdmin) == true)
```
stellake, line 26: We should have User.HasCentreManagerPermissions method on master now :)

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Shared/_CentreSideNavMenu.cshtml
DanBloxham-sw: Perhaps. It's a toss up though, as I think having them as a VC would mean losing some nice validation functionality on the asp-action and asp-controller attributes. Probably worth it as that can probably be easily tested anyway.

DigitalLearningSolutions.Web/Helpers/TrackingSystem/Centre/CentrePage.cs
stellake: Oh so I meant more under Web/Models or potentially even Web/Models/Enums -- I guess we can have the TrackingSystem folder, but I don't think we'll need the Centre etc because we won't have that many models/enums in the web project. :)

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Dashboard/Index.cshtml
```diff
@@ -1,31 +1,41 @@
...
-    <p class="nhsuk-body-m">Welcome <span class="nhsuk-u-font-weight-bold">@Model.FirstName</span>. You are managing <span class="nhsuk-u-font-weight-bold">@Model.CourseCategory</span> courses.</p>
-
-    <partial name="_DashboardTopCardGroup" model="Model"/>
-
-    <partial name="_DashboardCentreDetails" model="Model"/>
-
-    <partial name="_DashboardBottomCardGroup"/>
-  </div>
-</div>
+﻿@inject IConfiguration Configuration
```
stellake, line 1: Is this line endings? 👁️

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Shared/_CentreSideNavMenu.cshtml
stellake: I'll leave it up to you 👍

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Dashboard/Index.cshtml
DanBloxham-sw: Yep, I am sorting that along with the other changes.

stellake (approved):
Nice! I think it might be better to build separate components for href and asp-action/controller links because it will always be one or the other (I think in the end the href will only be used for one of the support pages, so it doesn't even have to be a component). Happy for you to do it as part of the "Support pages" ticket  so we can get this merged and tested! 👍

PR #401, opened 2021-06-10
HEEDLS-385 - implementation of centre coordinate saving.

DanBloxham-sw:
Added new API call when saving centre details so that postcode is validated.

Ran all unit tests. Tried with different postcodes in the UK, and random gibberish.


DigitalLearningSolutions.Web/Helpers/ExternalApis/MapsApiHelper.cs
```diff
@@ -0,0 +1,31 @@
...
+    using Newtonsoft.Json;
+
+    public interface IMapsApiHelper
+    {
+        Task<MapsResponse> GetPostcodeCoordinates(string postcode);
+    }
+
+    public class MapsApiHelper : IMapsApiHelper
+    {
+        private static readonly HttpClient Client = new HttpClient();
```
AlexJacksonDS, line 15: Is this (or the MapsApiHelper) scoped properly? I know HttpClient is something you have to be careful with. We don't want to be creating multiple instances of HttpClient (or multiple instances of MapsApiHelper containing multiple)

DigitalLearningSolutions.Web/appsettings.Production.json
```diff
@@ -1,12 +1,12 @@
-{
```
AlexJacksonDS, line 1: Was there supposed to be a change here?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/CentreConfigurationController.cs
```diff
@@ -80,13 +90,36 @@ public IActionResult EditCentreWebsiteDetails(EditCentreWebsiteDetailsViewModel
                 return View(model);
             }
 
+            model.CentrePostcode = model.CentrePostcode!.Trim();
+            var mapsResponse = mapsApiHelper.GetPostcodeCoordinates(model.CentrePostcode).Result;
+
+            if (mapsResponse.Status == "ZERO_RESULTS")
+            {
+                ModelState.AddModelError("CentrePostcode", "Enter a valid postcode.");
```
AlexJacksonDS, line 98: can we use nameof here instead of the property name string

AlexJacksonDS (commented):
Just one actual change and a couple of questions.

DigitalLearningSolutions.Web/Helpers/ExternalApis/MapsApiHelper.cs
DanBloxham-sw: From what I've read up, I suspect that it's fine. But just to be safe I've changed it about to make it definitely rely on an injected httpClient.

DigitalLearningSolutions.Web/appsettings.Production.json
DanBloxham-sw: There was a change at one point, but that was then reverted. The line endings changed to CRLF as part of it, and haven't been reverted. Might as well keep them in my opinion.

DigitalLearningSolutions.Web/Helpers/ExternalApis/MapsApiHelper.cs
AlexJacksonDS: Yeah, I thought it might be but wasn't sure. Nice change to use one of the patterns recommended by the asp net core docs though.

AlexJacksonDS (approved):
Looks good

DigitalLearningSolutions.Web/appsettings.Production.json
```diff
@@ -1,12 +1,12 @@
...
-    "DefaultConnection": ""
-  },
-  "CurrentSystemBaseUrl": "https://www.dls.nhs.uk",
-  "AppRootPath": "https://www.dls.nhs.uk/v2",
-  "LegacyLearningMenu": false,
-  "FeatureManagement": {
-    "Login": false
-  }
-}
+{
```
stellake, line 1: Could we add the new property here with the empty string value, like we do for "DefaultConnection" (and for other appsettings), so it's easy to find what the new property name is from appsettings. Then, the devs only need to populate it from what they can get in Keeper.

DigitalLearningSolutions.Web/Helpers/ExternalApis/MapsApiHelper.cs
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    public class MapsApiHelper : IMapsApiHelper
+    {
+        private readonly HttpClient client;
+        private readonly string mapsApiBaseUrl;
+
+        public MapsApiHelper(HttpClient httpClient)
+        {
+            var mapsApiKey = ConfigHelper.GetAppConfig()["MapsAPIKey"];
```
stellake, line 20: I wonder if it's worth putting the "MapsAPIKey" string into ConfigHelper and add a method "GetMapsApiKey" into the config helper so if we ever change the key name, we'll only need to change it in two places.. 🤔

DigitalLearningSolutions.Web/Helpers/ExternalApis/MapsApiHelper.cs
```diff
@@ -0,0 +1,34 @@
...
+            mapsApiBaseUrl =
+                $"https://maps.googleapis.com/maps/api/geocode/json?key={mapsApiKey}&components=country:GB|postal_code:";
+            httpClient.DefaultRequestHeaders.Accept.Clear();
```
stellake, line 23: I'm not familiar with this - why this needed?

DigitalLearningSolutions.Web/Helpers/ExternalApis/MapsApiHelper.cs
```diff
@@ -0,0 +1,34 @@
+﻿namespace DigitalLearningSolutions.Web.Helpers.ExternalApis
+{
+    using System.Net.Http;
+    using System.Threading.Tasks;
+    using DigitalLearningSolutions.Data.Models.External.Maps;
+    using Newtonsoft.Json;
+
+    public interface IMapsApiHelper
+    {
+        Task<MapsResponse> GetPostcodeCoordinates(string postcode);
```
stellake, line 10: It looks like instead of getting coordinates, we're getting MapsResponse. It might be worth making the current method private and then add a public method that returns a lat/long tuple?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/CentreConfigurationController.cs
```diff
@@ -80,13 +90,36 @@ public IActionResult EditCentreWebsiteDetails(EditCentreWebsiteDetailsViewModel
                 return View(model);
             }
 
+            model.CentrePostcode = model.CentrePostcode!.Trim();
+            var mapsResponse = mapsApiHelper.GetPostcodeCoordinates(model.CentrePostcode).Result;
+
+            if (mapsResponse.Status == "ZERO_RESULTS")
+            {
+                ModelState.AddModelError(nameof(model.CentrePostcode), "Enter a valid postcode.");
```
stellake, line 98: I think the new consistent style of error messages is without a full stop in the end (but do double check on latest master just in case) 🙏

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/CentreConfigurationController.cs
```diff
@@ -80,13 +90,36 @@ public IActionResult EditCentreWebsiteDetails(EditCentreWebsiteDetailsViewModel
                 return View(model);
             }
 
+            model.CentrePostcode = model.CentrePostcode!.Trim();
+            var mapsResponse = mapsApiHelper.GetPostcodeCoordinates(model.CentrePostcode).Result;
+
+            if (mapsResponse.Status == "ZERO_RESULTS")
```
stellake, line 96: Could we pull these status codes into the maps helper or models, so we don't need to reference the strings everywhere?
For example, could we do mapsResponse.Status.isNoResults() or something similar here? :)

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/CentreConfigurationController.cs
```diff
@@ -80,13 +90,36 @@ public IActionResult EditCentreWebsiteDetails(EditCentreWebsiteDetailsViewModel
...
 
-            centresDataService.UpdateCentreWebsiteDetails
-            (
+            centresDataService.UpdateCentreWebsiteDetails(
                 centreId,
-                model.CentrePostcode!,
+                model.CentrePostcode,
                 model.ShowCentreOnMap,
+                longitude,
+                latitude,
```
stellake, line 122: This is super minor, but my intuition is that latitude comes before longitude 😄

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/CentreConfigurationController.cs
```diff
@@ -64,7 +74,7 @@ public IActionResult EditCentreWebsiteDetails()
         {
```
stellake, line 74: Could we write some controller tests for this where we mock the maps api helper? 🙏

DigitalLearningSolutions.Web/Helpers/ExternalApis/MapsApiHelper.cs
```diff
@@ -0,0 +1,34 @@
...
+        public MapsApiHelper(HttpClient httpClient)
+        {
+            var mapsApiKey = ConfigHelper.GetAppConfig()["MapsAPIKey"];
+            mapsApiBaseUrl =
+                $"https://maps.googleapis.com/maps/api/geocode/json?key={mapsApiKey}&components=country:GB|postal_code:";
+            httpClient.DefaultRequestHeaders.Accept.Clear();
+            client = httpClient;
+        }
+
+        public async Task<MapsResponse> GetPostcodeCoordinates(string postcode)
```
stellake, line 27: Should we either call it something less specific than "PostcodeCoordinates" or maybe add a new public method that calls this and does the mapping to coordinates?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/CentreConfigurationController.cs
DanBloxham-sw: Typically, yes. However, in the database longitude comes before latitude, so when writing this I kept everything in the same order for consistency sake. Do we want the code to be consistent with general standards and thus have latitude before longitude?

DigitalLearningSolutions.Web/Helpers/ExternalApis/MapsApiHelper.cs
DanBloxham-sw: It is probably not "necessary" per se, but as the HttpClient can be used for other requests, it just clears the headers on any request to make sure we aren't sending old headers to where they are not necessary. It should probably be performed before actually making each individual call (as it is done in the existing FilteredApiHelper), and I accidentally moved it up here.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/CentreConfigurationController.cs
stellake: I think we should just to avoid anyone accidentally writing lat, then long. The unexpected order (long, lat) is much easier to spot than that lat -> long order when reviewing for example :)

DigitalLearningSolutions.Web/appsettings.Production.json
stellake: Just a minor thing - it looks like we've missed plain appsettings.json

DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/CentreConfiguration/CentreConfigurationControllerTests.cs
```diff
@@ -0,0 +1,130 @@
...
+                        Status = "OK",
+                        Results = new[]
+                        {
+                            new Map
+                            {
+                                Geometry = new Geometry
+                                    { Location = new Coordinates { Latitude = "50.123", Longitude = "-3.01" } }
+                            }
+                        }
+                    }
```
stellake, line 104: Could we pull this into a private method below to keep the test more concise. Maybe something like CreateSuccessfulMapsResponse(lat, long)

DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/CentreConfiguration/CentreConfigurationControllerTests.cs
```diff
@@ -0,0 +1,130 @@
...
+        private CentreConfigurationController controller = null!;
+        private ILogger<CentreConfigurationController> logger = null!;
+        private IMapsApiHelper mapsApiHelper = null!;
+
+        [SetUp]
+        public void Setup()
+        {
+            centresDataService = A.Fake<ICentresDataService>();
+            mapsApiHelper = A.Fake<IMapsApiHelper>();
+            logger = A.Fake<ILogger<CentreConfigurationController>>();
```
stellake, line 26: Could we define these fakes outside the setup, at the top of the test file?

stellake (approved):
Looks good! 👍

PR #402, opened 2021-06-10
Heedls 415 admin page pagination

AlexJacksonDS:
Refactored all the search/sort/paginate functionality to work for more than just Courses and Frameworks and added this functionality to the Centre Administrators page.

Ran all unit tests and checked that the functionality still works with javascript and without javascript on the existing pages and new page. We'll need to check the existing pages more thoroughly later.

image

image

image


DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/_SearchableAdminCard.cshtml
```diff
@@ -0,0 +1,15 @@
+﻿@using DigitalLearningSolutions.Web.ViewModels.TrackingSystem.Centre.Administrator
+@model SearchableAdminViewModel
+
+<div class="searchable-element nhsuk-panel" id="@Model.Id-card">
```
DanBloxham-sw, line 4: Getting IDE warning about nhsuk-panel being an unknown css class.

DigitalLearningSolutions.Web/Views/Shared/SearchablePage/_SearchResultsCount.cshtml
```diff
@@ -1,14 +1,11 @@
...
-    var count = Model.MatchingSearchResults;
-    <span role="alert" id="results-count">@count matching @(count == 1 ? "result" : "results")</span>
-  }
-</p>
+﻿@using DigitalLearningSolutions.Web.ViewModels.Common
+@model BaseSearchablePageViewModel
+
+<p class="nhsuk-u-margin-top-4 nhsuk-u-margin-bottom-0">
+  @if (Model.SearchString == null) {
+    <span hidden aria-hidden="true" aria-live="polite" id="results-count">0 matching results</span>
```
DanBloxham-sw, line 6: I am assuming that Model.SearchString is only null when nothing has been searched for - if this is the case, do we want this hidden span? Is it necessary for js?

AlexJacksonDS: I would assume so, but I'm not sure. This isn't new, it's refactored from the deleted /LearningPortal/Shared/_SearchResultCount.cshtml and /Frameworks/Shared/_SearchResultCount.cshtml without changing the content

DanBloxham-sw (commented):
Few questions and minor things to maybe change.

DigitalLearningSolutions.Web/Views/Shared/Components/SearchBox/Default.cshtml
```diff
@@ -0,0 +1,13 @@
+﻿@using DigitalLearningSolutions.Web.ViewModels.Common
+@model SearchBoxViewModel
+
+<div class="nhsuk-grid-column-one-half" id="search">
+  <label class="nhsuk-u-visually-hidden" for="search-field">@Model.Label</label>
+  <input class="search-box" id="search-field" name="searchString" type="search" placeholder="Search" autocomplete="off" value="@Model.ContainingPageViewModel.SearchString">
+  <button class="nhsuk-search__submit non-js-only" type="submit" asp-action="@ViewContext.RouteData.Values["action"].ToString()" asp-route-page="1">
```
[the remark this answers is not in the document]
AlexJacksonDS: Yes, if you hit enter after typing in the search box, the controller action does not get hit with JS enabled

DigitalLearningSolutions.Web/Views/LearningPortal/Current/CurrentCourseCard/_CurrentCourseCard.cshtml
```diff
@@ -1,11 +1,11 @@
 ﻿@using DigitalLearningSolutions.Web.ViewModels.LearningPortal.Current
 @model CurrentCourseViewModel
 
-<div class="course-card nhsuk-panel @Model.DateStyle()" id="@Model.Id-card">
+<div class="searchable-element nhsuk-panel @Model.DateStyle()" id="@Model.Id-card">
```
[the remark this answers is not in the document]
AlexJacksonDS: Yes, it's still used on some views under LearningPortal/SelfAssessments. We could probably replace them, and get rid of, but I didn't see any search stuff there, so wasn't sure I should remove them yet and I kept the coursesShared.scss that essentially duplicates. I suspect the best answer is going to be to refactor some of the scss so it applies to both classes, then just delete the coursesShared file.

DigitalLearningSolutions.Web/ViewModels/Common/BaseSearchablePageViewModel.cs
```diff
@@ -1,69 +1,73 @@
...
+    using DigitalLearningSolutions.Web.Helpers;
+    using Microsoft.AspNetCore.Mvc;
+    using Microsoft.AspNetCore.Mvc.Rendering;
+
+    public abstract class BaseSearchablePageViewModel
+    {
+        public const string DescendingText = "Descending";
+        public const string AscendingText = "Ascending";
+
+        private const int ItemsPerPage = 10;
```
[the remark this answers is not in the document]
AlexJacksonDS: I don't see any reason it would do, the searchable frameworks are displayed in a table unless I've missed something

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/AllAdmins.cshtml
```diff
@@ -0,0 +1,19 @@
...
+
+@{
+  Layout = null;
+}
+
+<!DOCTYPE html>
+
+<html>
+<head>
+  <title></title>
```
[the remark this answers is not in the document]
AlexJacksonDS: Yes it's to load in all the data across all the pages of the pagination for the JS search

DigitalLearningSolutions.Web/Styles/shared/searchableElements.scss
```diff
@@ -0,0 +1,325 @@
...
+  }
+}
+
+@include mq($from: 1024px) {
+  .asset-card-header {
+    min-height: 163px;
+  }
+}
+
+.tag.tag-article {
```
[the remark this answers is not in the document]
AlexJacksonDS: The initial idea was to have this as the newly renamed css file, but I agree css needs refactoring

DigitalLearningSolutions.Web/Scripts/learningPortal/searchSortAndPaginate.ts
```diff
@@ -1,105 +0,0 @@
...
-    cards.forEach((card) => courseCardsContainer.appendChild(card.element));
-    // This is required to polyfill the new elements in IE
-    Details();
-  }
-
-  private static getPathForEndpoint(endpoint: string): string {
-    const currentPath = window.location.pathname;
-    const indexOfBaseUrl = currentPath.indexOf('LearningPortal');
-    return `${currentPath.substring(0, indexOfBaseUrl)}LearningPortal/${endpoint}`;
-  }
```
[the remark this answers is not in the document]
AlexJacksonDS: Nope, I've instead replaced the places where the endpoint string was being specified with the full /LearningPortal/endpoint path. This change was to allow the search to use paths outside the Learning Portal area e.g. see change in completed.ts

DanBloxham-sw (approved):
I think this looks good to me (wouldn't be surprised if I missed a few things, given the size). Minor merge conflict with the newly merged bottom nav bar that should be easy to sort.

DigitalLearningSolutions.Data/Models/BaseSearchableItem.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace DigitalLearningSolutions.Data.Models
+{
+    public abstract class BaseSearchableItem
+    {
+        public string? SearchableNameValue;
+
+        public abstract string SearchableName { get; set; }
```
stellake, line 7: Is there a reason this is not an interface forcing the class to have a get-only SearchableName/SearchString? It seems like we never really use the SearchableNameValue and usually it's just a function of some other class properties that need to be kept in sync with the SearchableNameValue?

DigitalLearningSolutions.Web/Controllers/FrameworksController/Frameworks.cs
```diff
@@ -1,555 +1,659 @@
...
+
+    public partial class FrameworksController
+    {
+        private const string CookieName = "DLSFrameworkService";
+
+        [Route("/Frameworks/MyFrameworks/{page=1:int}")]
+        public IActionResult FrameworksDashboard(
+            string? searchString = null,
+            string? sortBy = null,
+            string sortDirection = BaseSearchablePageViewModel.DescendingText,
```
AlexJacksonDS, line 26: All changes in this file are either changing the sort direction string, the sort by property, or the SortingHelper

DigitalLearningSolutions.Web/Controllers/FrameworksController/Frameworks.cs
```diff
@@ -1,555 +1,659 @@
...
+                isFrameworkDeveloper
+            );
+            return View("Developer/MyFrameworks", model);
+        }
+
+        [Route("/Frameworks/AllFrameworks/{page=1:int}")]
+        public IActionResult FrameworksViewAll(
+            string? searchString = null,
+            string? sortBy = null,
+            string sortDirection = BaseSearchablePageViewModel.AscendingText,
```
AlexJacksonDS, line 56: Changed

DigitalLearningSolutions.Web/Controllers/FrameworksController/Frameworks.cs
```diff
@@ -1,555 +1,659 @@
...
+            int page = 1
+        )
+        {
+            sortBy ??= FrameworkSortByOptionTexts.FrameworkName.PropertyName;
```
AlexJacksonDS, line 60: Adding these sortBy default options

DigitalLearningSolutions.Web/Controllers/FrameworksController/Frameworks.cs
```diff
@@ -1,555 +1,659 @@
...
+            }
+
+            var adminId = GetAdminID();
+            var sameItems = frameworkService.GetFrameworkByFrameworkName(frameworkname, adminId);
+            var frameworks = frameworkService.GetAllFrameworks(adminId);
+            var sortedItems = GenericSortingHelper.SortAllItems(
+                frameworks.AsQueryable(),
+                FrameworkSortByOptionTexts.FrameworkName.PropertyName,
+                BaseSearchablePageViewModel.AscendingText
+            );
```
AlexJacksonDS, line 233: Change of helper

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/Index.cshtml
```diff
@@ -20,10 +22,49 @@
...
+
+      <div class="nhsuk-grid-row">
+        <vc:search-box label="Search administrators" asp-controller="Administrator" asp-action="Index" containing-page-model="@Model" />
+      </div>
+
+      <div class="nhsuk-grid-row">
+        <div class="nhsuk-grid-column-full">
+          @if (!Model.Admins.Any()) {
+            <p class="nhsuk-u-margin-top-4" role="alert">
+              <b>No data to display</b>
```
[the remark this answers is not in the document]
AlexJacksonDS: This is the same as the other existing ones. Change those too?

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/AllAdmins.cshtml
```diff
@@ -0,0 +1,19 @@
+﻿@using DigitalLearningSolutions.Web.ViewModels.TrackingSystem.Centre.Administrator
```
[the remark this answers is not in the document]
AlexJacksonDS: The javascript search doesn't build any html, it just searches in the html that it has been given and displays the matching elements

DigitalLearningSolutions.Web/Views/Shared/SearchablePage/_Pagination.cshtml
```diff
@@ -1,33 +1,33 @@
...
-      </li>
-    </div>
-  </ul>
-</nav>
+﻿@using DigitalLearningSolutions.Web.ViewModels.Common
+@model BaseSearchablePageViewModel
+
+<nav class="nhsuk-pagination" role="navigation" aria-label="Pagination">
+  <ul class="nhsuk-list nhsuk-pagination__list">
+    <div class="pagination-button-container">
```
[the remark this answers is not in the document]
AlexJacksonDS: I think this has been done because the .nhsuk-pagination-item--next/previous elements have width 50% and we need ~33%, but yes I agree the html isn't correct.

DigitalLearningSolutions.Web/ViewModels/Common/BaseSearchablePageViewModel.cs
```diff
@@ -1,69 +1,73 @@
...
+
+        [BindProperty] public string SortBy { get; set; }
+        
+        public int Page { get; protected set; }
+
+        public int TotalPages { get; protected set; }
+
+        public IEnumerable<SelectListItem> SortByOptions => SelectListHelper.MapOptionsToSelectListItems(SortOptions);
+
+        public abstract IEnumerable<(string, string)> SortOptions { get; }
```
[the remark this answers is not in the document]
AlexJacksonDS: The SortOptions are used to define the sort options on the classes that implement this so that we don't have to define a list of SelectListItems every time. The SelectListItems one is what we then use in the sort. Renaming.

DigitalLearningSolutions.Web/ViewModels/Common/BaseSearchablePageViewModel.cs
```diff
@@ -1,69 +1,73 @@
...
+        protected void SetTotalPages()
+        {
+            TotalPages = (int)Math.Ceiling(MatchingSearchResults / (double)ItemsPerPage);
+            if (Page < 1 || Page > TotalPages)
+            {
+                Page = 1;
+            }
+        }
+
+        private int OffsetFromPageNumber(int pageNumber)
```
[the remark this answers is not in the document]
AlexJacksonDS: No as it references the non-static itemsPerPage (that now cannot be static as I need it to be configurable)

DigitalLearningSolutions.Web/Styles/learningPortal/selfAssessment.scss
```diff
@@ -135,3 +136,81 @@ details.nhsuk-details {
...
+  @extend .nhsuk-u-margin-bottom-3;
+  padding-top: 8px;
+  padding-bottom: 16px;
+  padding-left: 32px;
+  padding-right: 32px;
+}
+
+.course-card .nhsuk-expander {
+  border: none
+}
```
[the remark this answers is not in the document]
AlexJacksonDS: They're used for the selfAssessment cards that don't have any search - hence moving them from the common to the selfAssessment.scss file

DigitalLearningSolutions.Web/Scripts/spec/sort.spec.ts
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion,@typescript-eslint/ban-ts-comment */
 import { JSDOM } from 'jsdom';
```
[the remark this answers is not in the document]
AlexJacksonDS: No and it doesn't make sense to here, especially since the logic is now not tied to course-cards and there is no sort for admins

DigitalLearningSolutions.Data/Models/BaseSearchableItem.cs
AlexJacksonDS: This needs to be a settable property on BaseSearchableItem for the FuzzySearch query. I needed to be able to set this property while forcing the classes which implement it to implement it's own get when it hasn't been explicitly set and this was the best pattern I could come up with.
Ideally I would want something like 
`public string SearchableName { abstract get; set; }`
but that isn't valid.

DigitalLearningSolutions.Web/ViewModels/LearningPortal/Current/CurrentPageViewModel.cs
```diff
@@ -33,32 +19,47 @@ public class CurrentPageViewModel : BaseCoursePageViewModel
...
-            var sortedItems = SortingHelper.SortAllItems(
-                allItems,
+
+            var sortedItems = GenericSortingHelper.SortAllItems(
+                allItems.AsQueryable(),
                 sortBy,
                 sortDirection
             );
-            var filteredItems = SearchHelper.FilterLearningItems(sortedItems, SearchString).ToList();
+            var filteredItems = GenericSearchHelper.SearchItems(sortedItems, SearchString).ToList();
```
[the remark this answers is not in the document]
AlexJacksonDS: Kept the Generic versions since there was new content merged that uses the old versions. Made the newly merged frameworks code compatible with the new helpers, but left the other new search/sort in the old helpers

DigitalLearningSolutions.Data/Models/BaseSearchableItem.cs
stellake: I see - in that case, could we add a comment about this to avoid confusion above the SearchableNameValue property. And maybe name it SearchableNameOverrideForFuzzySearch or something similar to make it clear that it's what the property does? :)

DigitalLearningSolutions.Web.Tests/ViewModels/TrackingSystem/Centre/Administrator/CentreAdministratorsViewModelTests.cs
```diff
@@ -0,0 +1,62 @@
...
+            UserTestHelper.GetDefaultAdminUser(firstName: "i", lastName: "Surname"),
+            UserTestHelper.GetDefaultAdminUser(firstName: "j", lastName: "Surname"),
+            UserTestHelper.GetDefaultAdminUser(firstName: "k", lastName: "Surname"),
+            UserTestHelper.GetDefaultAdminUser(firstName: "l", lastName: "Surname"),
+            UserTestHelper.GetDefaultAdminUser(firstName: "m", lastName: "Surname"),
+            UserTestHelper.GetDefaultAdminUser(firstName: "n", lastName: "Surname"),
+            UserTestHelper.GetDefaultAdminUser(firstName: "o", lastName: "Surname"),
+        };
+
+    [Test]
```
stellake, line 30: Formatting

DigitalLearningSolutions.Web/Controllers/FrameworksController/Frameworks.cs
```diff
@@ -1,603 +1,607 @@
...
+            return View(model);
+        }
+        [Route("/Frameworks/View/{tabname}/{page=1:int}")]
+        public IActionResult ViewFrameworks(string? searchString = null,
+            string? sortBy = null,
+            string sortDirection = BaseSearchablePageViewModel.Ascending,
+            int page = 1,
+            string tabname = "All")
+        {
+            sortBy ??= FrameworkSortByOptions.FrameworkName.PropertyName;
```
AlexJacksonDS, line 50: Change the sortBy & sortDirection defaults

DigitalLearningSolutions.Web/Controllers/FrameworksController/Frameworks.cs
```diff
@@ -1,603 +1,607 @@
...
+            }
+            var adminId = GetAdminID();
+            var sameItems = frameworkService.GetFrameworkByFrameworkName(frameworkname, adminId);
+            var frameworks = frameworkService.GetAllFrameworks(adminId);
+            var sortedItems = GenericSortingHelper.SortAllItems(
+               frameworks.AsQueryable(),
+               FrameworkSortByOptions.FrameworkName.PropertyName,
+               BaseSearchablePageViewModel.Ascending
+           );
+            var similarItems = GenericSearchHelper.SearchItems(sortedItems, frameworkname, 55, true);
```
AlexJacksonDS, line 252: Change the helpers being used

AlexJacksonDS (commented):
Frameworks.cs changes. All other changes in the file are as a result of fixing the mixed line endings

DigitalLearningSolutions.Web/Helpers/SelectListHelper.cs
```diff
@@ -20,5 +20,15 @@ public static IEnumerable<SelectListItem> MapOptionsToSelectListItems(IEnumerabl
         {
             return options.Select(o => new SelectListItem(o, o, o == selected)).ToList();
         }
+
+        public static IEnumerable<SelectListItem> MapOptionsToSelectListItems(IEnumerable<(string, string)> options)
+        {
+            return options.Select(NewSelectListItemFromTupleString);
```
stellake, line 26: Could we just do:
`return options.Select(option => new SelectListItem(option.text, option.value));`

(I don't think we get much benefit from the NewSelectListItemFromTupleString but can be convinced otherwise :P

DigitalLearningSolutions.Web/Helpers/SearchHelper.cs
```diff
@@ -1,163 +1,107 @@
 ﻿namespace DigitalLearningSolutions.Web.Helpers
```
stellake, line 1: I'm assuming this is just formatted and not changed? :)

stellake: If so, could we add a comment at the top of this saying that we're trying to move away from the non-generic version, pointing the user to the new file? 🙏

AlexJacksonDS: I've also removed the unused methods for courses + frameworks. Same for SortingHelper

DigitalLearningSolutions.Web/Scripts/searchSortAndPaginate/searchSortAndPaginate.ts
```diff
@@ -0,0 +1,100 @@
+import Details from 'nhsuk-frontend/packages/components/details/details';
```
stellake, line 1: I'm assuming this is basically copy-paste of the old version - is there anything particular that I should focus on if it has been changed?

AlexJacksonDS: Only major change other than the renaming is removing the GetPathFromEndpoint logic that was tying it to the /LearningPortal/ endpoints

DigitalLearningSolutions.Web/Scripts/spec/sort.spec.ts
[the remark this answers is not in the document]
stellake: I think it's still worth it to make sure the admin HTML is set up appropriately and works (all the mappings etc), but I agree that let's add it when we do sorting 👍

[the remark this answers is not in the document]
AlexJacksonDS: This doesn't test against any of the cshtml files. If I were to add tests as you suggest, I would be writing new html in the spec file to test against. I agree that we will probably need tests to confirm that the filtering javascript works.

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/Index.cshtml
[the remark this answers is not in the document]
stellake: No need - thank you! 🙏

stellake (approved):
A few minor comments (e.g. adding comments to the old files etc) + we should amend the documentation after merging, but these can be done without rereview! Looks good! 👍

DigitalLearningSolutions.Web/Helpers/SelectListHelper.cs
AlexJacksonDS: I think it was more of a left over as I was refactoring. Changed to just do it all in the public one.

PR #403, opened 2021-06-10
HEEDLS-357 Registration for particular centre

ibrahimmunir14:
- Added support for centre-specific registration links with centreId specified by a query parameter.
- Centre-specific registration does not have a centre selector on the first page, or a change centre button on the summary page.
- A user can switch to general registration by visiting the standard Register url (e.g. by clicking the Register button in the navbar), or by clicking the change centre link on the first page.
- An invalid centreId parameter will return 404.
- Registration functionality behaves as expected.

- Registration data is cleared if a user restarts the registration process (from the normal register link or a centre-specific link).
- The first page the user sees is now the Personal Information page. Index redirects to PersonalInformation after checking auth status and centreId query parameter.

All tests passing. Manually tested different scenarios.

**UI changes:**
image
  👇 
image

image


DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -120,6 +162,8 @@ public IActionResult LearnerInformation()
             var centreId = (int)data.RegisterViewModel.Centre;
 
             SetLearnerInformationViewBag(model, centreId);
+            ViewBag.IsCentreSpecific = data.RegisterViewModel.IsCentreSpecific;
+            ViewBag.CentreId = data.RegisterViewModel.Centre.Value;
```
AlexJacksonDS, line 166: I'm not sure why either of these need to go in the viewbag if you are getting their values off a model? 

Also they should be done in the SetLearnerInformationViewBag method if they do. (Or at least pulled into another method since these 2 lines of code are repeated a few times)

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -89,6 +115,22 @@ public IActionResult Index(RegisterViewModel model)
...
+                    {
+                        model.Centre = null;
+                        model.IsCentreSpecific = false;
+                    }
+                    else
+                    {
+                        // otherwise continue centre-specific registration process
+                        ViewBag.CentreName = centreName;
+                    }
+                }
```
AlexJacksonDS, line 133: I'm not sure the logic here is quite right either. In the POST action, I don't see how we hit this code.

Either we are getting our set value from the hidden (which can assume has a value so we won't get to this code as ModelState will be valid) or we are in the non-centre specific journey and they really have tried to post without selecting a centre.

If we are concerned about the hidden field being changed, I think it would be better do a different check for this in the centre specific case outside (before) the ModelState.IsValid check, but I don't think that we should be handling this case anyway. If this value has been changed from its set value in the centre specific journey, this suggests to me that the user is maliciously trying to change centre. (not that that would do them much good since they'd just get rejected anyway)

DigitalLearningSolutions.Web/Views/Register/Index.cshtml
```diff
@@ -15,25 +15,40 @@
...
-          to access all Digital Learning Solutions services.
+        @if (Model.IsCentreSpecific)
+        {
+          <p class="nhsuk-body-l">
+            Please enter your personal details to start the registration process
+            to access all Digital Learning Solutions services at <b>@ViewBag.CentreName</b>.
+            <a class="nhsuk-action-link__link nhsuk-link--no-visited-state nhsuk-u-padding-left-0 nhsuk-body-m nhsuk-u-margin-top-2" asp-action="Index">
+              Not the right centre? Click here to change.
+            </a>
         </p>
```
AlexJacksonDS, line 28: Minor, but the indentation seems to have gone wrong here. Should just be able to run the cleanup code to fix it.

AlexJacksonDS (changes requested):
A couple of places where the logic seems slightly off to me, but happy to discuss them and decide that they are OK.

Also a couple of minor refactoring changes.

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
ibrahimmunir14: They only exist in the RegisterViewModel, so for other pages, I've put them into the viewbag if their values are needed. I'll pull them into another method.

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
ibrahimmunir14: I see. The only way the centreName==null conditional is true is if the user maliciously changes the centreId to a non-existing one, or the centre for some reason becomes inactive or invalid. And this can just give a 500 error. So I will remove that check. We do however need to check isCentreSpecific, and set the CentreName if it is, so I will leave that in.

ibrahimmunir14: I put the check in because we have to GetCentreName anyway so might as well make sure it worked, but I see how that unnecessarily complicates the code.

AlexJacksonDS: My point was more that even if a user maliciously changes it to another value (whether it exists or not), this doesn't currently invalidate the model, and so none of this code even runs. Happy to talk about it if that would be easier that going back and forth in comments on the pull request

ibrahimmunir14: Ah yes, I get it.

AlexJacksonDS (approved):
Looks good now.

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -1,4 +1,4 @@
-﻿namespace DigitalLearningSolutions.Web.Controllers
+namespace DigitalLearningSolutions.Web.Controllers
```
stellake, line 1: Could we write some controller tests for this change?

DigitalLearningSolutions.Web.Tests/Controllers/Register/RegisterControllerTests.cs
```diff
@@ -34,57 +35,114 @@ public void Setup()
...
             A.CallTo(() => userService.GetUsersByEmailAddress(duplicateUser.EmailAddress!)).MustHaveHappened();
             result.Should().BeRedirectToActionResult().WithActionName("LearnerInformation");
         }
+
+        [Test]
+        public void IndexPost_with_invalid_centreId_param_shows_error()
+        {
+            // Given
+            controller.TempData.Set(new DelegateRegistrationData());
+            var centreId = CentreTestHelper.GetDefaultCentre().CentreId;
```
stellake, line 102: Do we need to get the id off a default centre or can it be any int given that we're mocking the centresDataService anyway?

DigitalLearningSolutions.Web.Tests/Controllers/Register/RegisterControllerTests.cs
```diff
@@ -34,57 +35,114 @@ public void Setup()
...
+
+            // When
+            var result = controller.Index(centreId);
+
+            // Then
+            A.CallTo(() => centresDataService.GetCentreName(centreId)).MustHaveHappened(1, Times.Exactly);
+            var data = controller.TempData.Peek<DelegateRegistrationData>()!;
+            data.Centre.Should().Be(centreId);
+            data.IsCentreSpecificRegistration.Should().Be(true);
```
stellake, line 127: You should be able to use .BeTrue() (and .BeFalse() for false assertions)

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -40,66 +42,62 @@ CustomPromptHelper customPromptHelper
...
+                return NotFound();
             }
 
-            ViewBag.CentreOptions = SelectListHelper.MapOptionsToSelectListItems(
-                centresDataService.GetActiveCentresAlphabetical(),
-                delegateRegistrationData.RegisterViewModel.Centre
-            );
+            var delegateRegistrationData = CreateDelegateRegistrationData(centreId);
+            delegateRegistrationData.IsCentreSpecificRegistration = centreId.HasValue;
+            delegateRegistrationData.Centre = centreId;
```
stellake, line 59: Doesn't this get set in the "CreateDelegateRegistrationData"? Also, if we named it "SetDelegateRegistrationData", we could move the TempData.Set into the method

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -265,7 +253,38 @@ public IActionResult Confirmation()
...
+        }
+
+        private bool CheckCentreIdValid(int? centreId)
+        {
+            if (centreId == null)
+            {
+                return true;
+            }
+
+            return centresDataService.GetCentreName(centreId.Value) != null;
```
stellake, line 284: Could be simplified to `centreId == null || centresDataService.GetCentreName(centreId.Value) != null`;

DigitalLearningSolutions.Web/ViewModels/Register/SummaryViewModel.cs
```diff
@@ -2,33 +2,27 @@
...
-            Centre = centre;
-            JobGroup = jobGroup;
-        }
         public string? FirstName { get; set; }
         public string? LastName { get; set; }
         public string? Email { get; set; }
         public string? Centre { get; set; }
         public string? JobGroup { get; set; }
         public bool Terms { get; set; }
+        public IEnumerable<EditCustomFieldViewModel>? CustomFields { get; set; }
```
stellake, line 15: Ah - do we want to pull the EditCustomFieldVM from MyAccount to somewhere shared/do we need all the fields you need for editing (isn't there a whole list of options)?

DigitalLearningSolutions.Web/Helpers/RegistrationMappingHelper.cs
```diff
@@ -8,33 +8,86 @@ public static class RegistrationMappingHelper
...
+            data.JobGroup = model.JobGroup;
+            data.Answer1 = model.Answer1;
+            data.Answer2 = model.Answer2;
+            data.Answer3 = model.Answer3;
+            data.Answer4 = model.Answer4;
+            data.Answer5 = model.Answer5;
+            data.Answer6 = model.Answer6;
+            return data;
         }
     }
```
stellake, line 92: Using the mapper to get ViewModel is great (I think we could do some of it in the ViewModels' constructors but David mentioned that there was some weird behaviour around this back in the day) 

I wonder if instead of the ViewModel to Data methods, we could declare some public methods on the data object - e.g. SetPersonalInformation(model), SetLearnerInformation(model), ClearCustomPromptAnswers() (we can just move the relevant tests to view model tests).

DigitalLearningSolutions.Web/Models/DelegateRegistrationData.cs
```diff
@@ -1,20 +1,29 @@
...
         {
             Id = new Guid();
-            RegisterViewModel = new RegisterViewModel();
-            LearnerInformationViewModel = new LearnerInformationViewModel();
         }
 
         public Guid Id { get; set; }
-        public RegisterViewModel RegisterViewModel { get; set; }
-        public LearnerInformationViewModel LearnerInformationViewModel { get; set; }
         public string? PasswordHash { get; set; }
```
stellake, line 13: I really like how you've grouped the fields per form section - do you think it's worth moving the passwordHash to the end to match the page order?

stellake (commented):
Just had a proper look & I think it looks great - just need to move the CustomPromptsViewModel to common and I wrote a few extra comments that are really minor.

DigitalLearningSolutions.Web/Controllers/RegisterController.cs
```diff
@@ -278,43 +290,79 @@ private void ValidateEmailAddress(RegisterViewModel model)
...
+                data.Centre!.Value,
+                data.Answer1,
+                data.Answer2,
+                data.Answer3,
+                data.Answer4,
+                data.Answer5,
+                data.Answer6
+            );
+            return customFields;
+        }
```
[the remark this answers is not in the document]
ibrahimmunir14: We need GetCustomFieldsFromModel to get the answers from the model, when they are not stored in data. In particular, when LearnerInformation POST has validation errors, the user's typed answers may not be in the data but will be in the model.
Conversely, we need GetCustomFieldsFromData for the case where the answers are stored in data but not in a model, specifically, when populating a SummaryViewModel

stellake (approved):
Looks good! 👍

PR #404, opened 2021-06-10
HEEDLS-471 Make default size of the placeholder image smaller

AlexJacksonDS:
Make the largest size that the centre configuration placeholder image can be 300 x 150px

image


stellake (approved):
Looks good! 👍

PR #405, opened 2021-06-10
HEEDLS-383 Make default action on enter Add prompt

AlexJacksonDS:
Made the default action on enter on both the Edit Prompt and Add prompt configure answers screen Add prompt. This only gets triggered when an input is selected and enter is pressed. This is fine on the Add prompt screen as the Add Answer is only input.

For the Edit Prompt screen, this may cause some unexpected form posts if you hit enter when you have the Mandatory checkbox selected, but I think that is probably relatively unlikely compared to the fairly common behaviour to enter something in a textbox and hit enter to submit it.


stellake (approved):
Looks good! 👍

PR #407, opened 2021-06-10
HEEDLS-481 Handle delegate id from old site claims

ibrahimmunir14:
Modify GetCandidateId method to return null if the delegate id claim is 0, as the old site sets delegate id claim to 0 for admin-only users.
Refactor usage of GetClaimAsInt for LearnCandidateId into GetCandidateId.
Add a test. All tests are passing.

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/Helpers/CustomClaimHelperTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/Helpers/CustomClaimHelperTests.cs
@@ -154,7 +154,7 @@
         }
 
         [Test]
-        public void GetCandidateIdHandlesZeroAsNull()
+        public void GetCandidateId_handles_zero_as_null()
         {
             // Given
             var user = new ClaimsPrincipal(new ClaimsIdentity(new[]
--- a/DigitalLearningSolutions.Web.Tests/Helpers/CustomClaimHelperTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/Helpers/CustomClaimHelperTests.cs
@@ -152,5 +152,21 @@
             // Then
             Assert.Throws<FormatException>(() => user.GetCustomClaimAsRequiredInt(CustomClaimTypes.LearnCandidateId));
         }
+
+        [Test]
+        public void GetCandidateIdHandlesZeroAsNull()
+        {
+            // Given
+            var user = new ClaimsPrincipal(new ClaimsIdentity(new[]
+            {
+                new Claim(CustomClaimTypes.LearnCandidateId, "0"),
+            }, "mock"));
+
+            // When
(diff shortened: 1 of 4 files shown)


DigitalLearningSolutions.Web.Tests/Helpers/CustomClaimHelperTests.cs
```diff
@@ -152,5 +152,21 @@ public void GetCustomClaimAsRequiredInt_handles_claim_invalid()
             // Then
             Assert.Throws<FormatException>(() => user.GetCustomClaimAsRequiredInt(CustomClaimTypes.LearnCandidateId));
         }
+
+        [Test]
+        public void GetCandidateIdHandlesZeroAsNull()
```
AlexJacksonDS, line 157: Test looks good but can you change the name to match the format of the rest of them

DigitalLearningSolutions.Web/Helpers/CustomClaimHelper.cs
```diff
@@ -11,7 +11,8 @@ public static class CustomClaimHelper
 
         public static int? GetCandidateId(this ClaimsPrincipal user)
         {
-            return user.GetCustomClaimAsInt(CustomClaimTypes.LearnCandidateId);
+            var id = user.GetCustomClaimAsInt(CustomClaimTypes.LearnCandidateId);
+            return id == 0 ? null : id;
         }
 
         public static int GetCandidateIdKnownNotNull(this ClaimsPrincipal user)
```
AlexJacksonDS, line 18: I wonder if we also need to sort out this method. When I initially created the nullable int version because of the admin users with no delegate ID causing unhandled exceptions, it was purposely left so we didn't have to go and handle the nullable int everywhere. Probably beyond the scope of this ticket but something to think about @stellake

AlexJacksonDS (approved):
Just one minor naming comment and a question for Stella. Will pass it on without re-review for the naming.

Agreed on your comments in the MR channel, we should probably try to refactor some of the other things accessing claims directly in the future, but it is beyond the scope of this ticket

DigitalLearningSolutions.Web.Tests/Helpers/CustomClaimHelperTests.cs
ibrahimmunir14: Fixed

DigitalLearningSolutions.Web/Helpers/CustomClaimHelper.cs
stellake: Yes, you're right! This is outside the scope and on my TODO! 👍

stellake (approved):
Looks good! 👍

PR #408, opened 2021-06-11
HEEDLS-429 enumeration model binder

ibrahimmunir14:
[HEEDLS-429](https://softwiretech.atlassian.net/browse/HEEDLS-429)

- Add enumeration model binder so we can use enumeration classes as endpoint parameter types (work in progress on branch enumeration-model-binder)
- Return 404 instead of 500 from edit notification preference endpoints when user type not recognised
- Change string userType params of edit notification preferences endpoints to UserType userType.

Tested the following:
- `/edit/AdminUser` works for an admin user
- `/edit/DelegateUser` works for a delegate user
- `/edit/AdminUser` returns 404 for a delegate-only user
- `/edit/DelegateUser` returns 404 for admin-only user
- `/edit/skjdhgfsfdg` returns 404 for any user
- `/edit/` returns 404 for any user


DigitalLearningSolutions.Web/Controllers/NotificationPreferencesController.cs
```diff
@@ -61,11 +68,17 @@ public IActionResult UpdateNotificationPreferences(string userType)
...
+            var userId = Equals(userType, UserType.AdminUser)
                 ? User.GetAdminId()
-                : User.GetCandidateIdKnownNotNull();
+                : Equals(userType, UserType.DelegateUser)
+                    ? User.GetCandidateId()
+                    : null;
+            if (userId == null)
+            {
+                return NotFound();
+            }
```
SteveJacksonSoft, line 81: I find it's a bit difficult to read multiple ternary statements chained together.
Is there a reason you've not gone for a switch statement?
That would allow you to put the `return NotFound()` closer to the condition in which it gets run too.
Same on other endpoint

DigitalLearningSolutions.Web/Controllers/NotificationPreferencesController.cs
```diff
@@ -61,11 +68,17 @@ public IActionResult UpdateNotificationPreferences(string userType)
         [Authorize]
         [HttpPost]
         [Route("/NotificationPreferences/Edit/{userType}")]
-        public IActionResult SaveNotificationPreferences(string userType, IEnumerable<int> notificationIds)
+        public IActionResult SaveNotificationPreferences(UserType userType, IEnumerable<int> notificationIds)
```
SteveJacksonSoft, line 71: Can we have a `UserType?` instead actually? That makes the nullability much more obvious to readers.
Same on other endpoint.

SteveJacksonSoft (changes requested):
Nice, just a couple of changes that should make it a bit more readable.

DigitalLearningSolutions.Web/Controllers/NotificationPreferencesController.cs
ibrahimmunir14: From my understanding a switch statement needs constant values to compare against; we're using object instances since UserType is a subclass of Enumeration not an actual enum. So I don't think it's possible (please enlighten if I'm wrong). But I've refactored and used if statements for clarity.

SteveJacksonSoft: It looks like they made some improvements in C# 7.0 (we're using 8.0): https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/switch#the-match-expression

I'd say it's worth giving it a go if you haven't already. If it doesn't work, what you've got looks great.

DigitalLearningSolutions.Web/Controllers/NotificationPreferencesController.cs
SteveJacksonSoft: Does it work to stick a `?` on `UserType` here?

DigitalLearningSolutions.Web/Controllers/NotificationPreferencesController.cs
ibrahimmunir14: The match expression can be anything but the case label patterns have to be constants or types. I'm afraid I still don't see how to get that to work.

DigitalLearningSolutions.Web/Controllers/NotificationPreferencesController.cs
```diff
@@ -46,30 +46,48 @@ public IActionResult Index()
...
-                ? User.GetAdminId()
-                : User.GetCandidateIdKnownNotNull();
+            var userId = GetUserId(userType);
+            if (userId == null)
+            {
+                return NotFound();
+            }
 
-            notificationPreferencesService.SetNotificationPreferencesForUser(userType, userId, notificationIds);
+            notificationPreferencesService.SetNotificationPreferencesForUser(userType!, userId, notificationIds);
```
SteveJacksonSoft, line 75: Ah, yeah, so now that you've extracted `GetUserId()` this `!` scares me a bit, because it's relying on the implementation of a method you're calling - someone might come along and change the implementation of `GetUserId()` later and let a null `userType` through.
I think extracting `GetUserId()` is a good shout. And I think you could get around this null check by modifying `GetUserId()` so that it returns a `UserRef?` instead, which should be null if the user doesn't have an account for the `userType` provided. Then you can reject with `NotFound` if the `userRef` is null, and use `userRef.UserType` here, which the compiler will know is non-null.
I think that's worth changing, but I'll pass you onto @stellake - she might have thoughts on this too.

DigitalLearningSolutions.Web/Controllers/NotificationPreferencesController.cs
SteveJacksonSoft: Ah, sorry, yep, I was barking up the wrong tree there. Fair enough!

DigitalLearningSolutions.Web/Controllers/NotificationPreferencesController.cs
```diff
@@ -66,28 +71,35 @@ public IActionResult UpdateNotificationPreferences(UserType? userType)
...
             }
-            if (Equals(userType, UserType.DelegateUser))
+            else if (Equals(userType, UserType.DelegateUser))
             {
-                return User.GetCandidateId();
+                userId = User.GetCandidateId();
             }
-            return null;
+
+            return userId != null ? new UserReference(userId.Value, userType!) : null;
```
stellake: This is really minor, but this might be easier to follow if we return immediately when reaching a true condition (overriding variables always makes the logic a bit trickier than immediate return):

```
 if (Equals(userType, UserType.AdminUser))
{
return new UserReference(User.GetAdminId().Value, UserType.AdminUser) 
}
else if (Equals(userType, UserType.DelegateUser))
{
return new UserReference(User.GetCandidateId().Value, UserType.DelegateUser) 
}
return null;
}
```

stellake: Actually, ignore this - we could have null id even if the usertype is different! 👍 🤦

stellake (approved):
Looks good! 👍

PR #409, opened 2021-06-11
Heedls 424 database delete rule change

JBsoftwire:


DigitalLearningSolutions.Data/DataServices/UserDataService.cs
```diff
@@ -351,18 +351,6 @@ public void RemoveDelegateUser(int delegateId)
...
-                    new { delegateId }
-                );
-
-                if (existingSessions.Any())
-                {
-                    throw new UserAccountInvalidStateException(
-                        $"Delegate user id {delegateId} cannot be removed as they have already started a session."
-                        );
-                }
-
```
stellake, line 365: Ah - should we Dispose in the finally block?

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests.cs
```diff
@@ -318,7 +319,7 @@ public void RemoveDelegateUser_cannot_remove_delegate_user_with_started_session(
                     Action action = () => userDataService.RemoveDelegateUser(id);
 
                     // Then
-                    action.Should().Throw<UserAccountInvalidStateException>();
+                    action.Should().Throw<SqlException>();
```
stellake, line 322: Nice - does this give a reasonable error in the logs so it's clear where/why the error came from?

DigitalLearningSolutions.Data/DataServices/UserDataService.cs
JBsoftwire: The dispose is handled in the catch block - dispose reverts the transaction so we don't want to fire it of there's no exception.

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests.cs
JBsoftwire: It does, the SQL error identifies the foreign key constraint that was violated.

stellake (approved):
Looks great! 👍

PR #410, opened 2021-06-11
HEEDLS-382 Clear temp data on cancel

AlexJacksonDS:
Clear the temp data when the cancel link on the first page of the add prompt journey is clicked.


AlexJacksonDS:
Will sort the issue this deals with under 384 that is changing the same area

PR #411, opened 2021-06-11
HEEDLS-482 Remove clickable link from admin card when user doesn't ha…

AlexJacksonDS:
…ve permissions

Added a new claim helper method for the CentreManager permission (same permissions as the Administrators page)

Used this to disable the link in the Admins card on the dashboard page.

Refactored each section of the Dashboard view into a partial view.

image


davidm-m (approved):
Nice use of partials, looks good!

stellake:
Can you post an image of this version to the hee external channel to see if HEE is happy with the design?

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Dashboard/Index.cshtml
```diff
@@ -22,209 +22,10 @@
     <h1 class="nhsuk-heading-xl" id="app-page-heading">Centre dashboard</h1>
     <p class="nhsuk-body-m">Welcome <span class="nhsuk-u-font-weight-bold">@Model.FirstName</span>. You are managing <span class="nhsuk-u-font-weight-bold">@Model.CourseCategory</span> courses.</p>
 
-  <ul class="nhsuk-grid-row nhsuk-card-group dashboard-4-1-card-group">
+    <partial name="_Dashboard4x1CardGroup" model="Model"/>
```
stellake, line 25: Nice! I like the partials split but I wonder if we could rename it into something more intuitive - I would probably find it easier to find the partial I want to modify without having to go to the Index page, if we had DashboardTopCardGroup/BottomCardGroup partials? Then, if the style of these section changes to say 3x2 we wouldn't have to rename these either 🤔

stellake (approved):
Looks good - a small comment about naming + running this past HEE! 👍

AlexJacksonDS:
Confirmation from Kevin in slack that it is OK.

PR #413, opened 2021-06-15
HEEDLS-384 Add bulk registration prompt answer page

AlexJacksonDS:
Added new bulk registration page that is accessible from both the Edit prompt journey and Add prompt journey. Changed the text area view component to allow multiple error messages in the same way as the text input component

Checked that the screen reader works as expected and mobile layout.

image

image

image


AlexJacksonDS:
Note that this also fixes the outstanding issue with temp data for 382

davidm-m (approved):
Happy with this in general, just want to check it's been tested with the full 4000 characters. Cookies have a max size of 4096 bytes so we're cutting it fine, and (while this is very unlikely) if someone tries to enter a lot of options with non-ASCII characters then we could go over that limit

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
```diff
@@ -22,7 +22,9 @@ public class RegistrationPromptsController : Controller
         public const string AddPromptAction = "addPrompt";
         public const string NextAction = "next";
         public const string SaveAction = "save";
+        public const string BulkAction = "bulk";
         private const string CookieName = "AddRegistrationPromptData";
```
stellake, line 26: Should this now be "AddPromptCookieName" (and the other one might need "Prompt" in the name too) :)

DigitalLearningSolutions.Web.Tests/ServiceFilter/RedirectEmptySessionDataTests.cs
```diff
@@ -60,5 +63,96 @@ public void Does_not_set_action_result_if_there_is_temp_data_matching_model()
...
+        {
+            // Given
+            var context = new ActionExecutingContext(
+                new ActionContext(
+                    new DefaultHttpContext(),
+                    new RouteData(new RouteValueDictionary()),
+                    new ActionDescriptor()),
+                new List<IFilterMetadata>(),
+                new Dictionary<string, object>(),
+                new RegistrationPromptsController(A.Fake<ICustomPromptsService>(), A.Fake<IUserDataService>()).WithDefaultContext().WithMockTempData());
```
stellake, line 78: Can we pull this controller into var like you've done below? 🙏

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
```diff
@@ -55,7 +66,15 @@ public IActionResult EditRegistrationPrompt(int promptNumber)
             var customPrompt = customPromptsService.GetCustomPromptsForCentreByCentreId(centreId).CustomPrompts
                 .Single(cp => cp.CustomPromptNumber == promptNumber);
 
-            return View(new EditRegistrationPromptViewModel(customPrompt));
+            var data = TempData.Peek<EditRegistrationPromptData>();
+
+            var model = data != null
```
stellake, line 71: We can use TempData Get instead of Peek here as it clears the tempdata by default. :)

stellake:
Re styling - could we make sure that the spacing is: 2 <= 1 <= 3 (see image for number references)
Might be worth checking what styles we use for normal text inputs.. 😄 

Also, the Cancel button looks quite low to me - is that spacing consistent with the rest of the site?

image

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/AddCentreCustomPromptsAccessibilityTests.cs
```diff
@@ -31,6 +31,10 @@ public void AddRegistrationPrompt_journey_has_no_accessibility_errors()
             ValidatePageHeading("Configure answers");
             var configureAnswerWithAnswersResult = new AxeBuilder(Driver).Analyze();
 
+            Driver.ClickButtonByText("Bulk add");
+            var bulkAdditionResult = new AxeBuilder(Driver).Analyze();
```
stellake, line 35: Do you think it's worth checking page titles here too to ensure we're scanning the expected page? 🤔

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
```diff
@@ -302,6 +388,31 @@ private IActionResult AddRegistrationPromptConfigureAnswersPostNext(Registration
...
+
+            Response.Cookies.Append(
+                EditCookieName,
+                id.ToString(),
+                new CookieOptions
+                {
+                    Expires = DateTimeOffset.UtcNow.AddDays(1)
+                }
+            );
+            TempData.Set(data);
```
stellake, line 411: Nice - two questions about this.
1. Could we pass "model" into the EditRegistrationPromptData constructor and set the EditModel when creating the object?
2. Could we pull the L399-411 into a private method "SetEditRegistrationPromptTempData" or similar?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
```diff
@@ -71,32 +90,83 @@ public IActionResult EditRegistrationPrompt(EditRegistrationPromptViewModel mode
...
+                TempData.Set(addData);
+
+                return RedirectToAction("AddRegistrationPromptConfigureAnswers");
             }
 
+            var editData = TempData.Peek<EditRegistrationPromptData>()!;
+            editData.EditModel!.OptionsString = model.OptionsString;
+            TempData.Set(editData);
+
+            return RedirectToAction("EditRegistrationPrompt", new { promptNumber = model.PromptNumber });
```
stellake, line 138: Now looking at this I wonder whether it would be worth creating two endpoints: `/Add/Bulk` and `/Edit/Bulk` but still use the same BulkRegistrationPromptAnswersViewModel. Given that you've pulled the ValidateBulkOptionsString(model.OptionsString); into a private method, the duplication between the methods would be minimised. What do you think?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
```diff
@@ -71,32 +90,83 @@ public IActionResult EditRegistrationPrompt(EditRegistrationPromptViewModel mode
...
+
+            var addRegistrationPromptData = new AddRegistrationPromptData();
+            var id = addRegistrationPromptData.Id;
+
+            Response.Cookies.Append(
+                CookieName,
+                id.ToString(),
+                new CookieOptions
+                {
+                    Expires = DateTimeOffset.UtcNow.AddDays(1)
```
stellake, line 155: Could we pull the cookie expiry time in days to the top of the file and use it for both add & edit? (then it can be changed easily too)
Also, I now wonder whether we should increase the time a bit as we clear it at the beginning of the journey anyway (I'm thinking of maybe a week)🤔

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/CentreConfiguration/BulkRegistrationPromptAnswersViewModel.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace DigitalLearningSolutions.Web.ViewModels.TrackingSystem.CentreConfiguration
+{
+    public class BulkRegistrationPromptAnswersViewModel
+    {
+        public BulkRegistrationPromptAnswersViewModel(){}
```
stellake, line 5: Formatting 👀

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
AlexJacksonDS: There is no such strongly typed method for this. This will result in something like this:
`var model = TempData[nameof(EditRegistrationPromptData)] is EditRegistrationPromptData data
                ? data.EditModel!
                : new EditRegistrationPromptViewModel(customPrompt);`
Or should I try and write one in a similar pattern to the Peek<T> method?

AlexJacksonDS: Nevermind, that just completely doesn't work. The options are either keep the existing Peek/Clear flow or write a new extension. Will try to write the extension as it'll probably be useful later.

AlexJacksonDS:
New error message formatting (fixed for text inputs too)
image

I've also tried with ~3800 characters and it worked fine. I'm assuming TempData is silently handling this for us as I see 4 .AspNetCore.Mvc.CookieTempDataProviderC1-4 cookies

davidm-m:
> I've also tried with ~3800 characters and it worked fine. I'm assuming TempData is silently handling this for us as I see 4 .AspNetCore.Mvc.CookieTempDataProviderC1-4 cookies

Fun! That's really good to know for the future, thanks for testing 😄

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
stellake: Yes, a new extension sounds great! 👍

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/EditRegistrationPrompt.cshtml
```diff
@@ -82,11 +81,18 @@
...
                      hint-text=""
-                     css-class="nhsuk-u-width-full nhsuk-u-font-weight-bold"/>
+                     css-class="nhsuk-u-width-full nhsuk-u-font-weight-bold" />
       <button name="action" class="nhsuk-button nhsuk-button--secondary" value="@RegistrationPromptsController.AddPromptAction">Add</button>
     </div>
   </div>
 
+  <div class="nhsuk-grid-row divider">
+    <div class="nhsuk-grid-column-one-half">
+      <p class="nhsuk-label"> Want to add answers in bulk?</p>
```
stellake, line 91: Is there meant to be whitespace before the "Want.."?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
stellake: Can we not add the conditional for the asp-action rather than than the whole form tag?

DigitalLearningSolutions.Web/Extensions/TempDataExtension.cs
```diff
@@ -22,5 +26,18 @@ private static bool TryPeek(this ITempDataDictionary tempData, string key, out s
                     return false;
             }
         }
+
+        private static bool TryGet(this ITempDataDictionary tempData, string key, out string value)
```
stellake, line 30: Would it make sense to return a deserialised object or a null in this method rather than return a bool and assign an empty string to value that's never used?

AlexJacksonDS: I've just followed the same pattern as the existing Peek/TryPeek method. I'm not really sure why they were written that way in the first place. Do you think I should change both or neither?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
AlexJacksonDS: My only thinking there was we lose the validation that the endpoint exists if we do that.

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/EditCentreCustomPromptsAccessibilityTests.cs
```diff
@@ -0,0 +1,32 @@
...
+
+            // When
+            Driver.Navigate().GoToUrl(BaseUrl + startUrl);
+            ValidatePageHeading("Edit delegate registration prompt");
+            var editPageResult = new AxeBuilder(Driver).Analyze();
+
+            Driver.ClickButtonByText("Bulk add");
+            ValidatePageHeading("Configure answers in bulk");
+            var bulkAdditionResult = new AxeBuilder(Driver).Analyze();
+
```
stellake, line 27: Can we also click "Next" on bulk page and then validate that the page heading is "Edit delegate registration prompt again" to ensure the button is correctly configured (given that the IDE doesn't check this automatically)

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
```diff
@@ -71,32 +89,92 @@ public IActionResult EditRegistrationPrompt(EditRegistrationPromptViewModel mode
...
+                false,
+                promptNumber
+            );
+
+            return View("BulkRegistrationPromptAnswers", model);
+        }
+
+        [HttpPost]
+        [Route("Edit/Bulk")]
+        [ServiceFilter(typeof(RedirectEmptySessionDataXor<AddRegistrationPromptData, EditRegistrationPromptData>))]
```
stellake, line 115: Do we still need the Xor service filter?

AlexJacksonDS: Ah no, probably not anymore

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CentreConfiguration/RegistrationPromptsController.cs
```diff
@@ -71,32 +89,92 @@ public IActionResult EditRegistrationPrompt(EditRegistrationPromptViewModel mode
...
+            editData.EditModel!.OptionsString = model.OptionsString;
+            TempData.Set(editData);
+
+            return RedirectToAction("EditRegistrationPrompt", new { promptNumber = model.PromptNumber });
+        }
+
+        [HttpPost]
+        [Route("Add/Bulk")]
+        [ServiceFilter(typeof(RedirectEmptySessionDataXor<AddRegistrationPromptData, EditRegistrationPromptData>))]
+        public IActionResult AddRegistrationPromptBulkPost(BulkRegistrationPromptAnswersViewModel model)
```
stellake, line 134: Super minor, but do you think it's worth grouping this with the Add/Bulk GET endpoint?

DigitalLearningSolutions.Web/Extensions/TempDataExtension.cs
stellake: Nice change!

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/EditCentreCustomPromptsAccessibilityTests.cs
stellake: Minor, but we probably want the same for "Add" journey :)

stellake (approved):
Quick comment about the "Add prompt" test, otherwise looks good! 👍

PR #415, opened 2021-06-15
HEEDLS-485 Centre link in delegate approval email

ibrahimmunir14:


DigitalLearningSolutions.Data/Services/DelegateApprovalsService.cs
```diff
@@ -183,20 +196,22 @@ private static Email GenerateDelegateApprovalEmail
...
+                                <p>Your Digital Learning Solutions (DLS) registration at the centre {centreName} has been rejected by an administrator.</p>
+                                <p>There are several reasons that this may have happened including:
                                     <ul>
                                         <li>You registered with a non-work email address which was not recognised by the administrator</li>
                                         <li>Your DLS centre chooses to manage delegate registration internally</li>
                                         <li>You have accidentally chosen the wrong centre during the registration process.</li>
                                     </ul>
-                                If you need access to the DLS platform, please use the <a href=""{findCentreUrl}"">Find Your Centre page</a> to locate your local DLS centre and use the contact details provided there to ask for help with registration.</p>
```
AlexJacksonDS, line 199: I missed this in the review of the previous version, but \<ul> tags shouldn't be in \<p> tags according to the html spec. Can we close the p above the ul and re-open a new one for the subsequent line, making sure the indentation is updated.

For reference [p](https://html.spec.whatwg.org/multipage/grouping-content.html#the-p-element) can only contain [Phrasing content](https://html.spec.whatwg.org/multipage/dom.html#phrasing-content-2) elements. [ul](https://html.spec.whatwg.org/multipage/grouping-content.html#the-ul-element) is [Flow content](https://html.spec.whatwg.org/multipage/dom.html#flow-content-2)

AlexJacksonDS (commented):
Just a minor comment on the ordering of tags in the html version of the rejection email

AlexJacksonDS (approved):

DigitalLearningSolutions.Data.Tests/Services/DelegateApprovalsServiceTests.cs
```diff
@@ -20,6 +20,7 @@ public class DelegateApprovalsServiceTests
         private IUserDataService userDataService = null!;
         private ICustomPromptsService customPromptsService = null!;
         private IEmailService emailService = null!;
+        private ICentresDataService centresDataService = null!;
```
stellake, line 23: Could this just be `A.Fake` rather than `null!` (+ some others too)

DigitalLearningSolutions.Data/Services/DelegateApprovalsService.cs
```diff
@@ -100,8 +103,15 @@ private void SendDelegateApprovalEmails(params DelegateUser[] delegateUsers)
                 }
                 else
                 {
-                    var delegateApprovalEmail = GenerateDelegateApprovalEmail
-                        (delegateUser.CandidateNumber, delegateUser.EmailAddress, LoginUrl);
+                    var centreUrl = centresDataService.GetCentreDetailsById(delegateUser.CentreId)?.ShowOnMap == true
```
stellake, line 106: Super minor, but could this be "centreInformationUrl"?

DigitalLearningSolutions.Data.Tests/Services/DelegateApprovalsServiceTests.cs
ibrahimmunir14: Their values are set (most of them to `A.Fake`) in `SetUp()`.

stellake: But is there a reason we don't set the value here (to make the setup a bit neater)?

ibrahimmunir14: Not that I know of, but it's done in SetUp in other test files so I just did it that way for consistency.

ibrahimmunir14: We need new mock objects for each test case; if we initialise them at declaration they'll be reused as the fixture class is only instantiated once.

stellake: Ah, I see! And I guess clearing the mock at the end is equivalent to creating the mock in the setup! That's fair - happy to leave as is! 🙌

stellake (approved):
Looks great!

PR #417, opened 2021-06-15
HEEDLS-367 Add option to select list view component to make default optional selectable

davidm-m:
Fixes the bug reported by Jonathan on the ticket: "When registering, if you have an optional field with a specified list of options, once you've selected one you can't then unselect it and blank the field without going back a page and starting over."

Tested by going through the registration journey with a centre that has some mandatory and some optional select custom prompts, all were displayed and only the optional prompts could be deselected.


JBsoftwire (approved):
All looks good to me.

stellake (approved):
Looks good! I'd just double check that the line endings are correct before merging :)

PR #421, opened 2021-06-17
HEEDLS-482 Add style for non clickable card to look like clickable card

AlexJacksonDS:
Added a style nhsuk-card--not-clickable matching the nhsuk-card--clickable except it's hover properties so that the admin card looks the same as the other cards but has no hover/clickable properties

image


DanBloxham-sw (approved):

stellake (approved):
Looks good! 👍

stellake:
Just a quick question actually -- what would this look like with a normal unclickable card (no bottom border)?

PR #423, opened 2021-06-17
HEEDLS-487 await SignInAsync in login POST

AlexJacksonDS:
Changed the login method that was calling HttpContext.SignInAsync to await that and propagated up the async/awaits in the login controller and unit tests of it.

Ran all unit tests and did a quick sanity check that login still works.


davidm-m (approved):
Good stuff! I never realised LoginController was so comprehensively tested

stellake (approved):
Looks good! 👍

PR #429, opened 2021-06-18
HEEDLS-483 - Home and Iframe accessibility bug fixes.

DanBloxham-sw:
Updated Learning Content iframes to use new titles, with new breadcrumb. 
Changes iframes to have a min-height of 500px on desktops so they are a little more useable. Mobile has been left as it is to minimise confusion on the reliance of two scrollers.
Other minor changes also performed.

Ran all unit tests,
Tested in chrome,
Checked with screenreader

Example screen shot of learning content page:

image

### DIFF
--- a/DigitalLearningSolutions.Web/ViewModels/LearningContent/LearningContentViewModel.cs
+++ b/DigitalLearningSolutions.Web/ViewModels/LearningContent/LearningContentViewModel.cs
@@ -3,10 +3,12 @@
     public class LearningContentViewModel
     {
         public readonly string Brand;
+        public readonly string Title;
 
-        public LearningContentViewModel(string brand)
+        public LearningContentViewModel(string brand, string title)
         {
             Brand = brand;
+            Title = title;
         }
     }
 }
--- a/DigitalLearningSolutions.Web/Views/FindYourCentre/Index.cshtml
+++ b/DigitalLearningSolutions.Web/Views/FindYourCentre/Index.cshtml
@@ -1,10 +1,15 @@
-﻿@using DigitalLearningSolutions.Web.Helpers
-@using DigitalLearningSolutions.Web.ViewModels.FindYourCentre
+﻿@using DigitalLearningSolutions.Web.ViewModels.FindYourCentre
 @model FindYourCentreViewModel
 @{
   ViewData["Title"] = "Find Your Centre";
 }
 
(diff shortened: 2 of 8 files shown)


DigitalLearningSolutions.Web/ViewModels/LearningContent/LearningContentViewModel.cs
```diff
@@ -1,12 +1,14 @@
...
-        public readonly string Brand;
-
-        public LearningContentViewModel(string brand)
-        {
-            Brand = brand;
-        }
-    }
-}
+﻿namespace DigitalLearningSolutions.Web.ViewModels.LearningContent
+{
```
DanBloxham-sw, line 2: To clarify, this is definitely now CRLF line endings

DigitalLearningSolutions.Web/Views/FindYourCentre/Index.cshtml
```diff
@@ -1,10 +1,15 @@
-﻿@using DigitalLearningSolutions.Web.Helpers
-@using DigitalLearningSolutions.Web.ViewModels.FindYourCentre
+﻿@using DigitalLearningSolutions.Web.ViewModels.FindYourCentre
 @model FindYourCentreViewModel
 @{
   ViewData["Title"] = "Find Your Centre";
```
AlexJacksonDS, line 4: Should this be capitalised like this? Doesn't match the heading capitalisation

AlexJacksonDS (commented):
Just one question about capitalisation

DigitalLearningSolutions.Web/Views/FindYourCentre/Index.cshtml
DanBloxham-sw: According to @stellake , the capitalisation here is correct, but the consistency isn't too important. 

"So H1 should be only the first letter capitalised
But titles are opposite and usually have every first letter capitalised
But I wouldn't worry too much about it, the titles are only in the tab and mainly for screenreader users"

AlexJacksonDS: OK so we're just inconsistent with uncapitalised titles on half the pages then, but this one is correct.

AlexJacksonDS (approved):

stellake (approved):
Looks good 👍

PR #435, opened 2021-06-22
HEEDLS-450 - minor wording changes on some centre pages and the side…

DanBloxham-sw:
… nav menu.

Also made slightly change to the ViewData["Title"] value on the reports page to make it consistent with other centre pages.


stellake (approved):
Looks good! 👍

PR #440, opened 2021-06-22
HEEDLS-524 Refactor prompt management pages for consistent spacing

AlexJacksonDS:
Changed all the pages in the manage registration prompt area to have approximately equal spacing between the h1 elements and the page content and the bottom of the content and the back link view component.

image

image

image

image

image

image

### DIFF
--- a/DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/AddRegistrationPromptConfigureAnswers.cshtml
+++ b/DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/AddRegistrationPromptConfigureAnswers.cshtml
@@ -13,70 +13,50 @@
 }
 
 @section NavMenuItems {
-  <partial name="~/Views/TrackingSystem/Shared/_NavMenuItems.cshtml"/>
-}
-
-@if (errorHasOccurred) {
-  <div class="nhsuk-grid-row">
-    <div class="nhsuk-grid-column-full">
-      <vc:error-summary order-of-property-names="@new []{ nameof(EditRegistrationPromptViewModel.OptionsString), nameof(EditRegistrationPromptViewModel.Answer)}"/>
-    </div>
-  </div>
+  <partial name="~/Views/TrackingSystem/Shared/_NavMenuItems.cshtml" />
 }
 
 <div class="nhsuk-grid-row">
   <div class="nhsuk-grid-column-full">
+    @if (errorHasOccurred) {
+      <vc:error-summary order-of-property-names="@new []{ nameof(EditRegistrationPromptViewModel.OptionsString), nameof(EditRegistrationPromptViewModel.Answer)}" />
+    }
+
     <h1 class="nhsuk-heading-xl">Configure answers</h1>
-  </div>
-</div>
 
-<form class="nhsuk-u-margin-bottom-8" method="post" novalidate asp-action="AddRegistrationPromptConfigureAnswers">
-  <div class="hidden-submit">
-    <button name="action" class="nhsuk-button" value="@RegistrationPromptsController.AddPromptAction">Add</button>
-  </div>
+    <form class="nhsuk-u-margin-bottom-3" method="post" novalidate asp-action="AddRegistrationPromptConfigureAnswers">
+      <div class="hidden-submit">
+        <button name="action" class="nhsuk-button" value="@RegistrationPromptsController.AddPromptAction">Add</button>
+      </div>
 
-  <input type="hidden" asp-for="OptionsString"/>
+      <input type="hidden" asp-for="OptionsString" />
 
-  <div class="nhsuk-grid-row">
-    <div class="nhsuk-grid-column-full">
       @if (string.IsNullOrEmpty(Model.OptionsString)) {
-        <partial name="_NoConfiguredAnswers"/>
+        <partial name="_NoConfiguredAnswers" />
       } else {
-        <partial name="_RegistrationPromptAnswerTable" model="Model"/>
+        <partial name="_RegistrationPromptAnswerTable" model="Model" />
       }
-    </div>
-  </div>
 
-  <div class="nhsuk-grid-row divider">
-    <div class="nhsuk-grid-column-one-half">
-      <vc:text-input asp-for="Answer"
-                     label="Add a new answer?"
-                     populate-with-current-value="true"
-                     type="text"
-                     spell-check="true"
-                     autocomplete=""
-                     hint-text=""
-                     css-class="nhsuk-u-width-full nhsuk-u-font-weight-bold"/>
-      <button name="action" class="nhsuk-button nhsuk-button--secondary" value="@RegistrationPromptsController.AddPromptAction">Add</button>
-    </div>
-  </div>
+      <div class="divider">
+        <vc:text-input asp-for="Answer"
+                       label="Add a new answer?"
+                       populate-with-current-value="true"
+                       type="text"
+                       spell-check="true"
+                       autocomplete=""
+                       hint-text=""
+                       css-class="nhsuk-u-width-one-half nhsuk-u-font-weight-bold" />
+        <button name="action" class="nhsuk-button nhsuk-button--secondary" value="@RegistrationPromptsController.AddPromptAction">Add</button>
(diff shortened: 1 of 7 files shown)


SteveJacksonSoft:
It feels like there's quite a big gap below the headings on the Add and Remove pages, probably because of the margins on text elements. I'm guessing it's the same distance between the headers and next elements on all pages. I think I'd be inclined to make the space a bit smaller, but happy for @stellake to overrule me!

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/AddRegistrationPromptConfigureAnswers.cshtml
```diff
@@ -13,70 +13,50 @@
...
+        <partial name="_NoConfiguredAnswers" />
       } else {
-        <partial name="_RegistrationPromptAnswerTable" model="Model"/>
+        <partial name="_RegistrationPromptAnswerTable" model="Model" />
       }
-    </div>
-  </div>
-
-  <div class="nhsuk-grid-row divider">
-    <div class="nhsuk-grid-column-one-half">
```
SteveJacksonSoft, line 51: Does removing this `nhsuk-grid-column-one-half` make any difference to the appearance of the page (inc. on mobile)?

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/Index.cshtml
```diff
@@ -8,100 +8,48 @@
...
-                Remove
-              </a>
-            </div>
-          </div>
-        </div>
-      </details>
-    </div>
-  </div>
-}
+    <div class="@(hasSixPrompts ? "nhsuk-u-margin-bottom-7" : "nhsuk-u-margin-bottom-3")">
```
SteveJacksonSoft, line 37: What's special about 6 prompts?

SteveJacksonSoft (commented):
Looks nice, just a couple of questions!

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/AddRegistrationPromptConfigureAnswers.cshtml
AlexJacksonDS: Not significantly, I've change from a half-column with a full width textbox to a full width column with a half width textbox.

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/Index.cshtml
AlexJacksonDS: The Add new prompt button goes away. This margin + the card margins collapse, but this margin + the button margin don't collapse because the button has padding.

AlexJacksonDS:
The spacing after all the titles should be 48px for desktop, 40 for mobile. This comes from the nhsuk-heading-xl style and is consistent across all the files.

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/Index.cshtml
SteveJacksonSoft: Ah, okay. What about renaming `hasSixPrompts` to something like `canAddNewPrompt`? I'm guessing we don't let them put more than 6 prompts on the registration form? Or is it just that we only have 6 prompts that we can put on the form?

AlexJacksonDS: There can never be more that 6 custom prompts defined for a centre (currently) as that is the number of columns in the database for the centre. (CustomPrompt?Id, F?Options, F?Mandatory columns). But yes, I think changing the name would be better.

SteveJacksonSoft: Ah, I guess it's actually `can'tAddNewPrompt`. But yeah.
I sort of feel like the logic `canAddNewPrompt = promptCount < 6` belongs somewhere a bit higher up as well, perhaps as a property on the model? Or maybe as a helper method? Because it's not page-specific - it's more than just `you're not allowed to add another prompt on this page`; it's `no centre should ever have more than 6 prompts`. But I'm not sure whether it would actually have any significant affect on future development. I'll pass you onto TL review anyway.
What do you think, @stellake?

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/Index.cshtml
```diff
@@ -8,7 +8,7 @@
   ViewData["Application"] = "Tracking System";
   ViewData["HeaderPath"] = $"{Configuration["AppRootPath"]}/TrackingSystem/Centre/Dashboard";
   ViewData["HeaderPathName"] = "Tracking System";
-  var hasSixPrompts = Model.CustomFields.Count == 6;
+  var canAddNewPrompt = Model.CustomFields.Count == 6;
```
SteveJacksonSoft, line 11: Sorry, my renaming suggestion also needs the logic negated 🤦 .

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/Index.cshtml
AlexJacksonDS: I personally don't think it is worth it. There are only really 2 places it is relevant, here to remove a button, and in the Add prompt save logic (if they somehow manage to get into there through our redirects and temp data service filters) where it is handled in a different way (if no empty slots to save into, don't save and go to error page).

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/Index.cshtml
SteveJacksonSoft: Should be strictly less than 6 right?

AlexJacksonDS: I should have never renamed it 😢

DigitalLearningSolutions.Web/Views/TrackingSystem/CentreConfiguration/RegistrationPrompts/Index.cshtml
stellake: I think ideally, we'd have a "shouldDisplayAddNewPromptsButton" on the ViewModel as that's what the VCs are meant for, but I don't think it's too important and I'm inclined to leave it, because I think the current version of the code is readable and I'm inclined not to spend too long on this ticket. 👍

stellake (approved):
I think the spacing looks ok with the expanders and agree that if there is no expander directly after the heading, it does look a bit spacious. However, if this gap comes from the nhs-heading class as Alex mentioned, I think we should keep the default, but make sure it's consistent across pages. 👍

stellake:
> I think the spacing looks ok with the expanders and agree that if there is no expander directly after the heading, it does look a bit spacious. However, if this gap comes from the nhs-heading class as Alex mentioned, I think we should keep the default, but make sure it's consistent across pages. 👍

Just looking at the Centre ranking review and the spacing looks different there :P

AlexJacksonDS:
> > I think the spacing looks ok with the expanders and agree that if there is no expander directly after the heading, it does look a bit spacious. However, if this gap comes from the nhs-heading class as Alex mentioned, I think we should keep the default, but make sure it's consistent across pages. 👍
> 
> Just looking at the Centre ranking review and the spacing looks different there :P

id="page-heading" reduces the margin (and nothing else as far as I can see). I've been removing it as I've fixed these. I guess I need to remove it over there too.

PR #441, opened 2021-06-22
HEEDLS-525 Refactor my account views to fix inconsistent spacing

AlexJacksonDS:
Refactored the html in the my account edit details screen and the 2 manage notification screens to fix the spacing before the back links. I've changed the html in the update notification preferences relatively significantly as VS was complaining about invalid html (legend inside div). This means that the page does look a bit different to before, but it now looks more like the rest of the pages in the system.

All the automated accessibility tests still pass + checked with chrome lighthouse for 100 rating on accessibility on the update notification page.

image
image
image

### DIFF
--- a/DigitalLearningSolutions.Web/Views/MyAccount/EditDetails.cshtml
+++ b/DigitalLearningSolutions.Web/Views/MyAccount/EditDetails.cshtml
@@ -12,9 +12,11 @@
 
 @if (User.IsDelegateOnlyAccount()) {
   ViewData["Application"] = "Learning Portal";
-@section NavMenuItems {
-  <partial name="../LearningPortal/Shared/_NavMenuItems" />
-}}
+
+  @section NavMenuItems {
+    <partial name="../LearningPortal/Shared/_NavMenuItems" />
+  }
+}
 
 <div class="nhsuk-grid-row">
   <div class="nhsuk-grid-column-full">
--- a/DigitalLearningSolutions.Web/Views/MyAccount/EditDetails.cshtml
+++ b/DigitalLearningSolutions.Web/Views/MyAccount/EditDetails.cshtml
@@ -12,26 +12,21 @@
 
 @if (User.IsDelegateOnlyAccount()) {
   ViewData["Application"] = "Learning Portal";
-  @section NavMenuItems {
-  <partial name="../LearningPortal/Shared/_NavMenuItems"/>
-  }
-}
-
-@if (errorHasOccurred) {
-  <div class="nhsuk-grid-row">
-    <div class="nhsuk-grid-column-full">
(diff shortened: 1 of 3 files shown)


DigitalLearningSolutions.Web/Views/MyAccount/EditDetails.cshtml
```diff
@@ -12,26 +12,21 @@
...
-@if (errorHasOccurred) {
-  <div class="nhsuk-grid-row">
-    <div class="nhsuk-grid-column-full">
-      <vc:error-summary order-of-property-names="@(new[] { nameof(Model.FirstName), nameof(Model.LastName), nameof(Model.Email), nameof(Model.Password) })"/>
-    </div>
-  </div>
-}
+@section NavMenuItems {
+  <partial name="../LearningPortal/Shared/_NavMenuItems" />
+}}
```
SteveJacksonSoft, line 17: Formatting here looks a bit funky.

DigitalLearningSolutions.Web/Views/NotificationPreferences/UpdateNotificationPreferences.cshtml
```diff
@@ -17,40 +17,41 @@
   }
 }
 
-<form method="post" asp-controller="NotificationPreferences" asp-action="SaveNotificationPreferences" asp-route-userType="@Model.UserType">
-  <div class="nhsuk-form-group">
-    <div class="nhsuk-fieldset" aria-describedby="notification-hint">
-      <legend class="nhsuk-fieldset__legend nhsuk-fieldset__legend--l">
```
SteveJacksonSoft, line 23: The legend doesn't do anything useful, no?

SteveJacksonSoft (changes requested):
Formatting change and a question only, I'm afraid.

DigitalLearningSolutions.Web/Views/NotificationPreferences/UpdateNotificationPreferences.cshtml
AlexJacksonDS: It was invalid html because we weren't doing the fieldset properly (div is marked as nhsuk-fieldset). Maybe I should have instead changed the div to a fieldset? The page style is more consistent with the rest of the site now though

SteveJacksonSoft: Hmm. I guess the the `legend` came from the page originally being modeled on the nhs-checkboxes component. But I feel like we've overhauled the appearance of the component so much that it's not really an nhs-checkboxes element anymore (or a fieldset for that matter), so I agree with you, I don't think there's any point sticking with fieldset/legend.

SteveJacksonSoft (approved):

stellake (approved):
Looks good! 👍

PR #442, opened 2021-06-22
HEEDLS-381 Move hidden input for prompt name inside form

AlexJacksonDS:
Moved the hidden input used to preserve the prompt name on failed validation of the checkbox inside the form tag.


stellake (approved):
Looks good! 👍

PR #443, opened 2021-06-22
Heedls 459 usage stats report

JBsoftwire:
I think it's probably the case that the reports want to be in an expandable, 12 months is a lot.

image
image

### DIFF
--- a/DigitalLearningSolutions.Data/Helpers/DateHelper.cs
+++ b/DigitalLearningSolutions.Data/Helpers/DateHelper.cs
@@ -12,7 +12,12 @@
             var monthEnumerable = Enumerable.Range(startDate.Month, diffInMonths + 1);
 
             return monthEnumerable.Select(
-                m => ((m - 1) % 12 + 1, startDate.Year + (m - 1) / 12)
+                m =>
+                {
+                    var month = (m - 1) % 12 + 1;
+                    var yearsToAdd = (m - 1) / 12;
+                    return (month, startDate.Year + yearsToAdd);
+                }
             );
         }
     }
--- a/DigitalLearningSolutions.Data/Helpers/DateHelper.cs
+++ b/DigitalLearningSolutions.Data/Helpers/DateHelper.cs
@@ -1,0 +1,19 @@
+﻿namespace DigitalLearningSolutions.Web.Helpers
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    public static class DateHelper
+    {
+        public static IEnumerable<(int Month, int Year)> GetMonthsAndYearsBetweenDates(DateTime startDate, DateTime endDate)
+        {
+            var diffInMonths = (endDate.Year - startDate.Year) * 12 + (endDate.Month - startDate.Month);
+            var monthEnumerable = Enumerable.Range(startDate.Month, diffInMonths + 1);
+
+            return monthEnumerable.Select(
+                m => ((m - 1) % 12 + 1, startDate.Year + (m - 1) / 12)
+            );
+        }
+    }
+}
--- a/DigitalLearningSolutions.Data/Services/ActivityService.cs
+++ b/DigitalLearningSolutions.Data/Services/ActivityService.cs
@@ -14,15 +14,17 @@
     public class ActivityService : IActivityService
     {
         private readonly IActivityDataService activityDataService;
+        private readonly IClockService clockService;
 
-        public ActivityService(IActivityDataService activityDataService)
+        public ActivityService(IActivityDataService activityDataService, IClockService clockService)
         {
             this.activityDataService = activityDataService;
+            this.clockService = clockService;
         }
 
         public IEnumerable<MonthOfActivity> GetRecentActivity(int centreId)
         {
-            var date = DateTime.Now;
+            var date = clockService.UtcNow;
             var currentYear = date.Year;
             var monthsToDate = Enumerable.Range(1, date.Month);
 
--- a/DigitalLearningSolutions.Data/Services/ActivityService.cs
+++ b/DigitalLearningSolutions.Data/Services/ActivityService.cs
@@ -1,4 +1,4 @@
-namespace DigitalLearningSolutions.Data.Services
+﻿namespace DigitalLearningSolutions.Data.Services
 {
     using System;
     using System.Collections.Generic;
@@ -26,12 +26,12 @@ namespace DigitalLearningSolutions.Data.Services
             var currentYear = date.Year;
             var monthsToDate = Enumerable.Range(1, date.Month);
 
-            var activity = activityDataService.GetActivityForMonthsInYear(currentYear, monthsToDate);
+            var activity = activityDataService.GetActivityForMonthsInYear(centreId, currentYear, monthsToDate);
 
             if (date.Month < 12)
             {
                 var monthsLastYear = Enumerable.Range(date.Month + 1, 12 - date.Month);
-                var lastYearActivity = activityDataService.GetActivityForMonthsInYear(currentYear - 1, monthsLastYear);
+                var lastYearActivity = activityDataService.GetActivityForMonthsInYear(centreId, currentYear - 1, monthsLastYear);
                 activity = activity.Concat(lastYearActivity);
             }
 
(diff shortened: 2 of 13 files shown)


DigitalLearningSolutions.Data/Services/ActivityService.cs
```diff
@@ -0,0 +1,43 @@
...
+        private readonly IActivityDataService activityDataService;
+        private readonly IClockService clockService;
+
+        public ActivityService(IActivityDataService activityDataService, IClockService clockService)
+        {
+            this.activityDataService = activityDataService;
+            this.clockService = clockService;
+        }
+
+        public IEnumerable<MonthOfActivity> GetRecentActivity(int centreId)
```
AlexJacksonDS, line 25: Given in the future we will be passing in some parameter from the filters, would it be worth setting this up so it can work with that and doing any default date logic that will be replaced where this is called?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Centre/Reports/ReportsController.cs
```diff
@@ -9,9 +11,23 @@
...
+        }
+
         public IActionResult Index()
         {
-            var model = new ReportsViewModel();
+            var centreId = User.GetCentreId();
+            var monthsOfActivity = activityService.GetRecentActivity(centreId);
+            var model = new ReportsViewModel
+            {
+                ActivityTableViewModel = new ActivityTableViewModel
```
AlexJacksonDS, line 26: Can we do this in the constructor instead?

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/Centre/Reports/ReportsViewModel.cs
```diff
@@ -1,6 +1,34 @@
...
+    }
+
+    public class ActivityTableViewModel
+    {
+        public IEnumerable<ActivityTableRow> Rows { get; set; }
+    }
+
+    public class ActivityTableRow
+    {
+        private static string[] MonthNames = { "Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"};
```
AlexJacksonDS, line 19: I don't think this is used anywhere

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Reports/_ActivityTable.cshtml
```diff
@@ -0,0 +1,40 @@
...
+        <th role="columnheader" class="" scope="col">
+          Registrations
+        </th>
+      </tr>
+    </thead>
+    <tbody class="nhsuk-table__body">
+    @foreach (var activityMonth in Model.Rows) {
+      <tr role="row" class="nhsuk-table__row">
+        <td role="cell" class="nhsuk-table__cell">
+          <span class="nhsuk-table-responsive__heading">Month </span>@activityMonth.Period
```
AlexJacksonDS, line 26: Should this be Period rather than Month to match the header

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Reports/Index.cshtml
```diff
@@ -24,7 +24,8 @@
   </div>
 
   <div class="nhsuk-grid-column-three-quarters">
-    <h1 id="page-heading" class="nhsuk-heading-xl">Centre reports</h1>
+    <h1 id="page-heading" class="nhsuk-heading-xl">Reports</h1>
```
AlexJacksonDS, line 27: This was updated to be 'Centre reports' as part of HEEDLS-450

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Reports/_ActivityTable.cshtml
```diff
@@ -0,0 +1,40 @@
...
+        </td>
+        <td role="cell" class="nhsuk-table__cell">
+          <span class="nhsuk-table-responsive__heading">Completions </span>@activityMonth.Completions
+        </td>
+        <td role="cell" class="nhsuk-table__cell">
+          <span class="nhsuk-table-responsive__heading">Evaluations </span>@activityMonth.Evaluations
+        </td>
+        <td role="cell" class="nhsuk-table__cell">
+          <span class="nhsuk-table-responsive__heading">Evaluations </span>@activityMonth.Registrations
```
AlexJacksonDS, line 35: And should this be Registrations?

AlexJacksonDS (changes requested):
A few things to change and a few questions

DigitalLearningSolutions.Data/Services/ActivityService.cs
JBsoftwire: I think the processing of the filter parameters is likely to be done in the service, so for the time being I'm inclined to leave it like this until the parameters are set up in the relevant ticket. It may not even use a DateTime object in the eventual implementation.

AlexJacksonDS: Fair enough, especially if it is not yet clear what the parameters will be.

AlexJacksonDS (approved):
Looks good now

DigitalLearningSolutions.Data/Services/ActivityService.cs
```diff
@@ -0,0 +1,64 @@
...
+                }
+            );
+            var monthsLastYear = Enumerable.Range(startTime.Month, 13 - startTime.Month).Select(
+                m => new MonthOfActivity
+                {
+                    Year = startTime.Year,
+                    Month = m
+                }
+            );
+
```
stellake, line 45: I think we should try and make this more generic and support creating monthly slots between any startTime & endTime. This could live in a separate helper, which we can use to grab slots that we can then map to the Usage stats data we need.

Maybe something like getting months between start date & end date and then creating an Enumerable.Range(startTime.Month, monthsBetweenDates) that we can then handle. (I'm assuming the default DateTime library has no ways of doing this). Or alternatively, we could keep adding 1 month until date > endDate?

DigitalLearningSolutions.Data/Services/ActivityService.cs
```diff
@@ -0,0 +1,64 @@
...
+                    month.Completions = monthData.Completions;
+                    month.Registrations = monthData.Registrations;
+                    month.Evaluations = monthData.Evaluations;
+                }
+            }
+
+            return months;
+        }
+
+        private IEnumerable<MonthOfActivity> GenerateBlankMonthsBetweenDates(DateTime startDate, DateTime endDate)
```
stellake, line 50: I think this is great, but I'd probably return a `IEnumerable<(int year, int month)>` from this method and pull it into a helper, where we can write very specific unit tests for this and reuse it in the future.

We could also do something like this above (a simple map to the object we need):
```
var monthSlots = DateTimeHelper.GenerateMonthSlots(..);
return monthSlots.Select(slot => {
  var monthlyUsageStats = ...
  return new MonthOfActivity(slot, monthlyUsageStats ) // Need to add a constructor
});
```

DigitalLearningSolutions.Data/Helpers/DateHelper.cs
```diff
@@ -0,0 +1,19 @@
...
+
+    public static class DateHelper
+    {
+        public static IEnumerable<(int Month, int Year)> GetMonthsAndYearsBetweenDates(DateTime startDate, DateTime endDate)
+        {
+            var diffInMonths = (endDate.Year - startDate.Year) * 12 + (endDate.Month - startDate.Month);
+            var monthEnumerable = Enumerable.Range(startDate.Month, diffInMonths + 1);
+
+            return monthEnumerable.Select(
+                m => ((m - 1) % 12 + 1, startDate.Year + (m - 1) / 12)
```
stellake, line 15: Minor, but it might be nice to pull these into var month and var yearsToAdd, e.g.
```
m => {
  var month = (m - 1) % 12 + 1;
  var yearsToAdd = (m - 1) / 12;
  return (month, startDate.Year + yearsToAdd);
}
```

stellake (approved):
Looks good - added a minor comment about readability but no need for a rereview 👍

PR #446, opened 2021-06-23
HEEDLS-415 Fix path so it works with site hosted in subdirectory

AlexJacksonDS:
Added back a method to figure out an endpoint from the specified endpoint so that it works with the site hosted in a subdirectory.


DanBloxham-sw (approved):
Looks fine to me.

stellake (approved):
Looks good! 👍

PR #452, opened 2021-06-25
HEEDLS-523 Changed h3 to h2 to not have out of order headings

davidm-m:
Google Lighthouse was complaining about having an h3 directly beneath an h1, so I've changed it to an h2. Hasn't had any effect on styling since that was all being set by the `nhsuk-warning-callout__label` class anyway.

Appearance before:
image

Appearance after:
image


DanBloxham-sw (approved):

stellake (approved):
Looks good! 👍

PR #456, opened 2021-06-30
HEEDLS-527 Admin Registration Backend

ibrahimmunir14:
Saving admin details and notification preferences in the database.


DigitalLearningSolutions.Data/DataServices/RegistrationDataService.cs
```diff
@@ -50,5 +51,38 @@ public string RegisterDelegate(DelegateRegistrationModel delegateRegistrationMod
...
+                values
+            );
+            connection.Execute(
+                @"INSERT INTO NotificationUsers (NotificationId, AdminUserId)
+                SELECT N.NotificationId, @adminUserId
+                FROM Notifications N INNER JOIN NotificationRoles NR 
+                ON N.NotificationID = NR.NotificationID 
+                WHERE RoleID IN (1,2) AND AutoOptIn = 1",
+                new { adminUserId }
+            );
```
AlexJacksonDS, line 83: This should go in a transaction

DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -15,6 +17,12 @@ public interface IRegistrationService
             string baseUrl,
             string userIp
         );
+
+        string RegisterAdminDelegate(
+            RegistrationModel registrationModel
+        );
```
AlexJacksonDS, line 23: I'm a little surprised this is formatted like this when there are other much longer lines in the file.

DigitalLearningSolutions.Data/DataServices/RegistrationDataService.cs
AlexJacksonDS: Ah never mind, the calling code has the transaction. In that case, these should go in separate methods (maybe even in separate data services if there is an appropriate place to put the second query?)

DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -60,17 +71,78 @@ string userIp
...
+            {
+                var candidateNumber = RegisterAdminDelegate(registrationModel);
+                if (candidateNumber == "-1")
+                {
+                    throw new Exception("Delegate account could not be created");
+                }
+
+                registrationDataService.RegisterCentreManagerAdmin(registrationModel);
+
+                centresDataService.SetCentreAutoRegistered(registrationModel.Centre);
```
AlexJacksonDS, line 132: I know that is what the ticket says, but does this not affect any subsequent delegate registrations? Seems a bit weird to change properties of the centre when registering an admin.

DigitalLearningSolutions.Web/Controllers/Register/RegisterAdminController.cs
```diff
@@ -161,7 +164,18 @@ public IActionResult Summary(SummaryViewModel model)
...
-            // TODO: (HEEDLS-527) register admin details and notification preferences in database
+            if (!data.Centre.HasValue || data.Email == null || !IsRegisterAdminAllowed(data.Centre.Value) ||
+                !CheckEmailMatchesCentre(data.Email, data.Centre.Value) || !CheckEmailUnique(data.Email))
+            {
+                return new StatusCodeResult(500);
+            }
+            
+            var registrationModel = RegistrationMappingHelper.MapToRegistrationModel(data);
+            var success = registrationService.RegisterCentreManager(registrationModel);
+            if (!success)
```
AlexJacksonDS, line 175: This doesn't really need to be put into a variable

DigitalLearningSolutions.Data/Services/RegistrationService.cs
AlexJacksonDS: Also should this be public if the only place it seems to be called is another method in this service?

DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -15,6 +17,12 @@ public interface IRegistrationService
             string baseUrl,
             string userIp
         );
+
+        string RegisterAdminDelegate(
+            RegistrationModel registrationModel
+        );
+
+        bool RegisterCentreManager(RegistrationModel registrationModel);
```
AlexJacksonDS, line 25: I'd probably call this TryRegisterCentreManager since it can fail and return false. (Like a TryParse does)

DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -60,17 +71,78 @@ string userIp
...
+        )
+        {
+            using var transaction = new TransactionScope();
+            try
+            {
+                var candidateNumber = RegisterAdminDelegate(registrationModel);
+                if (candidateNumber == "-1")
+                {
+                    throw new Exception("Delegate account could not be created");
+                }
```
AlexJacksonDS, line 128: Why do this here and not in the RegisterAdminDelegate method?

AlexJacksonDS: And what about the case -4?

DigitalLearningSolutions.Web/Controllers/Register/RegisterAdminController.cs
```diff
@@ -161,7 +164,18 @@ public IActionResult Summary(SummaryViewModel model)
                 return View(viewModel);
             }
 
-            // TODO: (HEEDLS-527) register admin details and notification preferences in database
+            if (!data.Centre.HasValue || data.Email == null || !IsRegisterAdminAllowed(data.Centre.Value) ||
+                !CheckEmailMatchesCentre(data.Email, data.Centre.Value) || !CheckEmailUnique(data.Email))
```
AlexJacksonDS, line 168: Might be worth pulling all these into a method

DigitalLearningSolutions.Web/Controllers/Register/RegisterAdminController.cs
```diff
@@ -201,26 +215,35 @@ private void SetAdminRegistrationData(int centreId)
...
+            var adminUser = userDataService.GetAdminUserByEmailAddress(email);
+            return adminUser == null;
+        }
+
+        private bool CheckEmailMatchesCentre(string email, int centreId)
+        {
+            var autoRegisterManagerEmail =
+                centresDataService.GetCentreAutoRegisterValues(centreId).autoRegisterManagerEmail;
+            return email.Equals(autoRegisterManagerEmail);
+        }
```
AlexJacksonDS, line 229: Minor naming thing. I'd call these something that signifies they return bool a bit more clearly e.g. IsEmailUnique & DoesEmailMatchCentre

DigitalLearningSolutions.Web/Helpers/RegistrationMappingHelper.cs
```diff
@@ -5,6 +5,18 @@
...
     public static class RegistrationMappingHelper
     {
+        public static RegistrationModel MapToRegistrationModel(RegistrationData data)
+        {
+            return new RegistrationModel(
+                data.FirstName!,
+                data.LastName!,
+                data.Email!,
+                (int)data.Centre!,
+                (int)data.JobGroup!,
```
AlexJacksonDS, line 15: Is there a reason not to do Centre!.Value ?

DigitalLearningSolutions.Data.Tests/TestHelpers/UserTestHelper.cs
```diff
@@ -156,5 +177,25 @@ FROM Candidates
 
             return users.Single();
         }
+
+        public static async Task<AdminUser> GetAdminUserByEmailAddressAsync(
```
AlexJacksonDS, line 181: We don't need this method.

We should always prefer to use the existing method in the data service that serves the same purpose rather than defining new queries for tests since they should already have unit tests so we know they work correctly. The above delegate users one has no equivalent data service method.

AlexJacksonDS (changes requested):
Some naming changes and refactoring

DigitalLearningSolutions.Data/DataServices/RegistrationDataService.cs
ibrahimmunir14: Initially I'd put them in separate data services but Stella said adding the default notification preferences should always happen when adding an admin to the database, so they can be together (like how the stored procedure for registering delegate sets default notification preferences).

AlexJacksonDS: I'm not keen on the fact that it wasn't clear that there was a transaction going on here with 2 separate executions (as seen from my comments). But if @stellake thinks this is OK that's fine I guess.

DigitalLearningSolutions.Data/Services/RegistrationService.cs
ibrahimmunir14: We need to set AutoRegistered to true to indicate that the centre manager has registered. I don't quite follow how this is unexpected or affects delegate registrations?

AlexJacksonDS: Huh I guess it is just a misleading column name then. Really it means HasManagerRegistered. From the naming I assumed it meant something to do with allowing automatic future registrations.

DigitalLearningSolutions.Data/DataServices/UserDataService.cs
```diff
@@ -235,7 +235,10 @@ public List<DelegateUser> GetDelegateUsersByUsername(string username)
                         Surname AS LastName,
                         Email AS EmailAddress,
                         Password,
-                        ResetPasswordID
+                        ResetPasswordID,
+                        CentreID,
+                        IsCentreManager,
+                        CentreAdmin AS IsCentreAdmin
                     FROM AdminUsers
                     WHERE (Email = @emailAddress)",
```
AlexJacksonDS, line 243: This is likely the cause of one of the merge conflicts. I updated this query to return a lot more values (including the ones you've added it it here).

The idea is that going forwards, we should fully populate all non-nullable properties of the object we are querying into to ease unit testing (we can just check equivalence to defined default objects from the test helpers) and it prevents unexpected errors in the future when someone else uses the method expecting non-nullable properties not to not have null values. If we really don't want to fully populate the object we should define a new object that only has the properties we want.

AlexJacksonDS (approved):
Looks good now. Just some merge conflicts to resolve.

DigitalLearningSolutions.Data/DataServices/RegistrationDataService.cs
stellake: I think these should be in one method because these should always be done hand-in-hand. I do agree with Alex that the 2 separate executions without transaction wrapping can be confusing - we should be able to nest transactions as per: https://docs.microsoft.com/en-us/previous-versions/dotnet/articles/ms973865(v=msdn.10)?redirectedfrom=MSDN#working-with-systemtransactions

stellake (commented):
Looking good - just a few comments/questions! 👍

DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -60,17 +64,66 @@ string userIp
...
+                registrationDataService.RegisterCentreManagerAdmin(registrationModel);
+
+                centresDataService.SetCentreAutoRegistered(registrationModel.Centre);
+
+                transaction.Complete();
+            }
+            catch (Exception e)
+            {
+                logger.LogWarning($"Centre Manager registration failed for the following reason: {e.Message}");
+                throw;
```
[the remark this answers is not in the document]
ibrahimmunir14: I don't see why?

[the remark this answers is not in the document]
stellake: All the errors that we don't catch are logged in V2LogEvents by default. If we logWarning & then throw, we should be able to see a warning message in the logs, but also a stack trace of the error that was thrown as a separate entry. That stack trace will include the error message and the method where it was thrown, so LogWarning in this case shouldn't be needed as all this information is already available in the default log.

stellake (approved):
Looks good! 👍

PR #458, opened 2021-06-30
HEEDLS-429 Notification preferences endpoints case insensitivity

ibrahimmunir14:
- Add flag to ignore case in TryGetFromIdOrName method in Enumeration
- Ignore case when matching with enumerations in model binder


AlexJacksonDS (approved):
Looks OK, I assume you checked it works for the example that I gave that failed testing?

ibrahimmunir14:
> Looks OK, I assume you checked it works for the example that I gave that failed testing?

Yep.

stellake (approved):
Looks good! 👍

PR #459, opened 2021-06-30
Heedls 415 admin page pagination

AlexJacksonDS:
Round 2 with actually correct filter text. Should have checked more closely that they were actually working.


DanBloxham-sw (commented):
This is looking good, but I think the diagnostic one will need to be changed from "HasDiagnostic,DiagnosticsScore" to "HasDiagnostic,DiagnosticScore" (making "Diagnostics" singular before the score)

AlexJacksonDS:
Yep, good spot. Missed the extra s

DanBloxham-sw (approved):

stellake:
Just double checking if we've made sure all refactored properties work as expected?

DanBloxham-sw:
When I tested previously, each one in that list worked except for the three being changed.

AlexJacksonDS:
And I've checked that these ones I've just changed work locally

stellake (approved):
Given that all the refactored search properties are tested, happy for merge! 👍

PR #460, opened 2021-06-30
HEEDLS-465 - Made Support page redirect to home if permissions don't …

DanBloxham-sw:
…match application

Ran all unit tests.

Also changed the spacing a little. It seemed to work locally for me before, but didn't appear on the test site for some reason, so I've changed the css slightly.

Here is a screen shot of the new spacing:
image


AlexJacksonDS (approved):
Looks fine

stellake (approved):
Looks good! 👍

PR #461, opened 2021-06-30
HEEDLS-416 Add details to admin cards

AlexJacksonDS:
Populated the admin cards with more details. I found an extra dependency that will partially rely on whether the filtering work changes at all, marked this with a TODO. The structure of the html should hopefully be OK for the JS filtering work, but it might change slightly.

Checked the screen reader reads these as expected and ran all unit tests. Note that the pagination is still currently invalid html here, so it fails lighthouse. There are no other accessibility errors currently.

image

image

### DIFF
--- a/DigitalLearningSolutions.Web/Helpers/FilterableTagHelper.cs
+++ b/DigitalLearningSolutions.Web/Helpers/FilterableTagHelper.cs
@@ -1,0 +1,51 @@
+﻿namespace DigitalLearningSolutions.Web.Helpers
+{
+    using System.Collections.Generic;
+    using DigitalLearningSolutions.Data.Models.User;
+
+    public static class FilterableTagHelper
+    {
+        public static IEnumerable<(string, string)> GetCurrentTagsForAdminUser(AdminUser adminUser)
+        {
+            var tags = new List<(string, string)>();
+
+            // TODO: Either change these when HEEDLS-532 is merged first or change them on HEEDLS-532 if HEEDLS-416 is merged first to the AdminFilterOptions
+            if (adminUser.IsLocked)
+            {
+                tags.Add(("Locked", nameof(AdminUser.IsLocked) + "|true"));
+            }
+
+            if (adminUser.IsCentreAdmin)
+            {
+                tags.Add(("Centre administrator", nameof(AdminUser.IsCentreAdmin) + "|true"));
+            }
+
+            if (adminUser.IsSupervisor)
+            {
+                tags.Add(("Supervisor", nameof(AdminUser.IsSupervisor) + "|true"));
+            }
+
+            if (adminUser.IsTrainer)
+            {
+                tags.Add(("Trainer", nameof(AdminUser.IsTrainer) + "|true"));
+            }
+
+            if (adminUser.IsContentCreator)
+            {
+                tags.Add(("Content Creator license", nameof(AdminUser.IsContentCreator) + "|true"));
+            }
+
+            if (adminUser.IsCmsAdministrator)
+            {
+                tags.Add(("CMS administrator", nameof(AdminUser.IsCmsAdministrator) + "|true"));
+            }
+
+            if (adminUser.IsContentManager)
+            {
+                tags.Add(("CMS manager", nameof(AdminUser.IsContentManager) + "|true"));
+            }
+
+            return tags;
+        }
+    }
+}
--- a/DigitalLearningSolutions.Web/Views/LearningPortal/Completed/Completed.cshtml
+++ b/DigitalLearningSolutions.Web/Views/LearningPortal/Completed/Completed.cshtml
@@ -5,6 +5,7 @@
 
 <link rel="stylesheet" href="@Url.Content("~/css/shared/searchableElements.css")" asp-append-version="true">
 <link rel="stylesheet" href="@Url.Content("~/css/learningPortal/completed.css")" asp-append-version="true">
+<link rel="stylesheet" href="@Url.Content("~/css/learningPortal/courses.css")" asp-append-version="true">
 
 @{
     ViewData["Application"] = "Learning Portal";
--- a/DigitalLearningSolutions.Web/Views/LearningPortal/Completed/Completed.cshtml
+++ b/DigitalLearningSolutions.Web/Views/LearningPortal/Completed/Completed.cshtml
@@ -5,7 +5,6 @@
 
 <link rel="stylesheet" href="@Url.Content("~/css/shared/searchableElements.css")" asp-append-version="true">
 <link rel="stylesheet" href="@Url.Content("~/css/learningPortal/completed.css")" asp-append-version="true">
-<link rel="stylesheet" href="@Url.Content("~/css/learningPortal/courses.css")" asp-append-version="true">
 
 @{
     ViewData["Application"] = "Learning Portal";
--- a/DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/_SearchableAdminCard.cshtml
+++ b/DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/_SearchableAdminCard.cshtml
@@ -59,7 +59,7 @@
           </a>
         }
         <a class="nhsuk-button delete-button nhsuk-button--secondary nhsuk-u-margin-bottom-1" role="button">
-          Inactive admin account
+          Deactivate admin account
         </a>
       </div>
     </div>
--- a/DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/_SearchableAdminCard.cshtml
+++ b/DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/_SearchableAdminCard.cshtml
@@ -4,12 +4,64 @@
 <div class="searchable-element nhsuk-panel" id="@Model.Id-card">
   <details class="nhsuk-details nhsuk-expander nhsuk-u-margin-bottom-0">
     <summary class="nhsuk-details__summary nhsuk-u-padding-top-0 nhsuk-u-padding-left-0">
-      <span class="nhsuk-details__summary-text searchable-element-title" id="@Model.Id-name" name="name" role="heading" aria-level="2">
+      <span class="nhsuk-details__summary-text searchable-element-title" id="@Model.Id-name" name="name">
         @Model.Name
       </span>
     </summary>
     <div class="nhsuk-details__text">
-
+      <div class="nhsuk-grid-row">
+        <div class="tags">
+          @if (!Model.IsLocked) {
+            <div class="tag nhsuk-u-font-size-14" value="@("IsLocked|false")" hidden>
+              Not locked
+            </div>
+          }
+          @foreach (var (displayText, filter) in Model.Tags) {
+            <div class="tag nhsuk-u-font-size-14" value="@filter">
+              @displayText
+            </div>
+          }
+        </div>
+      </div>
+      <div class="nhsuk-grid-row">
+        <dl class="nhsuk-summary-list">
+          <div class="nhsuk-summary-list__row">
+            <dt class="nhsuk-summary-list__key">
+              Email:
+            </dt>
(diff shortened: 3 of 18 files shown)


DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/_SearchableAdminCard.cshtml
```diff
@@ -4,12 +4,64 @@
...
+          <a class="nhsuk-button left-button-mobile-margin-bottom nhsuk-u-margin-right-2 nhsuk-u-margin-bottom-1" role="button">
+            Unlock account
+          </a>
+        } else {
+          <a class="nhsuk-button left-button-mobile-margin-bottom nhsuk-u-margin-right-2 nhsuk-u-margin-bottom-1" role="button">
+            Manage Roles
+          </a>
+        }
+        <a class="nhsuk-button delete-button nhsuk-button--secondary nhsuk-u-margin-bottom-1" role="button">
+          Inactive admin account
```
DanBloxham-sw, line 62: This should be "Inactivate" rather than "Inactive"

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/_SearchableAdminCard.cshtml
```diff
@@ -4,12 +4,64 @@
...
+          @if (!Model.IsLocked) {
+            <div class="tag nhsuk-u-font-size-14" value="@("IsLocked|false")" hidden>
+              Not locked
+            </div>
+          }
+          @foreach (var (displayText, filter) in Model.Tags) {
+            <div class="tag nhsuk-u-font-size-14" value="@filter">
+              @displayText
+            </div>
+          }
```
DanBloxham-sw, line 23: The "Locked" tag is included here in the collection of tags for all roles. The wireframes have the Locked tag separate from the role tags, at the same level as the Name in the expandable. The wireframe Locked tag is also a brown colour. Has it been moved into here on purpose?

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/_SearchableAdminCard.cshtml
```diff
@@ -4,12 +4,64 @@
...
     </summary>
     <div class="nhsuk-details__text">
-
+      <div class="nhsuk-grid-row">
+        <div class="tags">
+          @if (!Model.IsLocked) {
+            <div class="tag nhsuk-u-font-size-14" value="@("IsLocked|false")" hidden>
+              Not locked
+            </div>
+          }
```
DanBloxham-sw, line 18: I presume this is purely for filtering purposes?

DanBloxham-sw (commented):
Few minor comments on the SearchableAdminCard page

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/_SearchableAdminCard.cshtml
AlexJacksonDS: Yes I discussed this with Stella and we updated the ticket description accordingly.

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/_SearchableAdminCard.cshtml
AlexJacksonDS: Yep. I think I will need it for JS filtering. If it turns out I don't I'll remove it then.

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/_SearchableAdminCard.cshtml
AlexJacksonDS: Discussed with Stella since Inactivate felt wrong. Changed to Deactivate since it is the more common English usage for this type of action.

DanBloxham-sw (approved):

stellake:
Re design, I think we should aim for the same button spacing as My account page (both desktop & mobile):
image

I also wonder if there's a way to make the "this account is locked" text more visible..

There are no screenshots of the closed expanders, but could we try and make sure they have the NHS expander styling (I think they were using expander with buttons previously, which for some reason doesn't have the bottom border):
image

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/_SearchableAdminCard.cshtml
```diff
@@ -4,12 +4,64 @@
...
+        @if (Model.IsLocked) {
+          <a class="nhsuk-button left-button-mobile-margin-bottom nhsuk-u-margin-right-2 nhsuk-u-margin-bottom-1" role="button">
+            Unlock account
+          </a>
+        } else {
+          <a class="nhsuk-button left-button-mobile-margin-bottom nhsuk-u-margin-right-2 nhsuk-u-margin-bottom-1" role="button">
+            Manage Roles
+          </a>
+        }
+        <a class="nhsuk-button delete-button nhsuk-button--secondary nhsuk-u-margin-bottom-1" role="button">
```
stellake, line 61: Just out of curiosity - is the secondary button style needed if the button is already delete-button?

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/_SearchableAdminCard.cshtml
```diff
@@ -4,12 +4,64 @@
...
+            <dt class="nhsuk-summary-list__key">
+              Email:
+            </dt>
+            <dd class="nhsuk-summary-list__value">
+              @Model.EmailAddress
+            </dd>
+          </div>
+          <div class="nhsuk-summary-list__row">
+            <dt class="nhsuk-summary-list__key">
+              Course category:
```
stellake, line 38: Could we remove the ":" at the end of the properties (we currently have a mixture, but it looks like the NHS examples don't add ":" at the end of the words)

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/_SearchableAdminCard.cshtml
```diff
@@ -4,12 +4,64 @@
...
-
+      <div class="nhsuk-grid-row">
+        <div class="tags">
+          @if (!Model.IsLocked) {
+            <div class="tag nhsuk-u-font-size-14" value="@("IsLocked|false")" hidden>
+              Not locked
+            </div>
+          }
+          @foreach (var (displayText, filter) in Model.Tags) {
+            <div class="tag nhsuk-u-font-size-14" value="@filter">
```
stellake, line 20: I wonder if we should leave adding these values until we determine that they're needed for JS to keep the MRs more concise and relevant to the ticket and to avoid forgetting to remove the code if the JS version approach ends up being different.. 🤔

stellake: (unless you think it will be more work than worth it as I don't feel particularly strongly)

AlexJacksonDS: I don't see any point removing them now I've already added them until I find they aren't needed. I think they'll be used given the approximate plan I have for the JS filtering

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/_SearchableAdminCard.cshtml
AlexJacksonDS: It was just there on the buttons I copied across the format from (manage registration prompt Edit/Remove buttons) It's not actually needed any more, but I think when I made that page it was because an older version of the delete-button styles changed the padding or something

AlexJacksonDS:
Change the layouts by taking out all the custom styles from the course cards that were messing them up. These are now in a common courses css file.

image

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/_SearchableAdminCard.cshtml
```diff
@@ -2,14 +2,62 @@
...
     <div class="nhsuk-details__text">
 
+      <div class="tags">
+        @if (!Model.IsLocked) {
+          <div class="tag nhsuk-u-font-size-14" value="@("IsLocked|false")" hidden>
+            Not locked
+          </div>
+        }
+        @foreach (var (displayText, filter) in Model.Tags) {
+          <div class="tag nhsuk-u-font-size-14" value="@filter">
```
stellake, line 20: Can you use value on divs? I thought it was only a property of HTML inputs..

https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement
VS
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement

(but I might be missing something)

DigitalLearningSolutions.Web/Helpers/FilterableTagHelper.cs
```diff
@@ -0,0 +1,51 @@
+﻿namespace DigitalLearningSolutions.Web.Helpers
+{
+    using System.Collections.Generic;
+    using DigitalLearningSolutions.Data.Models.User;
+
+    public static class FilterableTagHelper
+    {
+        public static IEnumerable<(string, string)> GetCurrentTagsForAdminUser(AdminUser adminUser)
```
stellake, line 8: Would it be worth writing some simple unit tests for this method?

DigitalLearningSolutions.Web/Views/LearningPortal/Completed/Completed.cshtml
```diff
@@ -5,6 +5,7 @@
 
 <link rel="stylesheet" href="@Url.Content("~/css/shared/searchableElements.css")" asp-append-version="true">
 <link rel="stylesheet" href="@Url.Content("~/css/learningPortal/completed.css")" asp-append-version="true">
+<link rel="stylesheet" href="@Url.Content("~/css/learningPortal/courses.css")" asp-append-version="true">
```
stellake, line 8: Could we not just import the courses.css in the completed.css with `@import "courses.scss"` to have the completed page css built in one stylesheet?

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/_SearchableAdminCard.cshtml
AlexJacksonDS: Technically no but also kind of yes. 
`document.getElementById("id").value;` won't work for this since it isn't a property of a div
But there isn't really anything to stop you putting random attributes on a div and using `document.getElementById("id").getAttribute('value');`

I guess I should probably just name it something else so it isn't confusing. e.g. filter-value

DigitalLearningSolutions.Web/Views/LearningPortal/Completed/Completed.cshtml
AlexJacksonDS: Yes, and we probably could be importing the searchableElements in various places as well. Overall there's probably quite a lot of scss that is poorly set up in the project based on the amount of changes I was able to make when refactoring searchableElements.scss in 532.

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/_SearchableAdminCard.cshtml
AlexJacksonDS: Changed to data-filter-value

AlexJacksonDS:
Grey NHS tags

image

stellake (approved):
Looks good! 👍

PR #462, opened 2021-07-01
HEEDLS-542 Replace break-word with break-all

AlexJacksonDS:
Changed all usages of word-break: break-word to word-break: break-all

IE screens the ticket was raised for:
image

image

### DIFF
--- a/DigitalLearningSolutions.Web/Styles/index.scss
+++ b/DigitalLearningSolutions.Web/Styles/index.scss
@@ -1,5 +1,12 @@
 @import "~nhsuk-frontend/packages/core/all";
 
+@mixin word-break-ie-fix {
+  // IE11 hack
+  @media screen and (-ms-high-contrast: active), (-ms-high-contrast: none) {
+    word-break: break-all;
+  }
+}
+
 ul > li > ul > li {
   list-style-type: circle;
 }
@@ -131,7 +138,8 @@ h1#page-heading {
 
 .nhsuk-summary-list__value {
   @include govuk-media-query($until: tablet) {
-    word-break: break-all;
+    word-break: break-word;
+    @include word-break-ie-fix;
   }
 }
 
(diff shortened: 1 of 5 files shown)


ibrahimmunir14 (approved):

DigitalLearningSolutions.Web/Styles/index.scss
```diff
@@ -131,7 +131,7 @@ h1#page-heading {
 
 .nhsuk-summary-list__value {
   @include govuk-media-query($until: tablet) {
-    word-break: break-word;
+    word-break: break-all;
```
stellake, line 134: Doesn't break-all break the second word instead of putting it on a new line when there's two words in one line?

Example here:
https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

AlexJacksonDS:
I've added the IE only fix using the same pattern we use elsewhere in the css.

Since there was only one instance of each of the other css classes with word-break being used, I've instead added the .word-break css class to the html elements and removed it from the specific css classes so that we don't have to define the IE fix multiple times.

DigitalLearningSolutions.Web/Styles/index.scss
stellake, line 134: Ah - forgot to change here. Also, does the hack work ok on other main browsers?

AlexJacksonDS: Yes, this adds an extra declaration for IE only, other browsers only see break-word. IE gets both and ignores the one it doesn't support.

AlexJacksonDS: There are too many instances of this to add a css class to each of them so I've defined it using a mixin instead

stellake (approved):
Very nice - looks good! 👍

PR #467, opened 2021-07-02
HEEDLS-468 Increase spacing on admin table and add screen reader

AlexJacksonDS:
Increased the spacing between the table and the no activity text on the admin table. Also added aria-current to the highlighted row (which I think is the right attribute?)

image

### DIFF
--- a/DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Ranking/Index.cshtml
+++ b/DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Ranking/Index.cshtml
@@ -35,7 +35,7 @@
     }
 
     @if (Model.CentreHasNoActivity) {
-      <p class="current-centre nhsuk-body-l nhsuk-u-padding-left-2">Your centre overall rank: no activity</p>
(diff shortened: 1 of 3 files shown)


DanBloxham-sw (approved):
Looks good to me. I agree that aria-current should be used here. We've used it on the side nav menus and the home pages, so it should be good.

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Ranking/Index.cshtml
```diff
@@ -35,7 +35,7 @@
     }
 
     @if (Model.CentreHasNoActivity) {
-      <p class="current-centre nhsuk-body-l nhsuk-u-padding-left-2">Your centre overall rank: no activity</p>
+      <p class="current-centre nhsuk-body-l nhsuk-u-padding-left-2" aria-current="true">Your centre overall rank: no activity</p>
```
stellake, line 38: This looks good - do you think we could add a little bit of padding to the "Your centre overall rank: no activity" bit - maybe similar top/bottom padding as the table cells?

AlexJacksonDS:
image

stellake (approved):
Looks great!

PR #469, opened 2021-07-05
HEEDLS-350 Validate password if not null before ModelState validation

AlexJacksonDS:
Moved the validation of the password if it isn't null before the ModelState.IsValid check for both the ChangePassword and MyAccount EditDetails so that the user is shown the invalid current password error even if they also have other validation errors.

### DIFF
--- a/DigitalLearningSolutions.Data/Services/UserService.cs
+++ b/DigitalLearningSolutions.Data/Services/UserService.cs
@@ -18,7 +18,7 @@ namespace DigitalLearningSolutions.Data.Services
 
         public List<CentreUserDetails> GetUserCentres(AdminUser? adminUser, List<DelegateUser> delegateUsers);
 
-        public bool TryUpdateUserAccountDetails(
+        public void UpdateUserAccountDetails(
             AccountDetailsData accountDetailsData,
             CentreAnswersData? centreAnswersData = null
         );
@@ -26,6 +26,8 @@ namespace DigitalLearningSolutions.Data.Services
         public bool NewEmailAddressIsValid(string emailAddress, int? adminUserId, int? delegateUserId, int centreId);
 
         UserAccountSet GetVerifiedLinkedUsersAccounts(int? adminId, int? delegateId, string password);
+
+        public bool IsPasswordValid(int? adminId, int? delegateId, string password);
     }
 
     public class UserService : IUserService
@@ -103,7 +105,7 @@ namespace DigitalLearningSolutions.Data.Services
             return availableCentres.OrderByDescending(ac => ac.IsAdmin).ThenBy(ac => ac.CentreName).ToList();
         }
 
-        public bool TryUpdateUserAccountDetails(
+        public void UpdateUserAccountDetails(
             AccountDetailsData accountDetailsData,
             CentreAnswersData? centreAnswersData = null
         )
@@ -114,11 +116,6 @@ namespace DigitalLearningSolutions.Data.Services
                     accountDetailsData.DelegateId,
                     accountDetailsData.Password
                 );
-
-            if (verifiedAdminUser == null && verifiedDelegateUsers.Count == 0)
-            {
-                return false;
-            }
 
             if (verifiedAdminUser != null)
             {
@@ -156,8 +153,6 @@ namespace DigitalLearningSolutions.Data.Services
                     );
                 }
             }
-
-            return true;
         }
 
         public bool NewEmailAddressIsValid(string emailAddress, int? adminUserId, int? delegateUserId, int centreId)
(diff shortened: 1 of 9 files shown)


DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -201,5 +196,12 @@ string password
 
             return loginService.VerifyUsers(password, adminUser, delegateUsers);
         }
+
+        public bool IsPasswordValid(int? adminId, int? delegateId, string password)
+        {
+            var verifiedLinkedUsersAccounts = GetVerifiedLinkedUsersAccounts(adminId, delegateId, password);
+
+            return verifiedLinkedUsersAccounts.Any();
+        }
```
DanBloxham-sw, line 205: This new method makes sense, but it seems to make a call to GetVerifiedLinkedUsersAccounts, which is also performed in the UpdateUserAccountDetails method (called from MyAccountController) and in the ChangePasswordController's Post Index method, so it feels like these methods could be refactored to reduce potentially redundant calls. 

Off the top of my head, it feels like it will be more complicated than it seems. Do you reckon its worth investigating?

DanBloxham-sw (commented):
Looks all good for now, but I have left a question about refactoring to reduce potentially redundant duplicated calls.

DigitalLearningSolutions.Data/Services/UserService.cs
AlexJacksonDS: There isn't really a sensible way to do this without the duplicate calls. We don't want to be going anywhere near the actual bulk of the logic if there are other validation error so we need a basic password check before the ModelState.IsValid check.

DanBloxham-sw: Fair enough. Was just double checking that it had been considered and dismissed for a good reason.

DanBloxham-sw (approved):

DigitalLearningSolutions.Web.Tests/Controllers/ChangePasswordControllerTests.cs
```diff
@@ -49,10 +49,10 @@ public async Task Post_returns_form_if_model_invalid()
         public async Task Post_returns_form_if_current_password_does_not_match_user_ids()
         {
             // Given
-            GivenPasswordVerificationFails();
+            GivenPasswordIsNotValid();
 
             // When
-            var result = await authenticatedController.Index(new ChangePasswordViewModel());
+            var result = await authenticatedController.Index(new ChangePasswordViewModel{CurrentPassword = "test"});
```
stellake, line 55: Minor, but could we run the formatter?

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -118,11 +123,7 @@ private IActionResult EditDetailsPostSave(EditDetailsViewModel model)
...
             var (accountDetailsData, centreAnswersData) = MapToUpdateAccountData(model, userAdminId, userDelegateId);
 
-            if (!userService.TryUpdateUserAccountDetails(accountDetailsData, centreAnswersData))
-            {
-                ModelState.AddModelError(nameof(EditDetailsViewModel.Password), CommonValidationErrorMessages.IncorrectPassword);
-                return View(model);
-            }
+            userService.UpdateUserAccountDetails(accountDetailsData, centreAnswersData);
 
             return RedirectToAction("Index");
```
stellake, line 128: Is it worth writing tests for the functionality change?

DigitalLearningSolutions.Web/Controllers/ChangePasswordController.cs
```diff
@@ -28,30 +28,24 @@ public IActionResult Index()
...
                     CommonValidationErrorMessages.IncorrectPassword
                 );
+            }
+
+            if (!ModelState.IsValid)
+            {
                 return View(model);
             }
 
+            var verifiedLinkedUsersAccounts =
```
stellake, line 46: Given that we need the verifiedLinkedUserAccounts anyway, could we do something like:
```
var verifiedLinkedUsersAccounts = string.IsNullOrEmpty(model.CurrentPassword)
                ? new UserAccountSet()
                : userService.GetVerifiedLinkedUsersAccounts(adminId, delegateId, model.CurrentPassword!);
if (verifiedLinkedUsersAccounts.Any()) {
// Add errors
}
```
Then we can continue without having to make 2x calls to getting linked accounts. Will probably be more work than worth to avoid double calls on the MyAccountController though.

DigitalLearningSolutions.Web/Controllers/ChangePasswordController.cs
```diff
@@ -30,7 +31,12 @@ public async Task<IActionResult> Index(ChangePasswordViewModel model)
         {
             var adminId = User.GetAdminId();
             var delegateId = User.GetCandidateId();
-            if (model.CurrentPassword != null && !userService.IsPasswordValid(adminId, delegateId, model.CurrentPassword))
+
+            var verifiedLinkedUsersAccounts = string.IsNullOrEmpty(model.CurrentPassword)
+                ? new UserAccountSet(null, null)
```
stellake, line 36: Super minor, but is it worth adding an empty constructor to UserAccountSet?

stellake (approved):
Looks good - a single comment about UserAccountSet (as it's user in tests too), but no need for rereview 👍

PR #470, opened 2021-07-06
HEEDLS-545 All delegates - register delegate - Page 1,2

ibrahimmunir14:
`/TrackingSystem/Delegates/Register/PersonalInformation`
localhost_44363_TrackingSystem_Delegates_Register_PersonalInformation

`/TrackingSystem/Delegates/Register/LearnerInformation`
localhost_44363_TrackingSystem_Delegates_Register_LearnerInformation


DigitalLearningSolutions.Web/Views/RegisterDelegateByCentre/PersonalInformation.cshtml
```diff
@@ -0,0 +1,70 @@
+@using DigitalLearningSolutions.Web.ViewModels.Register
+@model PersonalInformationViewModel
+@{
+  var errorHasOccurred = !ViewData.ModelState.IsValid;
+  ViewData["Title"] = errorHasOccurred ? "Error: Register Delegate" : "Register Delegate";
```
DanBloxham-sw, line 5: As this is part of the Tracking System, I think we'll also want to define on these ViewData variables on these two new pages:
  ViewData["Application"] = "Tracking System";
  ViewData["HeaderPath"] = $"{Configuration["AppRootPath"]}/TrackingSystem/Centre/Dashboard";
  ViewData["HeaderPathName"] = "Tracking System";
like the other Tracking System pages so that the layout can function properly.

DigitalLearningSolutions.Web/Views/RegisterDelegateByCentre/PersonalInformation.cshtml
```diff
@@ -0,0 +1,70 @@
...
+
+@section NavMenuItems {
+  <partial name="~/Views/TrackingSystem/Shared/_NavMenuItems.cshtml" />
+}
+
+<div class="nhsuk-grid-row">
+  <div class="nhsuk-grid-column-full">
+    <form class="nhsuk-u-margin-bottom-8" method="post" asp-action="PersonalInformation">
+      @if (errorHasOccurred) {
+        <vc:error-summary order-of-property-names="@(new[] { nameof(Model.Centre), nameof(Model.FirstName), nameof(Model.LastName), nameof(Model.Email), nameof(Model.Alias) })" />
```
DanBloxham-sw, line 16: As the centre is fixed for these pages, I think the "nameof(Model.Centre)" can be safely removed from this error summary.

DigitalLearningSolutions.Web/Controllers/Register/RegisterDelegateByCentreController.cs
```diff
@@ -0,0 +1,195 @@
...
+            var centreId = User.GetCentreId();
+
+            SetCentreDelegateRegistrationData(centreId);
+
+            return RedirectToAction("PersonalInformation");
+        }
+
+        [ServiceFilter(typeof(RedirectEmptySessionData<DelegateRegistrationByCentreData>))]
+        [HttpGet]
+        [Route("PersonalInformation")]
```
DanBloxham-sw, line 49: A lot of these methods have these Route attributes that just restate the name of the method. Can these be removed safely, or does it mess everything up?

DigitalLearningSolutions.Web/Controllers/Register/RegisterDelegateByCentreController.cs
```diff
@@ -0,0 +1,195 @@
...
+                new CookieOptions
+                {
+                    Expires = DateTimeOffset.UtcNow.AddDays(30)
+                }
+            );
+
+            TempData.Set(centreDelegateRegistrationData);
+        }
+
+        private void ValidateEmailAddress(PersonalInformationViewModel model)
```
DanBloxham-sw, line 148: This method seems to be identical to the one in the RegisterController. Might be worth refactoring it into a helper method that returns true/false, and then the methods that call it can add the model error.

DigitalLearningSolutions.Web/Controllers/Register/RegisterDelegateByCentreController.cs
```diff
@@ -0,0 +1,195 @@
...
+            if (duplicateUsers.Count() != 0)
+            {
+                ModelState.AddModelError(
+                    nameof(PersonalInformationViewModel.Email),
+                    "A user with this email address is already registered at this centre"
+                );
+            }
+        }
+
+        private IEnumerable<EditCustomFieldViewModel> GetEditCustomFieldsFromModel(
```
DanBloxham-sw, line 167: This method can probably also be refactored into a helper method.

DigitalLearningSolutions.Web/Controllers/Register/RegisterDelegateByCentreController.cs
```diff
@@ -0,0 +1,195 @@
...
+                model.Answer1,
+                model.Answer2,
+                model.Answer3,
+                model.Answer4,
+                model.Answer5,
+                model.Answer6
+            );
+        }
+
+        private void PopulateLearnerInformationExtraFields(
```
DanBloxham-sw, line 183: This method can probably also be refactored into a helper method.

DanBloxham-sw (commented):
Looking pretty good, few minor comments about some refactoring.

DigitalLearningSolutions.Web/Controllers/Register/RegisterDelegateByCentreController.cs
ibrahimmunir14: It messes everything up. I think if there's a Route on the class then you need to add Routes to the methods otherwise it won't know which method to use when navigating to the class Route.

ibrahimmunir14: Okay I've cleaned it up a bit by adding `{action}` to the class Route, and only specifying the root Route for the Index action. The other action routes get inferred automatically.

DanBloxham-sw (approved):
Looks good to me.

DigitalLearningSolutions.Web/Controllers/Register/RegisterDelegateByCentreController.cs
```diff
@@ -0,0 +1,192 @@
...
+            return new OkResult();
+        }
+
+        private void SetCentreDelegateRegistrationData(int centreId)
+        {
+            var centreDelegateRegistrationData = new DelegateRegistrationByCentreData
+            {
+                IsCentreSpecificRegistration = true,
+                Centre = centreId
+            };
```
stellake, line 130: Could we add a constructor that takes centreId and sets the IsCentreSpecificRegistration to true?

DigitalLearningSolutions.Web/Views/RegisterDelegateByCentre/PersonalInformation.cshtml
```diff
@@ -0,0 +1,75 @@
...
+                     spell-check="false"
+                     hint-text=""
+                     autocomplete=""
+                     css-class="nhsuk-u-width-one-half" />
+
+      <button class="nhsuk-button" type="submit">Next</button>
+    </form>
+
+    <vc:back-link asp-controller="Dashboard" asp-action="Index" link-text="Cancel" />
+    @*TODO: set cancel link to return to all delegates page*@
```
stellake, line 73: Can do this TODO now 🎉

DigitalLearningSolutions.Web/Controllers/Register/RegisterDelegateByCentreController.cs
```diff
@@ -0,0 +1,192 @@
+namespace DigitalLearningSolutions.Web.Controllers.Register
```
stellake, line 1: Could we some tests for these additions?

stellake (commented):
Nice! A few comments + it looks like the "Cancel" button spacing isn't consistent with the other pages 👍

stellake (approved):
Looks good! 👍

PR #472, opened 2021-07-07
HEEDLS-430 - Implemented basic centre course setup page

DanBloxham-sw:
Ran all unit tests. Tested in Chrome, Edge, Firefox, IE. Tested with mobile/zoom, screen reader.

Desktop view:

Desktop view

Mobile view:


DigitalLearningSolutions.Web/Views/TrackingSystem/CourseSetup/_CentreCourseCard.cshtml
```diff
@@ -0,0 +1,27 @@
...
+
+  <div class="nhsuk-grid-row">
+    <div class="nhsuk-grid-column-full nhsuk-u-margin-left-4">
+      <a class="nhsuk-button nhsuk-u-margin-right-2" role="button"
+         asp-controller="LearningMenu" asp-action="Index" asp-route-customisationId="@Model.CustomisationId">
+        Launch course
+      </a>
+      <a class="nhsuk-button nhsuk-button--secondary " role="button" asp-controller="TopCourses" asp-action="Index">
+        Manage course
+      </a>
```
DanBloxham-sw, line 24: One thing I've noticed when we have two buttons/links next to each other on the same row is that NVDA seems to read them together in one quick burst, and the second button cannot be highlighted/stopped on when just using the arrow keys. Tabbing works fine. I don't know if this is just specific to my set up of NVDA, but there is a setting in Browse Mode that fixes this that I've enabled.

DigitalLearningSolutions.Web/Views/TrackingSystem/CourseSetup/_CentreCourseCard.cshtml
```diff
@@ -0,0 +1,27 @@
...
+    <div class="nhsuk-details__text">
+
+    </div>
+  </details>
+
+  <div class="nhsuk-grid-row">
+    <div class="nhsuk-grid-column-full nhsuk-u-margin-left-4">
+      <a class="nhsuk-button nhsuk-u-margin-right-2" role="button"
+         asp-controller="LearningMenu" asp-action="Index" asp-route-customisationId="@Model.CustomisationId">
+        Launch course
```
DanBloxham-sw, line 20: This button takes the user to the LearningMenu, but if the admin user is not also currently logged in to a delegate account, that page will redirect them to the home page due to a lack of authorisation. If they are logged into a delegate account, the page then redirects them to the old site, which seemed to crash locally (I suspect due to a lack of Session cookie data).

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CourseSetup/CourseSetupController.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    using System.Linq;
+    using DigitalLearningSolutions.Data.Services;
+    using DigitalLearningSolutions.Web.Helpers;
+    using DigitalLearningSolutions.Web.ViewModels.TrackingSystem.CourseSetup;
+    using Microsoft.AspNetCore.Authorization;
+    using Microsoft.AspNetCore.Mvc;
+
+    [Authorize(Policy = CustomPolicies.UserCentreAdmin)]
+    [Route("/TrackingSystem/CourseSetup")]
```
AlexJacksonDS, line 11: This route is different to the one on the ticket?

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/CourseSetup/CourseSetupViewModel.cs
```diff
@@ -0,0 +1,15 @@
...
+    using DigitalLearningSolutions.Data.Models.Courses;
+
+    public class CourseSetupViewModel
+    {
+        public CourseSetupViewModel(IEnumerable<CourseStatistics> courses)
+        {
+            Courses = courses;
+        }
+
+        public IEnumerable<CourseStatistics> Courses { get; set; }
```
AlexJacksonDS, line 13: If this is going to be a page with search/sort/filter in the future (like the wireframes suggest), this is going to need a separate view model rather than using the database entity. Doesn't necessarily need to change now, but it will likely need changing in the future

AlexJacksonDS (commented):
Some comments and questions. Definitely need to get latest master into the branch and sort out the css files used.

DigitalLearningSolutions.Data/Services/CourseService.cs
```diff
@@ -24,5 +25,11 @@ public IEnumerable<CourseStatistics> GetTopCourseStatistics(int centreId, int ca
             var allCourses = courseDataService.GetCourseStatisticsAtCentreForCategoryId(centreId, categoryId);
             return allCourses.Where(c => c.Active).OrderByDescending(c => c.InProgressCount);
         }
+
+        public IEnumerable<CourseStatistics> GetCentreSpecificCourseStatistics(int centreId, int categoryId)
+        {
+            var allCourses = courseDataService.GetCourseStatisticsAtCentreForCategoryId(centreId, categoryId);
+            return allCourses.Where(c => c.CentreId == centreId && c.ArchivedDate == null);
```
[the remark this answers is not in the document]
DanBloxham-sw: Exactly correct. Across the course and dashboard pages, the old site uses a couple of very messy and very different looking queries that effectively end up being identical - the only differences are some of what is put in the WHERE statement. I decided to make the similarities obvious here by making a single, neater query to tackle both cases (the other case being for the Top Courses page for the dashboard), and then just apply that different filtering in the courseService.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CourseSetup/CourseSetupController.cs
DanBloxham-sw: It is. After a little discussion (can be found in the comments of the HLD), the folder setup was preferred to be TrackingSystem/CourseSetup. The original TrackingSystem/CentreCourseSetup is still feasible, but I thought it would be best to keep it consistent with the folders, and avoid confusion with the TrackingSystem/Centre/... pages.

AlexJacksonDS (approved):

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/BasicAccessibilityTests.cs
```diff
@@ -1,57 +1,58 @@
...
+        [InlineData("/TrackingSystem/Centre/Ranking", "Centre ranking")]
+        [InlineData("/TrackingSystem/Centre/ContractDetails", "Contract details")]
+        [InlineData("/TrackingSystem/CentreConfiguration", "Centre configuration")]
+        [InlineData("/TrackingSystem/CentreConfiguration/EditCentreManagerDetails", "Edit centre manager details")]
+        [InlineData("/TrackingSystem/CentreConfiguration/EditCentreWebsiteDetails", "Edit centre content on DLS website")]
+        [InlineData("/TrackingSystem/CentreConfiguration/RegistrationPrompts", "Manage delegate registration prompts")]
+        [InlineData("/TrackingSystem/CentreConfiguration/RegistrationPrompts/1/Remove", "Remove delegate registration prompt")]
+        [InlineData("/TrackingSystem/Centre/Reports", "Centre reports")]
+        [InlineData("/TrackingSystem/Centre/TopCourses", "Top courses")]
+        [InlineData("/TrackingSystem/CourseSetup", "Centre course setup")]
```
DanBloxham-sw, line 40: For clarity, this file is definitely now CRLF, and the above line is the only line new here.

DigitalLearningSolutions.Web/Views/TrackingSystem/CourseSetup/_CentreCourseCard.cshtml
```diff
@@ -0,0 +1,27 @@
...
+      </a>
+      <a class="nhsuk-button nhsuk-button--secondary " role="button" asp-controller="TopCourses" asp-action="Index">
```
stellake, line 22: Minor, but is the space after the class needed?

stellake: Also - per scope this button should do nothing for now (it will take the user to the course specific page later)

DigitalLearningSolutions.Web/Views/TrackingSystem/CourseSetup/_CentreCourseCard.cshtml
stellake: I think it's ok but worth looking out for - can you add a comment on ticket to test this is ok on UAT. :)

DigitalLearningSolutions.Web/Views/TrackingSystem/CourseSetup/_CentreCourseCard.cshtml
```diff
@@ -0,0 +1,27 @@
...
+      </span>
+    </summary>
+    <div class="nhsuk-details__text">
+
+    </div>
+  </details>
+
+  <div class="nhsuk-grid-row">
+    <div class="nhsuk-grid-column-full nhsuk-u-margin-left-4">
+      <a class="nhsuk-button nhsuk-u-margin-right-2" role="button"
```
stellake, line 18: If we need to put some attributes on new line, could we put all of them on new lines for consistency with method params style + ease of reading

DigitalLearningSolutions.Data/Services/CourseService.cs
```diff
@@ -24,5 +25,11 @@ public IEnumerable<CourseStatistics> GetTopCourseStatistics(int centreId, int ca
             var allCourses = courseDataService.GetCourseStatisticsAtCentreForCategoryId(centreId, categoryId);
             return allCourses.Where(c => c.Active).OrderByDescending(c => c.InProgressCount);
         }
+
+        public IEnumerable<CourseStatistics> GetCentreSpecificCourseStatistics(int centreId, int categoryId)
```
stellake, line 29: Do we want any simple unit tests for this service?

DigitalLearningSolutions.Data/Services/CourseService.cs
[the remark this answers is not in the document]
DanBloxham-sw: I've changed things around so the query itself only recovers non-archived courses. This might lead to some changes on the Top Courses page compared to the old site, as it doesn't look like the Top Courses query for the old site did not filter on ArchivedDate.

DigitalLearningSolutions.Data.Tests/Services/CourseServiceTests.cs
```diff
@@ -0,0 +1,95 @@
...
+            A.CallTo(() => courseDataService.GetCourseStatisticsAtCentreForCategoryId(CentreId, AdminCategoryId))
+                .Returns(GetSampleCourses());
+
+            courseService = new CourseService(courseDataService);
+        }
+
+        [Test]
+        public void GetTopCourseStatistics_should_return_correctly_ordered_active_course_statistics()
+        {
+            using (new TransactionScope())
```
stellake, line 33: Do we need to wrap this in Transaction if the data services are mocked?

DigitalLearningSolutions.Data.Tests/Services/CourseServiceTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            courseDataService = A.Fake<ICourseDataService>();
+            A.CallTo(() => courseDataService.GetCourseStatisticsAtCentreForCategoryId(CentreId, AdminCategoryId))
+                .Returns(GetSampleCourses());
+
+            courseService = new CourseService(courseDataService);
+        }
+
+        [Test]
+        public void GetTopCourseStatistics_should_return_correctly_ordered_active_course_statistics()
```
stellake, line 31: Could we specify here: "ordered by In progress field" or something similar to make "Correctly" unambiguous.

stellake (approved):
Looks good! 👍

PR #473, opened 2021-07-07
HEEDLS-550 Add check junk mail text to confirmation pages

AlexJacksonDS:
Added 'Check your Junk folder if the email doesn't arrive.' to the reset password confirmation page and the register confirmation page.
image
image


ibrahimmunir14 (approved):

stellake (approved):
Looks good! 👍

PR #474, opened 2021-07-07
HEEDLS-532 Make filters in the same category OR not AND

AlexJacksonDS:
Changed filters so that filters in the same group work like an OR rather than an AND.

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/Helpers/FilteringHelperTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/Helpers/FilteringHelperTests.cs
@@ -42,11 +42,11 @@
         }
 
         [Test]
-        public void FilterItems_with_filters_in_same_category_returns_union_of_filters()
+        public void FilterItems_with_filters_in_same_group_returns_union_of_filters()
         {
             // Given
             var expectedItems = new[] { ItemA1, ItemA3, ItemB2 }.AsQueryable();
-            var filterBy = "Category|Name|a\r\nCategory|Name|b";
+            var filterBy = "Group|Name|a\r\nGroup|Name|b";
 
             // When
             var result = FilteringHelper.FilterItems(InputItems, filterBy);
@@ -56,11 +56,11 @@
         }
 
         [Test]
-        public void FilterItems_with_filters_in_same_category_returns_no_duplicates()
+        public void FilterItems_with_filters_in_same_group_returns_no_duplicates()
         {
             // Given
             var expectedItems = new[] { ItemA1, ItemA3 }.AsQueryable();
-            var filterBy = "Category|Name|a\r\nCategory|Name|a";
+            var filterBy = "Group|Name|a\r\nGroup|Name|a";
 
             // When
             var result = FilteringHelper.FilterItems(InputItems, filterBy);
@@ -70,11 +70,11 @@
         }
 
         [Test]
-        public void FilterItems_with_some_filters_in_same_category_returns_expected_items()
+        public void FilterItems_with_some_filters_in_same_group_returns_expected_items()
         {
             // Given
             var expectedItems = new[] { ItemA3 }.AsQueryable();
-            var filterBy = "Category|Name|a\r\nCategory|Name|b\r\nNumber|Number|3";
+            var filterBy = "Group|Name|a\r\nGroup|Name|b\r\nNumber|Number|3";
 
             // When
             var result = FilteringHelper.FilterItems(InputItems, filterBy);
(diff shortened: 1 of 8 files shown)


DigitalLearningSolutions.Web/Helpers/FilteringHelper.cs
```diff
@@ -20,70 +20,106 @@ public static class FilteringHelper
...
+        private class AppliedFilter
+        {
+            public AppliedFilter(string filterOption)
+            {
+                var splitFilter = filterOption.Split(Separator);
+                Group = splitFilter[0];
+                PropertyName = splitFilter[1];
+                PropertyValue = splitFilter[2];
+            }
+            public string Group { get; }
```
DanBloxham-sw, line 57: The FilterOptions defined below define this as "Category". Should this be the same for consistency or am I misunderstanding what it represents?

DanBloxham-sw (approved):
Looks good to me. Minor comment about a property name, but its good either way.

DigitalLearningSolutions.Web/Helpers/FilteringHelper.cs
```diff
@@ -20,70 +20,106 @@ public static class FilteringHelper
...
+            string propertyName,
+            string propertyValueString
+        )
+        {
+            var propertyType = typeof(T).GetProperty(propertyName)!.PropertyType;
+            var propertyValue = TypeDescriptor.GetConverter(propertyType).ConvertFromString(propertyValueString);
+            return items.Where(propertyName, propertyValue);
+        }
+
+        private class AppliedFilter
```
stellake, line 48: Looking at this, I think this file will get quite big quite quickly. Could we move this class to models and/or separate out the filteroptions?

DigitalLearningSolutions.Web.Tests/Helpers/FilteringHelperTests.cs
```diff
@@ -32,7 +32,21 @@ public void FilterItems_returns_expected_items_with_multiple_filters()
...
+
+            // When
+            var result = FilteringHelper.FilterItems(InputItems, filterBy);
+
+            // Then
+            result.Should().BeEquivalentTo(expectedItems);
+        }
+
+        [Test]
+        public void FilterItems_with_filters_in_same_category_returns_union_of_filters()
```
stellake, line 45: I think we should have a few extra tests for this functionality:
* Filtering two different categories not taking union
* Filtering twice per category not returning duplicates if one item matches both filters in that category
* Case where we test it all together - e.g. two filters in same category and one in a different category

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/Centre/Administrator/CentreAdministratorsViewModel.cs
```diff
@@ -50,7 +50,11 @@ int page
             Admins = paginatedItems.Select(adminUser => new SearchableAdminViewModel(adminUser));
             IEnumerable<FilterOptionViewModel> categoryOptions =
                 categories.Select(
-                    c => new FilterOptionViewModel(c, $"{nameof(AdminUser.CategoryName)}|{c}", FilterStatus.Default)
+                    c => new FilterOptionViewModel(
+                        c,
+                        $"{nameof(AdminUser.CategoryName)}|{nameof(AdminUser.CategoryName)}|{c}",
```
stellake, line 55: Should we use the "Separator" reference here?

DigitalLearningSolutions.Web.Tests/Helpers/FilteringHelperTests.cs
AlexJacksonDS: Filters in 2 categories already had a test FilterItems_returns_expected_items_with_multiple_filters

DigitalLearningSolutions.Web.Tests/Helpers/FilteringHelperTests.cs
```diff
@@ -32,7 +32,49 @@ public void FilterItems_returns_expected_items_with_multiple_filters()
...
+
+            // When
+            var result = FilteringHelper.FilterItems(InputItems, filterBy);
+
+            // Then
+            result.Should().BeEquivalentTo(expectedItems);
+        }
+
+        [Test]
+        public void FilterItems_with_filters_in_same_category_returns_no_duplicates()
```
stellake, line 59: in the same group?

stellake (approved):
Looks good - just need to modify the test names 👍  No need for rereview though

PR #476, opened 2021-07-08
HEEDLS-465 - Made help and github links open in new tabs

DanBloxham-sw:
Ran all unit tests and tested manually.

### DIFF
--- a/DigitalLearningSolutions.Web/ViewModels/Common/ViewComponents/SideMenuLinkWithHrefViewModel.cs
+++ b/DigitalLearningSolutions.Web/ViewModels/Common/ViewComponents/SideMenuLinkWithHrefViewModel.cs
@@ -8,20 +8,17 @@
         public readonly string? Href;
         public readonly bool IsCurrentPage;
         public readonly string LinkText;
-        public readonly bool OpenInNewTab;
 
         public SideMenuLinkWithHrefViewModel(
             string hRef,
             string linkText,
             bool isCurrentPage,
-            bool openInNewTab,
             Dictionary<string, string>? aspAllRouteData
         )
         {
             Href = hRef;
             LinkText = linkText;
             IsCurrentPage = isCurrentPage;
-            OpenInNewTab = openInNewTab;
             AspAllRouteData = aspAllRouteData;
         }
     }
--- a/DigitalLearningSolutions.Web/ViewModels/Common/ViewComponents/SideMenuLinkWithHrefViewModel.cs
+++ b/DigitalLearningSolutions.Web/ViewModels/Common/ViewComponents/SideMenuLinkWithHrefViewModel.cs
@@ -4,17 +4,19 @@
 
     public class SideMenuLinkWithHrefViewModel
     {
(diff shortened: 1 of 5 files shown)


ibrahimmunir14 (approved):

DigitalLearningSolutions.Web/ViewModels/Common/ViewComponents/SideMenuLinkWithHrefViewModel.cs
```diff
@@ -10,13 +10,16 @@ public class SideMenuLinkWithHrefViewModel
 
         public readonly bool IsCurrentPage;
 
+        public readonly bool OpenInNewTab;
+
         public readonly Dictionary<string, string>? AspAllRouteData;
 
-        public SideMenuLinkWithHrefViewModel(string hRef, string linkText, bool isCurrentPage,  Dictionary<string, string>? aspAllRouteData)
+        public SideMenuLinkWithHrefViewModel(string hRef, string linkText, bool isCurrentPage, bool openInNewTab,  Dictionary<string, string>? aspAllRouteData)
```
ibrahimmunir14, line 17: Maybe run a code cleanup here?

DigitalLearningSolutions.Web/Views/Shared/Components/SideMenuLinkWithHref/Default.cshtml
```diff
@@ -7,6 +7,8 @@
   </li>
 } else {
   <li class="side-nav__item">
-    <a class="side-nav__link" href="@Model.Href" asp-all-route-data="@Model.AspAllRouteData">@Model.LinkText</a>
+    <a class="side-nav__link" href="@Model.Href" asp-all-route-data="@Model.AspAllRouteData" target="@(Model.OpenInNewTab ? "_blank" : null)">
```
stellake, line 10: Ah - we should probably add "noopener" to the true external links (GitHub):
https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noopener

I wonder if it might be worth not reusing this component for these links in that case, so we can keep the component simple by not adding two extra properties just for two extra use cases.. Thoughts? 🤔

DanBloxham-sw: Makes sense. Do we think there won't be many other side menus with links that need to open in new tabs then?

DanBloxham-sw: Also, that link says it is not compatible with IE, I presume we are okay with that?

stellake: Ah - could we also add noreferrer to handle IE (I think that's supported on newer IE versions)? 🙏  
Re side menus and new tabs - we're not expecting to have any other external links. If it ever changes, we can rethink whether to commonise them :)

stellake (approved):
Looks good! 👍

PR #479, opened 2021-07-09
HEEDLS-436 - Added new AdminFields page for CourseSetup and applied c…

DanBloxham-sw:
…ommon styling to similar cards

Ran all unit tests. Tested on Chrome, Edge, IE, Firefox. Tested with Screenreader, mobile/zoom.

I've also changed a few other places where expandables with ever-present buttons exist to make the styling consistent (mainly borders and spacing). I've marked these in the changed files, and provided screen shots of the pages below.

Desktop AdminFields (populated):
Desktop populated admin fields

Mobile AdminFields (populated):

Desktop AdminFields (no fields):
Desktop no admin fields

Approve Delegates before:
image

Approve Delegates after:
Approve Delegates styling change

Centre Course Setup before:
image

Centre Course Setup after:
Centre Course setup basic page styling change


DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateApprovals/_unapprovedDelegateExpandable.cshtml
```diff
@@ -7,12 +7,12 @@
```
DanBloxham-sw, line 7: This file is one of the ones that has had minor styling changes (Approve Delegates)

DigitalLearningSolutions.Web/Views/TrackingSystem/CourseSetup/_CentreCourseCard.cshtml
```diff
@@ -15,14 +15,14 @@
 
   <div class="nhsuk-grid-row">
     <div class="nhsuk-grid-column-full nhsuk-u-margin-left-4">
-      <a class="nhsuk-button nhsuk-u-margin-right-2"
+      <a class="nhsuk-button nhsuk-u-margin-right-2 nhsuk-u-margin-bottom-2"
```
DanBloxham-sw, line 18: This file has also had minor styling changes (Centre Course Setup)

DigitalLearningSolutions.Data.Tests/DataServices/CustomPromptsDataServiceTests.cs
```diff
@@ -106,5 +105,26 @@ public void GetPromptNameForCentreAndPromptNumber_returns_expected_prompt_name()
...
+        {
+            // Given
+            var expectedCourseCustomPromptsResult =
+                CustomPromptsTestHelper.GetDefaultCourseCustomPromptsResult(
+                    1379,
+                    customField1Prompt: null,
+                    customField1Options: "Yes\nNo\nNot sure",
+                    customField1Mandatory: true,
+                    customField2Prompt: null,
+                    customField2Options:"Yes\nNo\nNot sure"
```
AlexJacksonDS, line 120: \r\n? Or is the data different for that setup?

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/BasicAccessibilityTests.cs
```diff
@@ -1,58 +1,59 @@
-namespace DigitalLearningSolutions.Web.AutomatedUiTests.AccessibilityTests
```
AlexJacksonDS, line 1: Line endings again?...

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/CourseSetup/AdminFieldsViewModel.cs
```diff
@@ -0,0 +1,15 @@
...
+    using DigitalLearningSolutions.Data.Models.CustomPrompts;
+
+    public class AdminFieldsViewModel
+    {
+        public AdminFieldsViewModel(CourseCustomPrompts courseCustomPrompts)
+        {
+            AdminFields = courseCustomPrompts.CourseAdminFields;
+        }
+
+        public IEnumerable<CustomPrompt> AdminFields { get; set; }
```
AlexJacksonDS, line 13: If we're setting this up in a similar way to the delegate registration prompts we should probably be using the view models in the same reason unless there is a good reason not to.

DigitalLearningSolutions.Web/Views/TrackingSystem/CourseSetup/AdminFields/AdminFields.cshtml
```diff
@@ -0,0 +1,58 @@
...
+
+<div class="nhsuk-grid-row nhsuk-u-margin-top-7">
+  <div class="nhsuk-grid-column-full">
+    <vc:back-link asp-controller="" asp-action="" link-text="Go back"/>
+  </div>
+</div>
+
+
+
+
```
AlexJacksonDS, line 58: Bunch of extra lines

DigitalLearningSolutions.Web/Views/TrackingSystem/CourseSetup/AdminFields/AdminFields.cshtml
```diff
@@ -0,0 +1,58 @@
...
+      <a class="nhsuk-button" role="button" href="#">Add admin field</a>
+    </div>
+  </div>
+}
+
+<div class="nhsuk-grid-row nhsuk-u-margin-top-7">
+  <div class="nhsuk-grid-column-full">
+    <vc:back-link asp-controller="" asp-action="" link-text="Go back"/>
+  </div>
+</div>
```
AlexJacksonDS, line 54: Can we structure the html here in a similar way to the delegate prompts index page for consistency or is there a good reason not to?

AlexJacksonDS (commented):
A couple of questions/things that I think could probably be changed

DigitalLearningSolutions.Data.Tests/DataServices/CustomPromptsDataServiceTests.cs
DanBloxham-sw: This specific record uses \n rather than \r\n. Seems there's a lot of inconsistency between how existing records separate new line data (I think because this is obtained from a textarea element, and if people enter data on a Mac then it will use different ones to a Windows pc).

AlexJacksonDS: Oh no, that could break a lot of things...

DanBloxham-sw: Would it? I haven't used custom prompts much, but there was (for a certain amount of time before it was refactored) a similar issue with the centre configuration details. The OrganisationsCovered and GeneralInformation fields were originally (and still are) entered in a text area. They have no formatting suggestions whatsoever, sometimes people would use new lines, sometimes they wouldn't, sometimes it was \r\n, sometimes just \n. That didn't seem to cause any issues there.

DanBloxham-sw: Just double checked, think it used the same logic as is used in the CustomPrompt constructor. Looks like it should be good to me as it should split on either, unless I am missing something:        
 ```
private List<string> SplitOptionsString(string? options)
        {
            var optionsList = new List<string>();
            if (options != null)
            {
                optionsList.AddRange(options.Split(new [] { '\r', '\n' },
                    System.StringSplitOptions.RemoveEmptyEntries));
            }

            return optionsList;
        }
```

**Review by AlexJacksonDS — APPROVED** [review]
Looks OK now. Can you just confirm that the BasicAccessibilityTests file is definitely the right line endings now.

**Comment by DanBloxham-sw** [comment]
BasicAccessibilityTests file is now definitely CRLF. The second commit fixed the entire file, except for the final line for some reason, which had to be corrected in the third commit.

File: DigitalLearningSolutions.Web/ViewModels/Common/DisplayPromptsViewModel.cs [hunk 8]
```diff
@@ -0,0 +1,18 @@
...
+    using System.Collections.Generic;
+    using System.Linq;
+    using DigitalLearningSolutions.Data.Models.CustomPrompts;
+
+    public class DisplayPromptsViewModel
+    {
+        public DisplayPromptsViewModel(IEnumerable<CustomPrompt> customPrompts)
+        {
+            CustomFields = customPrompts.Select(cp =>
+                    new CustomPromptManagementViewModel(cp.CustomPromptNumber, cp.CustomPromptText, cp.Mandatory,cp.Options))
```
stellake, line 12: Could we run the formatter? :)

DigitalLearningSolutions.Web/Styles/shared/cardWithButtons.scss
```diff
@@ -4,10 +4,9 @@
   background-color: #fff;
   @extend .nhsuk-u-margin-top-3;
   @extend .nhsuk-u-margin-bottom-4;
-  padding-top: 8px;
   padding-bottom: 16px;
-  padding-left: 32px;
-  padding-right: 32px;
+  border: 1px solid $nhsuk-border-color;
+  border-bottom-width: 4px;
```
stellake, line 9: Might be worth listing on the ticket which cards are affected so the tester can sanity check they're all ok :)

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CourseSetup/AdminFieldsController.cs
```diff
@@ -0,0 +1,37 @@
+﻿namespace DigitalLearningSolutions.Web.Controllers.TrackingSystem.CourseSetup
+{
+    using DigitalLearningSolutions.Data.Services;
+    using DigitalLearningSolutions.Web.Helpers;
+    using DigitalLearningSolutions.Web.ViewModels.TrackingSystem.CourseSetup;
+    using Microsoft.AspNetCore.Authorization;
+    using Microsoft.AspNetCore.Mvc;
+
+    [Authorize(Policy = CustomPolicies.UserCentreAdmin)]
+    [Route("/TrackingSystem/CourseSetup")]
```
stellake, line 10: Need to add the new feature flag, which is now on master :)

DigitalLearningSolutions.Data/DataServices/CustomPromptsDataService.cs
```diff
@@ -132,5 +133,39 @@ public string GetPromptNameForCentreAndPromptNumber(int centreId, int promptNumb
...
+	                    cu.CentreID, 
+	                    cp1.CoursePrompt AS CustomField1Prompt,
+	                    cu.Q1Options AS CustomField1Options, 
+	                    cu.Q1Mandatory AS CustomField1Mandatory, 
+	                    cp2.CoursePrompt AS CustomField2Prompt,
+	                    cu.Q2Options AS CustomField2Options, 
+	                    cu.Q2Mandatory AS CustomField2Mandatory,
+	                    cp3.CoursePrompt AS CustomField3Prompt,
+	                    cu.Q3Options AS CustomField3Options, 
+	                    cu.Q3Mandatory AS CustomField3Mandatory
```
stellake, line 151: Is there a reason these have different indentation than L141?

DigitalLearningSolutions.Data/Services/CustomPromptsService.cs
```diff
@@ -156,6 +157,17 @@ public string GetPromptNameForCentreAndPromptNumber(int centreId, int promptNumb
             return customPromptsDataService.GetPromptNameForCentreAndPromptNumber(centreId, promptNumber);
         }
 
+        public CourseCustomPrompts? GetCustomPromptsForCourse(int customisationId, int centreId, int categoryId)
```
stellake, line 160: Can we run the formatter on this file? 🙏

DigitalLearningSolutions.Data/Services/CustomPromptsService.cs
```diff
@@ -156,6 +157,17 @@ public string GetPromptNameForCentreAndPromptNumber(int centreId, int promptNumb
             return customPromptsDataService.GetPromptNameForCentreAndPromptNumber(centreId, promptNumber);
         }
 
+        public CourseCustomPrompts? GetCustomPromptsForCourse(int customisationId, int centreId, int categoryId)
+        {
+            var result = customPromptsDataService.GetCourseCustomPrompts(customisationId, centreId, categoryId);
+            return result == null ? null :
+                new CourseCustomPrompts(
+                result.CustomisationId,
+                result.CentreId,
```
stellake, line 166: Do we need the customisationID and centre id on the result if these are the params we specify when fetching them?

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/CourseSetup/AdminFieldsViewModel.cs
```diff
@@ -0,0 +1,12 @@
...
+{
+    using DigitalLearningSolutions.Data.Models.CustomPrompts;
+    using DigitalLearningSolutions.Web.ViewModels.Common;
+
+    public class AdminFieldsViewModel : DisplayPromptsViewModel
+    {
+        public AdminFieldsViewModel(CourseCustomPrompts courseCustomPrompts) : base(courseCustomPrompts.CourseAdminFields)
+        {
+        }
+    }
```
stellake, line 11: Do we need this + DisplayRegistrationPromptsViewModel or could we use the DisplayPromptsViewModel directly?

stellake (commented):
Looks good - only a few comments, mainly around formatting. The cards look great, but I would double check that the "Go back" button top spacing is consistent with the rest of the site. :)

DanBloxham-sw:
I've had a look at the other pages on the site that also use the "Go back" view component. The spacing does indeed seem to match. It might just be my eyes, but it looks like the screen shots in the description for this PR have slightly more spacing than when I just checked now (despite not having changed anything). I've added some screen shots below of what it looks like now, is this satisfactory?

Admin Fields no cards

Admin fields with cards

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/CourseSetup/AdminFieldsViewModel.cs
DanBloxham-sw: In theory, no. I decided to make these like this to allow for common properties to be carried over, but while keeping the view models specific to the view. I was expecting that we might get more pages in future that would make use of the DisplayPromptsViewModel, but require additional properties specific to their own models. So I set it up like this as an example of what should be done. Would we prefer to just use the DisplayPromptsViewModel?

stellake: I agree that doing this when/if we have additional properties makes sense, but no reason to create an extra layer of abstraction now that we don't need to specify anything extra. :)

stellake (commented):
Back button spacing looks great.. Not sure why the screenshots looked a bit different :)

stellake (approved):
Looks great! 👍

PR #481, opened 2021-07-12
Heedls 533 admin js filtering

AlexJacksonDS:
Added javascript filtering of admins using the same logic as the no-js version. This is enabled conditionally in the constructor call on the various pages that use the search/sort/paginate.
Also refactored out the common filters stuff into separate partials.

I've sanity checked the other pages that use the javascript search and they seem OK.

image

image


DigitalLearningSolutions.Web/Helpers/FilteringHelper.cs
```diff
@@ -9,13 +9,34 @@
...
         public const char Separator = '|';
+        public const char FilterSeparator = '╡';
+
+        public static string? AddNewFilterToFilterBy(string? filterBy, string? filterValue)
+        {
+            if (filterValue != null)
+            {
+                if (filterBy != null)
+                {
+                    if (!filterBy.Contains(filterValue))
```
DanBloxham-sw, line 20: In what situations do we expect there to be a null filterBy and a non-null filterValue?

DigitalLearningSolutions.Web/Helpers/FilteringHelper.cs
```diff
@@ -9,13 +9,34 @@
...
+                    {
+                        filterBy = filterBy + FilteringHelper.FilterSeparator + filterValue;
```
DanBloxham-sw, line 22: Should this be FilteringHelper.Separator rather than FilteringHelper.FilterSeparator? I think specifying FilteringHelper is redundant as well.

DanBloxham-sw (commented):
Looking pretty good. A few minor questions that might just be me not understanding things entirely.

DigitalLearningSolutions.Web/Helpers/FilteringHelper.cs
AlexJacksonDS: First filter being added

DigitalLearningSolutions.Web/Helpers/FilteringHelper.cs
AlexJacksonDS: Nope, we have a double separated list. FilteringHelper.Separator is the separator for within the filters i.e. Role|IsCentreAdmin|true. FilteringHelper.FilterSeparator is the separator for between each filters i.e. Role|IsCentreAdmin|true╡Role|IsCmsManager|true

DanBloxham-sw: Nevermind on the Separator differences, think I initially misunderstood what it was doing.

AlexJacksonDS: Maybe they need renaming...

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/Centre/Administrator/CentreAdministratorsViewModel.cs
```diff
@@ -31,17 +31,15 @@ public class CentreAdministratorsViewModel : BaseSearchablePageViewModel
...
-        ) : base(searchString, sortBy, sortDirection, page, filterString)
+        ) : base(searchString, page, false, true, filterString: filterString)
         {
             CentreId = centreId;
             var sortedItems = GenericSortingHelper.SortAllItems(
                 adminUsers.AsQueryable(),
-                sortBy,
-                sortDirection
+                DefaultSortByOptions.Name.PropertyName,
+                Ascending
```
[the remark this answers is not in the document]
AlexJacksonDS: I don't think it really matters because of Linq. They were in the same order for the courses etc. pre-admin search ticket. I've just kept the order of calls and stuck in a filter after the search.

DanBloxham-sw (approved):
Looks good to me.

stellake (commented):
I didn't have a chance to scan over everything, but I ran this locally and noticed a few things we should have a look and fix:
•	Flash of filters can be seen on page load with JS
•	First dropdown could be much smaller and we could put the second dropdown with the button next to it on desktop to be more compact with space
•	I can add no filter, which is unexpected (select category, don't choose filter, but press the Add filter button).
•	Course category of “All” seems to not be handled..
•	Could we name “Filters” to “Filter by” to match "Sort by" on other pages

DigitalLearningSolutions.Web/ViewModels/Common/SearchablePage/BaseSearchablePageViewModel.cs
```diff
@@ -18,12 +18,17 @@ public abstract class BaseSearchablePageViewModel
         public readonly string? SearchString;
 
         public int MatchingSearchResults;
+        
+        public readonly bool FilterEnabled;
+        public readonly bool SortEnabled;
```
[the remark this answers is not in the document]
AlexJacksonDS: Sort disabled is what defaults it to Name/Asc for the javascript version - setting up hidden inputs without having to specify them explicitly correctly for pages that don't have the sort inputs

AlexJacksonDS:
> I didn't have a chance to scan over everything, but I ran this locally and noticed a few things we should have a look and fix:
 • Flash of filters can be seen on page load with JS

I'm not really sure what I can do about this, the javascript has to hide them rather than doing it in the cshtml because the no js version has to have them loaded.

> • First dropdown could be much smaller and we could put the second dropdown with the button next to it on desktop to be more compact with space

Again I'm not really sure this is a good idea, I can make the first dropdown a lot smaller, but the second dropdown is just the same dropdowns as we are displaying for the no javascript version, which I think would look strange if we reduced their size with no first dropdown to put them alongside. I think keeping the styling consistent between the no-js and js version is probably a good idea.

> • I can add no filter, which is unexpected (select category, don't choose filter, but press the Add filter button).

It doesn't actually add anything, but will change it prevent it re-hiding the dropdown.

AlexJacksonDS:
@stellake Not sure what to do about the merge conflicts in _Pagination.cshtml, it is layout improvements in a different way to the ones I already did as part of the other ticket, but it has reintroduced the invalid html.

AlexJacksonDS:
For reference it looks like this to me now on master
image

AlexJacksonDS:
Resolved everything except the pagination.

Javascript filters now look like this and the hide is done with display:none;
image

I've kept the no JS filters like this
image

DigitalLearningSolutions.Data/DataServices/UserDataService.cs
```diff
@@ -82,7 +82,7 @@ public UserDataService(IDbConnection connection)
                         au.SummaryReports,
                         au.UserAdmin AS IsUserAdmin,
                         au.CategoryID,
-                        cc.CategoryName,
+                        ISNULL(cc.CategoryName, 'All') AS CategoryName,
```
DanBloxham-sw, line 85: I presume this is because if an AdminUser has the rights for all categories, they have a categoryId of 0, and thus the CategoryName will be NULL as no category will match the Id of 0. 

I'm wondering if it might be better do some something a little more specific, like a CASE that returns the CategoryName if au.CategoryId != 0, and "All" if it is 0. My worry is that while we know CategoryId = 0 definitely returns a NULL CategoryName, it might also be returned in other circumstances, like if some data gets messed up and a non-zero but non-existent categoryid is used. But I'm not sure if its worth it, as I don't think that's a particularly likely scenario.

DigitalLearningSolutions.Web/Helpers/FilteringHelper.cs
```diff
@@ -9,13 +9,29 @@
     public static class FilteringHelper
     {
         public const char Separator = '|';
+        public const char FilterSeparator = '╡';
+
+        public static string? AddNewFilterToFilterBy(string? filterBy, string? newFilterValue)
```
DanBloxham-sw, line 14: filterBy hasn't been renamed to currentFilters

DigitalLearningSolutions.Web/ViewModels/Common/SearchablePage/BaseSearchablePageViewModel.cs
```diff
@@ -33,17 +38,16 @@ public abstract class BaseSearchablePageViewModel
             SearchString = searchString;
             FilterString = filterString;
             Page = page;
+            SortEnabled = sortEnabled;
+            FilterEnabled = filterEnabled;
             Filters = new List<FilterViewModel>();
```
DanBloxham-sw, line 43: Did we want this renamed to FilterBy to match the SortBy, or are we keeping it as Filters to highlight the fact it is a list? Same for the BaseJavascriptFilterableViewModel

DigitalLearningSolutions.Web/Views/Shared/SearchablePage/_AppliedFilterTag.cshtml
```diff
@@ -0,0 +1,6 @@
+﻿@using DigitalLearningSolutions.Web.ViewModels.Common.SearchablePage
+@model AppliedFilterViewModel
+
+<div class="nhsuk-tag nhsuk-tag--blue filter-tag" data-filter-value="@Model.FilterValue">
+  @Model.FilterCategory: @Model.DisplayText<span class="nhsuk-u-visually-hidden">.</span>
```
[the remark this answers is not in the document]
DanBloxham-sw: Only things I can think of involve changing _DisplayText_ to _FilterValueForDisplay_ or _ReadableFilterValue_ or _FormattedFilterValue_, but I don't think they are that much clearer than _DisplayText_

DanBloxham-sw (commented):
Pretty good, I think all of the markups have been tackled. Some parts have not been renamed as suggested, not sure if that is a deliberate decision as I haven't seen any discussion, so I've highlighted the places I spotted. Also left a brief comment on the UserDataService changes that might not need anything doing.

DigitalLearningSolutions.Web/Helpers/FilteringHelper.cs
AlexJacksonDS: The filter string is called filterBy, including in the method name. Either we need to change it to currentFilters absolutely everywhere (and maybe update sortBy to match) or we keep it as filterBy here so I went with the easier option.

DigitalLearningSolutions.Web/ViewModels/Common/SearchablePage/BaseSearchablePageViewModel.cs
AlexJacksonDS: Not sure what you mean here, the corresponding sort property to Filters is SortOptions. FilterString should perhaps be changed to FilterBy though.

DigitalLearningSolutions.Web/Helpers/FilteringHelper.cs
DanBloxham-sw: Fair point. In which case, I agree that filterBy should be retained as the name here.

DanBloxham-sw (approved):
Think that's satisfactory. Will hand it back to Stella.

DigitalLearningSolutions.Web/Scripts/searchSortFilterAndPaginate/filter.ts
```diff
@@ -0,0 +1,253 @@
...
+    propertyValue: splitFilter[2],
+    filterValue: filter,
+  };
+}
+
+function filterElements(
+  searchableElements: ISearchableElement[],
+  appliedFilter: IAppliedFilter,
+): ISearchableElement[] {
+  return _.filter<ISearchableElement>(
```
stellake, line 106: This is minor, but could we just use JS native filter (as the syntax is a bit nicer) & for find on L236

DigitalLearningSolutions.Web/ViewModels/Common/SearchablePage/BaseSearchablePageViewModel.cs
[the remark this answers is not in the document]
stellake: Sorry - I'm not sure what you mean by:
`- setting up hidden inputs without having to specify them explicitly correctly for pages that don't have the sort inputs`

Could we not specify the default values in the constructor of the base VM and therefore avoid doing the defaults of "Name" and "Ascending" in two places (JS and C#)?

DigitalLearningSolutions.Web/Scripts/searchSortFilterAndPaginate/filter.ts
```diff
@@ -0,0 +1,253 @@
...
+
+function hideAppliedFilters(): void {
+  const element = getAppliedFiltersElement();
+  element.hidden = true;
+  element.setAttribute('aria-hidden', 'true');
+}
+
+function clearAppliedFilters() {
+  const appliedFilterContainer = getAppliedFilterContainer();
+  appliedFilterContainer.textContent = '';
```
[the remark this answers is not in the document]
AlexJacksonDS: Makes sure the container is empty as well as hiding the parent so we don't have any old filters hanging around

DigitalLearningSolutions.Web/Scripts/searchSortFilterAndPaginate/filter.ts
```diff
@@ -0,0 +1,253 @@
...
+
+function hideAllFilterDropdowns(): void {
+  const allDropdowns = Array.from(
+    document.getElementsByClassName('filter-dropdown-container'),
+  );
+  allDropdowns.forEach((dropdown) => {
+    const dropdownElement = <HTMLElement>dropdown;
+    dropdownElement.hidden = true;
+    dropdownElement.style.display = 'none';
+    dropdownElement.setAttribute('aria-hidden', 'true');
```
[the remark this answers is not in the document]
AlexJacksonDS: It's recommended that you do include it on elements that are sometimes visible.

DigitalLearningSolutions.Web/Scripts/searchSortFilterAndPaginate/filter.ts
```diff
@@ -0,0 +1,253 @@
...
+  clearAppliedFilters();
+  hideAppliedFilters();
+}
+
+function hideAllFilterDropdowns(): void {
+  const allDropdowns = Array.from(
+    document.getElementsByClassName('filter-dropdown-container'),
+  );
+  allDropdowns.forEach((dropdown) => {
+    const dropdownElement = <HTMLElement>dropdown;
```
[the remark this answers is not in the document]
AlexJacksonDS: I've sorted out the typing stuff, but this line still needs to be present (in a slightly altered form) otherwise the lint fails

DigitalLearningSolutions.Web/Scripts/searchSortFilterAndPaginate/filter.ts
```diff
@@ -0,0 +1,253 @@
...
+  const filterGroups = _.groupBy(appliedFilters, (filter) => filter.group);
+
+  _.forEach(filterGroups, (appliedFiltersInGroup) => {
+    const itemsToFilter = filteredSearchableElements;
+    const setOfFilteredLists = _.map(
+      appliedFiltersInGroup,
+      (af) => filterElements(itemsToFilter, af),
+    );
+    const flattenedElements = _.flatten(setOfFilteredLists);
+    filteredSearchableElements = _.uniq(flattenedElements);
```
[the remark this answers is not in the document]
AlexJacksonDS: Rider and Visual Studio without Resharper both agree it is fine. Resharper seems to have issues with typescript, I get loads of errors that definitely aren't errors due to failed property resolution. Will investigate the issue, but not fix this line for now as it works and isn't an error without Resharper.

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests/AdminUserDataServiceTests.cs
```diff
@@ -0,0 +1,172 @@
...
+
+            // When
+            var returnedAdminUser = userDataService.GetAdminUserById(7);
+
+            // Then
+            returnedAdminUser.Should().BeEquivalentTo(expectedAdminUser);
+        }
+
+        [Test]
+        public void GetAdminUserById_Returns_admin_user_with_category_name_all()
```
AlexJacksonDS, line 27: new unit test

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests/AdminUserDataServiceTests.cs
```diff
@@ -0,0 +1,172 @@
...
+
+            // When
+            var returnedAdminUser = userDataService.GetAdminUserByUsername("Username");
+
+            // Then
+            returnedAdminUser.Should().BeEquivalentTo(expectedAdminUser);
+        }
+
+        [Test]
+        public void GetAdminUserByUsername_Returns_admin_user_category_name_all()
```
AlexJacksonDS, line 53: new unit test

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests/AdminUserDataServiceTests.cs
```diff
@@ -0,0 +1,172 @@
...
+                returnedAdminUser!.FirstName.Should().BeEquivalentTo(expectedAdminUser.FirstName);
+                returnedAdminUser!.LastName.Should().BeEquivalentTo(expectedAdminUser.LastName);
+                returnedAdminUser!.EmailAddress.Should().BeEquivalentTo(expectedAdminUser.EmailAddress);
+                returnedAdminUser!.Password.Should().BeEquivalentTo(expectedAdminUser.Password);
+                returnedAdminUser!.ResetPasswordId.Should().Be(expectedAdminUser.ResetPasswordId);
+            }
+        }
+
+        [Test]
+        public void GetAdminUserByEmailAddress_Returns_admin_user_category_name_all()
```
AlexJacksonDS, line 88: new unit test

DigitalLearningSolutions.Data.Tests/TestHelpers/UserTestHelper.cs
```diff
@@ -100,6 +100,29 @@ public static class UserTestHelper
             };
         }
 
+        public static AdminUser GetDefaultCategoryNameAllAdminUser()
```
AlexJacksonDS, line 103: new helper

AlexJacksonDS (commented):
Marking the new stuff from the UserDataService split commit

ibrahimmunir14 (changes requested):
I think all the previous comments have been addressed.
Literally just some formatting comments below, mainly some unused imports from the UserDataService refactoring.

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests/AdminUserDataServiceTests.cs
```diff
@@ -0,0 +1,172 @@
...
+            using (new AssertionScope())
+            {
+                returnedAdminUser.Should().NotBeNull();
+                returnedAdminUser!.Id.Should().Be(expectedAdminUser.Id);
+                returnedAdminUser!.FirstName.Should().BeEquivalentTo(expectedAdminUser.FirstName);
+                returnedAdminUser!.LastName.Should().BeEquivalentTo(expectedAdminUser.LastName);
+                returnedAdminUser!.EmailAddress.Should().BeEquivalentTo(expectedAdminUser.EmailAddress);
+                returnedAdminUser!.Password.Should().BeEquivalentTo(expectedAdminUser.Password);
+                returnedAdminUser!.ResetPasswordId.Should().Be(expectedAdminUser.ResetPasswordId);
+            }
```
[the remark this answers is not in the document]
AlexJacksonDS: I think it's because the old query here didn't fully populate the AdminUser (and I just copied it). Will update it so that it is correct.

DigitalLearningSolutions.Data.Tests/Services/LoginServiceTests.cs
```diff
@@ -246,4 +247,4 @@ public void
             Assert.AreEqual(associatedAdminUser, returnedAdminUser);
         }
     }
-}
```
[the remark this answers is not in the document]
AlexJacksonDS: Probably a line ending change

ibrahimmunir14 (approved):

PR #487, opened 2021-07-13
HEEDLS-467 Change units and remove decimal place below gigabytes

AlexJacksonDS:
Changed the units on the byte string and removed the decimal place from the lower sizes.

image
image
image

### DIFF
--- a/DigitalLearningSolutions.Web/Helpers/DisplayStringHelper.cs
+++ b/DigitalLearningSolutions.Web/Helpers/DisplayStringHelper.cs
@@ -5,7 +5,7 @@
     public static class DisplayStringHelper
     {
         private const string Divider = " / ";
-        private static readonly string[] Units = { "B", "KiB", "MiB", "GiB", "TiB", "PiB", "EiB" };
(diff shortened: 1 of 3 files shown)


DigitalLearningSolutions.Web/Helpers/DisplayStringHelper.cs
```diff
@@ -30,7 +30,9 @@ private static string GenerateBytesDisplayString(long byteCount)
             }
 
             var place = Convert.ToInt32(Math.Floor(Math.Log(byteCount, 1024)));
-            var number = Math.Round(byteCount / Math.Pow(1024, place), 1);
+            // Do not include decimal place below GB
+            var decimalPlaces = place <= 2 ? 0 : 1;
+            var number = Math.Round(byteCount / Math.Pow(1024, place), decimalPlaces);
```
stellake, line 35: Could we write unit tests for the rounding? 
Also, I'm going to check with HEE whether we should only round only below GB or always (I don't think it's clear from the comment on the ticket)

stellake (approved):
Looks good! 👍

PR #488, opened 2021-07-13
HEEDLS-470 - Corrected course statistics count from query

DanBloxham-sw:
The different Count values in the course statistics were previously not restricted to candidates the specified centre. Now, the candidates should only count towards the Count values when the record's CentreID matches. This should correct some courses with the "AllCentres" flag from appearing to have more delegates taking the course then are at the centre.


JBsoftwire (approved):
LGTM

stellake (approved):
Looks good! 👍

PR #489, opened 2021-07-13
HEEDLS-547 All delegates - register delegate - registration and confirmation

ibrahimmunir14:
**Confirmation page (welcome email was SET)**
localhost_44363_TrackingSystem_Delegates_Register_Confirmation

**Confirmation page (welcome email was NOT SET, password was NOT SET)**
localhost_44363_TrackingSystem_Delegates_Register_Confirmation (1)

**Confirmation page (welcome email was NOT SET, password was SET)**
localhost_44363_TrackingSystem_Delegates_Register_Confirmation (2)

### DIFF
--- a/DigitalLearningSolutions.Data/Models/Register/RegistrationModel.cs
+++ b/DigitalLearningSolutions.Data/Models/Register/RegistrationModel.cs
@@ -8,7 +8,7 @@
             string email,
             int centre,
             int jobGroup,
-            string passwordHash
+            string? passwordHash
         )
         {
             FirstName = firstName;
@@ -30,7 +30,7 @@
 
         public int JobGroup { get; set; }
 
-        public string PasswordHash { get; set; }
+        public string? PasswordHash { get; set; }
 
         public bool Approved { get; set; }
     }
--- a/DigitalLearningSolutions.Data/Services/RegistrationService.cs
+++ b/DigitalLearningSolutions.Data/Services/RegistrationService.cs
@@ -65,7 +65,7 @@ namespace DigitalLearningSolutions.Data.Services
 
             passwordDataService.SetPasswordByCandidateNumber(
                 candidateNumber,
-                delegateRegistrationModel.PasswordHash
+                delegateRegistrationModel.PasswordHash!
             );
             if (!delegateRegistrationModel.Approved)
             {
@@ -123,7 +123,7 @@ namespace DigitalLearningSolutions.Data.Services
                 registrationModel.Email,
                 registrationModel.Centre,
                 registrationModel.JobGroup,
-                registrationModel.PasswordHash
+                registrationModel.PasswordHash!
             ) { Approved = true };
             var candidateNumber =
                 registrationDataService.RegisterDelegate(delegateRegistrationModel);
@@ -136,7 +136,7 @@ namespace DigitalLearningSolutions.Data.Services
 
             passwordDataService.SetPasswordByCandidateNumber(
                 candidateNumber,
-                delegateRegistrationModel.PasswordHash
+                delegateRegistrationModel.PasswordHash!
             );
         }
 
--- a/DigitalLearningSolutions.Data/Services/RegistrationService.cs
+++ b/DigitalLearningSolutions.Data/Services/RegistrationService.cs
@@ -17,6 +17,8 @@ namespace DigitalLearningSolutions.Data.Services
             bool refactoredTrackingSystemEnabled
         );
 
+        string RegisterDelegateByCentre(DelegateRegistrationModel delegateRegistrationModel);
+
         void RegisterCentreManager(RegistrationModel registrationModel);
     }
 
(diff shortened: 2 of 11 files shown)


DigitalLearningSolutions.Web/Controllers/Register/RegisterDelegateByCentreController.cs
```diff
@@ -202,7 +207,41 @@ public IActionResult Summary()
...
+                    return RedirectToAction("Error", "LearningSolutions");
+                case "-4":
+                    return RedirectToAction("Index");
+            }
+
+            TempData.Clear();
+            TempData.Add("delegateNumber", candidateNumber);
+            TempData.Add("emailSent", data.ShouldSendEmail);
+            TempData.Add("passwordSet", data.IsPasswordSet);
+            return RedirectToAction("Confirmation");
```
DanBloxham-sw, line 229: Is there a particular reason we don't just perform whatever is done in the Confirmation action here instead? I'm just thinking it would save having to worry about TempData as much.

DigitalLearningSolutions.Web/Helpers/RegistrationMappingHelper.cs
```diff
@@ -34,5 +34,23 @@ public static DelegateRegistrationModel MapToDelegateRegistrationModel(DelegateR
...
+                data.Centre!.Value,
+                data.JobGroup!.Value,
+                data.IsPasswordSet ? data.PasswordHash! : null,
+                data.Answer1,
+                data.Answer2,
+                data.Answer3,
+                data.Answer4,
+                data.Answer5,
+                data.Answer6
+            ) {AliasId = data.Alias, NotifyDate = data.WelcomeEmailDate};
```
DanBloxham-sw, line 53: Should these be added to the constructor as optional parameters?

DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -81,6 +83,25 @@ bool refactoredTrackingSystemEnabled
             return (candidateNumber, delegateRegistrationModel.Approved);
         }
 
+        public string RegisterDelegateByCentre(DelegateRegistrationModel delegateRegistrationModel)
```
DanBloxham-sw, line 86: Do we want a couple of unit tests to check that the SetPasswordByCandidateNumber method is called when it should be, and not called when it shouldn't, or is that a little too basic?

DigitalLearningSolutions.Data/Models/Register/RegistrationModel.cs
```diff
@@ -30,7 +30,7 @@ string passwordHash
 
         public int JobGroup { get; set; }
 
-        public string PasswordHash { get; set; }
+        public string? PasswordHash { get; set; }
```
DanBloxham-sw, line 33: I think this change has caused a few lines in the RegistrationService to display some IDE warnings complaining about potentially null values. Might be worth going through that and supressing the warnings.

DanBloxham-sw (commented):
Pretty good. Few minor questions.

DigitalLearningSolutions.Web/Views/RegisterDelegateByCentre/Confirmation.cshtml
```diff
@@ -0,0 +1,43 @@
...
+}
+
+<div class="nhsuk-grid-row">
+  <div class="nhsuk-grid-column-full">
+    <h1 id="page-heading" class="nhsuk-heading-xl">
+      Delegate registration complete
+    </h1>
+    @if (Model.EmailSent) {
+      <p class="nhsuk-body-l">
+        A welcome message will be sent to the email address provided, on the date specified, inviting the delegate to set their password and confirm their registration.
```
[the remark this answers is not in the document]
ibrahimmunir14: In this case the email will be sent in the future (on a date specified by the user). I'll rename EmailSent.

DigitalLearningSolutions.Web/Controllers/Register/RegisterDelegateByCentreController.cs
ibrahimmunir14: Oh yeah that would simplify things.

ibrahimmunir14: But then it would have the Summary URL wouldn't it.

DanBloxham-sw: Oh yeah that's true. Seeing as how this method only redirects to the confirmation page or some other pages in the case of an error, you might be able to get away with changing the Route to be Confirmation (might also need to make the Summary page form specify the action name). Although I think we have some other pages that use this TempData -> Redirect methodology, so this might be fine as is. I'll approve the PR anyway, and you can decide what you think would be better.

DanBloxham-sw (approved):

PR #492, opened 2021-07-15
HEEDLS-431 - Populated course setup cards

DanBloxham-sw:
Ran all unit tests. Tested in Chrome, Edge, Firefox, IE. Tested with screen reader, mobile view, and no JS.

Screen shots:
Desktop:
image

Mobile:

Desktop no JS: 
image


DigitalLearningSolutions.Web/Scripts/trackingSystem/centreCourseSetup.ts
```diff
@@ -0,0 +1,54 @@
...
+  const succeeded = copyTextToClipboard(link);
+  let alertMessage: string;
+
+  if (succeeded) {
+    alertMessage = `Copied the text: ${link}`;
+  } else {
+    alertMessage = `Copy not supported or blocked. Try manually selecting and copying: ${link}`;
+  }
+
+  alert(alertMessage);
```
DanBloxham-sw, line 36: Alerts were used on the old site as confirmation that the copying was either completed successfully or was blocked in some form. It seems pretty convenient to me, so I've introduced it here. A quick run through with the screen reader seems okay to me, but I'm happy to change this if we have a preference for an alternative method.

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/CourseSetup/SearchableCourseStatisticsViewModel.cs
```diff
@@ -24,5 +30,14 @@ public SearchableCourseStatisticsViewModel(CourseStatistics courseStatistics)
         public int InProgressCount { get; set; }
         public string CourseName { get; set; }
         public double PassRate { get; set; }
+        public bool HideInLearnerPortal { get; set; }
```
JBsoftwire, line 33: I feel like I'm missing something, but is this used anywhere?

JBsoftwire (commented):
I have one question, which may have an obvious answer. But I'm going to move this across anyway so we can potentially get it merged quickly.

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/CourseSetup/SearchableCourseStatisticsViewModel.cs
DanBloxham-sw: Good spot. I'd added it that thinking it would be used, got it factored into the Tags property, and then forgot to remove it when it ended up being an unnecessary property. There's a few other properties that are unused, so I'll remove them as well.

DigitalLearningSolutions.Web/Scripts/trackingSystem/centreCourseSetup.ts
DanBloxham-sw: It was decided to keep the alert as is at the Show & Tell on 16th July 2021. The current system uses this, and they have received no complaints about it, so it shall do.

DigitalLearningSolutions.Web/Scripts/trackingSystem/centreCourseSetup.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+function copyTextToClipboard(textToCopy: string) : boolean {
+  const hiddenInput = document.body.appendChild(document.createElement('input'));
```
stellake, line 40: Could we just use launch course button instead of creating and removing a new element?

DigitalLearningSolutions.Web/Views/TrackingSystem/CourseSetup/_CentreCourseCard.cshtml
```diff
@@ -8,7 +8,62 @@
...
+      <div class="nhsuk-grid-row">
+        <div class="nhsuk-grid-column-full">
+          <p class="nhsuk-u-margin-bottom-0">Want to share the course?</p>
+          <a href="@Model.GenerateEmailHref">Generate email</a>
+        </div>
+
+      </div>
+      <div class="nhsuk-grid-row">
+        <div class="nhsuk-grid-column-full">
+          <a href="#" class="js-only-inline copy-course-link nhsuk-link--no-visited-state" id="copy-course-@Model.CustomisationId">Copy course link</a>
```
stellake, line 63: Just wanted to double check if you've checked that this is the most appropriate HTML for this functionality (for clarity - I haven't looked into it at all)

DigitalLearningSolutions.Web/Scripts/trackingSystem/centreCourseSetup.ts
```diff
@@ -0,0 +1,54 @@
+const copyCourseLinkClass = 'copy-course-link';
+const copyLinkIdPrefix = 'copy-course-';
+const launchCourseButtonIdPrefix = 'launch-course-';
+
+setupCourseLinkClipboardCopiers();
+
+function setupCourseLinkClipboardCopiers() {
+  const copyCourseLinks = document.getElementsByClassName(copyCourseLinkClass);
+
+  for (let i = 0; i < copyCourseLinks.length; i += 1) {
```
stellake, line 10: Minor, but it would be much nicer to use JS forEach here (might have to do Array.from(..)) first like elsewhere where we deal with the elements.

stellake (commented):
Screenshots look great - just a few comments! ⭐

DigitalLearningSolutions.Web/Views/TrackingSystem/CourseSetup/_CentreCourseCard.cshtml
```diff
@@ -8,7 +8,62 @@
...
+
+      <div class="nhsuk-grid-row">
+        <div class="nhsuk-grid-column-full">
+          <p class="nhsuk-u-margin-bottom-0">Want to share the course?</p>
+          <a href="@Model.GenerateEmailHref">Generate email</a>
+        </div>
+
+      </div>
+      <div class="nhsuk-grid-row">
+        <div class="nhsuk-grid-column-full">
```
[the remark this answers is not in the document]
DanBloxham-sw: This was to keep the Generate Email and Copy Course link on separate lines. Do we have an alternative preferred method for this?

DigitalLearningSolutions.Web/Scripts/trackingSystem/centreCourseSetup.ts
DanBloxham-sw: Unfortunately it seems that for document.ExecCommand("copy") to work correctly, it needs an element that could have the text to copy selectable/highlightable. That's why an input or textarea seems to be used typically.

I did do a bit of further research though, and it looks like navigator.clipboard.writeText works on all browsers except IE, so I've moved this method to be a fallback. Mostly stolen from this https://stackoverflow.com/a/30810322

DigitalLearningSolutions.Web/Scripts/trackingSystem/centreCourseSetup.ts
```diff
@@ -0,0 +1,59 @@
+const copyCourseLinkClass = 'copy-course-button';
+const copyLinkIdPrefix = 'copy-course-';
+const launchCourseButtonIdPrefix = 'launch-course-';
+
+setupCourseLinkClipboardCopiers();
```
stellake, line 5: Minor, but setUp as it's a verb 👁️

DigitalLearningSolutions.Web/Scripts/trackingSystem/centreCourseSetup.ts
```diff
@@ -0,0 +1,59 @@
...
+
+function setupCourseLinkClipboardCopiers() {
+  const copyCourseLinks = Array.from(document.getElementsByClassName(copyCourseLinkClass));
+
+  copyCourseLinks.forEach(
+    (currentLink) => {
+      const linkId = currentLink.id;
+      const customisationId = linkId.slice(copyLinkIdPrefix.length);
+      currentLink.addEventListener('click', () => { copyLaunchCourseLinkToClipboard(customisationId); });
```
stellake, line 14: Could we just do:
`currentLink.addEventListener('click', () => copyLaunchCourseLinkToClipboard(customisationId));`

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/CourseSetup/SearchableCourseStatisticsViewModel.cs
```diff
@@ -1,28 +1,39 @@
...
+
+        public string EmailHref => GenerateEmailHref();
+
+        private string GenerateEmailHref()
+        {
+            var launchCourseLink =
+                $"{ConfigHelper.GetAppConfig()[ConfigHelper.AppRootPathName]}/LearningMenu/{CustomisationId}";
+            var subject = Uri.EscapeDataString($"Digital Learning Solutions Course Link - {CourseName}");
+            var content = Uri.EscapeDataString($"To start your {CourseName} course, go to {launchCourseLink}");
+            return $"mailto:?subject={subject}&body={content}";
```
stellake, line 36: Is there a reason we don't do UrlEncode on the whole url instead here (not sure how it interacts with mailto links though - just interested 🤔)?

DigitalLearningSolutions.Web/Views/TrackingSystem/CourseSetup/_CentreCourseCard.cshtml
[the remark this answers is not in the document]
stellake: Isn't a button/p element on a new line automatically?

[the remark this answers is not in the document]
DanBloxham-sw: Normally, yes, but when the button was present, it had the js-only-inline class which put everything inline. I've changed it to js-only-block, and removed the divs, and it achieves the same effect.

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/CourseSetup/SearchableCourseStatisticsViewModel.cs
DanBloxham-sw: The reason the UrlEncode is done in the first place is because mailto requires a few specific characters to function correctly so it knows when to split things (e.g. "&body=" specifies the email body). If you don't encode your subject/content and it contains one of the reserved characters, it can break things.

For instance, if a course name was something like "Excel & Word Office Training" then the email subject would appear as "Digital Learning Solutions Course Link - Excel " as it interprets "&..." as a new header. 

You also can't encode the entire href because it will then encode those specific characters used to specify the different sections. It no longer becomes recognisable as a mailto, and instead is interpreted as a site URL. 

So you have to encode the actual stuff that goes in the email and nothing else basically.

stellake (approved):
Looks great! 👍

PR #497, opened 2021-07-20
HEEDLS-553 Date Input Validation

ibrahimmunir14:
Refactored/rewrote/tidied up DateValidator, DateInputVC, and WelcomeEmail. Date input errors are now model state errors, compatible with error summary and treated in a similar way to normal text input errors. Errors are identified and indicated correctly. Error messages are more appropriate and correct.

- Run the formatter and made sure there are no IDE errors.
- Written tests for the changes (controller, data services, services, view models etc) and manually tested my work with and without JavaScript.
- Updated my Jira ticket with information about other parts of the system that were touched as part of the MR and have to be sanity tested to ensure nothing’s broken.
- Scanned over my own MR to ensure everything is as expected.


DigitalLearningSolutions.Web/ViewComponents/DateInputViewComponent.cs
```diff
@@ -35,6 +34,13 @@ string hintText
             var dayValue = dayProperty?.GetValue(model)?.ToString();
             var monthValue = monthProperty?.GetValue(model)?.ToString();
             var yearValue = yearProperty?.GetValue(model)?.ToString();
+            var dayErrors = ViewData.ModelState[dayProperty?.Name]?.Errors;
+            var monthErrors = ViewData.ModelState[monthProperty?.Name]?.Errors;
+            var yearErrors = ViewData.ModelState[yearProperty?.Name]?.Errors;
+
+            var errorMessage = dayErrors?.Count > 0 ? dayErrors[0].ErrorMessage :
+                               monthErrors?.Count > 0 ? monthErrors[0].ErrorMessage :
+                               yearErrors?.Count > 0 ? yearErrors?[0]?.ErrorMessage : null;
```
DanBloxham-sw, line 43: yearErrors?[0]? could be simplified to yearErrors[0]? (remove first question mark).

DigitalLearningSolutions.Web/ControllerHelpers/DateValidator.cs
```diff
@@ -1,172 +1,135 @@
...
-            public string ErrorMessage { get; set; }
-            public bool DayValid { get; set; }
-            public bool MonthValid { get; set; }
-            public bool YearValid { get; set; }
-            public int? Day { get; }
-            public int? Month { get; }
-            public int? Year { get; }
-        }
-
-        public static ValidationResult ValidateRequiredDate(int? day, int? month, int? year, string name)
```
DanBloxham-sw, line 42: It looks like this file was ported over into OldDateValidator, but OldDateValidator seems to be missing this method. Was it never used previously?

DigitalLearningSolutions.Web/ControllerHelpers/DateValidator.cs
```diff
@@ -1,172 +1,135 @@
...
+                return required ? new DateValidationResult(name + " is required") : new DateValidationResult();
+            }
+
+            if (!day.HasValue || !month.HasValue || !year.HasValue)
+            {
+                var errorMessage = GetMissingValuesErrorMessage(day, month, year, name);
+                return new DateValidationResult(!day.HasValue, !month.HasValue, !year.HasValue, errorMessage);
+            }
+
+            return ValidateDate(day.Value, month.Value, year.Value, name);
```
DanBloxham-sw, line 28: I have a slight preference to invert this to make the if condition more readable so that it becomes:

```
            if (day.HasValue && month.HasValue && year.HasValue)
            {
                return ValidateDate(day.Value, month.Value, year.Value, name);
            }

            var errorMessage = GetMissingValuesErrorMessage(day, month, year, name);
            return new DateValidationResult(!day.HasValue, !month.HasValue, !year.HasValue, errorMessage);
```

DigitalLearningSolutions.Web/ControllerHelpers/DateValidator.cs
```diff
@@ -1,172 +1,135 @@
...
+            {
+                missingValues.Add("month");
+            }
+
+            if (!year.HasValue)
+            {
+                missingValues.Add("year");
+            }
+
+            return name + " must have a " + string.Join(" and ", missingValues);
```
DanBloxham-sw, line 76: Do we want the string.Join() separator " and " to instead be " and a " to be consistent with the earlier part of the sentence?

DanBloxham-sw (commented):
Few minor comments, but pretty good overall.

DigitalLearningSolutions.Web/ControllerHelpers/DateValidator.cs
ibrahimmunir14: I added this method for use in delegate registration by centre (parent ticket of this one), it didn't exist beforehand, so OldDateValidator doesn't need it. The combined functionality of OldDateValidator and this method that was added is all available in the ValidateDate method of the new DateValidator anyway since its refactored. But we need to keep OldDateValidator around until those LearningPortal pages gets refactored.

DanBloxham-sw (approved):

DigitalLearningSolutions.Web/ViewModels/Register/RegisterDelegateByCentre/WelcomeEmailViewModel.cs
```diff
@@ -22,7 +24,17 @@ public WelcomeEmailViewModel(DelegateRegistrationByCentreData data)
...
+
+        public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
+        {
+            if (!ShouldSendEmail)
+            {
+                return new List<ValidationResult>();
+            }
+
+            var dateValidationResult = DateValidator.ValidateDate(Day, Month, Year, "Email delivery date", true);
+            return DateValidator.ToValidationResultList(dateValidationResult, "Day", "Month", "Year");
```
stellake, line 36: If these always go hand-in-hand would it make sense to always do these together in DateValidator?

stellake, line 36: Instead of hardcoded strings, should we use property names for "Day", "Month", and "Year"?

DigitalLearningSolutions.Web/ViewComponents/DateInputViewComponent.cs
```diff
@@ -35,6 +34,13 @@ string hintText
             var dayValue = dayProperty?.GetValue(model)?.ToString();
             var monthValue = monthProperty?.GetValue(model)?.ToString();
             var yearValue = yearProperty?.GetValue(model)?.ToString();
+            var dayErrors = ViewData.ModelState[dayProperty?.Name]?.Errors;
+            var monthErrors = ViewData.ModelState[monthProperty?.Name]?.Errors;
+            var yearErrors = ViewData.ModelState[yearProperty?.Name]?.Errors;
+
+            var errorMessage = dayErrors?.Count > 0 ? dayErrors[0].ErrorMessage :
+                               monthErrors?.Count > 0 ? monthErrors[0].ErrorMessage :
+                               yearErrors?.Count > 0 ? yearErrors[0].ErrorMessage : null;
```
stellake, line 43: Just trying to think if we'd ever want to display more than one error, but I don't think we would. :)

DigitalLearningSolutions.Web/ControllerHelpers/DateValidator.cs
```diff
@@ -1,172 +1,135 @@
...
+        private static DateValidationResult ValidateDate(int day, int month, int year, string name)
+        {
+            var invalidDay = day < 1 || day > 31;
+            var invalidMonth = month < 1 || month > 12;
+            var invalidYear = year < 1;
+
+            if (invalidDay || invalidMonth || invalidYear)
+            {
+                return new DateValidationResult(invalidDay, invalidMonth, invalidYear, name + " must be a real date");
+            }
```
stellake, line 40: Given that we're not highlighting all the inputs here, I wonder if we want to make the error more specific like it is for the required fields. Maybe something like Date must include real month and year?

DigitalLearningSolutions.Web/ControllerHelpers/DateValidator.cs
```diff
@@ -1,172 +1,135 @@
...
+            {
+                results.Add(new ValidationResult(result.ErrorMessage, new[] { dayId }));
+                errorMessageAdded = true;
+            }
+
+            if (result.HasMonthError)
+            {
+                results.Add(new ValidationResult(!errorMessageAdded ? result.ErrorMessage : "", new[] { monthId }));
+                errorMessageAdded = true;
+            }
```
stellake, line 99: Could we maybe pull out the errorMessage and add a comment above that. Something like:
```
// Should only add error message once per date to avoid duplicates in error summary component, but still highlight all inputs with errors
var errorMessage = ...
```

stellake (commented):
Would be good to share this at knowledge sharing next week so the team is aware of the existence of this component 👍

DigitalLearningSolutions.Web/ControllerHelpers/DateValidator.cs
```diff
@@ -1,172 +1,135 @@
...
+
+            var errorMessage = GetMissingValuesErrorMessage(day, month, year, name);
+            return new DateValidationResult(!day.HasValue, !month.HasValue, !year.HasValue, errorMessage);
+        }
+
+        private static DateValidationResult ValidateDate(int day, int month, int year, string name)
+        {
+            var invalidDay = day < 1 || day > 31;
+            var invalidMonth = month < 1 || month > 12;
+            var invalidYear = year < 1;
```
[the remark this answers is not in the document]
ibrahimmunir14: Minimum year is handled by the fact that date must be in future. Maximum year gives the 'real date' error.

DigitalLearningSolutions.Web/ViewComponents/DateInputViewComponent.cs
```diff
@@ -1,20 +1,19 @@
...
 {
-    using DigitalLearningSolutions.Web.ControllerHelpers;
     using DigitalLearningSolutions.Web.ViewModels.Common.ViewComponents;
     using Microsoft.AspNetCore.Mvc;
 
     public class DateInputViewComponent : ViewComponent
     {
         /// <summary>
-        /// Render DateInput view component.
+        ///     Render DateInput view component.
```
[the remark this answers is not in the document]
ibrahimmunir14: This was done by the formatter. Should I undo it?

DigitalLearningSolutions.Web/ViewModels/Register/RegisterDelegateByCentre/WelcomeEmailViewModel.cs
ibrahimmunir14: I thought since it is a validator, ValidateDate should simply validate a date and return an object which could easily tell you the validity and error. Forcing it to always be a list of ValidationResult seems unexpected. This way its more flexible? Not sure if one way is better than the other?

DigitalLearningSolutions.Web/ControllerHelpers/DateValidator.cs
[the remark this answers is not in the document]
ibrahimmunir14: From another comment, year > 10000 now gives a more specific error. Left minimum year as 0 otherwise the error message indicating that 1752 is not a real year is a bit strange.

DigitalLearningSolutions.Web/ViewModels/Register/RegisterDelegateByCentre/WelcomeEmailViewModel.cs
ibrahimmunir14: ToValidationResultList moved into DateValidationResult as discussed

DigitalLearningSolutions.Web.Tests/ControllerHelpers/DateValidatorTests.cs
```diff
@@ -1,149 +1,250 @@
...
+        [TestCase(false, false, true)]
+        [TestCase(true, true, false)]
+        [TestCase(true, false, true)]
+        [TestCase(false, true, true)]
+        [TestCase(true, true, true)]
+        public void ToValidationResultList_includes_one_error_for_each_erroneous_part_of_date(
+            bool hasDayError,
+            bool hasMonthError,
+            bool hasYearError
+        )
```
stellake, line 165: I think this test has too much logic in it - it's nice to have parameterised tests, but if we need to construct an array with if statements, we're adding logic that technically should be tested itself. Could we simplify this (feel free to break it up into separate tests) such that the additional logic on constructing the errorMemberNames is not required? 🙏

stellake (approved):
Looks good! 👍

PR #498, opened 2021-07-21
HEEDLS-555 View delegate page - course section

ibrahimmunir14:


DigitalLearningSolutions.Web/ViewModels/TrackingSystem/Delegates/DelegateCourseInfoViewModel.cs
```diff
@@ -0,0 +1,74 @@
...
+        public string? Completed { get; set; }
+        public string? Evaluated { get; set; }
+        public string EnrollmentMethod { get; set; }
+        public int LoginCount { get; set; }
+        public string LearningTime { get; set; }
+        public int? DiagnosticScore { get; set; }
+        public bool IsAssessed { get; set; }
+        public List<string?> CustomAnswers { get; set; }
+
+        public List<CustomPrompt> CustomPrompts { get; set; }
```
DanBloxham-sw, line 68: There is a CustomPromptWithAnswer class that is probably best to use here. It is a better way to keep the answers and questions ordered together, with less worry about getting the indexes correct. 

It will probably require a new method in the CustomPromptService that maps them together (see PopulateCustomPromptWithAnswerListFromCentreCustomPromptsResult as an example). I think those methods are normally private and called through the public method that gets the prompt questions in the first place.

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/Delegates/DelegateCourseInfoViewModel.cs
```diff
@@ -0,0 +1,74 @@
...
+            CustomisationId = info.CustomisationId;
+            ApplicationName = info.ApplicationName;
+            CustomisationName = info.CustomisationName;
+            Supervisor = info.Supervisor;
+            Enrolled = info.Enrolled.ToString("dd/MM/yyyy");
+            LastUpdated = info.LastUpdated.ToString("dd/MM/yyyy");
+            if (info.CompleteBy.HasValue)
+            {
+                CompleteBy = info.CompleteBy.Value.ToString("dd/MM/yyyy");
+            }
```
DanBloxham-sw, line 24: Could these be simplified to the following:
CompleteBy = info.CompleteBy?.ToString("dd/MM/yyyy"); 

Also you might be able to get away with .ToShortDateString().

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/ViewDelegate/Index.cshtml
```diff
@@ -120,4 +120,19 @@
       }
     </div>
   </div>
+
+  <div class="nhsuk-grid-column-full">
+    <h2 class="nhsuk-heading-l nhsuk-u-margin-bottom-2">Courses</h2>
+    @if (!Model.CourseInfoList.Any()) {
+      <p class="nhsuk-u-margin-top-4" role="alert">
```
DanBloxham-sw, line 127: Does this need role="alert"?

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/ViewDelegate/_DelegateCourseInfoCard.cshtml
```diff
@@ -0,0 +1,132 @@
...
+        </div>
+
+        <div class="nhsuk-summary-list__row details-list-with-button__row">
+          <dt class="nhsuk-summary-list__key">
+            Supervisor
+          </dt>
+          <partial name="_SummaryFieldValue" model="@Model.Supervisor" />
+          <dd class="nhsuk-summary-list__actions">
+            <a href="#">Edit<span class="nhsuk-u-visually-hidden"> supervisor</span></a>
+          </dd>
```
DanBloxham-sw, line 28: From the screenshots, it looks like the presence of the actions messes up the line separators between the fields. The wireframes have uniform lines that go all the way across on every field.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/ViewDelegate/_DelegateCourseInfoCard.cshtml
```diff
@@ -0,0 +1,132 @@
...
+            </dt>
+            <partial name="_SummaryFieldValue" model="@Model.TotalAttempts.ToString()" />
+          </div>
+
+          <div class="nhsuk-summary-list__row details-list-with-button__row">
+            <dt class="nhsuk-summary-list__key">
+              Assessment attempts
+            </dt>
+            <partial name="_SummaryFieldValue" model="@Model.AttemptsPassed.ToString()" />
+          </div>
```
DanBloxham-sw, line 108: I think the model values displayed for these two are mixed up.

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/Delegates/DelegateCourseInfoViewModel.cs
```diff
@@ -0,0 +1,74 @@
...
+        public string LearningTime { get; set; }
+        public int? DiagnosticScore { get; set; }
+        public bool IsAssessed { get; set; }
+        public List<string?> CustomAnswers { get; set; }
+
+        public List<CustomPrompt> CustomPrompts { get; set; }
+        public int TotalAttempts { get; set; }
+        public int AttemptsPassed { get; set; }
+
+        public double? PassRatio => TotalAttempts != 0 ? AttemptsPassed / (double)TotalAttempts : (double?)null;
```
DanBloxham-sw, line 72: There was another page where we had a similar PassRatio (I think it was the Top Courses page) that was indeed called "PassRatio" on the old site, but displayed a percentage.

For the top courses page, we ended up renaming it "PassRate" and still displayed it as a percentage. Might be worth double checking with Stella, but I think the same should be done here.

DanBloxham-sw (commented):
Few minor things, might require some refactoring to use CustomPromptWithAnswer class.

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/Delegates/DelegateCourseInfoViewModel.cs
ibrahimmunir14: Oh yeah.

Can't do ToShortDateString because that depends on locale (e.g. fails on github runner because it uses MM/dd/yyyy format).

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/ViewDelegate/_DelegateCourseInfoCard.cshtml
ibrahimmunir14: Whoops 🤭

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/ViewDelegate/_DelegateCourseInfoCard.cshtml
```diff
@@ -0,0 +1,143 @@
...
+            <dt class="nhsuk-summary-list__key">
+              Assessment attempts
+            </dt>
+            <partial name="_SummaryFieldValue" model="@Model.TotalAttempts.ToString()" />
+            <dd class="nhsuk-summary-list__actions"></dd>
+          </div>
+
+          <div class="nhsuk-summary-list__row details-list-with-button__row">
+            <dt class="nhsuk-summary-list__key">
+              Pass ratio
```
DanBloxham-sw, line 121: Going to want to rename the field to Pass Rate as well.

DanBloxham-sw (approved):
Looks good. Minor comment about renaming a display field, but no need for reapproval from me. 

I think the empty action on all the fields should fix the line issue, but it would be nice to have an updated screen shot as well.

DigitalLearningSolutions.Data/Services/CustomPromptsService.cs
```diff
@@ -4,6 +4,7 @@
     using System.Linq;
```
stellake, line 4: This is getting quite long - I wonder if we could refactor this into two separate services, one for registration prompts and the other for admin fields.. Olivia is working on this too atm, so she could potentially do the refactor later.

DigitalLearningSolutions.Data/DataServices/CourseDataService.cs
```diff
@@ -181,5 +183,57 @@ public IEnumerable<CourseStatistics> GetCourseStatisticsAtCentreForCategoryId(in
...
+        }
+
+        public (int totalAttempts, int attemptsPassed) GetDelegateCourseAttemptStats(
+            int delegateId,
+            int customisationId
+        )
+        {
+            return connection.QueryFirstOrDefault<(int, int)>(
+                @"SELECT COUNT(aa.Status) AS TotalAttempts,
+                        COUNT(CASE WHEN aa.Status=1 THEN 1 END) AS AttemptsPassed
```
stellake, line 228: I wonder if it makes sense pulling these two numbers together with other course data?

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/Delegates/DelegateCourseInfoViewModel.cs
```diff
@@ -0,0 +1,65 @@
...
+            CompleteBy = info.CompleteBy?.ToString("dd/MM/yyyy");
+            Completed = info.Completed?.ToString("dd/MM/yyyy");
+            Evaluated = info.Evaluated?.ToString("dd/MM/yyyy");
+            EnrolmentMethod = info.EnrolmentMethodId switch
+            {
+                1 => "Self enrolled",
+                2 => "Administrator",
+                3 => "Group",
+                4 => "System",
+                _ => ""
```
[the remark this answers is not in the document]
ibrahimmunir14: db has a couple of records with EnrollmentMethodID = 0

DigitalLearningSolutions.Data/DataServices/CustomPromptsDataService.cs
```diff
@@ -156,11 +158,9 @@ public string GetPromptNameForCentreAndPromptNumber(int centreId, int promptNumb
                     LEFT JOIN CoursePrompts AS cp3 
                         ON cu.CourseField3PromptID = cp3.CoursePromptID
                     INNER JOIN dbo.Applications AS ap ON ap.ApplicationID = cu.ApplicationID
-                    WHERE (ap.CourseCategoryID = @categoryId OR @categoryId = 0) 
-                        AND cu.CentreID = @centreId
-                        AND ap.ArchivedDate IS NULL
```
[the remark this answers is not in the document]
ibrahimmunir14: We decided to move these filters into the service.
CategoryId filter needed to be optional, so we added service methods to make it cleaner.
Then I wasn't sure whether or not archived courses were included in delegate courses, so pulled that filter out too for consistency. If it's more flexible than needed I can put the archived date filter back into the sql?

DigitalLearningSolutions.Data/DataServices/CourseDataService.cs
ibrahimmunir14: I thought it would make the query too complicated. Also these are not needed in many cases any where IsAssessed=0.

DigitalLearningSolutions.Data/Services/CustomPromptsService.cs
AlexJacksonDS: I agree with this, it's at a similar length to the UserDataService I've just split. @livzorn FYI, this split might come your way at the end of the admin fields work.

DigitalLearningSolutions.Data.Tests/DataServices/CourseDataServiceTests.cs
```diff
@@ -245,5 +245,47 @@ public void GetCourseStatisticsAtCentreForCategoryID_should_return_course_statis
...
+
+        [Test]
+        public void GetDelegateCoursesInfo_should_return_delegate_course_info_correctly()
+        {
+            // When
+            var results = courseDataService.GetDelegateCoursesInfo(20).ToList();
+
+            // Then
+            var enrollmentDate = new DateTime(2019, 04, 11, 14, 33, 37).AddMilliseconds(140);
+            var expected = new DelegateCourseInfo(
```
AlexJacksonDS, line 260: Do we want a helper for this?

DigitalLearningSolutions.Data/DataServices/CourseDataService.cs
AlexJacksonDS: I think there is a bit more refactoring needed in that case (also see my comment on the controller). If these aren't needed because IsAssessed=0, why are we still always calling it?

DigitalLearningSolutions.Data/DataServices/CustomPromptsDataService.cs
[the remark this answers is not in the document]
AlexJacksonDS: I think we should keep the Archived filter in the SQL for now, we can always bring it back if we need to in the future, but it seems unlikely that we would want to show archived courses.

DigitalLearningSolutions.Data/Services/CustomPromptsService.cs
```diff
@@ -157,16 +172,54 @@ public string GetPromptNameForCentreAndPromptNumber(int centreId, int promptNumb
...
+            var result = GetCourseCustomPromptsResultForCourse(customisationId, centreId, categoryId, includeArchived);
+
+            return PopulateCustomPromptWithAnswerListFromCourseCustomPromptsResult(result, delegateCourseInfo);
+        }
+
+        private CourseCustomPromptsResult? GetCourseCustomPromptsResultForCourse(
+            int customisationId,
+            int centreId,
+            int categoryId,
+            bool includeArchived
```
AlexJacksonDS, line 212: Just marking where this is so I don't have to find it again. This should be removed as we add it back to SQL

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/ViewDelegateController.cs
```diff
@@ -12,29 +14,54 @@
...
+                    {
+                        var courseCustomPromptsWithAnswers = customPromptsService.GetCustomPromptsWithAnswersForCourse(
+                            info,
+                            info.CustomisationId,
+                            centreId,
+                            includeArchived: true
+                        );
+                        var attemptStats =
+                            courseDataService.GetDelegateCourseAttemptStats(delegateId, info.CustomisationId);
+                        return new DelegateCourseInfoViewModel(info, courseCustomPromptsWithAnswers, attemptStats);
```
AlexJacksonDS, line 60: I don't think all this logic should be in the controller. I think it should be in the Service level, populate all the information we need for the view into a single entity (or tuple if necessary).

i.e. a new method in CourseService GetAllCoursesForDelegate that essentially does the content of the lambda. Then this call just becomes Select(x => new DelegateCourseInfoViewModel(x))

DigitalLearningSolutions.Web/Styles/trackingSystem/viewDelegate.scss
```diff
@@ -23,3 +23,9 @@
     width: 100%;
   }
 }
+
+.nhsuk-summary-list__value {
+  @media (min-width: 40.0625em) {
```
AlexJacksonDS, line 28: Random width? Presumably this is the point where something breaks. Could you clarify what? (and maybe change to px?)

AlexJacksonDS (changes requested):
A few things that I think need changing.

DigitalLearningSolutions.Data.Tests/DataServices/CourseDataServiceTests.cs
ibrahimmunir14: For creating the object? I didn't think so as it's only used once and none of the other tests in the file use helpers.

AlexJacksonDS: OK fine, I'll bring it up again if we start initialising multiple of these in the future.

DigitalLearningSolutions.Web/Styles/trackingSystem/viewDelegate.scss
ibrahimmunir14: This is the point where summary list changes from horizontal to vertical layout.

AlexJacksonDS: Is that not tied to an NHS breakpoint?

ibrahimmunir14: Yeah, I'm trying to find it.

DigitalLearningSolutions.Data/Services/CustomPromptsService.cs
livzorn: @AlexJacksonDS ah yeah thanks, I'll keep that in mind!

DigitalLearningSolutions.Data/Services/CourseService.cs
```diff
@@ -4,20 +4,29 @@
     using System.Linq;
     using DigitalLearningSolutions.Data.DataServices;
     using DigitalLearningSolutions.Data.Models.Courses;
+    using DigitalLearningSolutions.Data.Models.CustomPrompts;
 
     public interface ICourseService
     {
+        public IEnumerable<(DelegateCourseInfo info, List<CustomPromptWithAnswer> prompts, (int, int) stats)>
```
AlexJacksonDS, line 11: I think at this point we'd be better with a class to contain this data

DigitalLearningSolutions.Data.Tests/TestHelpers/CustomPromptsTestHelper.cs
```diff
@@ -125,7 +122,9 @@ public static CentreCustomPromptsResult GetDefaultCentreCustomPromptsResult
                 CustomField2Mandatory = customField2Mandatory,
                 CustomField3Prompt = customField3Prompt,
                 CustomField3Options = customField3Options,
-                CustomField3Mandatory = customField3Mandatory
+                CustomField3Mandatory = customField3Mandatory,
+                ArchivedDate = archivedDate,
```
AlexJacksonDS, line 126: Do we still need this?

DigitalLearningSolutions.Data/DataServices/CustomPromptsDataService.cs
```diff
@@ -146,7 +146,9 @@ public string GetPromptNameForCentreAndPromptNumber(int centreId, int promptNumb
                         cu.Q2Mandatory AS CustomField2Mandatory,
                         cp3.CoursePrompt AS CustomField3Prompt,
                         cu.Q3Options AS CustomField3Options, 
-                        cu.Q3Mandatory AS CustomField3Mandatory
+                        cu.Q3Mandatory AS CustomField3Mandatory,
+                        ap.CourseCategoryID,
+                        ap.ArchivedDate
```
AlexJacksonDS, line 151: This is always null now?

DigitalLearningSolutions.Data/Services/CourseService.cs
```diff
@@ -31,5 +40,24 @@ public IEnumerable<CourseStatistics> GetCentreSpecificCourseStatistics(int centr
             var allCourses = courseDataService.GetCourseStatisticsAtCentreForCategoryId(centreId, categoryId);
             return allCourses.Where(c => c.CentreId == centreId);
         }
+
+        public IEnumerable<(DelegateCourseInfo info, List<CustomPromptWithAnswer> prompts, (int, int) stats)>
```
AlexJacksonDS, line 44: Unit tests for this method?

AlexJacksonDS (changes requested):
Some remaining ArchivedDate to remove still.
I don't think an IEnumerable<(DelegateCourseInfo info, List<CustomPromptWithAnswer> prompts, (int, int) stats)> is a good idea really. Replace this with an IEnumerable of a new class
New CourseService method needs unit tests, unless I've missed them?

AlexJacksonDS (approved):
Looks OK to me now 👍

PR #500, opened 2021-07-22
HEEDLS-558 Show tracking system to UserAdmin

livzorn:
A small bug fix - I changed the tracking system access to accept users who are User Admins, Centre Managers, or Centre Admins, rather than assuming that if a user is a Centre Admin they have those other permissions.

### DIFF
--- a/DigitalLearningSolutions.Web/Controllers/ApplicationSelectorController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/ApplicationSelectorController.cs
@@ -13,7 +13,7 @@
         public IActionResult Index()
         {
             var learningPortalAccess = User.GetCustomClaimAsBool(CustomClaimTypes.LearnUserAuthenticated) ?? false;
-            var trackingSystemAccess = User.GetCustomClaimAsBool(CustomClaimTypes.UserCentreAdmin) ?? false;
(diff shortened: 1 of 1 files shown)


DigitalLearningSolutions.Web/Controllers/ApplicationSelectorController.cs
```diff
@@ -13,7 +13,7 @@ public class ApplicationSelectorController : Controller
         public IActionResult Index()
         {
             var learningPortalAccess = User.GetCustomClaimAsBool(CustomClaimTypes.LearnUserAuthenticated) ?? false;
-            var trackingSystemAccess = User.GetCustomClaimAsBool(CustomClaimTypes.UserCentreAdmin) ?? false;
+            var trackingSystemAccess = User.GetCustomClaimAsBool(CustomClaimTypes.UserUserAdmin) | User.GetCustomClaimAsBool(CustomClaimTypes.UserCentreManager) | User.GetCustomClaimAsBool(CustomClaimTypes.UserCentreAdmin) ?? false;
```
AlexJacksonDS, line 16: This should be able to be replaced with User.HasCentreAdminPermissions(), I think the resulting bool is the same.

If it can't be replaced by that smaller helper method, you need to run the formatter on this to break up the long line. If you don't know how to do that drop me a message.

AlexJacksonDS (changes requested):
I think the long line can be replaced with a claims helper method, otherwise it needs the formatter running.

AlexJacksonDS (approved):
Looks good.

stellake (approved):
Looks good! 👍

PR #501, opened 2021-07-22
HEEDLS-548 Exclude centres from self registration

livzorn:
Small fix - I added one line to exclude centres with kbSelfRegister=0 from the select options in the delegate self-registration journey.

### DIFF
--- a/DigitalLearningSolutions.Data/DataServices/CentresDataService.cs
+++ b/DigitalLearningSolutions.Data/DataServices/CentresDataService.cs
@@ -12,7 +12,7 @@
     {
         string? GetBannerText(int centreId);
         string? GetCentreName(int centreId);
-        IEnumerable<(int, string)> GetActiveCentresForDelegateSelfRegistrationAlphabetical();
+        IEnumerable<(int, string)> GetCentresForDelegateSelfRegistrationAlphabetical();
         Centre? GetCentreDetailsById(int centreId);
 
         void UpdateCentreManagerDetails(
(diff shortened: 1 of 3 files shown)


AlexJacksonDS (commented):
Change itself looks fine, can we just add a small unit test to for this case in the same pattern as Get_active_centres_should_not_contain_an_inactive_centre

livzorn:
@AlexJacksonDS Thanks! I added a unit test, let me know if it looks okay.

AlexJacksonDS (approved):
Looks good

DigitalLearningSolutions.Data/DataServices/CentresDataService.cs
```diff
@@ -99,6 +99,7 @@ public IEnumerable<(int, string)> GetActiveCentresAlphabetical()
                 @"SELECT CentreID, CentreName
                         FROM Centres
                         WHERE Active = 1
+                        AND kbSelfRegister = 1
```
stellake, line 102: Happy to put this in the db query level given that this is only used in registration, but I think we should modify the name of the method so it's clear that we're getting the centre options for delegate self registration (rather than a true active centres list, which should technically only have the Active = 1 constraint)

livzorn: Good idea! I changed the method name to account for this

stellake (approved):
Looks good! I'd probably remove the "Active" from the method name as it's unnecessary with the extra specification we've added, but no need for rereview 👍

PR #502, opened 2021-07-22
Heedls 506 delegate groups basic page

AlexJacksonDS:
Added a delegate groups page with delegate group expanders.

Buttons at the top collapse with the same breakpoints as the buttons on the all delegates page.

image

image

image

image

image


ibrahimmunir14 (changes requested):
Looks good. A few questions and comments mainly about formatting.

DigitalLearningSolutions.Web/DigitalLearningSolutions.Web.csproj
```diff
@@ -99,6 +99,10 @@
     <TypeScriptCompile Include="Scripts\trackingSystem\reports.ts" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="ViewModels\TrackingSystem\Delegates\DelegateGroups\" />
+  </ItemGroup>
```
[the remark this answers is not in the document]
AlexJacksonDS: It's a strange behaviour that sometimes happens when I've added a folder. It's safe to remove, I just didn't notice it this time.

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/Delegates/DelegateGroups/DelegateGroupsViewModel.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    using System.Collections.Generic;
+    using System.Linq;
+    using DigitalLearningSolutions.Data.Models.DelegateGroups;
+
+    public class DelegateGroupsViewModel
+    {
+        public DelegateGroupsViewModel(IEnumerable<Group> groups)
+        {
+            DelegateGroups = groups.Select(g => new SearchableDelegateGroupViewModel(g));
```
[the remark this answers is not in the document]
AlexJacksonDS: Yeah that is going to need to change once we start implementing the search/sort/filter functionality on that page. The helpers expect IEnumerable<T> where T : BaseSearchableItem (in the data project, see AdminUser for an example). We do all the filtering down then convert the resulting list to the list item view model and this logic lives in the constructor of the overall view model so that it can be unit tested more easily.

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/Delegates/DelegateGroups/SearchableDelegateGroupViewModel.cs
```diff
@@ -0,0 +1,44 @@
...
+        public SearchableDelegateGroupViewModel(Group group)
+        {
+            Id = group.GroupId;
+            Name = group.GroupLabel;
+            Description = group.GroupDescription;
+            DelegateCount = group.DelegateCount;
+            CourseCount = group.CoursesCount;
+            LinkedToField = group.LinkedToField;
+            LinkedField = group.LinkedToFieldName;
+            AddedBy = $"{group.AddedByFirstName} {group.AddedByLastName}";
```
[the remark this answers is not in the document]
AlexJacksonDS: Wasn't sure what we'd need later, figured it might be better to do it c# side just in case. Don't think it makes any difference at all where it happens though.

ibrahimmunir14 (approved):

stellake (commented):
Looks good, just a few comments. Styling wise could we:
1. Add a screenshot of an opened card on mobile.
2. Use "-" if the field is empty or null (I believe @ibrahimmunir14 wrote a component for it)
3. Make sure top and bottom spacing for buttons is equal.

image

ibrahimmunir14:
> Looks good, just a few comments. Styling wise could we:
> 
> 1. Add a screenshot of an opened card on mobile.
> 2. Use "-" if the field is empty or null (I believe @ibrahimmunir14 wrote a component for it)
> 3. Make sure top and bottom spacing for buttons is equal.
> 
> image

2. It's _SummaryFieldValue partial which takes in a `string?`

AlexJacksonDS:
image
image
image
image
image
image
image
image

AlexJacksonDS:
image

DigitalLearningSolutions.Web/Styles/trackingSystem/delegates/delegateGroups.scss
```diff
@@ -0,0 +1,12 @@
+﻿@import 'delegateCommon';
+
+.nhsuk-details__text > .nhsuk-button {
```
SteveJacksonSoft, line 3: I'm slightly scared by this because the HTML file will look like we're just using nhsuk's styles, and future devs probably wouldn't expect the button's styling to change just because they move it, or because they add a div between it and the `.nhsuk-details__text`.

I think a better approach would be to put an extra class on the buttons you're selecting here and select them as `.nhsuk-button.<new-class>` - that gets around the problems above, and may even get rid of the need for `!important`.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateGroups/Index.cshtml
```diff
@@ -0,0 +1,62 @@
...
+    <div class="nhsuk-grid-row">
+      <div class="nhsuk-grid-column-full">
+        @if (!Model.DelegateGroups.Any()) {
+          <p class="nhsuk-u-margin-top-4" role="alert">
+            <b>No delegate groups found.</b>
+          </p>
+        } else {
+          <div id="searchable-elements">
+            @foreach (var groupModel in Model.DelegateGroups) {
+              <partial name="_SearchableDelegateGroupCard" model="groupModel" />
```
SteveJacksonSoft, line 51: I don't understand where `Searchable` came from in this name, but I haven't been paying much attention to DLS recently. What was your reasoning?

DigitalLearningSolutions.Data/Models/DelegateGroups/Group.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        public string AddedByLastName { get; set; }
+
+        public int LinkedToField { get; set; }
+
+        public string LinkedToFieldName { get; set; }
+
+        public bool AddNewRegistrants { get; set; }
+
+        public bool SyncFieldChanges { get; set; }
```
SteveJacksonSoft, line 25: So, obviously this was the db creator's choice, but I don't know what `AddNewRegistrants=true` means - does it mean that new registrants should be added to the group? Or have been added? I have similar issues with `SyncFieldChanges=true`.

I would be inclined to give these boolean properties more helpful names and use a mapper (take inspiration from `CurrentCourseMap`) to map them to the db properties.

SteveJacksonSoft (changes requested):
Lots of nice stuff, just a couple of things I think are worth changing.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateGroups/Index.cshtml
AlexJacksonDS: There will be some combination of Search/Sort/Filter/Paginate added as part of 507/508. Just naming it appropriately now.

DigitalLearningSolutions.Web/Styles/trackingSystem/delegates/delegateGroups.scss
AlexJacksonDS: Yep, that is definitely better and fixes the specificity issue I was having resulting in this selector and all the !important declarations.

SteveJacksonSoft: Excellent! Maybe useful for the name of the new class to mention `nhsuk-button` or somehow make clear it should be used in conjunction with it and won't do anything if you put it on any old button, if that can be done without making it annoyingly lanky, but no big deal.

DigitalLearningSolutions.Data.Tests/DataServices/GroupsDataServiceTests.cs
```diff
@@ -34,8 +34,8 @@ public void GetGroupsForCentre_returns_expected_groups()
                 AddedByLastName = "Whittaker (Developer)",
                 LinkedToField = 4,
                 LinkedToFieldName = "Job group",
-                AddNewRegistrants = true,
-                SyncFieldChanges = true
+                AddNewRegistrantsToGroup = true,
+                SyncChangesToRegistrationInfoWithGroupMembership = true
```
SteveJacksonSoft: So yeah, that definitely gives much better context, but I'm still not certain what they mean.
The ideal would be if each variable's name were a predicate which is true or false, rather than a command (because commands can easily be misunderstood as a concise form of a predicate; predicates generally can't be misunderstood).
So e.g. ShouldAddNewRegistrantsToGroup / NewRegistrantsHaveBeenAddedToGroup / NewRegistrantsAreInGroup are probably clearer names.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateGroups/_SearchableDelegateGroupCard.cshtml
```diff
@@ -93,21 +93,21 @@
...
           <dd class="nhsuk-summary-list__value">
-            @Model.SyncFieldChanges
+            @Model.SyncChangesToRegistrationInfoWithGroupMembership
           </dd>
           <dd class="nhsuk-summary-list__actions" />
         </div>
       </dl>
 
-      <a class="nhsuk-button left-button" role="button">
+      <a class="nhsuk-button nhsuk-button__card-button" role="button">
```
SteveJacksonSoft, line 102: Nice, think it's better not to start it with `nhsuk` though - people will assume it's an nhs style (or nhs might update their library and it might become an nhs style!). I appreciate the adherence to BEM, but I think in this case it should be a `delegate-card__button`, or maybe `searchable-element__button`?

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateGroups/Index.cshtml
SteveJacksonSoft: Aye, if you're confident enough it gives a solid description of the component, all fine by me!

SteveJacksonSoft (changes requested):
Nice, just a couple of naming things.

SteveJacksonSoft (approved):
Beautiful! Merge away!

DigitalLearningSolutions.Data.Tests/DataServices/GroupsDataServiceTests.cs
AlexJacksonDS: Yeah, I'm not great at coming up with reasonably concise names for variables when they relate to a fairly complex concept. (the Sync one is definitely worse for this).

SteveJacksonSoft: Think you've come up with a lovely one for sync! But I reckon clarity generally trumps conciseness, especially now we have IDEs to autofill things for us.

PR #503, opened 2021-07-22
HEEDLS-469 Make region filter wider

AlexJacksonDS:
Increased the width of the region dropdown.

Note that this build failed unit tests because master is currently broken, will have to wait until that is resolved.

image


stellake (approved):
Looks good! Master should be ok now 👍

PR #508, opened 2021-07-23
Heedls 557 email length consistency

livzorn:
### JIRA link
[HEEDLS-557](https://softwiretech.atlassian.net/browse/HEEDLS-557)

### Description
I created a migration that changed the maximum length of the Email column on AdminUsers table from 250 to 255 characters, so it is consistent with the Candidates table. The migration also deletes and recreates the AdminUsers_Email index accordingly as it depends the Email column.

I also modified two view models so the form input for Email will have the same max length.

I will document these changes in the release notes after merge!

### DIFF
--- a/DigitalLearningSolutions.Data.Migrations/202107221605_ChangeEmailMaxLengthOnAdminUsers.cs
+++ b/DigitalLearningSolutions.Data.Migrations/202107221605_ChangeEmailMaxLengthOnAdminUsers.cs
@@ -8,7 +8,7 @@
         public override void Up()
         {
             Delete.Index("IX_AdminUsers_Email").OnTable("AdminUsers");
-            Alter.Column("Email").OnTable("AdminUsers").AsString(255);
+            Alter.Column("Email").OnTable("AdminUsers").AsString(255).Nullable();
             Create.Index("IX_AdminUsers_Email").OnTable("AdminUsers").OnColumn("Email").Ascending().WithOptions()
                 .Unique().WithOptions().NonClustered();
         }
@@ -16,7 +16,7 @@
         public override void Down()
         {
             Delete.Index("IX_AdminUsers_Email").OnTable("AdminUsers");
-            Alter.Column("Email").OnTable("AdminUsers").AsString(250);
+            Alter.Column("Email").OnTable("AdminUsers").AsString(250).Nullable();
             Create.Index("IX_AdminUsers_Email").OnTable("AdminUsers").OnColumn("Email").Ascending().WithOptions()
                 .Unique().WithOptions().NonClustered();
         }
--- a/DigitalLearningSolutions.Data.Migrations/202107221605_ChangeEmailMaxLengthOnAdminUsers.cs
+++ b/DigitalLearningSolutions.Data.Migrations/202107221605_ChangeEmailMaxLengthOnAdminUsers.cs
@@ -19,7 +19,6 @@
             Alter.Column("Email").OnTable("AdminUsers").AsString(250);
             Create.Index("IX_AdminUsers_Email").OnTable("AdminUsers").OnColumn("Email").Ascending().WithOptions()
                 .Unique().WithOptions().NonClustered();
-            ;
         }
     }
 }
--- a/DigitalLearningSolutions.Data.Migrations/202107221605_ChangeEmailMaxLengthOnAdminUsers.cs
+++ b/DigitalLearningSolutions.Data.Migrations/202107221605_ChangeEmailMaxLengthOnAdminUsers.cs
@@ -7,7 +7,7 @@
     {
         public override void Up()
         {
-            Delete.Index("IX_AdminUsers_Email");
+            Delete.Index("IX_AdminUsers_Email").OnTable("AdminUsers");
             Alter.Column("Email").OnTable("AdminUsers").AsString(255);
             Create.Index("IX_AdminUsers_Email").OnTable("AdminUsers").OnColumn("Email").Ascending().WithOptions()
                 .Unique().WithOptions().NonClustered();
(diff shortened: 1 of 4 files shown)


AlexJacksonDS (approved):
The change itself looks fine, but question for @stellake.

I can see that you've decided the centre fields are out of scope for this ticket in the HLD discussion, but now we are in the situation where the Centre manager can have an email that is 255 in the AdminUsers table, but only 250 in the Centres table. Do we have a ticket in the future to fix this? (Centre website email is 100 characters too, but doesn't necessarily match an admin like the CentreManager field should)

DigitalLearningSolutions.Data.Migrations/202107221605_ChangeEmailMaxLengthOnAdminUsers.cs
```diff
@@ -0,0 +1,25 @@
...
+                .Unique().WithOptions().NonClustered();
+        }
+
+        public override void Down()
+        {
+            Delete.Index("IX_AdminUsers_Email").OnTable("AdminUsers");
+            Alter.Column("Email").OnTable("AdminUsers").AsString(250);
+            Create.Index("IX_AdminUsers_Email").OnTable("AdminUsers").OnColumn("Email").Ascending().WithOptions()
+                .Unique().WithOptions().NonClustered();
+            ;
```
stellake, line 22: Do we need ";" here?

DigitalLearningSolutions.Data.Migrations/202107221605_ChangeEmailMaxLengthOnAdminUsers.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    [Migration(202107221605)]
+    public class ChangeEmailMaxLengthOnAdminUsers : Migration
+    {
+        public override void Up()
+        {
+            Delete.Index("IX_AdminUsers_Email").OnTable("AdminUsers");
+            Alter.Column("Email").OnTable("AdminUsers").AsString(255);
+            Create.Index("IX_AdminUsers_Email").OnTable("AdminUsers").OnColumn("Email").Ascending().WithOptions()
+                .Unique().WithOptions().NonClustered();
```
stellake, line 13: Do we need "WithOptions" twice?

DigitalLearningSolutions.Data.Migrations/202107221605_ChangeEmailMaxLengthOnAdminUsers.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace DigitalLearningSolutions.Data.Migrations
+{
+    using FluentMigrator;
+
+    [Migration(202107221605)]
+    public class ChangeEmailMaxLengthOnAdminUsers : Migration
```
stellake, line 6: Could you please include a screenshot of the before and after Index properties in the db for easy comparison? (you should be able to just right click on the index and choose "properties" in SQL Server Management Studio. The old index can be found on our VM, but make sure to use the non-test db)

stellake:
> I can see that you've decided the centre fields are out of scope for this ticket in the HLD discussion, but now we are in the situation where the Centre manager can have an email that is 255 in the AdminUsers table, but only 250 in the Centres table. Do we have a ticket in the future to fix this? (Centre website email is 100 characters too, but doesn't necessarily match an admin like the CentreManager field should)

The centre manager details are completely independent of the admin user that has centre manager permissions. The reason we're changing the email lengths of the users is that it affects the logins/core functionality/how we've combined these users into one account. The centre manager details are there as additional contact details display for centre, so we're good to leave it (and this way we don't need to introduce extra risk or spend unnecessary budget either). I suspect in reality, no one's going to specify their centre contact details as something with 200+ characters :)

DigitalLearningSolutions.Data.Migrations/202107221605_ChangeEmailMaxLengthOnAdminUsers.cs
livzorn: Oops! just a typo

DigitalLearningSolutions.Data.Migrations/202107221605_ChangeEmailMaxLengthOnAdminUsers.cs
livzorn: I first had it only before Unique, but I was getting a squiggly on Nonclustered saying `CS1061 'ICreateIndexOnColumnSyntax' does not contain a definition for 'NonClustered' and no accessible extension method 'NonClustered' accepting a first argument of type 'ICreateIndexOnColumnSyntax' could be found (are you missing a using directive or an assembly reference?)`

If I remove WithOptions before Unique the same thing happens on Unique, so I'm assuming this is just the syntax? But I don't know, I've never written migrations in c sharp before 🙃

DigitalLearningSolutions.Data.Migrations/202107221605_ChangeEmailMaxLengthOnAdminUsers.cs
livzorn: image

livzorn: After: 
image

stellake: It looks like the "Allow NULLs" tab on the image is cut off - I'm pretty sure NULL values were allowed before, so this change could break the live DB

stellake: We've also changed varchar to nvarchar - do you mind double checking it's the same as in the Candidates table to make sure these are consistent 🙏

livzorn: Yes nvarchar is consistent with the Candidates table!

livzorn: @Stellake You're right, NULL values were allowed before. I think I've made the right change but I'm having trouble applying the migration to the database so I can see if the properties change correctly, what's the right way to do that?

livzorn: @stellake Got the migrations to apply, and the Index looks correct now!
image

stellake (approved):
Looks good! 👍

PR #512, opened 2021-07-26
HEEDLS-431 - Changed javascript to copy text before alert and updated…

DanBloxham-sw:
… button styling

### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-431

### Description
The previous javascript copying functions only copied the text to clipboard after the alert was dismissed. I've updated it to be copied first. I've also updated the styling of the button to look the same as other links when focussed (which it previously didn't). 

### Screenshots
image

-----
### Developer checks
I have:

### DIFF
--- a/DigitalLearningSolutions.Web/Scripts/trackingSystem/centreCourseSetup.ts
+++ b/DigitalLearningSolutions.Web/Scripts/trackingSystem/centreCourseSetup.ts
@@ -24,20 +24,22 @@ function copyLaunchCourseLinkToClipboard(customisationId: string) {
 }
 
 function copyTextToClipboard(textToCopy: string): void {
-  try {
-    navigator.clipboard.writeText(textToCopy).then(() => {
-      displaySuccessAlert(textToCopy);
-    }, () => {
-      displayFailureAlert(textToCopy);
-    });
-  } catch (e) {
+  if (!navigator.clipboard) {
     const succeeded = copyTextToClipboardFallback(textToCopy);
     if (succeeded) {
       displaySuccessAlert(textToCopy);
     } else {
       displayFailureAlert(textToCopy);
     }
+    return;
   }
+
+  navigator.clipboard.writeText(textToCopy).then(() => {
+    displaySuccessAlert(textToCopy);
+  },
+  () => {
+    displayFailureAlert(textToCopy);
+  });
 }
 
 function copyTextToClipboardFallback(textToCopy: string): boolean {
@@ -61,6 +63,6 @@ function displaySuccessAlert(text: string): void {
   alert(`Copied the text: ${text}`);
 }
 
-function displayFailureAlert(text: string) :void {
+function displayFailureAlert(text: string): void {
   alert(`Copy not supported or blocked. Try manually selecting and copying: ${text}`);
 }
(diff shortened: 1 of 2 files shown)


DigitalLearningSolutions.Web/Styles/trackingSystem/courseSetup.scss
```diff
@@ -2,17 +2,25 @@
...
-        text-decoration: none;
-        cursor: pointer;
-    }
+  &:hover {
+    color: $nhsuk-link-hover-color;
+    text-decoration: none;
+    cursor: pointer;
+  }
+
+  &:focus {
```
livzorn, line 19: Is there an NHS class that will do all this for us?

DigitalLearningSolutions.Web/Scripts/trackingSystem/centreCourseSetup.ts
```diff
@@ -20,20 +20,23 @@ function copyLaunchCourseLinkToClipboard(customisationId: string) {
...
 
-function copyTextToClipboard(textToCopy: string): boolean {
+function copyTextToClipboard(textToCopy: string): void {
   try {
-    navigator.clipboard.writeText(textToCopy);
-    return true;
+    navigator.clipboard.writeText(textToCopy).then(() => {
+      displaySuccessAlert(textToCopy);
+    }, () => {
+      displayFailureAlert(textToCopy);
```
livzorn, line 31: Will displayFailureAlert only occur is displaySuccessAlert fails?

DigitalLearningSolutions.Web/Styles/trackingSystem/courseSetup.scss
DanBloxham-sw: Good thinking. There is indeed for the :focus, so I've implemented that instead.

DigitalLearningSolutions.Web/Scripts/trackingSystem/centreCourseSetup.ts
DanBloxham-sw: displayFailureAlert only occurs if promise writeText() fails. I suppose it could be considered a little redundant here - if we removed this optional failure path, then the .then() should just throw an error in fail cases and that will get picked up by the try...catch, which would eventually display a failure.

I think it is necessary to keep the failure path on the promise in case someone has some security restrictions preventing access to the clipboard. However, I've realised there is a better way to deal with this code so we aren't relying on a try...catch. So I've refactored it to be a bit nicer.

livzorn (approved):
Looks good!

DigitalLearningSolutions.Web/Scripts/trackingSystem/centreCourseSetup.ts
```diff
@@ -20,21 +20,26 @@ function copyLaunchCourseLinkToClipboard(customisationId: string) {
...
+    }
+    return;
   }
+
+  navigator.clipboard.writeText(textToCopy).then(() => {
+    displaySuccessAlert(textToCopy);
+  },
+  () => {
+    displayFailureAlert(textToCopy);
+  });
```
stellake, line 42: That's odd.. I'd expect Promise syntax to be something like:
```
navigator.clipboard.writeText(textToCopy)
    .then(() => displaySuccessAlert(textToCopy))
    .catch(() => displayFailureAlert(textToCopy));
```

Also, I'm assuming clipboard doesn't work on IE and is just replaces with fallback?

DanBloxham-sw: First example I saw for Promise was as it was previously written, but your example looks neater so I've used that instead. And you are correct, IE has to use the fallback as navigator.clipboard doesn't exist in IE.

PR #525, opened 2021-07-30
Heedls 458 Chartist IE issues

JBsoftwire:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-458

### Description
The chartist chart was not displaying in IE. This turned out to be due to a problem parsing the data. However, once this problem was fixed it revealed that the label styling does not work in IE due to how it handles SVG text, rendering the labels illegible. After investigation, the solution adopted was to recreate as much of the styling as possible in JavaScript.

The solution has slight text alignment errors on tablet and desktop, and larger errors on mobile, but it was agreed with client in Show & Tell (30/07/21) that the use case of running IE in an extremely narrow window is marginal enough that it is not worth potentially significant development time to fix.

Rendering in IE should be specifically checked when future work is done on the usage charts (or any other charts).

### Screenshots
image
image
image

-----
### Developer checks
I have:
- [X] Written tests for the changes (controller, data services, services, view models etc) and manually tested my work with and without JavaScript. Full manual testing guidelines can be found here: https://softwiretech.atlassian.net/wiki/spaces/HEE/pages/6703648740/Testing
- [X] Updated/added documentation in Swiki and/or Readme.

### DIFF
--- a/DigitalLearningSolutions.Web/Scripts/trackingSystem/reports.ts
+++ b/DigitalLearningSolutions.Web/Scripts/trackingSystem/reports.ts
@@ -33,25 +33,32 @@ const options = {
 };
 
 request.onload = () => {
-  let response = request.response;
-  // IE11 does not support automatic parsing to JSON with XMLHttpRequest.responseType, so we need to manually parse the JSON string
+  let { response } = request;
+  // IE does not support automatic parsing to JSON with XMLHttpRequest.responseType
+  // so we need to manually parse the JSON string if not already parsed
   if (typeof request.response === 'string') {
     response = JSON.parse(response);
   }
   const data = constructChartistData(response);
-  // eslint-disable-next-line no-new
   const chart = new Chartist.Line('.ct-chart', data, options);
 
   chart.on('draw',
-    (foo: any) => {
-      const element = foo.element;
-      if (element.classes().indexOf('ct-horizontal') >= 0 && element.classes().indexOf('ct-label') >= 0) {
-        const xOrigin = Number(element.getNode().getAttribute("x"));
-        const yOrigin = element.getNode().getAttribute("y");
-        const width = Number(element.getNode().getAttribute("width"));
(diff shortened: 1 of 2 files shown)


DigitalLearningSolutions.Web/Scripts/trackingSystem/reports.ts
```diff
@@ -33,9 +33,35 @@ const options = {
...
+      const { element } = drawnElement;
+      // IE renders text in SVGs with 'text' tags that do not work with most CSS properties
+      // so we set the relevant attributes manually
+      if (element.getNode().tagName === 'text'
+        && element.classes().indexOf('ct-horizontal') >= 0
+        && element.classes().indexOf('ct-label') >= 0) {
+        const xOrigin = Number(element.getNode().getAttribute('x'));
+        const yOrigin = element.getNode().getAttribute('y');
+        const width = Number(element.getNode().getAttribute('width'));
+        const mediaQuery = window.matchMedia('(min-width: 641px)');
```
AlexJacksonDS, line 58: Might be worth putting a comment here that this should match the width of the NHS tablet breakpoint. Just in case it comes up in the future.

AlexJacksonDS (commented):
Overall looks OK to me, just an extra comment I think would be good to have.

Do these changes mean that we have the same styling for all browsers, or do these styling in javascript changes only affect IE?

JBsoftwire:
> Overall looks OK to me, just an extra comment I think would be good to have.
> 
> Do these changes mean that we have the same styling for all browsers, or do these styling in javascript changes only affect IE?

They affect any browser that uses SVG <text> tags for text in vector graphics, but IE is the only one I know of that does that. Maybe some other minor browsers but none of the major ones we check support for in development.

AlexJacksonDS (approved):

livzorn (approved):
Looks good!

PR #531, opened 2021-08-04
Heedls 437 edit course admin fields

livzorn:
### JIRA link
[_HEEDLS-437_](https://softwiretech.atlassian.net/browse/HEEDLS-437)

### Description
Adds a page to edit course admin field answers and a page to edit them in bulk. 

### Screenshots
localhost_44363_TrackingSystem_CourseSetup_100_AdminFields_1_Edit
localhost_44363_TrackingSystem_CourseSetup_100_AdminFields_1_Edit_Bulk
localhost_44363_TrackingSystem_CourseSetup_100_AdminFields_1_Edit (1)
localhost_44363_TrackingSystem_CourseSetup_100_AdminFields_1_Edit_Bulk (2)
localhost_44363_TrackingSystem_CourseSetup_100_AdminFields_1_Edit (2)
localhost_44363_TrackingSystem_CourseSetup_100_AdminFields_1_Edit_Bulk (1)

-----
### Developer checks
I have:

### DIFF
--- a/DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/EditCourseAdminFieldsAccessibilityTests.cs
+++ b/DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/EditCourseAdminFieldsAccessibilityTests.cs
@@ -25,6 +25,9 @@
             ValidatePageHeading("Configure answers in bulk");
             var bulkAdditionResult = new AxeBuilder(Driver).Analyze();
 
+            Driver.ClickButtonByText("Submit");
+            ValidatePageHeading("Edit course admin field");
+
             editPageResult.Violations.Should().BeEmpty();
             bulkAdditionResult.Violations.Should().BeEmpty();
         }
--- a/DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/EditCourseAdminFieldsAccessibilityTests.cs
+++ b/DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/EditCourseAdminFieldsAccessibilityTests.cs
@@ -1,0 +1,35 @@
+﻿namespace DigitalLearningSolutions.Web.AutomatedUiTests.AccessibilityTests
+{
+    using DigitalLearningSolutions.Web.AutomatedUiTests.TestHelpers;
+    using FluentAssertions;
+    using Selenium.Axe;
+    using Xunit;
+
+    public class EditCourseAdminFieldsAccessibilityTests : AccessibilityTestsBase
+    {
+        public EditCourseAdminFieldsAccessibilityTests(SeleniumServerFactory<Startup> factory) : base(factory) { }
+
+        [Fact]
+        public void EditRegistrationPrompt_journey_has_no_accessibility_errors()
+        {
+            // Given
+            Driver.LogUserInAsAdminAndDelegate(BaseUrl);
+            const string startUrl = "/TrackingSystem/CourseSetup/100/AdminFields/1/Edit";
+
+            // When
+            Driver.Navigate().GoToUrl(BaseUrl + startUrl);
+            ValidatePageHeading("Edit course admin field");
+            var editPageResult = new AxeBuilder(Driver).Analyze();
+
+            Driver.ClickButtonByText("Bulk edit");
+            ValidatePageHeading("Configure answers in bulk");
+            var bulkAdditionResult = new AxeBuilder(Driver).Analyze();
+
+            Driver.ClickButtonByText("Next");
+            ValidatePageHeading("Edit delegate registration prompt");
+
+            editPageResult.Violations.Should().BeEmpty();
+            bulkAdditionResult.Violations.Should().BeEmpty();
+        }
+    }
+}
--- a/DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/EditCourseAdminFieldsAccessibilityTests.cs
+++ b/DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/EditCourseAdminFieldsAccessibilityTests.cs
@@ -10,7 +10,7 @@
         public EditCourseAdminFieldsAccessibilityTests(SeleniumServerFactory<Startup> factory) : base(factory) { }
 
         [Fact]
-        public void EditRegistrationPrompt_journey_has_no_accessibility_errors()
+        public void EditAdminField_journey_has_no_accessibility_errors()
         {
             // Given
             Driver.LogUserInAsAdminAndDelegate(BaseUrl);
--- a/DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/EditCourseAdminFieldsAccessibilityTests.cs
+++ b/DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/EditCourseAdminFieldsAccessibilityTests.cs
@@ -25,9 +25,6 @@
             ValidatePageHeading("Configure answers in bulk");
             var bulkAdditionResult = new AxeBuilder(Driver).Analyze();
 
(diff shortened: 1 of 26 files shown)


DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/CourseSetup/AdminFieldsControllerTests.cs
```diff
@@ -51,7 +57,151 @@ public void AdminFields_returns_AdminFields_page_when_appropriate_course_found()
...
+                AssertNumberOfConfiguredAnswersOnView(result, 2);
+            }
+        }
+
+        [Test]
+        public void PostEditRegistrationPrompt_delete_removes_configured_answer()
+        {
+            // Given
+            var prompt = new CustomPrompt(1, "Test", "Test\r\nAnswer", false);
+            var model = new EditAdminFieldViewModel(prompt, 1);
```
AlexJacksonDS, line 126: I guess it's fine if the unit tests run, but I would have expected this to use the other constructor

DigitalLearningSolutions.Web/Views/TrackingSystem/CourseSetup/AdminFields/_NoConfiguredAnswers.cshtml
```diff
@@ -0,0 +1,5 @@
+﻿<p class="nhsuk-body-m">
+  There are no answers configured for this registration prompt.
```
AlexJacksonDS, line 2: This looks like it has been copied but the text not updated to be relevant to the current page. Either we need to use the same partial for both this and the registration prompts if we want the same text, or the text here needs updating. (I suspect the text here needs updating)

AlexJacksonDS (changes requested):
Mostly looks fine, a couple of minor comments.

You also need to set up the accessibility tests for the new pages. This will need to be done in the same way as EditCentreCustomPromptsAccessibilityTests (rather than the more simple BasicAccessibilityTests)

There is also the split of the CustomPromptsService into separate services for the different types of prompts (and then probably splitting the DataService too) because it is getting quite long that I mentioned on one of the other pull requests. This doesn't necessarily need doing now if it'll cause you issues with the next ticket you are working on. It's more of a nice to do at some point.

DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/CourseSetup/AdminFieldsControllerTests.cs
livzorn: Yeah I'm not sure why this constructor works here but not on the other ones, I've changed it to the other constructor for consistency.

DigitalLearningSolutions.Web/Views/TrackingSystem/CourseSetup/AdminFields/_NoConfiguredAnswers.cshtml
livzorn: Ah good catch, just an oversight on my part. I've changed it to say "There are no answers configured for this admin field."

livzorn:
@AlexJacksonDS I've set up the accessibility tests, let me know if it looks okay!

I agree the CustomPromptsService and CustomPromptsDataService are getting long. I will split them in my next ticket 439 as I've modified both those files on that branch.

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/EditCourseAdminFieldsAccessibilityTests.cs
```diff
@@ -0,0 +1,32 @@
...
+            // When
+            Driver.Navigate().GoToUrl(BaseUrl + startUrl);
+            ValidatePageHeading("Edit course admin field");
+            var editPageResult = new AxeBuilder(Driver).Analyze();
+
+            Driver.ClickButtonByText("Bulk edit");
+            ValidatePageHeading("Configure answers in bulk");
+            var bulkAdditionResult = new AxeBuilder(Driver).Analyze();
+
+            editPageResult.Violations.Should().BeEmpty();
```
AlexJacksonDS, line 28: Need a variation of the lines you deleted in the latest commit added back in here, checking the page title matches the Edit page title. (I presume this was removed to make the test pass since it was using the registration prompts title).

The purpose of the extra chunk was to verify that the submit button on the Bulk page redirected you back to the correct place, we've lost that check for the AdminFields journey by removing it. (There is an assertion in the ValidatePageHeading method).

AlexJacksonDS (changes requested):
Changes look fine, just need to bring back the final assertion in the accessibility test.

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/EditCourseAdminFieldsAccessibilityTests.cs
livzorn: @AlexJacksonDS Ah, I was confused about that chunk because the wording of my button on the bulk page is different, added it back in and it is passing.

Also, earlier I changed one field in the Customisations table in the remote test database (adding a CourseField1PromptID value to the customisation with ID 100), so the edit course admin field page would exist for the test to check. Is this okay and do I have to document this change somewhere?

AlexJacksonDS: The preferred way to do this would be for the tests to access a prompt with existing data that works for them. (I don't know if this exists in this case, but it is worth checking).

Otherwise, I think the process is as below, but there may be something extra I've missed:
- Add a script to the SQLScripts folder that will do whatever changes are needed
- Update the test database backup we use in the Zero to Hero
- Let everyone know this exists in slack
- (I assume tell Kevin at some point?)

AlexJacksonDS (approved):

JBsoftwire (commented):
A couple of minor comments/questions. No need to re-review if they work/don't work..

DigitalLearningSolutions.Web/Views/TrackingSystem/CourseSetup/AdminFields/BulkAdminFieldAnswers.cshtml
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    <h1 class="nhsuk-heading-xl">Configure answers in bulk</h1>
+
+    <form class="nhsuk-u-margin-bottom-3" method="post" novalidate asp-action="EditAdminFieldBulkPost">
+
+      <input type="hidden" asp-for="PromptNumber" />
+
+      <vc:text-area asp-for="OptionsString"
+                    label="Enter each response option on a separate line (max 1000 characters). The order of responses will be the order they appear on the dropdown for users."
```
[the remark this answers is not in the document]
livzorn: Oh yeah, that is a bit confusing - I changed the wording to (max 1000 characters in total). And there is a line underneath the textbox that says "You can enter up to 1000 characters" so I think it should be clear now.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CourseSetup/AdminFieldsController.cs
```diff
@@ -39,5 +51,225 @@ public IActionResult AdminFields(int customisationId)
...
+                    centreId,
+                    categoryId.Value
+                ).CourseAdminFields
+                .Single(cp => cp.CustomPromptNumber == promptNumber);
+
+            var data = TempData.Get<EditAdminFieldData>();
+
+            var model = data != null
+                ? data.EditModel!
+                : new EditAdminFieldViewModel(courseCustomPrompt, customisationId);
```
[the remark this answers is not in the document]
livzorn: Good idea, I've replaced it with null coalesce!

PR #540, opened 2021-08-11
HEEDLS-417 IE javascript event fix

AlexJacksonDS:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-417

### Description
IE fix for the javascript filtering since IE does stuff a different way for some reason.

### Screenshots
N/A

-----
### Developer checks
I have:

### DIFF
--- a/DigitalLearningSolutions.Web/Scripts/common.ts
+++ b/DigitalLearningSolutions.Web/Scripts/common.ts
@@ -4,3 +4,19 @@ export default function getPathForEndpoint(endpoint: string): string {
   const indexOfBaseUrl = currentPath.indexOf(endpointUrlParts[0]);
   return `${currentPath.substring(0, indexOfBaseUrl)}${endpoint}`;
 }
+
+/** This allows us to dispatch browser events in old IE and newer browsers */
+export const SendBrowserAgnosticEvent = <T extends HTMLElement | Window>(elem: T, eventName: string): Event => {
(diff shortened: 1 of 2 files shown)


DanBloxham-sw (approved):
Looks good to me.

DigitalLearningSolutions.Web/Scripts/common.ts
```diff
@@ -4,3 +4,22 @@ export default function getPathForEndpoint(endpoint: string): string {
   const indexOfBaseUrl = currentPath.indexOf(endpointUrlParts[0]);
   return `${currentPath.substring(0, indexOfBaseUrl)}${endpoint}`;
 }
+
+/** This allows us to dispatch browser events in old IE and newer browsers */
+export const SendBrowserAgnosticEvent = <T extends HTMLElement | Window>(
```
stellake, line 9: Nice just two quick questions:
Why is the type union of Window?
Is there a reason this is an arrow function rather than a normal function (the naming is function-like and the getPathForEndpoint above is a function)?

AlexJacksonDS: Because it totally wasn't taken from StackOverflow 🙂 
I assume it is because you can call T.dispatchEvent() on both HTMLElements and Window, but they don't have a nice common type to use.
I'll convert it to a regular function.

PR #557, opened 2021-08-20
HEEDLS-578 Add handling of FailedLoginCount for admin accounts

AlexJacksonDS:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-578

### Description
Adds logic to handle locked admin accounts, and increase the Failed Login Count on unsuccessful logins. New view to display when an admin tries to access a locked account. The structure of this page matches the AccountNotApproved view.

### Screenshots
image
image
image

-----
### Developer checks
I have:

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
@@ -4,6 +4,7 @@
     using System.Linq;
     using System.Security.Claims;
     using System.Threading.Tasks;
+    using DigitalLearningSolutions.Data.DataServices.UserDataService;
     using DigitalLearningSolutions.Data.Models.User;
     using DigitalLearningSolutions.Data.Services;
     using DigitalLearningSolutions.Data.Tests.TestHelpers;
@@ -26,16 +27,18 @@
         private ILoginService loginService = null!;
         private ISessionService sessionService = null!;
         private IUserService userService = null!;
+        private IUserDataService userDataService = null!;
 
         [SetUp]
         public void SetUp()
         {
             loginService = A.Fake<ILoginService>();
             userService = A.Fake<IUserService>();
+            userDataService = A.Fake<IUserDataService>();
             sessionService = A.Fake<ISessionService>();
             logger = A.Fake<ILogger<LoginController>>();
 
-            controller = new LoginController(loginService, userService, sessionService, logger)
+            controller = new LoginController(loginService, userService, sessionService, userDataService, logger)
                 .WithDefaultContext()
                 .WithMockUser(false)
                 .WithMockTempData()
@@ -61,7 +64,7 @@
         public void Index_should_redirect_if_user_is_authenticated()
         {
             // Given
-            var controllerWithAuthenticatedUser = new LoginController(loginService, userService, sessionService, logger)
+            var controllerWithAuthenticatedUser = new LoginController(loginService, userService, sessionService, userDataService, logger)
                 .WithDefaultContext()
                 .WithMockUser(true);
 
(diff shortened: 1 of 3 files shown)


DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
```diff
@@ -497,6 +506,104 @@ public async Task Leading_trailing_whitespaces_in_username_are_ignored()
...
+
+            // When
+            await controller.Index(LoginTestHelper.GetDefaultLoginViewModel());
+
+            // Then
+            A.CallTo(() => userDataService.UpdateAdminUserFailedLoginCount(7, 1)).MustHaveHappened();
+        }
+
+        [Test]
+        public async Task Admin_with_incorrect_password_increases_FailedLoginCount_and_returns_locked_view()
```
DanBloxham-sw, line 526: Could this test be renamed to make it obvious that the locked view is only returned when the locked threshold is exceeded?

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
```diff
@@ -497,6 +506,104 @@ public async Task Leading_trailing_whitespaces_in_username_are_ignored()
...
+            result.Should().BeRedirectToActionResult().WithActionName("AccountLocked");
+        }
+
+        [Test]
+        public async Task Admin_already_locked_increases_FailedLoginCount_and_returns_locked_view()
+        {
+            // Given
+            var adminUser = UserTestHelper.GetDefaultAdminUser(failedLoginCount: 6);
+            A.CallTo(() => userService.GetUsersByUsername(A<string>._)).Returns((adminUser, new List<DelegateUser>()));
+            A.CallTo(() => loginService.VerifyUsers(A<string>._, A<AdminUser>._, A<List<DelegateUser>>._))
```
DanBloxham-sw, line 548: Unless I'm mistaken, I think this A.CallTo can be removed.

DanBloxham-sw (commented):
Looks good, couple of very minor changes.

DanBloxham-sw (approved):

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -67,14 +67,15 @@ public async Task<IActionResult> Index(LoginViewModel model)
...
             {
                 userDataService.UpdateAdminUserFailedLoginCount(adminUser.Id, adminUser.FailedLoginCount + 1);
                 return RedirectToAction("AccountLocked", new { failedCount = adminUser.FailedLoginCount + 1 });
             }
 
-            var (verifiedAdminUser, verifiedDelegateUsers) =
-                loginService.VerifyUsers(model.Password!, adminUser, delegateUsers);
             if (verifiedAdminUser == null && verifiedDelegateUsers.Count == 0)
             {
                 if (adminUser != null && verifiedAdminUser == null)
```
DanBloxham-sw, line 81: Just realised we don't need this ".. && verifiedAdminUser == null" here as it is implicit from the encapsulating if statement.

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -67,14 +67,15 @@ public async Task<IActionResult> Index(LoginViewModel model)
                 return View("Index", model);
             }
 
-            if (adminUser != null && adminUser.IsLocked)
+            var (verifiedAdminUser, verifiedDelegateUsers) =
+                loginService.VerifyUsers(model.Password!, adminUser, delegateUsers);
+
+            if (adminUser != null && adminUser.IsLocked && verifiedDelegateUsers.Count == 0)
```
DanBloxham-sw, line 73: I'm not 100% sure, but I think this allows the user to effectively bypass the lock in the following example:

1. A user has an admin account and at least one delegate account. These accounts share the same password.
2. Only the admin account becomes locked from trying to log in to that account 5 times with the wrong password, while the delegate account(s) remain untouched. (This might get counteracted if the Delegate Account locking is implemented, but I think even then users can specifically log in to a single admin account, even if it has the same centre/password as a delegate account, by using the admin-specific Login column)
3. They then try to log in using the email address tied to all accounts, and they use the correct password.
4. When it reaches this if statement, verifiedDelegateUsers.Count > 0, so this is skipped.
5. Seeing as how the password matched their Admin account password, verifiedAdminUser is not null here.
6. Assuming the rest of their account details are suitable for logging in, I think they will be logged in as with their Delegate + Admin account (or at least it will be an option on the Choose a Centre Screen).

Does that example hold water? Or am I missing something here? If this is the case, the unit test will probably need changing as well.

DanBloxham-sw (commented):
Potential issue bypassing the admin account lock, but I might be forgetting how the system works.

DigitalLearningSolutions.Web/Controllers/LoginController.cs
AlexJacksonDS: Yes I think that might be true, but it might be worth having a quick chat about it in the morning to make sure we get the correct solution.

I think it might be as simple as 
```
if (adminUser != null && adminUser.IsLocked && verifiedDelegateUsers.Count != 0) 
{
    verifiedAdminUser = null;
}
```

DanBloxham-sw: Happy to have a chat whenever you are ready. I think that solution would work fine, don't know if its worth nesting it inside the above commented line and making it something like
```
if (adminUser != null && adminUser.IsLocked) 
{
    if (verifiedDelegateUsers.Count != 0) 
    {
        verifiedAdminUser = null;
    } else {
        // the other code for redirecting to failed login page
    }
}
```

DigitalLearningSolutions.Data/Services/LoginService.cs
```diff
@@ -55,7 +55,10 @@ List<DelegateUser> unverifiedDelegateUsers
             }
 
             var adminUserAssociatedWithDelegate = userDataService.GetAdminUserByUsername(delegateUser.EmailAddress);
-            return adminUserAssociatedWithDelegate?.CentreId == delegateUser.CentreId &&
+            
+            return adminUserAssociatedWithDelegate != null &&
+                   !adminUserAssociatedWithDelegate.IsLocked &&
```
DanBloxham-sw, line 60: I think these two lines can be can be simplified to 
`adminUserAssociatedWithDelegate?.IsLocked == false.`

DanBloxham-sw (approved):
Looks good. Minor comment about potential simplification, but doesn't need re-review from me.

SteveJacksonSoft (changes requested):
Think that's everything, happy wrangling!

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
```diff
@@ -497,6 +506,164 @@ public async Task Leading_trailing_whitespaces_in_username_are_ignored()
...
+            var result = await controller.Index(LoginTestHelper.GetDefaultLoginViewModel());
+
+            // Then
+            A.CallTo(() => userDataService.UpdateAdminUserFailedLoginCount(7, 0)).MustHaveHappened();
+            result.Should().BeRedirectToActionResult()
+                .WithControllerName("Home").WithActionName("Index");
+        }
+
+        [Test]
+        public async Task Admin_with_correct_password_and_no_failed_logins_logs_in_without_resetting_FailedLoginCount()
```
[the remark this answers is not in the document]
AlexJacksonDS: Technically no, other than an unnecessary call to update the record from 0 to 0. With the discussed refactor of this logic, this test becomes redundant anyway.

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
```diff
@@ -497,6 +506,164 @@ public async Task Leading_trailing_whitespaces_in_username_are_ignored()
...
+            // When
+            var result = await controller.Index(LoginTestHelper.GetDefaultLoginViewModel());
+
+            // Then
+            A.CallTo(() => userDataService.UpdateAdminUserFailedLoginCount(7, 7)).MustHaveHappened();
+            result.Should().BeRedirectToActionResult().WithActionName("AccountLocked");
+        }
+
+        [Test]
+        public async Task Unverified_locked_admin_with_verified_delegate_logs_into_delegate_only_and_doesnt_increment_FailedLoginCount()
```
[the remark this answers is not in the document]
AlexJacksonDS: This is already tested by this unit test with the A.CallTo(...IncrementCount...).MustNotHaveHappened(); (And still will be with the updated UserService version of the test).

DigitalLearningSolutions.Data.Tests/Services/UserServiceTests.cs
```diff
@@ -458,5 +458,48 @@ public void NewEmailAddressIsValid_returns_true_for_delegate_only_with_unchanged
             A.CallTo(() => userDataService.GetAdminUserByEmailAddress(email)).MustNotHaveHappened();
             A.CallTo(() => userDataService.GetDelegateUsersByEmailAddress(email)).MustNotHaveHappened();
         }
+
+        [Test]
+        public void ResetFailedLoginCount_resets_count()
+        {
+            // Given
+            var adminUser = UserTestHelper.GetDefaultAdminUser(failedLoginCount: 4);
+            A.CallTo(() => userDataService.UpdateAdminUserFailedLoginCount(adminUser.Id, 0)).DoesNothing();
```
SteveJacksonSoft, line 467: Does this mock need to be set up?

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
```diff
@@ -514,28 +524,31 @@ public async Task Admin_with_incorrect_password_increases_FailedLoginCount()
             A.CallTo(() => userService.GetUsersByUsername(A<string>._)).Returns((adminUser, new List<DelegateUser>()));
             A.CallTo(() => loginService.VerifyUsers(A<string>._, A<AdminUser>._, A<List<DelegateUser>>._))
                 .Returns(new UserAccountSet());
+            A.CallTo(() => userService.IncrementFailedLoginCount(adminUser)).DoesNothing();
```
SteveJacksonSoft, line 527: Does this mock need to be set up? Similar in other tests.

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -70,38 +67,38 @@ public async Task<IActionResult> Index(LoginViewModel model)
             var (verifiedAdminUser, verifiedDelegateUsers) =
                 loginService.VerifyUsers(model.Password!, adminUser, delegateUsers);
 
-            if (adminUser != null && adminUser.IsLocked)
+            var adminAccountFailedVerification = adminUser != null && verifiedAdminUser == null;
+            var adminAccountIsLocked = adminUser != null && adminUser.IsLocked ||
+                                       adminAccountFailedVerification && adminUser!.FailedLoginCount == 4;
+            var shouldLogIntoDelegateAccount = verifiedDelegateUsers.Any();
```
SteveJacksonSoft, line 73: I know I came up with these names, but looking at where they're used I think they could be improved.

I think this should be changed to `delegateAccountVerificationSuccessful`, and I'd change `adminAccountFailedVerification` to `adminAccountVerificationAttemptedAndFailed` to make it clear that it's not just "There's no admin account to log into". Does that sound reasonable?

SteveJacksonSoft (commented):
Nice, just some name changes and possible test simplifications

DigitalLearningSolutions.Web/Controllers/LoginController.cs
AlexJacksonDS: yeah those sound reasonable to me.

DigitalLearningSolutions.Data.Tests/Services/UserServiceTests.cs
AlexJacksonDS: I'm a lot more used to strict mocking from previous clients, never remember which I can drop safely.

PR #561, opened 2021-08-23
HEEDLS-579 Improve screen reader performance on Choose a Centre page

ibrahimmunir14:
### JIRA link
[HEEDLS-579](https://softwiretech.atlassian.net/browse/HEEDLS-579)

### Description
Screen reader - read roles as comma-separated values ("Admin, Delegate" instead of "Admin Delegate")
Screen reader - include centre name in choose button ("Choose - <centre name>")

### Screenshots
N/A

-----
### Developer checks
I have:


DigitalLearningSolutions.Web/Views/Login/ChooseACentre.cshtml
```diff
@@ -44,17 +44,18 @@
               <span class="nhsuk-table-responsive__heading">Roles: </span>
               @if (centreRow.IsAdmin)
               {
-                <strong class="nhsuk-tag nhsuk-tag--blue centre-role-tag">Admin</strong>
+                <strong class="nhsuk-tag nhsuk-tag--blue centre-role-tag" aria-hidden="true">Admin</strong>
               }
```
AlexJacksonDS, line 48: This approach doesn't really solve the problem in the correct way. We should be trying to achieve this by fixing the html structure rather than hiding things from the screen reader.

The reason these are being read as a single string is because the strong elements doesn't cause a break in content, it's for highlighting content as bold typically. The same is true for span elements.

The correct solution to this would be to either add a div surrounding the strong element like the tags on the expanders in the search/filter pages as a div will cause a pause or to add a visually hidden comma in a span between the two strong elements.
The way this is done for the tags on the various searchable cards is to put a div around the

DigitalLearningSolutions.Data/Models/User/CentreUserDetails.cs
```diff
@@ -1,18 +1,40 @@
-﻿namespace DigitalLearningSolutions.Data.Models.User
```
AlexJacksonDS, line 1: Are the line endings in this file correct?

AlexJacksonDS (changes requested):
The centre name on the choose button is fine, but the approach for the tags is wrong. I'd expect the only file changed to be the view, there is no need for any other code changes as incorrect reading by the screen reader is strictly a html structure issue.

DigitalLearningSolutions.Data/Models/User/CentreUserDetails.cs
ibrahimmunir14: It was LF, formatter made it CRLF when I was editing the file.

DigitalLearningSolutions.Web/Views/Login/ChooseACentre.cshtml
ibrahimmunir14: After our conversation, Stella said it was preferred to have the tags read separately, so I went for the divs. That "blank" issue seems to have fixed itself somehow after my laptop crashed and I rebooted 🤷‍♂️

AlexJacksonDS: Strange, we know it should work OK from the card tags, so I guess we'll just assume it is a one off.

AlexJacksonDS (approved):
Looks OK, have you checked the display: flex works fine in IE for reasonable window sizes? It only has partial support/is a bit buggy.

ibrahimmunir14:
> Looks OK, have you checked the display: flex works fine in IE for reasonable window sizes? It only has partial support/is a bit buggy.

It seems okay.

stellake (approved):
Looks good! 👍

PR #570, opened 2021-09-01
HEEDLS-519 - Change system notification cookie management

DanBloxham-sw:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-519

### Description
Updated the SystemNotificationController so that the cookie set when first visiting the page (Show Later link no longer needs to be clicked, allowing the user to immediately return to the dashboard). Also deleted the cookie when no more unacknowledged notifications exist.

### Screenshots
N/A

-----
### Developer checks
I have:

### DIFF
--- a/DigitalLearningSolutions.Web/Controllers/TrackingSystem/Centre/SystemNotificationsController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/TrackingSystem/Centre/SystemNotificationsController.cs
@@ -33,6 +33,19 @@
             var adminId = User.GetAdminId()!.Value;
             var unacknowledgedNotifications =
                 systemNotificationsDataService.GetUnacknowledgedSystemNotifications(adminId).ToList();
+
+            if (unacknowledgedNotifications.Count == 0)
+            {
+                if (Request.Cookies.HasSkippedNotificationsCookie(adminId))
+                {
+                    Response.Cookies.DeleteSkipSystemNotificationCookie();
+                }
+            }
+            else
+            {
+                Response.Cookies.SetSkipSystemNotificationCookie(adminId, clockService.UtcNow);
+            }
+
             var model = new SystemNotificationsViewModel(unacknowledgedNotifications, page);
             return View(model);
         }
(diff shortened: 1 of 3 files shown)


AlexJacksonDS (commented):
I can't raise against the lines, but does this change not make the SkipNotifications action redundant?

DanBloxham-sw:
In a most cases, SkipNotifications is pretty redundant. But if there are unacknowledged notifications, skipping will reset the expiry date. It will also set the cookie if the user visits the page when no notifications are present, in case the user definitely doesn't want to see any new notifications in the next 24 hours. But these are so minor that we could get rid of them if we think its unnecessary.

AlexJacksonDS:
It's not doing anything that the else branch isn't doing, and I assume the only page linking to the Skip action is the notification page which will have literally just set it in the else. I think the behaviour you've just described would be unexpected. If a user has no notification and then gets some, I think they should be forced to view them the next time.

AlexJacksonDS (approved):

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Centre/SystemNotificationsController.cs
```diff
@@ -33,31 +33,30 @@ public IActionResult Index(int page = 1)
...
+            {
+                if (Request.Cookies.HasSkippedNotificationsCookie(adminId))
+                {
+                    Response.Cookies.DeleteSkipSystemNotificationCookie();
+                }
+            }
+            else
+            {
+                Response.Cookies.SetSkipSystemNotificationCookie(adminId, clockService.UtcNow);
+            }
```
stellake, line 47: Could this be simplified with:

```
if (unacknowledgedNotifications.Count > 0) {
    Response.Cookies.SetSkipSystemNotificationCookie(adminId, clockService.UtcNow);
} else if (HasCookie) {
   Response.Cookies.DeleteSkipSystemNotificationCookie();
}

// continue
```

stellake (approved):
One minor improvement suggestion to avoid double "if" blocks, but no need for rereview :)

PR #572, opened 2021-09-01
HEEDLS-577 Add no store attribute for cache control

AlexJacksonDS:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-577

### Description
Add an ActionFilterAttribute to add Cache-Control 'no-store, max-age=0' headers to pages where the attribute is set against their action.

Note that this is likely to be overruled for some pages as there is some built in asp.net(?) functionality setting the header to no-store, no-cache on pages with forms. (I've made a note of this on the ticket too)

### Screenshots
image

-----
### Developer checks
I have:

### DIFF
--- a/DigitalLearningSolutions.Web/Attributes/NoStoreAttribute.cs
+++ b/DigitalLearningSolutions.Web/Attributes/NoStoreAttribute.cs
@@ -1,0 +1,12 @@
+﻿namespace DigitalLearningSolutions.Web.Attributes
+{
+    using Microsoft.AspNetCore.Mvc.Filters;
+
+    public class NoStoreAttribute : ActionFilterAttribute
+    {
+        public override void OnActionExecuted(ActionExecutedContext context)
+        {
+            context.HttpContext.Response.Headers.Add("Cache-Control", "no-store, max-age=0");
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


JBsoftwire (approved):
This looks good to me.

DigitalLearningSolutions.Web/Attributes/NoStoreAttribute.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace DigitalLearningSolutions.Web.Attributes
+{
+    using Microsoft.AspNetCore.Mvc.Filters;
+
+    public class NoStoreAttribute : ActionFilterAttribute
```
SteveJacksonSoft, line 5: I'd be inclined to call it something like `NoCaching` instead of `NoStore` for readability - `NoStore` is a bit more vague and developers might not work out what it's doing when they see it on an endpoint so quickly.
I get why you haven't gone for `NoCache`, but I think `NoCaching` or `NotCacheable` are sufficiently different from the "no-cache" directive not to cause confusion.

SteveJacksonSoft (changes requested):
Neat! Just think the attribute name could be improved.

SteveJacksonSoft (approved):
Nice, very neat. Merge away!

PR #575, opened 2021-09-03
Heedls 581 resolve account linking issues

AlexJacksonDS:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-581

### Description
Resolve outstanding login issues by refactoring the login process.

### Screenshots
N/A

-----
### Developer checks
I have:


DigitalLearningSolutions.Data.Tests/Services/LoginServiceTests.cs
```diff
@@ -1,239 +1,461 @@
...
+            AttemptLogin_find_multiple_delegates_and_returns_choose_a_centre_login_result()
         {
             // Given
-            var associatedAdminUser = UserTestHelper.GetDefaultAdminUser(centreId: 5);
-            A.CallTo(() => userDataService.GetAdminUserByUsername(A<string>._)).Returns(associatedAdminUser);
-            A.CallTo(() => cryptoService.VerifyHashedPassword(A<string>._, A<string>._)).Returns(true);
+            var delegateUser = UserTestHelper.GetDefaultDelegateUser();
+            var secondDelegateUser = UserTestHelper.GetDefaultDelegateUser(2, 3);
+            var delegateUsers = new List<DelegateUser> { delegateUser, secondDelegateUser };
+            A.CallTo(() => userService.GetUsersByUsername(Username))
```
DanBloxham-sw, line 382: The further we get through the AttemptLogin function, the more CallTo(...) we have to set up. Some of these can make the tests feel a bit cluttered and hard to read. How feasible would it be to make one or more private methods such as "GivenSuccessfulLoginUpToInitialPasswordVerification" that set up a bunch of calls necessary to reach that specific step (e.g. Initial password verification where VerifyUsers is called) by taking in the expected AdminUsers and DelegateUsers and setting up all the necessary CallTo()? Might be able to make several that use each other for multiple stages, and should help squash the tests down a bit.

DanBloxham-sw (commented):
Number of comments and questions.

DigitalLearningSolutions.Data/Services/LoginService.cs
```diff
@@ -1,65 +1,178 @@
...
+                return new LoginResult(LoginAttemptResult.InvalidUsername);
+            }
+
+            var (verifiedAdminUsers, verifiedDelegateUsers) = userVerificationService.VerifyUsers(
+                password,
+                unverifiedAdminUsers,
+                unverifiedDelegateUsers
+            );
+
+            var verifiedAdminUser = verifiedAdminUsers.SingleOrDefault();
```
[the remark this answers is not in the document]
AlexJacksonDS: SingleOrDefault will throw an exception if 2+ are found (in contrast to FirstOrDefault which won't).

And yeah, in theory Logins should be unique, but there is nothing to guarantee this currently, hence the change.

DigitalLearningSolutions.Data/Services/LoginService.cs
```diff
@@ -1,65 +1,178 @@
...
+
+            var adminAccountIsAlreadyLocked = doWeKnowWhichAdminAccountToRecordFailureAgainst &&
+                                              unverifiedAdminUsers.Single().IsLocked;
+            var adminAccountHasJustBecomeLocked = doWeKnowWhichAdminAccountToRecordFailureAgainst &&
+                                                  unverifiedAdminUsers.Single().FailedLoginCount == 4 &&
+                                                  adminAccountVerificationAttemptedAndFailed;
+
+            var adminAccountIsLocked = adminAccountIsAlreadyLocked || adminAccountHasJustBecomeLocked;
+            var delegateAccountVerificationSuccessful = verifiedDelegateUsers.Any();
+            var shouldIncreaseFailedLoginCount =
```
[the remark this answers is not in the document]
AlexJacksonDS: This is as per https://softwiretech.atlassian.net/browse/HEEDLS-578. If someone has admin + delegate accounts that can be found with the same username, if the password matches the delegate account but not the admin account, login count should not increase.

DigitalLearningSolutions.Data/Services/LoginService.cs
```diff
@@ -1,65 +1,178 @@
...
+            }
+
+            var (verifiedLinkedAdmin, verifiedLinkedDelegates) = GetVerifiedLinkedAccounts(
+                password,
+                approvedVerifiedDelegates,
+                verifiedAdminUser
+            );
+
+            var adminUserToLoginIfCentreActive = new[] { verifiedAdminUser, verifiedLinkedAdmin }
+                .Where(au => au != null).Distinct().SingleOrDefault();
```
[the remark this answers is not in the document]
AlexJacksonDS: They should either be the same or null, but this handles the case they are not. 
If we have 2 identical admins, but do not call Distinct before SingleOrDefault, it will cause an exception to occur.

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -186,16 +185,26 @@ string password
...
             }
 
             var (adminUser, delegateUsers) = GetUsersByEmailAddress(signedInEmailIfAny);
 
-            return loginService.VerifyUsers(password, adminUser, delegateUsers);
+            var adminUsers = adminUser != null
+                ? new List<AdminUser> { adminUser }
+                : new List<AdminUser>();
+
+            var temp = userVerificationService.VerifyUsers(password, adminUsers, delegateUsers);
```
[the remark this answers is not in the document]
AlexJacksonDS: Ah whoops, this was a debugging temp thing

DigitalLearningSolutions.Data/Services/UserVerificationService.cs
```diff
@@ -0,0 +1,86 @@
...
+        public AdminUser? GetVerifiedAdminUserAssociatedWithDelegateUser(DelegateUser? delegateUser, string password)
+        {
+            if (string.IsNullOrWhiteSpace(delegateUser?.EmailAddress))
+            {
+                return null;
+            }
+
+            var adminUserAssociatedWithDelegate = userDataService.GetAdminUserByEmailAddress(delegateUser.EmailAddress);
+
+            var isSuitableAdmin = adminUserAssociatedWithDelegate?.CentreId == delegateUser.CentreId &&
```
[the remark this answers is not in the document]
AlexJacksonDS: I think you're right that the logic is wrong here when going from the delegate users to an admin. I've checked with Steve what we want:
For delegates to admin, we only want to select an admin at the centre of the delegate accounts. The check is therefore incorrect, and should be if the found admin matches the centreId of ANY of the delegate users.
(Note that this only applies when logging in via CandidateNumber or AliasID, email will find the admin in step one)

DigitalLearningSolutions.Data/Services/UserVerificationService.cs
```diff
@@ -0,0 +1,86 @@
...
+            {
+                return new List<DelegateUser>();
+            }
+
+            var delegatesAssociatedWithAdmin = userDataService.GetDelegateUsersByEmailAddress(adminUser.EmailAddress!);
+
+            var suitableDelegates = delegatesAssociatedWithAdmin
+                .Where(du => du.Active)
+                .Where(du => du.Approved)
+                .Where(du => du.CentreId == adminUser.CentreId)
```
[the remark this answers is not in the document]
AlexJacksonDS: For the admin to delegates direction, this is also wrong. We should be getting all matching delegates by email regardless of what centre they are at. Will remove that check and combine the wheres

DigitalLearningSolutions.Data.Tests/Services/UserVerificationServiceTests.cs
```diff
@@ -0,0 +1,379 @@
...
+                adminUsers,
+                delegateUsers
+            );
+
+            // Then
+            Assert.IsEmpty(verifiedDelegateUsers);
+        }
+
+        [Test]
+        public void VerifyUsers_Returns_no_delegate_users_when_delegate_stored_password_is_empty()
```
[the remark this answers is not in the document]
AlexJacksonDS: Agreed, most of these tests are copied across from LoginServiceTests after refactoring out some methods to fix a circular dependency, but this test clearly serves no purpose.

DigitalLearningSolutions.Data/Services/UserVerificationService.cs
```diff
@@ -0,0 +1,89 @@
...
+
+            return suitableDelegates.ToList();
+        }
+
+        public AdminUser? GetVerifiedAdminUserAssociatedWithDelegateUsers(
+            List<DelegateUser> delegateUsers,
+            string password
+        )
+        {
+            var delegateEmail = delegateUsers.FirstOrDefault(du => du.EmailAddress != null)?.EmailAddress;
```
DanBloxham-sw, line 71: Is taking the first email address fine? Is there a situation that can occur where we have multiple delegate users with different emails here? I don't think there is - the only time where getting associated admin users would differ from the initial user retrieval would be if a candidate number was used to log in, in which case there would only be one delegate user anyway. Just worth double checking.

DigitalLearningSolutions.Data.Tests/Services/DiagnosticAssessmentDataServiceTests/DiagnosticAssessmentDataServiceNextTests.cs
```diff
@@ -1,328 +1,328 @@
...
-        public void Get_diagnostic_assessment_nextSection_skips_assessed_section_with_no_assessment_path()
-        {
-            using (new TransactionScope())
-            {
-                // Given
-                const int customisationId = 10820;
-                const int candidateId = 1;
-                const int sectionId = 104;
-
+﻿namespace DigitalLearningSolutions.Data.Tests.Services.DiagnosticAssessmentDataServiceTests
```
DanBloxham-sw, line 1: What's going on with the large number of supposed changes in this file?

DanBloxham-sw (commented):
Few quick points.

DigitalLearningSolutions.Data.Tests/Services/DiagnosticAssessmentDataServiceTests/DiagnosticAssessmentDataServiceNextTests.cs
AlexJacksonDS: Line endings, I did a replace all for //Given, //When and //Then. This file also had mixed line endings so I fixed that.

DigitalLearningSolutions.Data/Services/UserVerificationService.cs
AlexJacksonDS: The only place this is called is from within the login service when getting linked accounts. If the initial verified account set has multiple emails, we're throwing an exception (including when there is 1 email + null). It might become a problem if other methods start calling the same method (I had to move it into this new service to get rid of a circular dependency) so I'll just add a comment above.

DanBloxham-sw (approved):

DigitalLearningSolutions.Data.Tests/Services/UserVerificationServiceTests.cs
```diff
@@ -0,0 +1,340 @@
...
+            // Then
+            Assert.AreEqual(associatedAdminUser, returnedAdminUser);
+        }
+
+        [Test]
+        public void GetVerifiedDelegateUsersAssociatedWithAdminUser_Returns_empty_list_when_admin_is_null()
+        {
+            // When
+            var returnedDelegates = userVerificationService.GetVerifiedDelegateUsersAssociatedWithAdminUser(
+                null,
```
SteveJacksonSoft, line 245: Not a big deal, but in cases like this, I like having the test input value set in a `given` section - that makes it clear that the special case you're testing is where `adminUser = null` rather than, say, where `password = "password"`.

DigitalLearningSolutions.Data.Tests/Services/UserVerificationServiceTests.cs
```diff
@@ -0,0 +1,340 @@
...
+                "password"
+            );
+
+            // Then
+            Assert.IsNull(returnedAdminUser);
+        }
+
+        [Test]
+        public void
+            GetVerifiedAdminUserAssociatedWithDelegateUsers_Returns_nothing_when_admin_account_found_is_at_different_centre()
```
SteveJacksonSoft, line 198: Think the LoginService should be responsible for this really, and this test should move.

DigitalLearningSolutions.Data.Tests/Services/LoginServiceTests.cs
```diff
@@ -1,239 +1,556 @@
...
+                result.LoginAttemptResult.Should().Be(LoginAttemptResult.LogIntoSingleCentre);
+                result.LogInAdmin.Should().Be(adminUser);
+                result.LogInDelegates.Should().BeEmpty();
+            }
         }
 
         [Test]
         public void
-            GetVerifiedAdminUserAssociatedWithDelegateUser_Returns_nothing_when_no_admin_account_is_associated_with_delegate()
+            AttemptLogin_does_not_find_linked_admins_if_verified_admin_already_found_but_linked_delegates_are_found_and_returns_single_centre_login_result()
```
SteveJacksonSoft, line 251: This test name isn't generally true of our code - if I start with 1 admin and 2 delegates, I'll look for more linked admins, which is the behaviour we want.

I think an improvement would be to replace this test and the previous test with tests for these statements:
* AttemptLogin doesn't look for linked admins if there are no initial, approved, verified delegates.
* AttemptLogin returns single centre login result if 0 or 1 delegates are found

DigitalLearningSolutions.Data/Models/User/UserAccountSet.cs
```diff
@@ -5,34 +5,33 @@
...
     public class UserAccountSet
     {
-        public readonly AdminUser? AdminAccount;
+        public readonly List<AdminUser> AdminAccounts;
         public readonly List<DelegateUser> DelegateAccounts;
 
         public UserAccountSet() : this(null, null) { }
 
-        public UserAccountSet(AdminUser? adminAccount, IEnumerable<DelegateUser?>? delegateAccounts)
+        public UserAccountSet(List<AdminUser>? adminAccounts, IEnumerable<DelegateUser?>? delegateAccounts)
```
SteveJacksonSoft, line 13: I fancy you can split this in two.
Conceptually, the `UserAccountSet`, when structured as `(AdminUser?, List<DelegateUser>)`, was very different from `(List<AdminUser>, List<DelegateUser>)`, because it represented the set of accounts belonging to a particular user. So that's how it's used in ChangePasswordController (which also uses `Any()` and `GetUserRefs()`).

However, it's being used in the Login procedure as a wrapper around a tuple basically, because it's really working with the List<AU> and List<DU> separately and not using the methods. Perhaps it would make sense for the return value of AttemptLogin to contain a `UserAccountSet`, because at that point you do have a set of accounts belonging to a particular person.

The uses in the Login procedure are disjoint from the uses in the ChangePassword procedure, I think.

So I think it would be better to create a new type, something like `AccountCollection`, or even simply `AccountSet` (though that does imply having sets rather than lists), for the login procedure, then leave `UserAccountSet` as it was for the ChangePassword procedure to use.

SteveJacksonSoft (changes requested):
Impressive test coverage! Guess it comes with long methods.
Some things which I think could be improved.

DigitalLearningSolutions.Data.Tests/Services/LoginServiceTests.cs
AlexJacksonDS: The statement was correct for the case admin + delegates going into the account linking step. The call to get more admins only occurred when there was no already an admin (using ??). I've now refactored the private method due to the move of the centre check so that the call will always occur. However we will still always throw away any admin we find if we already have one. (They must be the same admin because of the Email uniqueness and previous checks)

DigitalLearningSolutions.Data/Models/User/UserAccountSet.cs
AlexJacksonDS: Reverted all the changes here back to AdminUser?. Kevin is OK with us throwing if we manage to get 2 unverified admins at step one, so I've reverted that step to take a first or default, which means all the Verify methods go back to taking a single admin.

DigitalLearningSolutions.Data/Services/LoginService.cs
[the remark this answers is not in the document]
AlexJacksonDS: Just going to trust the code then, and remove the check. Since we are taking verifiedAdminUser ?? verifiedLinkedAdmin in the get linked accounts method, we will no only have a single AdminUser? to deal with.

DigitalLearningSolutions.Data/Services/LoginService.cs
```diff
@@ -2,64 +2,170 @@
...
+                unverifiedAdminUsers,
+                unverifiedDelegateUsers
+            );
+
+            var verifiedAdminUser = verifiedAdminUsers.SingleOrDefault();
+
+            if (MultipleEmailsUsedAcrossAccounts(verifiedAdminUser, verifiedDelegateUsers))
+            {
+                throw new LoginWithMultipleEmailsException("Not all accounts have the same email");
+            }
```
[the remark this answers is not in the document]
AlexJacksonDS: Kevin is OK with us throwing if we manage to get 2 unverified admins at step one, so I've reverted that step to take a single or default, which means we will still get InvalidOperationExceptions, but not in the AttemptLogin method.

DigitalLearningSolutions.Data/Services/LoginService.cs
```diff
@@ -2,64 +2,170 @@
...
+
+            var adminAccountVerificationAttemptedAndFailed = unverifiedAdminUsers.Any() && verifiedAdminUser == null;
+            var adminAccountToRecordFailureAgainstIsKnown = unverifiedAdminUsers.Count == 1;
+
+            var adminAccountIsAlreadyLocked = adminAccountToRecordFailureAgainstIsKnown &&
+                                              unverifiedAdminUsers.Single().IsLocked;
+            var adminAccountHasJustBecomeLocked = adminAccountToRecordFailureAgainstIsKnown &&
+                                                  unverifiedAdminUsers.Single().FailedLoginCount == 4 &&
+                                                  adminAccountVerificationAttemptedAndFailed;
```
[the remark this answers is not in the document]
AlexJacksonDS: Now we are back to a single unverified admin at step 1, this logic should all be returned to as per the account locking ticket HEEDLS-578.

DigitalLearningSolutions.Data/Services/LoginService.cs
```diff
@@ -2,64 +2,170 @@
...
+
+            var verifiedAdminUser = verifiedAdminUsers.SingleOrDefault();
+
+            if (MultipleEmailsUsedAcrossAccounts(verifiedAdminUser, verifiedDelegateUsers))
+            {
+                throw new LoginWithMultipleEmailsException("Not all accounts have the same email");
+            }
+
+            var adminAccountVerificationAttemptedAndFailed = unverifiedAdminUsers.Any() && verifiedAdminUser == null;
+            var adminAccountToRecordFailureAgainstIsKnown = unverifiedAdminUsers.Count == 1;
```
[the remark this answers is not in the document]
AlexJacksonDS: Reverted to single unverified admin, so this no longer applies.

DigitalLearningSolutions.Data/Services/LoginService.cs
```diff
@@ -25,45 +25,39 @@ public LoginService(IUserService userService, IUserVerificationService userVerif
...
-            var adminAccountVerificationAttemptedAndFailed = unverifiedAdminUsers.Any() && verifiedAdminUser == null;
-            var adminAccountToRecordFailureAgainstIsKnown = unverifiedAdminUsers.Count == 1;
+            var adminAccountVerificationAttemptedAndFailed = unverifiedAdminUser != null && verifiedAdminUser == null;
 
-            var adminAccountIsAlreadyLocked = adminAccountToRecordFailureAgainstIsKnown &&
-                                              unverifiedAdminUsers.Single().IsLocked;
-            var adminAccountHasJustBecomeLocked = adminAccountToRecordFailureAgainstIsKnown &&
-                                                  unverifiedAdminUsers.Single().FailedLoginCount == 4 &&
+            var adminAccountIsAlreadyLocked = unverifiedAdminUser?.IsLocked == true;
+            var adminAccountHasJustBecomeLocked = unverifiedAdminUser?.FailedLoginCount == 4 &&
```
SteveJacksonSoft, line 49: This is still an issue - `adminAccountHasJustBecomeLocked` can be true when `shouldIncreaseFailedLoginCount` is false (when we have verified delegates), so the code is lying.

DigitalLearningSolutions.Data/Services/LoginService.cs
```diff
@@ -139,11 +135,18 @@ public LoginResult AttemptLogin(string username, string password)
...
-                                      userVerificationService.GetVerifiedAdminUserAssociatedWithDelegateUsers(
-                                          approvedVerifiedDelegates,
-                                          password
-                                      );
+            var verifiedAssociatedAdmin = userVerificationService.GetVerifiedAdminUserAssociatedWithDelegateUsers(
+                approvedVerifiedDelegates,
+                password
+            );
+
+            if (approvedVerifiedDelegates.All(du => du.CentreId != verifiedAssociatedAdmin?.CentreId))
```
SteveJacksonSoft, line 143: Can we have a comment here explaining why we discard the admin we've found in this case?

DigitalLearningSolutions.Data.Tests/Services/LoginServiceTests.cs
```diff
@@ -211,72 +198,67 @@ public void AttemptLogin_returns_inactive_centre_if_all_centres_are_inactive()
...
-                result.LogInDelegates.Should().BeEmpty();
+                result.Accounts.AdminAccount.Should().Be(null);
+                result.Accounts.DelegateAccounts.Should().BeEmpty();
             }
         }
 
         [Test]
         public void
-            AttemptLogin_does_not_find_linked_admins_if_verified_admin_already_found_and_no_linked_delegates_and_returns_single_centre_login_result()
+            AttemptLogin_finds_linked_delegate_and_returns_single_centre_login_result()
```
SteveJacksonSoft, line 208: It feels more natural to call this `AttemptLogin_returns_single_centre_login_result_when_single_delegate_found` - is there a reason not to call it that? Currently it sounds like we're testing that it finds a linked delegate, but also have a line in the test `GivenLinkedDelegateAccountsFound(<single delegate>)`.

DigitalLearningSolutions.Data.Tests/Services/LoginServiceTests.cs
```diff
@@ -211,72 +198,67 @@ public void AttemptLogin_returns_inactive_centre_if_all_centres_are_inactive()
...
-                result.LogInDelegates.Should().BeEmpty();
+                result.Accounts.AdminAccount.Should().Be(adminUser);
+                result.Accounts.DelegateAccounts.Should().BeEquivalentTo(linkedDelegateUsers);
             }
         }
 
         [Test]
         public void
-            AttemptLogin_does_not_find_linked_admins_if_verified_admin_already_found_but_linked_delegates_are_found_and_returns_single_centre_login_result()
+            AttemptLogin_finds_linked_delegates_and_returns_choose_a_centre_login_result()
```
SteveJacksonSoft, line 236: Similar to test above

DigitalLearningSolutions.Data/DataServices/UserDataService/AdminUserDataService.cs
```diff
@@ -51,7 +51,7 @@ ELSE cc.CategoryName
             return user;
         }
 
-        public List<AdminUser> GetAdminUsersByUsername(string username)
+        public AdminUser? GetAdminUserByUsername(string username)
```
SteveJacksonSoft, line 54: I think this method deserves an XML doc to explain that it will throw an InvalidOperationException if multiple admin users are found, and explaining that that will only happen if one admin's login matches another's email address, which is very unlikely.

SteveJacksonSoft (changes requested):

DigitalLearningSolutions.Data/Services/LoginService.cs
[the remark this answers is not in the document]
SteveJacksonSoft: Ah, I didn't realise that - did you just DM him?

DigitalLearningSolutions.Data/Models/LoginResult.cs
```diff
@@ -0,0 +1,30 @@
...
+            LoginAttemptResult result,
+            AdminUser? adminUser = null,
+            List<DelegateUser>? delegateUsers = null,
+            List<CentreUserDetails>? availableCentres = null
+        )
+        {
+            LoginAttemptResult = result;
+            LogInAdmin = adminUser;
+            LogInDelegates = delegateUsers ?? new List<DelegateUser>();
+            AvailableCentres = availableCentres;
```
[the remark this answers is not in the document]
SteveJacksonSoft: You can lose the ? on the property now.

DigitalLearningSolutions.Data/Services/LoginService.cs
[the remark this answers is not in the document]
AlexJacksonDS: From talking to Stella

DigitalLearningSolutions.Data.Tests/Services/LoginServiceTests.cs
AlexJacksonDS: I don't think the suggested names for these tests are better, we are testing the linked delegates cases. We start with a single admin then find linked delegate(s) and return the appropriate result now we have an admin + delegate(s).

i.e. we're testing exactly what you think it sounds like.

SteveJacksonSoft (approved):
Beautiful, merge away

PR #579, opened 2021-09-06
Heedls-438 Add course admin fields

livzorn:
### JIRA link
[_HEEDLS-438_](https://softwiretech.atlassian.net/browse/HEEDLS-438)

### Description
Created a new page to add a course admin field, and linked it to the "Add admin field" button on the Manage course admin fields page. The page has the same functionality as the  437 - Edit course admin fields page, with an additional Select field input. The page links to a Configure answers in bulk page identical to the one in 437.

I changed some of the existing admin field tests to use Customisation 100 rather than 1379 as 100 actually has admin fields configured.

Some of the files show the entire file as changed - I think this is because I corrected the line endings to CRLF and the File Encoding to UTF-8 BOM, but reviewers - please let me know if any files show up as different on your end.

### Screenshots
AddAdminField_desktop
AddAdminField_tablet
AddAdminField_mobile

-----
### Developer checks
I have:


DigitalLearningSolutions.Web/Views/TrackingSystem/CourseSetup/AdminFields/AddAdminField.cshtml
```diff
@@ -0,0 +1,92 @@
...
+                  value="@Model.CustomPromptId.ToString()"
+                  hint-text=""
+                  deselectable="false"
+                  css-class="nhsuk-u-width-one-half"
+                  default-option="Select a prompt name"
+                  select-list-options="@ViewBag.CoursePromptNameOptions" />
+
+  <div class="hidden-submit">
+    <button name="action" class="nhsuk-button" value="@AdminFieldsController.AddPromptAction">Add</button>
+  </div>
```
AlexJacksonDS, line 46: This button needs the changes from HEEDLS-488 adding to it so that it works in IE (see EditAdminField)

AlexJacksonDS (changes requested):
I've checked a few of the files showing as all changed, and they seem to be CRLF, so hopefully the rest are OK. There are a few strange empty files now, not sure where they've come from.

I think you have to many calls to the service methods that could easily be a single call in the controller. Note that by changing these, the service methods can likely become private, and will need to be unit tested differently.

There are also some changes to the hidden button from 488 you need to include now.

DigitalLearningSolutions.Data.Tests/DataServices/CustomPromptsDataServiceTests/CourseCustomPromptsDataServiceTests.cs
```diff
@@ -0,0 +1 @@
+﻿
```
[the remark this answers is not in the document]
livzorn: Oops, I think that's leftover from making the full split of the custom prompts data services. I've removed all those empty files.

DigitalLearningSolutions.Data/DataServices/CustomPromptsDataService/CustomPromptsDataService.cs
```diff
@@ -0,0 +1 @@
+﻿
```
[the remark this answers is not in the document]
livzorn: Deleted it

DigitalLearningSolutions.Data/Services/CustomPromptsService/CourseCustomPromptsService.cs
```diff
@@ -0,0 +1 @@
+﻿
```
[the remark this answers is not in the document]
livzorn: Deleted it

DigitalLearningSolutions.Web/Views/TrackingSystem/CourseSetup/AdminFields/AddAdminField.cshtml
livzorn: Ah yes, forgot about that. I've added the aria-hidden and tabindex

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CourseSetup/AdminFieldsController.cs
```diff
@@ -222,6 +323,60 @@ private void SetEditAdminFieldTempData(EditAdminFieldViewModel model)
...
+            );
+
+            var promptNumber = courseAdminFieldsService.GetNextPromptNumber(courseAdminFields);
+
+            if (courseAdminFieldsService.AddCustomPromptToCourse(
+                model.CustomisationId,
+                promptNumber,
+                model.CustomPromptId!.Value,
+                model.OptionsString
+            ))
```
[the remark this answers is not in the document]
livzorn: I've moved the logic into AddCustomPromptToCourse, made the GetNextPromptNumber method private, and adjusted the tests - let me know if that looks okay

AlexJacksonDS (approved):
Looks OK to me now.

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/CourseSetup/AddAdminFieldViewModel.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            CustomisationId = customisationId;
+            CustomPromptId = customPromptId;
+            OptionsString = options;
+            IncludeAnswersTableCaption = true;
+            Answer = answer;
+        }
+
+        [Required(ErrorMessage = "Select a prompt name")]
+
```
stellake, line 33: Minor, but unnecessary extra line

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CourseSetup/AdminFieldsController.cs
```diff
@@ -298,6 +501,26 @@ private static bool TryGetAnswerIndexFromDeleteAction(string action, out int ind
...
+        {
+            var coursePrompts = courseAdminFieldsService.GetCoursePromptsAlphabeticalList();
+            ViewBag.CoursePromptNameOptions =
+                SelectListHelper.MapOptionsToSelectListItems(coursePrompts, selectedId);
+        }
+
+        private void UpdateTempDataWithCoursePromptModelValues()
+        {
+            var data = TempData.Peek<AdminFieldAnswersViewModel>()!;
+            TempData.Set(data);
```
stellake, line 514: Why do we need this?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CourseSetup/AdminFieldsController.cs
```diff
@@ -146,6 +147,107 @@ BulkAdminFieldAnswersViewModel model
...
+
+            Response.Cookies.Append(
+                AddPromptCookieName,
+                id.ToString(),
+                new CookieOptions
+                {
+                    Expires = CookieExpiry
+                }
+            );
+            TempData.Set(addAdminFieldData);
```
stellake, line 168: Could pull this into a private method to make it easier to glance over the controller?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CourseSetup/AdminFieldsController.cs
```diff
@@ -146,6 +147,107 @@ BulkAdminFieldAnswersViewModel model
...
+
+        [HttpGet]
+        [Route("{customisationId:int}/AdminFields/Add")]
+        public IActionResult AddAdminField(int customisationId)
+        {
+            var addAdminFieldData = TempData.Peek<AddAdminFieldData>()!;
+
+            SetViewBagCoursePromptNameOptions(addAdminFieldData.AddModel.CustomPromptId);
+
+            var model = addAdminFieldData?.AddModel ?? new AddAdminFieldViewModel(customisationId);
```
stellake, line 181: When would we get here without the temp data?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CourseSetup/AdminFieldsController.cs
```diff
@@ -249,6 +398,43 @@ private IActionResult RemoveAdminFieldAndRedirect(int customisationId, int promp
...
+            {
+                UpdateTempDataWithCoursePromptModelValues();
+            }
+
+            return View(model);
+        }
+
+        private IActionResult AdminFieldAnswersPostAddPrompt(
+            AddAdminFieldViewModel model,
+            bool saveToTempData = false
```
stellake, line 411: When would you not save to temp data?

DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/CourseSetup/AdminFieldsControllerTests.cs
```diff
@@ -194,6 +191,212 @@ public void AdminFieldBulkPost_updates_temp_data_and_redirects_to_edit()
...
+            {
+                A.CallTo(
+                    () => courseAdminFieldsService.AddCustomPromptToCourse(
+                        100,
+                        101,
+                        0,
+                        1,
+                        "Test"
+                    )
+                ).MustHaveHappened();
```
stellake, line 293: We're already mocking this to return true/false in these two tests, which makes its inclusion in the assertion unnecessary (and also the tests will get more difficult to read/scan over). If you wanted you could make a separate test for checking that the mock gets called once with the correct values, but I think it would probably classify as "overtesting" the controller :)

DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/CourseSetup/AdminFieldsControllerTests.cs
```diff
@@ -194,6 +191,212 @@ public void AdminFieldBulkPost_updates_temp_data_and_redirects_to_edit()
...
+            // When
+            var result = controller.AddAdminField(1);
+
+            // Then
+            AssertAddAdminFieldViewModelIsExpectedModel(expectedPromptModel);
+            result.As<ViewResult>().Model.Should().BeOfType<AddAdminFieldViewModel>();
+        }
+
+        [Test]
+        public void AddAdminField_save_calls_service_and_redirects_to_index()
```
stellake, line 222: clears temp data and redirects to index?

DigitalLearningSolutions.Web.AutomatedUiTests/DigitalLearningSolutions.Web.AutomatedUiTests.csproj
```diff
@@ -1,31 +1,31 @@
...
-      <PrivateAssets>all</PrivateAssets>
-    </PackageReference>
-  </ItemGroup>
-
-    <ItemGroup>
-    <ProjectReference Include="..\DigitalLearningSolutions.Web\DigitalLearningSolutions.Web.csproj" />
-  </ItemGroup>
-
-</Project>
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
stellake, line 1: Is this just line endings?

DigitalLearningSolutions.Data/Services/CourseAdminFieldsService.cs
```diff
@@ -93,6 +145,17 @@ public string GetPromptName(int customisationId, int promptNumber)
...
 
+        private static int? GetNextPromptNumber(CourseAdminFields? courseAdminFields)
+        {
+            var existingPromptNumbers = courseAdminFields.AdminFields
+                .Select(c => c.CustomPromptNumber);
+
+            var promptNumbers = new List<int> { 1, 2, 3 };
+            var unusedPromptNumbers = promptNumbers.Except(existingPromptNumbers).ToList();
+            var nextPromptNumber = unusedPromptNumbers.Any() ? unusedPromptNumbers.Min() : (int?)null;
+            return nextPromptNumber;
```
stellake, line 156: Can return immediately rather than declaring the "next prompt number var"

DigitalLearningSolutions.Data.Tests/Services/CourseAdminFieldsServiceTests.cs
```diff
@@ -74,21 +78,103 @@ public void GetCustomPromptsWithAnswersForCourse_Returns_Populated_List_of_Custo
...
+            // Given
+            const string promptName = "Access Permissions";
+            A.CallTo(() => courseAdminFieldsDataService.GetCoursePromptsAlphabetical()).Returns
+                (new List<(int, string)> { (1, promptName) });
+            // When
+            var result = courseAdminFieldsService.GetCoursePromptsAlphabeticalList();
+
+            // Then
+            A.CallTo(() => courseAdminFieldsDataService.GetCoursePromptsAlphabetical()).MustHaveHappened();
+            result.Contains((1, promptName)).Should().BeTrue();
```
stellake, line 102: Can we just check that the result is what we return? Also, minor, but missing empty line before //when

DigitalLearningSolutions.Data.Tests/Services/CourseAdminFieldsServiceTests.cs
```diff
@@ -74,21 +78,103 @@ public void GetCustomPromptsWithAnswersForCourse_Returns_Populated_List_of_Custo
...
+            A.CallTo(() => courseAdminFieldsDataService.GetCourseAdminFields(100, 101, 0))
+                .Returns(
+                    CustomPromptsTestHelper.GetDefaultCourseAdminFieldsResult(
+                        "System Access Granted",
+                        "Test",
+                        false,
+                        "Priority Access",
+                        "",
+                        false,
+                        "Access Permissions"
```
stellake, line 144: Could we just pass in three default prompts to this helper to make this more intuitive? (so change the parameter types of CustomPromptsTestHelper.GetDefaultCourseAdminFieldsResult)

DigitalLearningSolutions.Data.Tests/DataServices/CourseAdminFieldsDataServiceTests.cs
```diff
@@ -75,6 +80,36 @@ public void GetPromptNameForCourseAndPromptNumber_returns_expected_prompt_name()
...
+            {
+                // Given
+                const string? options = "options";
+
+                // When
+                courseAdminFieldsDataService.UpdateCustomPromptForCourse(100, 3, 1, options);
+                var courseCustomPrompts = courseAdminFieldsDataService.GetCourseAdminFields(100, 101, 2);
+                var customPrompt = courseAdminFieldsDataService.GetCoursePromptsAlphabetical()
+                    .Single(c => c.Item1 == 1)
+                    .Item2;
```
stellake, line 97: Instead of Item1 & Item2, we should name these id and name

DigitalLearningSolutions.Data.Tests/DataServices/CourseAdminFieldsDataServiceTests.cs
```diff
@@ -75,6 +80,36 @@ public void GetPromptNameForCourseAndPromptNumber_returns_expected_prompt_name()
...
+
+                // Then
+                using (new AssertionScope())
+                {
+                    courseCustomPrompts.CustomField3Prompt.Should().BeEquivalentTo(customPrompt);
+                    courseCustomPrompts.CustomField3Mandatory.Should().BeFalse();
```
stellake, line 103: why do we still check mandatory?

DigitalLearningSolutions.Data.Tests/Services/CourseAdminFieldsServiceTests.cs
livzorn: We could, but then this helper will be constructed differently from the GetDefaultCentreCustomPromptsResult helper. Is that okay or should I change them both?

DigitalLearningSolutions.Data.Tests/DataServices/CourseAdminFieldsDataServiceTests.cs
stellake: In C# you can specify tuple field names - https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/value-tuples#tuple-field-names

We're doing this in various places, e.g. see GetUsersByUsername in userService :)

DigitalLearningSolutions.Web/Views/TrackingSystem/CourseSetup/AdminFields/BulkAdminFieldAnswers.cshtml
```diff
@@ -1,49 +1,57 @@
...
+    }
+
+    <h1 class="nhsuk-heading-xl">Configure answers in bulk</h1>
+
+    <form class="nhsuk-u-margin-bottom-3" method="post" novalidate asp-action="@(Model.IsAddPromptJourney ? "AddAdminFieldAnswersBulk" : "EditAdminFieldBulkPost")">
+
+      <input type="hidden" asp-for="IsAddPromptJourney" />
+      <input type="hidden" asp-for="PromptNumber" />
+
+      <vc:text-area asp-for="OptionsString"
```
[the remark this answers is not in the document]
stellake: For various inputs, we use e.g. `asp-for="@nameof(Model.ConfirmPassword)"` so if the "ConfirmPassword" field is renamed in the class, the IDE picks the change up in the view too. If we just hardcode it, it will need to be manually changed when field is renamed.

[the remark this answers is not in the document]
AlexJacksonDS: This isn't strictly true, we need to do this for the view components as they take a string, but we don't technically need to do it for default asp.net components (like the hidden inputs here) as they will produce build errors with an incorrect model property.

DigitalLearningSolutions.Web/Views/TrackingSystem/CourseSetup/AdminFields/AddAdminField.cshtml
```diff
@@ -0,0 +1,92 @@
...
+  <div class="nhsuk-grid-column-full">
+    @if (errorHasOccurred) {
+      <vc:error-summary order-of-property-names="@new []{ nameof(AddAdminFieldViewModel.CustomPromptId), nameof(AddAdminFieldViewModel.Answer) }" />
+    }
+
+    <h1 class="nhsuk-heading-xl">Add course admin field</h1>
+  </div>
+</div>
+
+<form method="post" novalidate asp-action="AddAdminField">
```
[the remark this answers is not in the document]
stellake: Not sure if I'm missing anything, but it doesn't look like `<form>` is wrapped with the row/column like the h1 above and I was just wondering what's the reason behind difference. 
Thinking about this, I think it makes sense to put things in one row/column unless we need to define a new section with several columns (due to layout). Investigating the NHS styles, it looks like row just sets margins and display type (block) and column sets padding, float and widths.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CourseSetup/AdminFieldsController.cs
```diff
@@ -146,6 +147,97 @@ BulkAdminFieldAnswersViewModel model
...
+            var model = addAdminFieldData?.AddModel ?? new AddAdminFieldViewModel(customisationId);
+
+            return View(model);
+        }
+
+        [HttpPost]
+        [Route("{customisationId}/AdminFields/Add")]
+        public IActionResult AddAdminField(int customisationId, AddAdminFieldViewModel model, string action)
+        {
+            if (!ModelState.IsValid)
```
stellake, line 180: What if the user wants to delete/add an answer but they haven't selected an admin field yet?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CourseSetup/AdminFieldsController.cs
```diff
@@ -146,6 +147,97 @@ BulkAdminFieldAnswersViewModel model
...
+        )
+        {
+            ValidateBulkOptionsString(model.OptionsString);
+            if (!ModelState.IsValid)
+            {
+                return View("BulkAdminFieldAnswers", model);
+            }
+
+            var addData = TempData.Peek<AddAdminFieldData>()!;
+            addData.AddModel!.OptionsString =
```
stellake, line 231: Should AddModel be non-nullable on the data?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CourseSetup/AdminFieldsController.cs
```diff
@@ -298,6 +481,20 @@ private static bool TryGetAnswerIndexFromDeleteAction(string action, out int ind
             return int.TryParse(action.Remove(0, DeleteAction.Length), out index);
         }
 
+        private void SetViewModelCoursePromptNameOptions(AddAdminFieldViewModel model)
+        {
+            var coursePrompts = courseAdminFieldsService.GetCoursePromptsAlphabeticalList();
+            model.CoursePromptNameOptions =
+                SelectListHelper.MapOptionsToSelectListItems(coursePrompts, model.CustomPromptId);
```
stellake, line 488: Just remembered why we didn't keep this on VM - it's because we don't want to save all these options on temp data (as we need to refetch them anyway). The options then were to:
- Have a different model for the data we store in TempData (I think we do this in registration).
- Not bother and use ViewBag

The first option is neater, but more dev, so I'll leave it up to you which method to use, but bare in mind that TempData is not meant for storing large quantities of data (we've had problems in the past for TempData exceeding the max and causing page to error)

stellake (changes requested):
Given that there are a few bugs with e.g. page erroring when a new answer is added without selecting the prompt first, we should probably add a few extra unit tests in the controller level and rescan over the controller logic. :)

DigitalLearningSolutions.Data.Tests/Services/CourseAdminFieldsServiceTests.cs
stellake: Ah, now that I look at it again, it's not too bad actually, especially when we remove the mandatory fields so I'd leave it 👍

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CourseSetup/AdminFieldsController.cs
livzorn: I'll just move it to the ViewBag as that's how registration prompts is done

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/CourseSetup/AddAdminFieldViewModel.cs
```diff
@@ -0,0 +1,39 @@
...
+            CustomPromptId = customPromptId;
+            OptionsString = options;
+            IncludeAnswersTableCaption = true;
+            Answer = answer;
+        }
+
+        [Required(ErrorMessage = "Select a prompt name")]
+        public int? CustomPromptId { get; set; }
+
+        public IEnumerable<SelectListItem>? CoursePromptNameOptions { get; set; }
```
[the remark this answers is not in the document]
livzorn: I agree the naming gets confusing, the admin field is of type CustomPrompt but stored in the CoursePrompt table but called Admin Field in the view, I'll change it to Admin field as I think that's the clearest, and the error message will be "Select a field name"

DigitalLearningSolutions.Web/Views/TrackingSystem/CourseSetup/AdminFields/AddAdminField.cshtml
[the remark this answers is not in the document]
livzorn: Ah right, I copied this layout from another page, looks like a lot of our views have this layout where the form isn't wrapped by anything. I've changed the nhsuk-grid-row and nhsuk-column-full divs to encompass the whole page and I've removed the divs that are then made redundant. Also changed this on the Edit admin field page

DigitalLearningSolutions.Web/Views/TrackingSystem/CourseSetup/AdminFields/BulkAdminFieldAnswers.cshtml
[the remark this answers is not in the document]
livzorn: I've changed the non-hidden inputs to take the name from the model

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CourseSetup/AdminFieldsController.cs
```diff
@@ -222,15 +313,62 @@ private void SetEditAdminFieldTempData(EditAdminFieldViewModel model)
             TempData.Set(data);
         }
 
+        private IActionResult AddAdminFieldPostSave(AddAdminFieldViewModel model)
+        {
+            ModelState.ClearAllErrors();
```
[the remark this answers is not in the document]
livzorn: Not sure actually, I copied it from how the Registration Prompts were set up. @AlexJacksonDS is this necessary here?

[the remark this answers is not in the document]
livzorn: Ah Alex nevermind, I understand how ModelState.ClearAllErrors() works now!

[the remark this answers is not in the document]
livzorn: I've changed it to clear all errors except the AdminFieldId so that only the Answer field errors are cleared before we check if the model state is valid

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CourseSetup/AdminFieldsController.cs
livzorn: I've made the Answer field required in the View model but cleared the Answer field error in appropriate places so the user can add or remove an answer or navigate to the bulk page without first selecting a field name

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CourseSetup/AdminFieldsController.cs
livzorn: I've removed the `?? new AddAdminFieldViewModel(customisationId);` part and added a service filter. Though when I navigate to /Add without clicking through the button, I get redirected to /AdminFields rather than /TrackingSystem/CourseSetup/24286/AdminFields and I'm not sure why?

DigitalLearningSolutions.Web.AutomatedUiTests/DigitalLearningSolutions.Web.AutomatedUiTests.csproj
livzorn: Yes just line endings ;)

DigitalLearningSolutions.Data.Tests/DataServices/CourseAdminFieldsDataServiceTests.cs
livzorn: renamed to id and name

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CourseSetup/AdminFieldsController.cs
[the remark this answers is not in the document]
livzorn: And I've added some extra unit tests to check these conditions :)

DigitalLearningSolutions.Web/ServiceFilter/RedirectEmptySessionData.cs
```diff
@@ -5,23 +5,30 @@
...
             if (context.Controller is Controller controller)
             {
                 var userSessionData = controller.TempData.Peek<T>();
-                if (userSessionData == null)
+                if (userSessionData == null && context.ActionArguments.ContainsKey("customisationId"))
+                {
+                    // ReSharper disable once Mvc.ActionNotResolved
+                    context.Result = controller.RedirectToAction(
+                        "Index",
+                        new { customisationId = context.ActionArguments["customisationId"] }
```
stellake, line 28: Hmm.. Could we just pass through all ActionArguments? I can't think of a case where we wouldn't want to do this.. 🤔

DigitalLearningSolutions.Data/DataServices/CourseAdminFieldsDataService.cs
```diff
@@ -9,13 +10,14 @@ public interface ICourseAdminFieldsDataService
     {
         public CourseAdminFieldsResult? GetCourseAdminFields(int customisationId, int centreId, int categoryId);
 
-        public void UpdateCustomPromptForCourse(int customisationId, int promptNumber, bool mandatory, string? options);
+        public void UpdateCustomPromptForCourse(int customisationId, int promptNumber, string? options);
```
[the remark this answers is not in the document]
livzorn: Doesn't this method need to be public so it can be used in CourseAdminFieldsService?

DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/CourseSetup/AdminFieldsControllerTests.cs
```diff
@@ -194,14 +191,293 @@ public void AdminFieldBulkPost_updates_temp_data_and_redirects_to_edit()
...
 
+        [Test]
+        public void AddAdminFieldNew_sets_new_temp_data()
+        {
+            // When
+            var result = controller.AddAdminFieldNew(1);
+
+            // Then
+            controller.TempData.Peek<AddAdminFieldData>().Should()
+                .NotBeNull();
```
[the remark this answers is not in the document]
livzorn: moved it to the same line

DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/CourseSetup/AdminFieldsControllerTests.cs
```diff
@@ -255,10 +531,22 @@ private static void AssertNumberOfConfiguredAnswersOnView(IActionResult result,
                 .Be(expectedCount);
         }
 
+        private void AssertAddAdminFieldViewModelIsExpectedModel(AddAdminFieldViewModel addModel)
+        {
+            controller.TempData.Peek<AddAdminFieldData>()!.AddModel.Should()
+                .BeEquivalentTo(addModel);
+        }
```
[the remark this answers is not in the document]
livzorn: Oops, they are duplicates, removed AssertAddAdminFieldViewModelIsExpectedModel

DigitalLearningSolutions.Web/ServiceFilter/RedirectEmptySessionData.cs
livzorn: Didn't realize you could do that, yes that's much neater

DigitalLearningSolutions.Web/Controllers/TrackingSystem/CourseSetup/AdminFieldsController.cs
```diff
@@ -23,7 +23,8 @@ public class AdminFieldsController : Controller
         public const string AddPromptAction = "addPrompt";
```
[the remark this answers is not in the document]
livzorn: I think when it is called within the Edit action we know at least one field exists but when we call it in Index the course may have none

stellake (approved):
Looks good! 👍

PR #582, opened 2021-09-07
HEEDLS-586 Add extra padding to iframes with long titles

livzorn:
### JIRA link
[HEEDLS-586](https://softwiretech.atlassian.net/browse/HEEDLS-586)

### Description
Added extra padding to iframes with long titles. Previously the "Terminology and classifications delivery service" page's iframe was running over the footer because its title goes onto a second line. I've added logic that checks if the page title is long, and if so, doubles the padding-bottom which makes the space between the iframe and the footer identical to those with shorter titles on desktop, with negligible differences on tablet and mobile. I also made the padding-bottom on all learning content iframes a bit wider as it looked a bit too close to me.

### Screenshots
Iframe with long title:
Terminology_laptop
Terminology_tablet
Terminology_mobile

Iframe with short title:
ItSkills_laptop
ItSkills_tablet
ItSkills_mobile

-----
### Developer checks
I have:

### DIFF
--- a/DigitalLearningSolutions.Web/Controllers/LearningContentController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/LearningContentController.cs
@@ -35,7 +35,7 @@
             var model = new LearningContentViewModel(
                 TerminologyAndClassificationsDeliveryServiceBrand,
                 TerminologyAndClassificationsDeliveryServiceTitle,
-                true
+                titleIsLong: true
             );
             return View("Index", model);
         }
--- a/DigitalLearningSolutions.Web/Controllers/LearningContentController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/LearningContentController.cs
@@ -11,24 +11,32 @@
         private const string ItSkillsPathwayTitle = "IT skills pathway";
         private const string ReasonableAdjustmentFlagBrand = "ReasonableAdjustmentFlag";
         private const string ReasonableAdjustmentFlagTitle = "Reasonable adjustment flag";
-        private const string TerminologyAndClassificationsDeliveryServiceBrand = "TerminologyandClassificationsDeliveryService";
-        private const string TerminologyAndClassificationsDeliveryServiceTitle = "Terminology and classifications delivery service";
-        
+
+        private const string TerminologyAndClassificationsDeliveryServiceBrand =
+            "TerminologyandClassificationsDeliveryService";
+
+        private const string TerminologyAndClassificationsDeliveryServiceTitle =
+            "Terminology and classifications delivery service";
+
         public IActionResult ItSkillsPathway()
         {
             var model = new LearningContentViewModel(ItSkillsPathwayBrand, ItSkillsPathwayTitle);
             return View("Index", model);
         }
-        
+
         public IActionResult ReasonableAdjustmentFlag()
         {
             var model = new LearningContentViewModel(ReasonableAdjustmentFlagBrand, ReasonableAdjustmentFlagTitle);
             return View("Index", model);
         }
-        
+
         public IActionResult TerminologyAndClassificationsDeliveryService()
         {
-            var model = new LearningContentViewModel(TerminologyAndClassificationsDeliveryServiceBrand, TerminologyAndClassificationsDeliveryServiceTitle);
+            var model = new LearningContentViewModel(
+                TerminologyAndClassificationsDeliveryServiceBrand,
+                TerminologyAndClassificationsDeliveryServiceTitle,
+                true
+            );
             return View("Index", model);
         }
     }
--- a/DigitalLearningSolutions.Web/Styles/index.scss
+++ b/DigitalLearningSolutions.Web/Styles/index.scss
@@ -110,7 +110,11 @@ h1#page-heading {
 }
 
 .responsive-iframe-wrapper-padding {
-  padding-bottom: nhsuk-spacing(7);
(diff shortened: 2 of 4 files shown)


DigitalLearningSolutions.Web/Views/LearningContent/Index.cshtml
```diff
@@ -30,6 +30,6 @@
   </div>
 </div>
 
-<div class="nhsuk-grid-row responsive-iframe-wrapper responsive-iframe-wrapper-padding">
+<div class="nhsuk-grid-row responsive-iframe-wrapper @(@Model.Title.Length > 35 ? "responsive-iframe-wrapper-extra-padding" : "responsive-iframe-wrapper-padding")">
```
AlexJacksonDS, line 33: This solution doesn't work, it is possible to generate a title that will have not trigger this extra padding. e.g. the below with 29 characters. I think we should instead add it as an optional parameter bool in the view model constructor (default to false, set to true for this page)
image

AlexJacksonDS (changes requested):

DigitalLearningSolutions.Web/Views/LearningContent/Index.cshtml
livzorn: I've added an optional LongTitle parameter to the view model and set it to true for this page

AlexJacksonDS (approved):

DigitalLearningSolutions.Web/Styles/index.scss
```diff
@@ -110,7 +110,11 @@ h1#page-heading {
 }
 
 .responsive-iframe-wrapper-padding {
-  padding-bottom: nhsuk-spacing(7);
+  padding-bottom: nhsuk-spacing(8);
+}
+
+.responsive-iframe-wrapper-extra-padding {
+  padding-bottom: nhsuk-spacing(8) * 2;
```
SteveJacksonSoft, line 117: I'd recommend pulling nhsuk-spacing(8) out into a variable at the top of this file, so that people don't miss one of them when changing the value.
Something like `$responsive-iframe-padding: nhsuk-spacing(8)`

DigitalLearningSolutions.Web/ViewModels/LearningContent/LearningContentViewModel.cs
```diff
@@ -3,12 +3,14 @@
     public class LearningContentViewModel
     {
         public readonly string Brand;
+        public readonly bool LongTitle;
```
SteveJacksonSoft, line 6: It's nice to have boolean variable names as statements, so it would be better to call this (and the constructor param) something like `titleIsLong`.

That's so that it's obvious 
- that they're booleans (`longTitle` sounds a bit like it refers to an extended version of the title)
- what it means if they're true or false (in this case it's pretty obvious)

DigitalLearningSolutions.Web/Controllers/LearningContentController.cs
```diff
@@ -1,35 +1,43 @@
...
+            var model = new LearningContentViewModel(ReasonableAdjustmentFlagBrand, ReasonableAdjustmentFlagTitle);
+            return View("Index", model);
+        }
+
+        public IActionResult TerminologyAndClassificationsDeliveryService()
+        {
+            var model = new LearningContentViewModel(
+                TerminologyAndClassificationsDeliveryServiceBrand,
+                TerminologyAndClassificationsDeliveryServiceTitle,
+                true
```
SteveJacksonSoft, line 38: You can label parameter values when calling a method.
It's generally so that you can specify parameters in whatever order you want, but it's also helpful when you're passing literals, because you don't have variable names to tell the reader what the value means.
Especially with boolean parameters, it's quite hard to work out what significance they have without reading the method declaration.

Could you add one on the `true`? It should end up looking like `longName: true` (or whatever new name you give `longName`).

SteveJacksonSoft (changes requested):

DigitalLearningSolutions.Web/Styles/index.scss
livzorn: Good idea, I've pulled it out into a variable called $iframe-padding-bottom

DigitalLearningSolutions.Web/ViewModels/LearningContent/LearningContentViewModel.cs
livzorn: Makes perfect sense, changed the name to titleIsLong

DigitalLearningSolutions.Web/Controllers/LearningContentController.cs
livzorn: Added the label titleIsLong, it gets removed when I run the formatter though, is that alright?

SteveJacksonSoft: Intriguing. When you say formatter, do you mean the autoformat function? Are you using VS or Rider?

livzorn: Yeah in Rider. It says `Inconsistent argument style: redundant name identifier` when I hover over the label.
But without the label the IDE does bring up a label automatically:
image

SteveJacksonSoft: I think it's useful to have the explicit label. We should change your configuration so that it doesn't auto-remove it, but let's look at that when I'm back on Tuesday.

PR #589, opened 2021-09-10
HEEDLS-556 - Implemented synchronising changed delegate details with …

DanBloxham-sw:
…groups

### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-556

### Description
When a delegate updates their details, if the job group or any answers to the custom prompts have changed, and these link to synchronised groups, we now update various group records. Note that the new GroupService has a large number of unit tests that I have split over a few partial classes to hopefully make it easier to follow.

### Screenshots
N/A

-----
### Developer checks
I have:


DigitalLearningSolutions.Data/DataServices/GroupsDataService.cs
```diff
@@ -22,6 +23,25 @@ public interface IGroupsDataService
...
+            int customisationId,
+            int customisationVersion,
+            DateTime submittedTime,
+            int enrollmentMethodId,
+            int? enrolledByAdminId,
+            DateTime? completeByDate,
+            int supervisorAdminId
+        );
+
+        void CreateNewAspProgress(int tutorialId, int progressId);
```
AlexJacksonDS, line 44: I think some of these are in the incorrect data service. Progress doesn't really relate to groups.

DigitalLearningSolutions.Data/DataServices/GroupsDataService.cs
```diff
@@ -177,7 +200,116 @@ public void DeleteGroupDelegatesRecordForDelegate(int groupId, int delegateId)
...
+	                    CustomisationID, 
+	                    Completed, 
+	                    RemovedDate, 
+	                    SystemRefreshed, 
+	                    SupervisorAdminID,
+                        CompleteByDate,
+                        EnrollmentMethodID,
+                        EnrolledByAdminID,
+                        SubmittedTime,
+                        CustomisationVersion
```
AlexJacksonDS, line 224: Some odd alignment here

DigitalLearningSolutions.Data/Services/GroupsService.cs
```diff
@@ -0,0 +1,234 @@
...
+        void SynchroniseUserChangesWithGroups(
+            DelegateUser oldDelegateDetails,
+            AccountDetailsData newDelegateDetails,
+            CentreAnswersData newCentreAnswers,
+            string baseUrl
+        );
+
+        IEnumerable<Group> GetSynchronisedGroupsForCentre(int centreId);
+
+        void EnrolDelegateOnGroupCourses(
```
AlexJacksonDS, line 25: Are these 2 only public for unit testing or are they going to be used in the future?

DigitalLearningSolutions.Data/Services/GroupsService.cs
```diff
@@ -0,0 +1,234 @@
...
+        );
+    }
+
+    public class GroupsService : IGroupsService
+    {
+        private const string AddedByProcess = "AddDelegateToGroup_Refactor";
+        private readonly IClockService clockService;
+        private readonly IEmailService emailService;
+
+        private readonly string EnrolEmailSubject = "New Learning Portal Course Enrolment";
```
AlexJacksonDS, line 40: Any particular reason one string is const, the other is readonly?

DigitalLearningSolutions.Web/Startup.cs
```diff
@@ -183,6 +183,7 @@ private static void RegisterServices(IServiceCollection services)
             services.AddScoped<ISessionService, SessionService>();
             services.AddScoped<ISupervisorService, SupervisorService>();
             services.AddScoped<IUserService, UserService>();
+            services.AddScoped<IGroupsService, GroupsService>();
```
AlexJacksonDS, line 186: I was tempted to ask you to move this alphabetically since I tidied them all up, but I know there is no way it'll stay like that 😢

AlexJacksonDS (changes requested):
Overall looks OK, a few odd minor things.

I think some of the data service methods are in the incorrect place, if they don't fit suitably in any of the other existing ones, make a new Progress one or something.

DigitalLearningSolutions.Data/Services/GroupsService.cs
DanBloxham-sw: EnrolDelegateOnGroupCourses should end up being used on https://softwiretech.atlassian.net/browse/HEEDLS-514 (which is mentions in the ticket for this), but it is also useful for unit testing. GetSynchronisedGroupsForCentre I think will only be used here, but is unit tested, though admittedly its a very simple method that could get incorporated into the logic if we'd prefer?

AlexJacksonDS: I think remove GetSynchronisedGroupsForCentre from the interface then and make it private. Makes it easier to bring back if we do decide we need it again outside this service.

AlexJacksonDS (approved):

DigitalLearningSolutions.Data.Tests/DataServices/GroupsDataServiceTests.cs
```diff
@@ -157,6 +157,31 @@ public async Task RemoveRelatedProgressRecordsForGroupDelegate_updates_progress_
             }
         }
 
+        [Test]
+        public async Task RemoveRelatedProgressRecordsForGroupDelegate_does_not_update_progress_record_when_course_is_shared_by_another_group()
```
DanBloxham-sw, line 161: NOTE TO SELF: this test requires a new record adding to the test DBs. When this is about to get merged, add the record to the test environment DBs and let everyone know in the internal chat.

DigitalLearningSolutions.Data/Services/GroupsService.cs
```diff
@@ -0,0 +1,234 @@
...
+        {
+            groupsDataService.DeleteGroupDelegatesRecordForDelegate(groupId, delegateId);
+            groupsDataService.RemoveRelatedProgressRecordsForGroupDelegate(groupId, delegateId, clockService.UtcNow);
+        }
+
+        private Email BuildEnrolmentEmail(
+            string emailAddress,
+            string fullName,
+            GroupCourse course,
+            string baseUrl,
```
SteveJacksonSoft, line 195: `baseUrl` gets passed down about 5 layers to get here, but comes from a static method on ConfigHelper. Is there a reason not to make the call directly in this method?

DigitalLearningSolutions.Data/Services/GroupsService.cs
```diff
@@ -0,0 +1,234 @@
...
+            {
+                DateTime? completeByDate = null;
+                if (groupCourse.CompleteWithinMonths != 0)
+                {
+                    completeByDate = clockService.UtcNow.AddMonths(groupCourse.CompleteWithinMonths);
+                }
+
+                var candidateProgressOnCourse =
+                    progressDataService.GetDelegateProgressForCourse(oldDetails.Id, groupCourse.CustomisationId);
+                var existingRecordToUpdate = candidateProgressOnCourse.FirstOrDefault(
```
SteveJacksonSoft, line 133: We should possibly be updating all we find, but I'll run it past Kevin tomorrow.

SteveJacksonSoft (changes requested):

DigitalLearningSolutions.Data.Tests/DataServices/GroupsDataServiceTests.cs
DanBloxham-sw: It is indeed just this test that requires the new data. There is a circumstance that could arise when a progress record is shared by multiple courses in different groups that one person is on, but we don't have any data that actually fits that category. 

It would be feasible to add this data here, but it would require using three of the new data service methods created as part of this ticket just to set up the record. Correct me if I'm wrong, but I feel like in the past we've only ever used simple tests for the data services, and whenever new data is needed we add them in scripts to prevent having a test that relies on other methods (that also need tests). But it should be simple enough to add it here if we do want that.

DigitalLearningSolutions.Data.Tests/Services/GroupServiceTests/SynchroniseGroupsTests.cs
```diff
@@ -0,0 +1,191 @@
+﻿namespace DigitalLearningSolutions.Data.Tests.Services.GroupServiceTests
+{
+    using System.Collections.Generic;
+    using DigitalLearningSolutions.Data.Models.DelegateGroups;
+    using DigitalLearningSolutions.Data.Tests.TestHelpers;
+    using FakeItEasy;
+    using FluentAssertions.Execution;
+    using NUnit.Framework;
+
+    public partial class GroupsServiceTests
```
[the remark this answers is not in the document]
DanBloxham-sw: This is already inside the GroupServiceTests folder, do we need to change the file names as well?

DigitalLearningSolutions.Data.Tests/DataServices/GroupsDataServiceTests.cs
SteveJacksonSoft: If you think using the data services to set up the data would be detrimental to the quality of the test (I think I'd agree) I think it's better to use the sql script you've created to set up the data - but at the start of the test rather than as a permanent change to the db.

There are a few downsides to adding permanent test data:
* It requires everyone on the team to run another sql script / adds another script to run when setting up the test db
* It adds another entity to possibly clash with when writing other data service tests.
* It means you need more contextual information to understand why the test works.
* It makes the test setup much harder to change, should that be required in the future.

So I think it's good to minimise it as much as possible.

DigitalLearningSolutions.Data.Tests/Services/GroupServiceTests/EnrolDelegateTests.cs
```diff
@@ -0,0 +1,667 @@
...
+            {
+                A.CallTo(
+                        () => emailService.ScheduleEmail(A<Email>._, "AddDelegateToGroup_Refactor", null)
+                    )
+                    .MustHaveHappened();
+            }
+        }
+
+        [Test]
+        public void EnrolDelegateOnGroupCourses_sends_correct_email_with_no_CompleteByDate()
```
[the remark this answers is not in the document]
DanBloxham-sw: Strange. They both pass for me when run individually, as a file, and when running all tests. The Jenkins builds from when I first made the commits all seemed to have passed. Not sure why it would be failing now for you, but I'll have a further investigation.

DigitalLearningSolutions.Data.Tests/TestHelpers/TutorialContentTestHelper.cs
```diff
@@ -84,5 +85,16 @@ DELETE CustomisationTutorials
                 new { customisationId, tutorialId }
             );
         }
+
+        public int? GetAspProgressFromTutorialAndProgressId(int tutorialId, int progressId)
```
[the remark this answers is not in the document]
DanBloxham-sw: It could easily go into the Data project, but it is only used for a single test currently, and I can't foresee any time it would need to be used outside of tests. Do we want it moved to the data project?

DigitalLearningSolutions.Data.Tests/TestHelpers/UserTestHelper.cs
```diff
@@ -120,9 +120,9 @@ public static class UserTestHelper
         public static AdminUser GetDefaultCategoryNameAllAdminUser()
         {
             return GetDefaultAdminUser(
-                id: 11,
-                centreId: 59,
-                centreName: "Guy's and St Thomas' NHS Foundation Trust",
+                11,
+                59,
+                "Guy's and St Thomas' NHS Foundation Trust",
```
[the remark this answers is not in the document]
DanBloxham-sw: These are not intentionally removed. Resharper finds the labels on these three redundant because they are in exactly the same order as the constructor's parameters. So it automatically removes them when cleaning up the code. When looking at this in VS, the parameter labels are inferred and displayed automatically (see picture), but I'm happy to add them back, maybe slightly reordering these so they don't vanish unexpectedly. 

image

DigitalLearningSolutions.Data/Models/LinkedFieldNumberWithValues.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace DigitalLearningSolutions.Data.Models
+{
+    public class LinkedFieldNumberWithValues
```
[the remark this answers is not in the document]
DanBloxham-sw: LinkedFieldChange sounds sensible to me, so I'll change the name to that. For context, when designing this I was indeed treating the linked field numbers as the main thing with the changed values being an aside. 

Before this class comes into play, you have the old details with the old answers, and the new details with the new answers. But you don't have a way of mapping the answers to the LinkedToField column in the DB. The helper method is what creates that link between answers and the LinkedToField. So the values are not the main focus, as you already have the values; what you need is the mapping of the answers to the corresponding linked field so the DB can be synced correctly.

The reason the values are included in this class is mostly for convenience. It is trivial to find which new answers are different to old ones. You could keep them separate to the field numbers, but bundling them with the linked field just makes it easier to keep track of. But that is why I named it as the LinkedFieldNumber being the main thing.

DigitalLearningSolutions.Data/Models/Progress.cs
```diff
@@ -0,0 +1,20 @@
...
+    using System;
+
+    public class Progress
+    {
+        public int ProgressId { get; set; }
+        public int CandidateId { get; set; }
+        public int CustomisationId { get; set; }
+        public DateTime? Completed { get; set; }
+        public DateTime? RemovedDate { get; set; }
+        public bool SystemRefreshed { get; set; }
```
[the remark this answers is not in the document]
DanBloxham-sw: Looked back at some old conversations with Stella to find out. In the context of another ticket that was looking at progress records:

> The SystemRefreshed is to do with automatic removals, but when it's set, the RemovedDate is always set at the same time, so actually we don't need both

So I reckon it can safely be removed here as we now should only rely on RemovedDate.

DigitalLearningSolutions.Data/Services/GroupsService.cs
DanBloxham-sw: ConfigHelper and the configuration data is defined in the Web project, which references the Data project. We can't reference the Web project from the Data project or else we start getting cyclical references. We have the same issue with setting up some emails (the ForgotPassword email comes to mind).

DigitalLearningSolutions.Data.Tests/Services/GroupServiceTests/SynchroniseGroupsTests.cs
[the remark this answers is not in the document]
SteveJacksonSoft: Yeah, I think it would be helpful, because it's more common to search for files by name than by location in the file tree.

DigitalLearningSolutions.Data/Services/GroupsService.cs
DanBloxham-sw: From what I recall, the other records that would be found are completed/removed records that need to not be updated. We otherwise only have 1 record per delegate-course combination. Happy to be corrected though.

DigitalLearningSolutions.Data.Tests/TestHelpers/TutorialContentTestHelper.cs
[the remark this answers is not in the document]
SteveJacksonSoft: Sorry, thought I'd deleted all the TODO comments 🤦 Yeah, it's fine in tests I think.

DigitalLearningSolutions.Data/Services/GroupsService.cs
```diff
@@ -0,0 +1,234 @@
...
+        {
+            var groupCourses = groupsDataService.GetGroupCourses(groupId, oldDetails.CentreId);
+
+            foreach (var groupCourse in groupCourses)
+            {
+                DateTime? completeByDate = null;
+                if (groupCourse.CompleteWithinMonths != 0)
+                {
+                    completeByDate = clockService.UtcNow.AddMonths(groupCourse.CompleteWithinMonths);
+                }
```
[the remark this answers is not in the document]
DanBloxham-sw: I had tried something like this:
```
DateTime? completeByDate = groupCourse.CompleteWithinMonths != 0
                    ? clockService.UtcNow.AddMonths(groupCourse.CompleteWithinMonths)
                    : null;
```

But get the error: Feature 'target-typed condition expression' is not available in C# 8.0. Please use language version 9.0 or greater. There is no implicit conversion between System.Datetime and null.

DigitalLearningSolutions.Data/Services/GroupsService.cs
SteveJacksonSoft: Yep, checked with Kevin. He confirmed we should update all matching, in-progress Progress records rather than just the first.

DanBloxham-sw: Well now I am very confused. Do we expect multiple in-progress Progress records per delegate-course combination?

DigitalLearningSolutions.Data/Services/GroupsService.cs
```diff
@@ -0,0 +1,234 @@
...
+    using DigitalLearningSolutions.Data.Helpers;
+    using DigitalLearningSolutions.Data.Models.DelegateGroups;
+    using DigitalLearningSolutions.Data.Models.Email;
+    using DigitalLearningSolutions.Data.Models.User;
+    using MimeKit;
+
+    public interface IGroupsService
+    {
+        void SynchroniseUserChangesWithGroups(
+            DelegateUser oldDelegateDetails,
```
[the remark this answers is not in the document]
DanBloxham-sw: Bit confused here. AccountDetailsData does not contain properties for any of the Answers, so we can't just replace the DelegateUser with it unless we want to changeAccountDetailsData to now contain those answers?
What does "Also, that's all it actually needs" mean?

DigitalLearningSolutions.Data.Tests/Services/GroupServiceTests/EnrolDelegateTests.cs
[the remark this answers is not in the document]
DanBloxham-sw: Yeah I've got no idea why this would be failing for you. I would normally suspect subtle differences in our test databases, but all data service queries here have been mocked. Is there any useful information about why it failed?

DigitalLearningSolutions.Data.Tests/Helpers/LinkedFieldHelperTests.cs
```diff
@@ -0,0 +1,164 @@
...
+                newAnswers,
+                jobGroupsDataService
+            );
+
+            // Then
+            result.Should().BeEquivalentTo(expectedResult);
+        }
+
+        [Test]
+        public void Changed_JobGroup_maps_to_linked_field_4()
```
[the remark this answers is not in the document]
DanBloxham-sw: I've implemented TestCaseSource as we already have a bunch of other tests that use TestCase vs Theories, and I found it easier to get working. If you think DatapointSource would be better, let me know.

DigitalLearningSolutions.Data/Models/Progress.cs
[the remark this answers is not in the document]
SteveJacksonSoft: Yeah, Stella said that when the system refreshes the course, it both sets the removed date and changes `SystemRefreshed` from 0 to 1. (So `SystemRefreshed` actually means `HasBeenSystemRefreshed`). But yeah, I think the implication there is that we don't need it.

DigitalLearningSolutions.Data/Services/GroupsService.cs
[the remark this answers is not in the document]
SteveJacksonSoft: Sorry, I think I meant to put this on `LinkedFieldHelper.GetLinkedFieldChanges()`.
Ignore the comment above and I'll put something correct in the place I meant to put it.

But here you do have the same issue as on `EnrolDelegateOnGroupCourses()`, where it would be nicer for the DelegateUser to be called something like `delegateAccountWithOldDetails`.

DigitalLearningSolutions.Data/Services/GroupsService.cs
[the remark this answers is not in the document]
SteveJacksonSoft: Oh, I think you just have to cast `clockService.UtcNow....)` to `(DateTime?)` - the compiler assumes it's a `DateTime`, which doesn't square with the alternative being `null`.

DigitalLearningSolutions.Data/Services/GroupsService.cs
SteveJacksonSoft: We don't, but the database doesn't constrain it. If we did end up with more than one, it's probably least likely to cause surprising behaviour if we just treat them all the same way.

DigitalLearningSolutions.Data/Services/GroupsService.cs
SteveJacksonSoft: Ah. That's frustrating. We should change that, but I'll put it in a tech debt ticket.
For now, could you just grab the configuration directly?
You can
- inject it into the service in the constructor as `IConfiguration`
- create a `ConfigHelper` class in the Data project (which will eventually become the only one) and copy the `GetAppRootPath()` method
- get the base url where you need it by using the new helper method.

You'll need to add a reference to Microsoft.Extensions.Configuration or somesuch.
It's possible having another class called `ConfigHelper` will cause problems. If so, I think just give it a different name. Maybe `ConfigurationHelper`.

Could you also stick an `Obsolete` attribute on the existing config helper with a description saying that we're moving to ConfigurationHelper, and that people should copy over the methods they need?

DigitalLearningSolutions.Data/Models/LinkedFieldNumberWithValues.cs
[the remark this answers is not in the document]
SteveJacksonSoft: Ah, I see, thanks. I think the new name works for both purposes and is a bit easier to understand, cheers.

DigitalLearningSolutions.Data/Helpers/LinkedFieldHelper.cs
```diff
@@ -7,60 +7,60 @@
 
     public static class LinkedFieldHelper
     {
-        public static List<LinkedFieldNumberWithValues> GetLinkedFieldNumbersWithValuesFromChangedAnswers(
+        public static List<LinkedFieldChange> GetLinkedFieldChanges(
             DelegateUser oldDelegateDetails,
```
SteveJacksonSoft, line 11: Yes, so the DelegateUser properties you use here are the same as the properties on CentreAnswersData right?
So you could change this parameter to `CentreAnswersData oldAnswers`, and this method would become easier to use (because you don't need an entire delegate to use it). The signature would also be less surprising to future devs.
Then you can add a method `DelegateUser.GetCentreAnswersData` to extract the answers from your delegate before you call this method (I previously suggested an implicit method, but on second thoughts the idea of casting a `DelegateUser` into a `CentreAnswersData` is a bit strange).

DigitalLearningSolutions.Data/Services/GroupsService.cs
```diff
@@ -0,0 +1,234 @@
...
+                DateTime? completeByDate = null;
+                if (groupCourse.CompleteWithinMonths != 0)
+                {
+                    completeByDate = clockService.UtcNow.AddMonths(groupCourse.CompleteWithinMonths);
+                }
+
+                var candidateProgressOnCourse =
+                    progressDataService.GetDelegateProgressForCourse(oldDetails.Id, groupCourse.CustomisationId);
+                var existingRecordToUpdate = candidateProgressOnCourse.FirstOrDefault(
+                    p => !p.SystemRefreshed && p.Completed == null && p.RemovedDate == null
```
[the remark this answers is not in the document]
SteveJacksonSoft: Ah, just realised this isn't necessarily dealing with automatic enrollment.
So just `ProgressShouldBeUpdatedOnEnrollment`? Sorry

DigitalLearningSolutions.Data.Tests/Services/GroupServiceTests/EnrolDelegateTests.cs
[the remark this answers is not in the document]
SteveJacksonSoft: They work now. 🤷

SteveJacksonSoft (approved):
Lovely jubbly, merge away!

PR #600, opened 2021-09-16
HEEDLS-600 Store query parameters when redirecting to login page

davidm-m:
### JIRA link
[HEEDLS-600](https://softwiretech.atlassian.net/browse/HEEDLS-600)

### Description
Preserved query parameters when redirecting to and from the login page. This was done in exactly the same way we already store the return URL.

-----
### Developer checks
I have:


JBsoftwire (approved):
This looks good to me.

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -34,14 +34,14 @@ ILogger<LoginController> logger
...
-        public IActionResult Index(string? returnUrl = null)
+        public IActionResult Index(string? returnUrl = null, string? queryString = null)
         {
             if (User.Identity.IsAuthenticated)
             {
                 return RedirectToAction("Index", "Home");
             }
 
-            var model = new LoginViewModel(returnUrl);
+            var model = new LoginViewModel(returnUrl, queryString);
```
SteveJacksonSoft, line 44: Is there a reason not to stick `returnUrl` and `queryString` together here, rather than passing them separately through TempData?

DigitalLearningSolutions.Web.Tests/Controllers/Login/LoginControllerTests.cs
```diff
@@ -123,6 +123,27 @@ public async Task Successful_sign_in_with_local_return_url_should_redirect_to_re
...
+
+            var returnUrl = "/some/other/page";
+            var queryString = "?query=true&other=false";
+            var urlHelper = controller.Url;
+            A.CallTo(() => urlHelper.IsLocalUrl(returnUrl)).Returns(true);
+
+            // When
+            var loginViewModel = LoginTestHelper.GetDefaultLoginViewModel();
+            loginViewModel.ReturnUrl = returnUrl;
+            loginViewModel.QueryString = queryString;
```
SteveJacksonSoft, line 140: I would put this in `Given` because it's not something you're testing - it's just setting up data.

DigitalLearningSolutions.Web/Startup.cs
```diff
@@ -274,7 +274,8 @@ private Task RedirectToLogin(RedirectContext<CookieAuthenticationOptions> contex
         {
             var applicationPath = new Uri(config["AppRootPath"]).AbsolutePath.TrimEnd('/');
             var url = HttpUtility.UrlEncode(applicationPath + context.Request.Path);
-            context.HttpContext.Response.Redirect(config["AppRootPath"] + $"/Login?returnUrl={url}");
+            var queryString = HttpUtility.UrlEncode(context.Request.QueryString.Value);
+            context.HttpContext.Response.Redirect(config["AppRootPath"] + $"/Login?returnUrl={url}&queryString={queryString}");
```
SteveJacksonSoft, line 278: Could you change these two config calls to use our recent helper method, `config.GetAppRootPath()`?

SteveJacksonSoft (changes requested):

SteveJacksonSoft:
Just to check, are the unticked tasks on the description because there are no docs worth adding and no other parts of the system need to be sanity checked?

davidm-m:
> 
> 
> Just to check, are the unticked tasks on the description because there are no docs worth adding and no other parts of the system need to be sanity checked?

Yes, I wasn't really sure what to do with this since ticking them wouldn't technically be correct (but maybe I'm just being overly literal)

DigitalLearningSolutions.Web/Controllers/LoginController.cs
davidm-m: Thinking about it, none that I can think of. I've combined them and tested that it still works.

SteveJacksonSoft (approved):
Nice, cheers!

PR #601, opened 2021-09-16
HEEDLS-575 - Implemented the limit of admin role editing if no availa…

DanBloxham-sw:
…ble space

### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-575

### Description
When the number of admins in specific roles at a centre reaches or exceeds the limits placed on them, a warning pops up on the Edit Roles page for users outside those roles, and the corresponding checkbox/radio button is removed from the page. This refactors the logic from the NumberOfAdministratorsViewComponent into its own helper so that it can be used here without duplicating a bunch of code.

### Screenshots
Desktop with some roles filled:
Desktop some missing fields

Mobile version:

Desktop with all content management roles filled
Desktop content management hidden

-----
### Developer checks
I have:


DigitalLearningSolutions.Web/Controllers/TrackingSystem/Centre/Administrator/AdministratorController.cs
```diff
@@ -82,24 +86,39 @@ public IActionResult AllAdmins()
...
+            var adminUser = userDataService.GetAdminUserById(adminId);
+
+            if (adminUser == null || adminUser.CentreId != centreId)
+            {
+                return NotFound();
+            }
+
+            if (NewUserRolesExceedAvailableSpots(model, adminUser))
+            {
+                return new StatusCodeResult(500);
```
AlexJacksonDS, line 119: Do we want to throw something more explicit here so we can track if this is occurring?

DigitalLearningSolutions.Data/Helpers/NumberOfAdministratorsHelper.cs
```diff
@@ -0,0 +1,35 @@
...
+    using DigitalLearningSolutions.Data.DataServices.UserDataService;
+    using DigitalLearningSolutions.Data.Models;
+
+    public interface INumberOfAdministratorsHelper
+    {
+        NumberOfAdministrators GetCentreAdministratorNumbers(int centreId);
+
+    }
+
+    public class NumberOfAdministratorsHelper: INumberOfAdministratorsHelper
```
AlexJacksonDS, line 13: Is there any particular reason this is a Helper not a Service? It seems more like a service to me, with helpers usually being static.

DigitalLearningSolutions.Data/Helpers/NumberOfAdministratorsHelper.cs
```diff
@@ -0,0 +1,35 @@
+﻿namespace DigitalLearningSolutions.Data.Helpers
+{
+    using DigitalLearningSolutions.Data.DataServices;
+    using DigitalLearningSolutions.Data.DataServices.UserDataService;
+    using DigitalLearningSolutions.Data.Models;
+
+    public interface INumberOfAdministratorsHelper
+    {
+        NumberOfAdministrators GetCentreAdministratorNumbers(int centreId);
+
```
AlexJacksonDS, line 10: extra line

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/EditAdminRoles.cshtml
```diff
@@ -34,20 +34,39 @@
...
+            </span>
+          </h3>
+          <p>
+            Some user roles already have the maximum number of members assigned and are not available for selection.
+            Check the "Number of administrators" information to see which roles have reached their limits.
+          </p>
+        </div>
+      }
+
+      <vc:number-of-administrators centre-id="Model.CentreId" />
```
AlexJacksonDS, line 54: Does all this stuff need to be inside the form?

AlexJacksonDS (changes requested):
I think the helper should probably be a service. Plus some minor formatting/layout issues/questions.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Centre/Administrator/AdministratorController.cs
DanBloxham-sw: At time of writing, we have decided to go with this. I believe Stella is still waiting to hear back if HEE would prefer a custom error page. I've added a logger though as its a good idea.

DigitalLearningSolutions.Data/Services/NumberOfAdministratorsHelper.cs
```diff
@@ -0,0 +1,34 @@
+﻿namespace DigitalLearningSolutions.Data.Services
```
AlexJacksonDS, line 1: Filename is wrong now.

AlexJacksonDS (commented):
Just need to fix the file name

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Centre/Administrator/AdministratorController.cs
AlexJacksonDS: I was more thinking of throwing an exception with a message, which will result in a 500 anyway, but I think an explicit log and return 500 is probably clearer.

AlexJacksonDS (approved):

DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/Centre/Administrator/AdministratorControllerTests.cs
```diff
@@ -188,5 +202,163 @@ public void UnlockAccount_unlocks_account_and_returns_to_page()
...
+            var numberOfAdmins = NumberOfAdministratorsTestHelper.GetDefaultNumberOfAdministrators();
+            A.CallTo(() => userDataService.GetAdminUserById(A<int>._)).Returns(currentAdminUser);
+            A.CallTo(() => numberOfAdministratorsService.GetCentreAdministratorNumbers(A<int>._))
+                .Returns(numberOfAdmins);
+
+            // When
+            var result = administratorController.EditAdminRoles(newRoles, currentAdminUser.Id);
+
+            // Then
+            A.CallTo(
```
stellake, line 233: These tests are quite long and hence more difficult to glance over. One way to make them a bit shorter and better readable would be to pull some stuff into private methods. What I can think of atm is that we could pull this method into AssertAdminPermissionsUpdatedCorrectly(id, newRoles). As a bonus, this can then be reused for the test below.

DigitalLearningSolutions.Web.Tests/ViewModels/TrackingSystem/Centre/Administrator/EditRolesViewModelTests.cs
```diff
@@ -76,12 +78,246 @@ public void EditRolesViewModel_sets_ContentManagementRole_none()
...
+            // When
+            var result = new EditRolesViewModel(adminUser, 1, categories, numberOfAdmins);
+
+            // Then
+            using (new AssertionScope())
+            {
+                result.Radios.Count.Should().Be(1);
+                result.Radios.Contains(AdminRoleInputs.CmsManagerRadioButton).Should().BeFalse();
+                result.Radios.Contains(AdminRoleInputs.CmsAdministratorRadioButton).Should().BeFalse();
+                result.Radios.Contains(AdminRoleInputs.NoCmsPermissionsRadioButton).Should().BeTrue();
```
stellake, line 317: If we check the count = 1 and "contains", do we need to check that it doesn't contain bunch of values?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Centre/Administrator/AdministratorController.cs
stellake: HEE has confirmed that throwing is OK here - I think Alex is right that it would make sense to just throw a custom error, (which will be handled by default with logging & redirect to 500)

stellake (changes requested):
Main improvement would be moving some logic into the service layer.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Centre/Administrator/AdministratorController.cs
```diff
@@ -138,5 +165,36 @@ private IEnumerable<string> GetCourseCategories(int centreId)
             categories = categories.Prepend("All");
             return categories;
         }
+
+        private bool NewUserRolesExceedAvailableSpots(EditRolesViewModel newRoles, AdminUser oldUserDetails)
```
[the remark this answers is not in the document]
DanBloxham-sw: I kept this check method in the controller, rather than making it part of a service method, as I was under the impression that Super Admins will be able to add to roles even when they are full. In which case it would make more sense not to bake this check in to the update procedure if it will eventually be side-stepped anyway. So the change is certainly feasible, and I can easily make it, but I just want to clarify if we think it is best?

DanBloxham-sw:
Here are some screenshots of the page with extra spacing around the Number Of Administrators component, one without the warning, and one with
Desktop number of admins extra space withw arning
Desktop number of admins extra space

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Centre/Administrator/AdministratorController.cs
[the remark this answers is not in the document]
stellake: We can always add a flag to skip the check as a param to the service method

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -242,9 +259,77 @@ public IEnumerable<DelegateUserCard> GetDelegateUserCardsForWelcomeEmail(int cen
...
 
+        public void UpdateAdminUserPermissions(
+            int adminId,
+            bool isCentreAdmin,
+            bool isSupervisor,
+            bool isTrainer,
+            bool isContentCreator,
+            bool isContentManager,
+            bool importOnly,
+            int categoryId
```
stellake, line 270: Could we wrap this in a model - e.g. `AdminRoles`. Then you can also do something like EditRolesViewModel.GetAdminRoles() in the controller when calling this method

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -242,9 +259,77 @@ public IEnumerable<DelegateUserCard> GetDelegateUserCardsForWelcomeEmail(int cen
...
+            {
+                return true;
+            }
+
+            if (isContentCreator && !oldUserDetails.IsContentCreator && currentNumberOfAdmins.CcLicencesAtOrOverLimit)
+            {
+                return true;
+            }
+
+            if (isContentManager && importOnly & !oldUserDetails.IsCmsAdministrator &&
```
stellake, line 320: If we wrap this in AdminRoles model, we can have a derived field cmsAdministrator on it (which will be nicer to read here over `isContentManager && importOnly` )

DigitalLearningSolutions.Data/Models/AdminRoles.cs
```diff
@@ -0,0 +1,31 @@
...
+    public class AdminRoles
+    {
+        public bool IsCentreAdmin { get; set; }
+        public bool IsSupervisor { get; set; }
+        public bool IsTrainer { get; set; }
+        public bool IsContentCreator { get; set; }
+        public bool IsContentManager { get; set; } 
+        public bool ImportOnly { get; set; }
+
+        public AdminRoles( bool isCentreAdmin = true,
```
stellake, line 12: Formatting

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Centre/Administrator/AdministratorController.cs
```diff
@@ -115,14 +116,18 @@ public IActionResult EditAdminRoles(EditRolesViewModel model, int adminId)
...
+            var adminRoles = new AdminRoles(
                 model.IsCentreAdmin,
                 model.IsSupervisor,
-                model.IsTrainer,
                 model.IsContentCreator,
+                model.IsTrainer,
                 model.ContentManagementRole.IsContentManager,
-                model.ContentManagementRole.ImportOnly,
+                model.ContentManagementRole.ImportOnly
+            );
```
stellake, line 126: Could we just add this as a method on the ViewModel so:
1. We can write a simple unit test for it.
2. We can just do model.GetAdminRoles() when calling UpdateAdminUserPermissions below?

DigitalLearningSolutions.Data.Tests/TestHelpers/UserTestHelper.cs
```diff
@@ -219,5 +220,25 @@ FROM Candidates
 
             return users.Single();
         }
+
+        public static AdminRoles GetDefaultAdminRoles
```
stellake, line 224: Could you just do new AdminRoles() instead of defining this new helper method as the constructor of AdminRoles sets everything to true by default?

stellake (approved):
Looks good!

PR #605, opened 2021-09-17
HEEDLS-581 Case insensitive email check and remove inactive unapprove…

AlexJacksonDS:
…d linked admins

### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-581

### Description
Make the multiple email check case insensitive as emails aren't. Filter out inactive, unapproved admins when looking for linked admins

### Screenshots
N/A

-----
### Developer checks
I have:


DigitalLearningSolutions.Data/Services/LoginService.cs
```diff
@@ -156,12 +156,12 @@ public LoginResult AttemptLogin(string username, string password)
 
         private static bool MultipleEmailsUsedAcrossAccounts(AdminUser? adminUser, List<DelegateUser> delegateUsers)
         {
-            var emails = delegateUsers.Select(du => du.EmailAddress)
+            var emails = delegateUsers.Select(du => du.EmailAddress?.ToLower())
```
davidm-m, line 159: Should we be using `ToLowerInvariant` or `Tolower` with a specified culture? We could conceivably see an issue with this if a person's locale is set to certain cultures ([StackOverflow example](https://stackoverflow.com/questions/6600954/c-tolower-is-sometimes-removing-dot-from-the-letter-i)).

davidm-m (commented):
Just one question about how we should handle making things lowercase, otherwise looks good 👍

DigitalLearningSolutions.Data/Services/LoginService.cs
AlexJacksonDS: I don't think it will actually change anything for us, assuming my understanding of how the culture stuff work is correct. I thought it was based of the device the application was running on, in this case the server that we (HEE) control and is therefore most likely to be English (UK) or English (US). It sounds like the example in that StackOverflow is some kind of client app (which could then be picking up any culture) calling a web service and failing. I've changed it anyway, because I don't think it makes any difference is actually correctly set to English.

davidm-m: Yeah, I feel like it's better to be safe than sorry here since if our locale is fixed to English there will be no difference between `ToLower` and `ToLowerInvariant` but if it isn't we're avoiding a hard to track down bug. Thanks for changing it 🙂

davidm-m (approved):
All good now!

DigitalLearningSolutions.Data/Services/UserVerificationService.cs
```diff
@@ -83,6 +83,8 @@ string password
             var adminUserAssociatedWithDelegates = userDataService.GetAdminUserByEmailAddress(delegateEmail);
 
             var isSuitableAdmin = adminUserAssociatedWithDelegates != null &&
+                                  adminUserAssociatedWithDelegates.Active &&
+                                  adminUserAssociatedWithDelegates.Approved &&
```
SteveJacksonSoft, line 87: Could we just tweak the names of this method and the delegate equivalent?
`Verified` says "username and password match" to me, but doesn't imply active / approved.
I think adding an adjective to the names of the methods would do the trick. Something that implies both `active` and `approved`. "Active" would be ideal if it weren't a meaningful name already. "InUse"?

SteveJacksonSoft (changes requested):
Nice, just two tweaks that would be good.

DigitalLearningSolutions.Data/Services/UserVerificationService.cs
AlexJacksonDS: Rather than trying to think of some other word to use, I've just gone with the simple approach of adding ActiveApproved

SteveJacksonSoft (approved):
Cheers, merge away.

PR #613, opened 2021-09-20
HEEDLS-602 Supervisor-Delegate relationship

ibrahimmunir14:
### JIRA link
[HEEDLS-602](https://softwiretech.atlassian.net/browse/HEEDLS-602)

### Description
- When a delegate starts the registration process, if they followed a link with a supervisor-delegate inviteId, pre-populate the email form field.
- When a delegate completes registration, find a pending supervisor-delegate relationship matching their email address (and inviteId when relevant). If one exists, mark user as approved, and update the CandidateId on the supervisor-delegate record. If there was an inviteId, also set the Confirmed date on the record.
- When an admin registers a delegate from the tracking system, find a pending supervisor-delegate relationship matching the email address. If one exists, update the CandidateId on the supervisor-delegate record.
- Note: no effect on registrations through `TrackingSystem/Delegates/BulkUpload`.

### Screenshots
N/A

-----
### Developer checks
I have:


AlexJacksonDS (approved):
LGTM

DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -46,30 +51,51 @@ IConfiguration config
...
         )
         {
-            var centreIpPrefixes =
-                centresDataService.GetCentreIpPrefixes(delegateRegistrationModel.Centre);
-            delegateRegistrationModel.Approved =
-                centreIpPrefixes.Any(ip => userIp.StartsWith(ip.Trim())) || userIp == "::1";
-
-            var candidateNumber =
-                registrationDataService.RegisterDelegate(delegateRegistrationModel);
+            var supervisorDelegateRecordIds = supervisorDelegateService.GetPendingSupervisorDelegateRecordsByEmail(
```
SteveJacksonSoft, line 65: `supervisorDelegateRecordIds` is a little vague... could we add `MatchingDelegate` or something on the end of the name?

SteveJacksonSoft: We could also extract this block into a method `GetPendingSupervisorDelegateIdsMatchingDelegate(delegateRegistrationModel)` to share with `RegistyerDelegateByCentre()`.

DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -46,30 +51,51 @@ IConfiguration config
...
+            passwordDataService.SetPasswordByCandidateNumber(candidateNumber, delegateRegistrationModel.PasswordHash!);
+
+            if (supervisorDelegateRecordIds.Any())
+            {
+                supervisorDelegateService.AddCandidateIdToSupervisorDelegateRecords(
+                    supervisorDelegateRecordIds,
+                    delegateRegistrationModel.Centre,
+                    candidateNumber
+                );
+                if (foundRecordForSupervisorDelegateId)
```
SteveJacksonSoft, line 92: This if doesn't need to be nested.

SteveJacksonSoft (changes requested):
Found some changes for you to make so you've got something to do on your last day. 😛

DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -62,12 +67,13 @@ IFrameworkNotificationService frameworkNotificationService
             int? supervisorDelegateId = null
         )
         {
-            var supervisorDelegateRecordIds = supervisorDelegateService.GetPendingSupervisorDelegateRecordsByEmail(
-                delegateRegistrationModel.Centre,
-                delegateRegistrationModel.Email
-            ).Select(record => record.ID).ToList();
+            var supervisorDelegateRecordIdsMatchingDelegate =
+                GetPendingSupervisorDelegateIdsMatchingDelegate(delegateRegistrationModel);
```
SteveJacksonSoft, line 71: Can we `ToList()` this so it doesn't get double-enumerated?

DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -150,21 +154,40 @@ public string RegisterDelegateByCentre(DelegateRegistrationModel delegateRegistr
...
+        }
 
-            if (supervisorDelegateRecordIds.Any())
+        private void UpdateSupervisorDelegateRecordsWithNewCandidateId(
+            string candidateNumber,
+            int centreId,
+            IEnumerable<int> recordIds
+        )
+        {
+            if (recordIds.Any())
```
SteveJacksonSoft, line 175: We should `ToList()` `recordIds` first to ensure it's not double-enumerated.

DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -150,21 +154,40 @@ public string RegisterDelegateByCentre(DelegateRegistrationModel delegateRegistr
...
             {
-                supervisorDelegateService.AddCandidateIdToSupervisorDelegateRecords(
-                    supervisorDelegateRecordIds,
-                    delegateRegistrationModel.Centre,
-                    candidateNumber
-                );
+                var delegateUser = userDataService.GetDelegateUserByCandidateNumber(candidateNumber, centreId)!;
```
SteveJacksonSoft, line 177: This ! is worrying me a bit. I don't think we should assume the candidate will be in the db because
- This method might be used by a future dev elsewhere, not straight after registration
- If there's a error connecting to the database, we'll get a NullReferenceException.
It would be nicer to handle the null case, log that we couldn't find the candidate in the database, and don't bother with adding them to the supervisor delegate records.
You could handle the error in here and change the name of the method from `Do this` to `Do this if candidate exists`. Or you could throw here and handle outside this method.

SteveJacksonSoft (changes requested):

DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -88,12 +93,26 @@ IUserDataService userDataService
...
 
-            UpdateSupervisorDelegateRecordsWithNewCandidateId(
+            var delegateUser = userDataService.GetDelegateUserByCandidateNumber(
                 candidateNumber,
-                delegateRegistrationModel.Centre,
-                supervisorDelegateRecordIdsMatchingDelegate
+                delegateRegistrationModel.Centre
             );
 
+            if (delegateUser == null)
```
AlexJacksonDS, line 101: I don't see how we can hit this case without some error with the code above? The stored procedure to register the delegate can only return a valid non-null candidateNumber, "-1" or "-4", the "-2" or "-3" cases will be caught by the c# code before even hitting the stored procedure. Since we immediately return for -1/-4, we should have a delegate from the get by candidate number.

DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -155,28 +174,29 @@ public string RegisterDelegateByCentre(DelegateRegistrationModel delegateRegistr
...
                 candidateNumber,
-                delegateRegistrationModel.Centre,
-                supervisorDelegateRecordIdsMatchingDelegate
+                delegateRegistrationModel.Centre
             );
 
-            return candidateNumber;
-        }
-
-        private void UpdateSupervisorDelegateRecordsWithNewCandidateId(
```
AlexJacksonDS: We've lost this method resulting in duplicate code in the 2 register delegate methods, is it not possible to keep this?
I suspect we'll end up in the situation where we have even more common code between the 2 methods after I've done some of the refactoring I was planning to do.

AlexJacksonDS (commented):
A few questions, not sure they necessarily need changing, but I think it's worth checking that we do want to do things this way.

DigitalLearningSolutions.Data/Services/RegistrationService.cs
SteveJacksonSoft: I don't see where -2 and -3 are caught...?

DigitalLearningSolutions.Data/Services/RegistrationService.cs
SteveJacksonSoft: I'll see where the other conversation goes before committing to anything on this.

DigitalLearningSolutions.Data/Services/RegistrationService.cs
AlexJacksonDS: -2 is missing parameters `len(@FirstName) = 0 or len(@LastName) = 0 or @JobGroupID < 1 or @JobGroupID > (SELECT     MAX(JobGroupID) AS JobGroupID FROM JobGroups)`

-3 is existing aliasId, which we aren't setting here

AlexJacksonDS (approved):

SteveJacksonSoft (approved):

PR #615, opened 2021-09-21
Heedls 460 usage stats download

JBsoftwire:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-460

### Description
A button has been added to the centre Reports page. If pressed, it will download 

### Screenshots
image
image

-----
### Developer checks
I have:
- [X] Written tests for the changes (controller, data services, services, view models etc) and manually tested my work with and without JavaScript. Full manual testing guidelines can be found here: https://softwiretech.atlassian.net/wiki/spaces/HEE/pages/6703648740/Testing
- [X] Updated/added documentation in Swiki and/or Readme.


DigitalLearningSolutions.Web/Controllers/TrackingSystem/Centre/Reports/ReportsController.cs
```diff
@@ -76,5 +77,41 @@ public IEnumerable<ActivityDataRowModel> GetGraphData()
...
+                adminUser.CategoryId == 0 ? courseCategoryId : adminUser.CategoryId,
+                customisationId,
+                reportInterval
+            );
+
+            var dataFile = activityService.GetActivityDataFileForCentre(centreId, filterData);
+
+            return File(
+                dataFile,
+                "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
```
davidm-m, line 112: Is this worth storing as a variable somewhere in case other methods need to return spreadsheets?

davidm-m (approved):
All looks good! Just got one small question.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Centre/Reports/ReportsController.cs
JBsoftwire: I copied it from somewhere else, so definitely yes. Should have pulled it out as a constant at the time.

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Reports/Index.cshtml
```diff
@@ -37,6 +37,14 @@
...
     </div>
     <partial name="_ActivityTable.cshtml" model="@Model.UsageStatsTableViewModel" />
+    
+    <a
+      class="nhsuk-button nhsuk-u-margin-top-6"
+      asp-controller="Reports"
+      asp-action="DownloadUsageData"
+      asp-all-route-data="@Model.ReportsFilterModel.FilterValues"
+      role="button"
+    >Download usage data</a>
```
stellake, line 47: I think we usually format this like (not sure if the formatter catches this):
```
<a
      class="nhsuk-button nhsuk-u-margin-top-6"
      asp-controller="Reports"
      asp-action="DownloadUsageData"
      asp-all-route-data="@Model.ReportsFilterModel.FilterValues"
      role="button">
      Download usage data
</a>
```

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Centre/Reports/ReportsController.cs
```diff
@@ -81,5 +82,41 @@ public IEnumerable<ActivityDataRowModel> GetGraphData()
...
+            int? jobGroupId,
+            int? courseCategoryId,
+            int? customisationId,
+            string startDate,
+            string endDate,
+            ReportInterval reportInterval
+        )
+        {
+            var parsedStartDate = DateTime.Parse(startDate);
+            var parsedEndDate = DateTime.Parse(endDate);
```
stellake, line 98: We probably want to do some validation on date times and ensure that:
* Start date is not before the earliest activityLog
* End date is after start date and not in the future

We could also use DateTime rather than string for automatic binding if we do validation.

Can we return 404 if that's not the case?

DigitalLearningSolutions.Data.Tests/Services/ActivityServiceTests.cs
```diff
@@ -167,6 +170,72 @@ public void GetFilteredActivity_requests_activity_with_correct_parameters()
...
+                    () => activityDataService.GetFilteredActivity(
+                        A<int>._,
+                        A<DateTime>._,
+                        A<DateTime>._,
+                        A<int?>._,
+                        A<int?>._,
+                        A<int?>._
+                    )
+                )
+                .Returns(expectedActivityResult);
```
stellake, line 203: Could we maybe pull this into a private method to make the test easier to read - e.g. GivenActivityDataServiceReturnsDataInExampleSheet() 

There is more potential for breaking this down a bit to make it easier to read and the assertion looks very generic so could potentially be used in several places/belong to the helper.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Centre/Reports/ReportsController.cs
```diff
@@ -97,19 +142,29 @@ public IEnumerable<ActivityDataRowModel> GetGraphData()
...
+            if (parsedStartDate < activityService.GetStartOfActivityForCentre(centreId))
+            {
+                return new NotFoundResult();
+            }
+
+            if (DateTime.TryParse(endDate, out var parsedEndDate) &&
+                (parsedEndDate < parsedStartDate || parsedEndDate > DateTime.Now))
+            {
+                return new NotFoundResult();
+            }
```
stellake, line 159: Could we put this into a service? Potentially something like `getValidatedUsageStatsDateRange(startDateString, endDateString)` that returns `(startDate, endDate?)?`
Then we can return NotFound if the tuple is null. And the date validation will also mean that we can just pass in endDate to the ActivityFilterData and keep all the relevant logic in service (with nice small unit tests).

DigitalLearningSolutions.Data.Tests/Services/ActivityServiceTests.cs
```diff
@@ -344,6 +338,83 @@ private void GivenDataServicesReturnData(int centreId, int categoryId)
...
+                    var expectedWorksheet = expectedWorkbook.Worksheets.Worksheet(resultWorksheet.Name);
+                    var cells = resultWorksheet.CellsUsed();
+                    cells.Count().Should().Be(expectedWorksheet.CellsUsed().Count());
+
+                    foreach (var cell in cells)
+                    {
+                        var expectedCell = expectedWorksheet.Cell(cell.Address);
+                        cell.Value.Should().BeEquivalentTo(expectedCell.Value);
+                    }
+                }
```
stellake, line 381: Could we pull this into a separate method or even a helper if it's useful in other Excel comparison tests.

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/Centre/Reports/ReportsViewModel.cs
```diff
@@ -71,6 +71,15 @@ bool userManagingAllCourses
             DateRange =
                 $"{filterData.StartDate.ToString(DateHelper.StandardDateFormat)} - {filterData.EndDate?.ToString(DateHelper.StandardDateFormat) ?? "Today"}";
             ShowCourseCategoryFilter = userManagingAllCourses;
+            FilterValues = new Dictionary<string, string>
+            {
+                { "jobGroupId", filterData.JobGroupId?.ToString() ?? "" },
+                { "courseCategoryId", filterData.CourseCategoryId?.ToString() ?? "" },
+                { "customisationId", filterData.CustomisationId?.ToString() ?? "" },
+                { "startDate", filterData.StartDate.ToString() },
+                { "endDate", filterData.EndDate.ToString() },
```
stellake, line 80: What if endDate is null?

DigitalLearningSolutions.Data/Services/ActivityService.cs
```diff
@@ -126,6 +132,28 @@ public byte[] GetActivityDataFileForCentre(int centreId, ActivityFilterData filt
...
         }
 
+        public (DateTime startDate, DateTime? endDate)? GetValidatedUsageStatsDateRange(
+            string startDateString,
+            string endDateString,
+            int centreId
+        )
+        {
+            var parsedStartDate = DateTime.Parse(startDateString);
+            var start = GetStartOfActivityForCentre(centreId);
```
stellake, line 142: Is this used?

DigitalLearningSolutions.Data/Services/ActivityService.cs
```diff
@@ -126,6 +132,28 @@ public byte[] GetActivityDataFileForCentre(int centreId, ActivityFilterData filt
             return stream.ToArray();
         }
 
+        public (DateTime startDate, DateTime? endDate)? GetValidatedUsageStatsDateRange(
+            string startDateString,
+            string endDateString,
+            int centreId
+        )
+        {
+            var parsedStartDate = DateTime.Parse(startDateString);
```
stellake, line 141: Could this just be startDate? Also what happens if parsing fails - we should just handle it as not found rather than throwing

DigitalLearningSolutions.Data/Services/ActivityService.cs
```diff
@@ -126,6 +132,28 @@ public byte[] GetActivityDataFileForCentre(int centreId, ActivityFilterData filt
...
+        {
+            var parsedStartDate = DateTime.Parse(startDateString);
+            var start = GetStartOfActivityForCentre(centreId);
+            if (parsedStartDate < GetStartOfActivityForCentre(centreId))
+            {
+                return null;
+            }
+
+            if (DateTime.TryParse(endDateString, out var parsedEndDate) &&
+                (parsedEndDate < parsedStartDate || parsedEndDate > DateTime.Now))
```
stellake, line 149: This is quite long and hard to read, maybe we can pull `var isEndDateSet `into a separate var?

DigitalLearningSolutions.Data/Services/ActivityService.cs
```diff
@@ -126,6 +132,28 @@ public byte[] GetActivityDataFileForCentre(int centreId, ActivityFilterData filt
...
+            {
+                return null;
+            }
+
+            return (parsedStartDate, parsedEndDate > DateTime.MinValue ? (DateTime?)parsedEndDate : null);
```
stellake, line 154: Do we need to check for minvalue if it has to be after start date anyway?

JBsoftwire: It only checks if it is after the start date if the end date is set. TryParse returns DateTime.MinValue for invalid or empty strings. I think it'll be clearer if I move this check to right after the parse however.

stellake (approved):
Looks good! 👍

PR #636, opened 2021-09-24
HEEDLS-585 All delegates - welcome emails - filter - fix

ibrahimmunir14:
### JIRA link
[HEEDLS-585](https://softwiretech.atlassian.net/browse/HEEDLS-585)

### Description
- Fix JS bug: If I add a filter, and then go to add a different filter, but leave it on the default value of “Select [filter name]” and hit the Add Filter button then nothing happens as expected. However, I then cannot add any new filters until I clear all the filters.
- Word break for really long delegate checkbox labels
- Read out results count when selecting/deselect all

### Screenshots
Word wrap:

-----
### Developer checks
I have:

### DIFF
--- a/DigitalLearningSolutions.Web/Scripts/trackingSystem/emailDelegates.ts
+++ b/DigitalLearningSolutions.Web/Scripts/trackingSystem/emailDelegates.ts
@@ -8,11 +8,18 @@ const route = `TrackingSystem/Delegates/Email/AllEmailDelegateItems?${queryParam
 // eslint-disable-next-line no-new
 new SearchSortFilterAndPaginate(route, false, false, true, 'EmailDelegateFilter');
 
+function alertResultCount(): void {
+  const resultCount = document.getElementById('results-count') as HTMLSpanElement;
+  resultCount.setAttribute('role', '');
+  resultCount.setAttribute('role', 'alert');
+}
+
 function selectAll(): void {
   const allCheckboxes = document.querySelectorAll('.delegate-checkbox') as NodeListOf<HTMLInputElement>;
   allCheckboxes.forEach((checkbox) => {
     if (!checkbox.checked) checkbox.click();
   });
+  alertResultCount();
 }
 
 function deselectAll(): void {
@@ -20,6 +27,7 @@ function deselectAll(): void {
   allCheckboxes.forEach((checkbox) => {
     if (checkbox.checked) checkbox.click();
   });
+  alertResultCount();
 }
 
 const selectAllForm = document.getElementById('select-all-form') as HTMLFormElement;
--- a/DigitalLearningSolutions.Web/Scripts/trackingSystem/emailDelegates.ts
+++ b/DigitalLearningSolutions.Web/Scripts/trackingSystem/emailDelegates.ts
@@ -7,7 +7,7 @@ const route = `TrackingSystem/Delegates/Email/AllEmailDelegateItems?${queryParam
 
 // eslint-disable-next-line no-new
 new SearchSortFilterAndPaginate(route, false, false, true, 'EmailDelegateFilter');
-setupSelectAndDeselectButtons();
+setUpSelectAndDeselectButtons();
 
 function alertResultCount(): void {
   // un-assign and re-assign the alert role to results-count so a screen-reader re-reads it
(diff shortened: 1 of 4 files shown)


DigitalLearningSolutions.Web/Styles/trackingSystem/delegates/emailDelegates.scss
```diff
@@ -5,3 +5,7 @@
   border-radius: 4px;
   margin-right: nhsuk-spacing(2);
 }
+
+.delegate-label {
```
AlexJacksonDS, line 9: This fix won't work as expected in IE, use the word-break class on the elements (This is defined in index.scss)

DigitalLearningSolutions.Web/Scripts/trackingSystem/emailDelegates.ts
```diff
@@ -8,18 +8,26 @@ const route = `TrackingSystem/Delegates/Email/AllEmailDelegateItems?${queryParam
...
+  resultCount.setAttribute('role', '');
+  resultCount.setAttribute('role', 'alert');
+}
+
 function selectAll(): void {
   const allCheckboxes = document.querySelectorAll('.delegate-checkbox') as NodeListOf<HTMLInputElement>;
   allCheckboxes.forEach((checkbox) => {
     if (!checkbox.checked) checkbox.click();
   });
+  alertResultCount();
```
AlexJacksonDS, line 22: I think these alert calls need to go in the event listener declaration. It's not a part of the select all, it belongs to the button press i.e. the button action becomes "Select all then alert the user"
I've moved the selectAll functions to a common place as part of 441, and not all pages will necessarily have the results count element to be able to do this and will need an alternative way to announce it if we want this functionality.

AlexJacksonDS (changes requested):
A couple of minor things that need modifying.

AlexJacksonDS (approved):

DigitalLearningSolutions.Web/Scripts/trackingSystem/emailDelegates.ts
```diff
@@ -27,5 +33,11 @@ const selectAllButton = document.getElementById('select-all-button') as HTMLButt
 const deselectAllButton = document.getElementById('deselect-all-button') as HTMLButtonElement;
 
 selectAllForm.addEventListener('submit', (e) => e.preventDefault());
-selectAllButton.addEventListener('click', selectAll);
-deselectAllButton.addEventListener('click', deselectAll);
+selectAllButton.addEventListener('click', () => {
+  selectAll();
+  alertResultCount();
+});
+deselectAllButton.addEventListener('click', () => {
```
stellake, line 40: Minor, but missing an empty line above.

Also, can we wrap all this with setUpSelectAndDeselectButtons() function and call that at the top of the page (so devs don't have to scroll through the whole file to see what is happening on page load).

DigitalLearningSolutions.Web/Scripts/trackingSystem/emailDelegates.ts
```diff
@@ -8,6 +8,12 @@ const route = `TrackingSystem/Delegates/Email/AllEmailDelegateItems?${queryParam
 // eslint-disable-next-line no-new
 new SearchSortFilterAndPaginate(route, false, false, true, 'EmailDelegateFilter');
 
+function alertResultCount(): void {
+  const resultCount = document.getElementById('results-count') as HTMLSpanElement;
+  resultCount.setAttribute('role', '');
+  resultCount.setAttribute('role', 'alert');
```
stellake, line 14: What does this do? Maybe worth a comment if this is a hack to let screen reader announce the message/or rename to announceNewResultCount()?
Alert usually means popup on a page in terms of JS (try the alert method in console):
image

ibrahimmunir14: If a html element has role=alert it gets read by the screen reader. In the filtering stuff, whenever a filter is changed, the text on the element changes which makes it re-read. Here the text isn't changing, so we can just unassign and reassign the role alert to make it re-read.

DigitalLearningSolutions.Web/Scripts/trackingSystem/emailDelegates.ts
```diff
@@ -7,8 +7,10 @@ const route = `TrackingSystem/Delegates/Email/AllEmailDelegateItems?${queryParam
 
 // eslint-disable-next-line no-new
 new SearchSortFilterAndPaginate(route, false, false, true, 'EmailDelegateFilter');
+setupSelectAndDeselectButtons();
```
stellake, line 10: Super minor, but given this is a verb, it should technically be "setUp" re capitalisation 👁️

stellake (approved):
Super minor comment about capitalisation, but no need for rereview

PR #648, opened 2021-09-30
HEEDLS-585 - fix inconsistent announcement of result counts

DanBloxham-sw:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-585

### Description
Changed the JS for the Select/Deselect All buttons so that the result count is actually read out consistently across browsers. I've also taken the liberty to fix a similar issue with the JS search/filter result announcements. See comments on individual files for details.

### Screenshots
N/A

-----
### Developer checks

I have:

### DIFF
--- a/DigitalLearningSolutions.Web/Scripts/searchSortFilterAndPaginate/searchSortFilterAndPaginate.ts
+++ b/DigitalLearningSolutions.Web/Scripts/searchSortFilterAndPaginate/searchSortFilterAndPaginate.ts
@@ -161,9 +161,15 @@ export class SearchSortFilterAndPaginate {
 
   static updateResultCount(count: number): void {
     const resultCount = <HTMLSpanElement>document.getElementById('results-count');
+    const oldResultCountMessage = resultCount.innerHTML;
     resultCount.hidden = false;
     resultCount.setAttribute('aria-hidden', 'false');
-    resultCount.textContent = count === 1 ? '1 matching result' : `${count.toString()} matching results`;
+    let newResultCountMessage = count === 1 ? '1 matching result' : `${count.toString()} matching results`;
+
+    if (newResultCountMessage === oldResultCountMessage) {
+      newResultCountMessage += '&nbsp';
+    }
+    resultCount.innerHTML = newResultCountMessage;
   }
 
   private static scrollToTop() : void {
--- a/DigitalLearningSolutions.Web/Scripts/trackingSystem/emailDelegates.ts
+++ b/DigitalLearningSolutions.Web/Scripts/trackingSystem/emailDelegates.ts
@@ -10,10 +10,17 @@ new SearchSortFilterAndPaginate(route, false, false, true, 'EmailDelegateFilter'
 setUpSelectAndDeselectButtons();
 
 function alertResultCount(): void {
-  // un-assign and re-assign the alert role to results-count so a screen-reader re-reads it
+  // change the content of the search results
+  // so that an "identical" result is announced by the aria-live attribute
   const resultCount = document.getElementById('results-count') as HTMLSpanElement;
-  resultCount.setAttribute('role', '');
-  resultCount.setAttribute('role', 'alert');
+  const resultCountMessage = resultCount.innerHTML;
+  const indexOfSpace = resultCountMessage.search('&nbsp');
+
+  if (indexOfSpace === -1) {
+    resultCount.innerHTML = `${resultCountMessage}&nbsp`;
+  } else {
+    resultCount.innerHTML = resultCountMessage.substring(0, indexOfSpace);
+  }
 }
 
 function selectAll(): void {
--- a/DigitalLearningSolutions.Web/Views/Shared/SearchablePage/_SearchResultsCount.cshtml
+++ b/DigitalLearningSolutions.Web/Views/Shared/SearchablePage/_SearchResultsCount.cshtml
@@ -5,5 +5,5 @@
   @{
     var count = Model.MatchingSearchResults;
   }
-  <span role="alert" id="results-count">@count matching @(count == 1 ? "result" : "results")</span>
+  <span aria-live="polite" id="results-count">@count matching @(count == 1 ? "result" : "results")</span>
 </p>


DigitalLearningSolutions.Web/Views/Shared/SearchablePage/_SearchResultsCount.cshtml
```diff
@@ -5,5 +5,5 @@
   @{
     var count = Model.MatchingSearchResults;
   }
-  <span role="alert" id="results-count">@count matching @(count == 1 ? "result" : "results")</span>
+  <span aria-live="polite" id="results-count">@count matching @(count == 1 ? "result" : "results")</span>
```
DanBloxham-sw, line 8: I've replaced the role="alert" with aria-live="polite" as it seems more appropriate given the description in the docs:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

DigitalLearningSolutions.Web/Scripts/trackingSystem/emailDelegates.ts
```diff
@@ -10,10 +10,17 @@ new SearchSortFilterAndPaginate(route, false, false, true, 'EmailDelegateFilter'
...
-  resultCount.setAttribute('role', '');
-  resultCount.setAttribute('role', 'alert');
+  const resultCountMessage = resultCount.innerHTML;
+  const indexOfSpace = resultCountMessage.search('&nbsp');
+
+  if (indexOfSpace === -1) {
+    resultCount.innerHTML = `${resultCountMessage}&nbsp`;
+  } else {
+    resultCount.innerHTML = resultCountMessage.substring(0, indexOfSpace);
+  }
```
DanBloxham-sw, line 23: Removing and reassigning the alert role seemed to be inconsistently picked up by browsers, if at all. As the alert role (and the potential new replacement `aria-live `) both announce any changes to the content, the reliable way to get them to announce the number of results would be changing the content. 

In this case, we are simply selecting check boxes, so there are no changes to the number of results. In order to force the button press to announce the number of results (so that it is obvious the button has actually done something), we add a no-break space to the end of the message (or remove it if it already exists) so that there is a clear change in the message so it is read out.

DigitalLearningSolutions.Web/Scripts/searchSortFilterAndPaginate/searchSortFilterAndPaginate.ts
```diff
@@ -161,9 +161,15 @@ export class SearchSortFilterAndPaginate {
...
+    const oldResultCountMessage = resultCount.innerHTML;
     resultCount.hidden = false;
     resultCount.setAttribute('aria-hidden', 'false');
-    resultCount.textContent = count === 1 ? '1 matching result' : `${count.toString()} matching results`;
+    let newResultCountMessage = count === 1 ? '1 matching result' : `${count.toString()} matching results`;
+
+    if (newResultCountMessage === oldResultCountMessage) {
+      newResultCountMessage += '&nbsp';
+    }
+    resultCount.innerHTML = newResultCountMessage;
```
DanBloxham-sw, line 172: This applies to all of the search/filter stuff where we have a results count. The results count was previously read out, but only when the content of the message (i.e. the count) changed. This change works in the same way as the other comment to make it the content "change" so it is read out when no difference in the count actually happens.

DigitalLearningSolutions.Web/Views/Shared/SearchablePage/_SearchResultsCount.cshtml
```diff
@@ -5,5 +5,5 @@
   @{
     var count = Model.MatchingSearchResults;
   }
-  <span role="alert" id="results-count">@count matching @(count == 1 ? "result" : "results")</span>
+  <span role="alert" aria-live="polite" id="results-count">@count matching @(count == 1 ? "result" : "results")</span>
```
DanBloxham-sw, line 8: I originally replaced the `role="alert"` with `aria-live="polite" `as it seems more appropriate given the description in the docs:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions
However, we need the `role="alert"` to announce the results when there is no javascript enabled. Seeing as how `role="alert" ` implements the same functionality as the `aria-live="polite"`, should I just get rid of the new addition and keep this how it was?

DigitalLearningSolutions.Web/Scripts/searchSortFilterAndPaginate/searchSortFilterAndPaginate.ts
SteveJacksonSoft: Looks good, can we put the new stuff in a function of its own? `modifyResultCountHtmlForScreenReaderIfUnchanged` or something?

DigitalLearningSolutions.Web/Scripts/trackingSystem/emailDelegates.ts
SteveJacksonSoft: As with the other one, can we put the new stuff in a function of its own?

DigitalLearningSolutions.Web/Views/Shared/SearchablePage/_SearchResultsCount.cshtml
SteveJacksonSoft: If this doesn't change the ux at all, I don't think we should bother adding it. Extra code to understand and maintain.

DigitalLearningSolutions.Web/Scripts/searchSortFilterAndPaginate/searchSortFilterAndPaginate.ts
```diff
@@ -161,15 +161,28 @@ export class SearchSortFilterAndPaginate {
 
   static updateResultCount(count: number): void {
     const resultCount = <HTMLSpanElement>document.getElementById('results-count');
-    const oldResultCountMessage = resultCount.innerHTML;
     resultCount.hidden = false;
     resultCount.setAttribute('aria-hidden', 'false');
     let newResultCountMessage = count === 1 ? '1 matching result' : `${count.toString()} matching results`;
```
SteveJacksonSoft, line 166: This can go in the new function.

DigitalLearningSolutions.Web/Scripts/searchSortFilterAndPaginate/searchSortFilterAndPaginate.ts
```diff
@@ -161,15 +161,28 @@ export class SearchSortFilterAndPaginate {
...
-    const oldResultCountMessage = resultCount.innerHTML;
     resultCount.hidden = false;
     resultCount.setAttribute('aria-hidden', 'false');
     let newResultCountMessage = count === 1 ? '1 matching result' : `${count.toString()} matching results`;
+    newResultCountMessage = this.modifyResultCountMessageForScreenReaderIfUnchanged(
+      newResultCountMessage,
+      resultCount,
+    );
+
+    resultCount.innerHTML = newResultCountMessage;
```
SteveJacksonSoft, line 172: Does the html not get modified if you do this in the new function?

DigitalLearningSolutions.Web/Scripts/searchSortFilterAndPaginate/searchSortFilterAndPaginate.ts
DanBloxham-sw: Sorry, which lines specifically? Just 166: let newResultCountMessage... ?

SteveJacksonSoft: Think so yep. You're only calculating newResultCoundMessage to pass it into a function right? Which means you can just calculate it in the function.

DigitalLearningSolutions.Web/Scripts/searchSortFilterAndPaginate/searchSortFilterAndPaginate.ts
SteveJacksonSoft: Hmm, you can either get `resultCount` from the dom again within the new method (and extract a `getResultCount()` function), or you can rename the method to `GetNewResultCountMessage` or something. I think I'm okay with either, but in the latter case it would be worth adding a comment to explain what you're doing with the nbsp.

DigitalLearningSolutions.Web/Scripts/searchSortFilterAndPaginate/searchSortFilterAndPaginate.ts
```diff
@@ -163,22 +163,23 @@ export class SearchSortFilterAndPaginate {
...
-    newResultCountMessage: string,
+  static getNewResultCountMessage(
+    count: number,
     resultCountElement: HTMLSpanElement,
   ): string {
     const oldResultCountMessage = resultCountElement.innerHTML;
+    const newResultCountMessage = count === 1 ? '1 matching result' : `${count.toString()} matching results`;
 
     if (newResultCountMessage === oldResultCountMessage) {
+      // add a non-breaking space if messages are the same to cause the screen reader to announce it
```
SteveJacksonSoft, line 182: You could explain more clearly that the screen reader doesn't make an announcement if the message doesn't change - it's only implied here. And I think that's all you need to say actually - what you're doing is obvious from the code.

PR #649, opened 2021-09-30
Heedls 615 indicate selected page

livzorn:
### JIRA link
[_HEEDLS-615_](https://softwiretech.atlassian.net/browse/HEEDLS-615)

### Description
I've added HTML helper extensions that ensure the correct tab is visually indicated as selected, even if the user clicks into another page or journey that is under the scope of that tab. 

Previously, the HTML helper was only checking for an action name. Since some of the tabs have multiple controllers within their scope, and some have different actions within the same controller as another tab, I've created two helpers, IsActionSelected and IsControllerSelected, which check whether the current action or current controller matches the one for that tab. IsControllerSelected has an optional exceptAction param to account for the Frameworks application having all the Frameworks controller content under the Frameworks tab with the exception of the Index action being under the Dashboard tab. 

Since the MyAccount tab is on every application, I've created an additional IsMyAccountTabSelected helper to simplify the code.

### Screenshots
localhost_5001_ChangePassword

-----
### Developer checks

I have:

### DIFF
--- a/DigitalLearningSolutions.Web/Extensions/HtmlHelperExtensions.cs
+++ b/DigitalLearningSolutions.Web/Extensions/HtmlHelperExtensions.cs
@@ -2,7 +2,6 @@
 {
     using System.Collections.Generic;
     using System.Linq;
-    using DigitalLearningSolutions.Web.ViewModels.Common.ViewComponents;
     using Microsoft.AspNetCore.Mvc.Rendering;
 
     public static class HtmlHelperExtensions
@@ -21,21 +20,6 @@
             }
 
             var returnStatement = selectedTab.ToString() == currentTab
-                ? selectedCssClass
-                : string.Empty;
-            return returnStatement;
-        }
-
-        public static string IsTabSelected(
-            this IHtmlHelper htmlHelper,
-            NavMenuAndViewDataViewModel model,
-            string currentTab,
-            string selectedCssClass = "selected"
-        )
-        {
-            var selectedTab = model.SelectedTab;
-
-            var returnStatement = selectedTab == currentTab
                 ? selectedCssClass
                 : string.Empty;
             return returnStatement;
--- a/DigitalLearningSolutions.Data/Enums/Tab.cs
+++ b/DigitalLearningSolutions.Data/Enums/Tab.cs
@@ -1,10 +1,26 @@
 ﻿namespace DigitalLearningSolutions.Data.Enums
 {
-    public class Tab
+    public enum Tab
     {
         Welcome,
         FindYourCentre,
         Pricing,
-        MyAccount
+        MyAccount,
+        ApplicationSelector,
+        Register,
+        LogIn,
+        Centre,
+        Delegates,
+        CourseSetup,
+        Admin,
+        Support,
+        Dashboard,
+        MyStaff,
+        ProfileAssessments,
+        Frameworks,
+        RoleProfiles,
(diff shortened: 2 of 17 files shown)


DigitalLearningSolutions.Web/Extensions/HtmlHelperExtensions.cs
```diff
@@ -1,20 +1,58 @@
...
+            string? exceptAction = null,
+            string selectedCssClass = "selected"
+        )
+        {
+            if (exceptAction != null)
+            {
+                var currentAction = htmlHelper.ViewContext.RouteData.Values["action"] as string;
+                if (currentAction == exceptAction)
+                {
+                    return "";
```
AlexJacksonDS, line 36: string.Empty

DigitalLearningSolutions.Web/Views/TrackingSystem/Shared/_NavMenuItems.cshtml
```diff
@@ -1,43 +1,48 @@
...
+@using DigitalLearningSolutions.Web.Helpers
 @using DigitalLearningSolutions.Web.Models.Enums
+@using Microsoft.AspNetCore.Mvc.TagHelpers
 @using Microsoft.Extensions.Configuration
 @inject IConfiguration Configuration
 
 @if (User.HasCentreAdminPermissions()) {
-  <li class="nhsuk-header__navigation-item">
+  <li class="nhsuk-header__navigation-item
+    @Html.IsControllerSelected(new[] { "Dashboard", "Configuration", "Administrator", "Reports" })">
```
AlexJacksonDS, line 10: This doesn't cover every controller in the dashboard area

DigitalLearningSolutions.Web/Views/TrackingSystem/Shared/_NavMenuItems.cshtml
```diff
@@ -1,43 +1,48 @@
-﻿@using DigitalLearningSolutions.Web.Helpers
+﻿@using DigitalLearningSolutions.Web.Extensions
+@using DigitalLearningSolutions.Web.Helpers
 @using DigitalLearningSolutions.Web.Models.Enums
+@using Microsoft.AspNetCore.Mvc.TagHelpers
```
AlexJacksonDS, line 4: I think this using statement keeps getting added but isn't actually needed?

DigitalLearningSolutions.Web/Extensions/HtmlHelperExtensions.cs
```diff
@@ -1,20 +1,58 @@
...
-            return (controller == null || controller == currentController) && action == currentAction
+            return (controller == null || controller == currentController) && actions.Contains(currentAction)
+                ? selectedCssClass
+                : string.Empty;
+        }
+
+        public static string IsControllerSelected(
+            this IHtmlHelper htmlHelper,
+            IEnumerable<string> controllers,
+            string? exceptAction = null,
```
AlexJacksonDS, line 27: Is this ever likely to need a list?

AlexJacksonDS (commented):
Just actually leaving the comment of the things I think need changing if we were to keep this approach. As I mentioned on slack, I do think it is worth considering other options to set this up, let me know if you want to discuss any different options further. There's a chance we could potentially tidy up a fair bit of stuff in the views depending on what approach we decide on.

PR #650, opened 2021-09-30
HEEDLS-603 - set up supervisor delegate relation when bulk uploading …

DanBloxham-sw:
…delegates

### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-603

### Description
Added some calls to the supervisor delegate service during the bulk upload process for registrations.

### Screenshots
N/A

-----
### Developer checks

I have:

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Services/DelegateUploadFileServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/DelegateUploadFileServiceTests.cs
@@ -653,7 +653,7 @@ namespace DigitalLearningSolutions.Data.Tests.Services
             catch (ArgumentOutOfRangeException ex)
             {
                 // Then
-                ex.Message.Should().Be($"Unknown return value when creating delegate record. (Parameter 'statusOrCandidateNumber')\r\nActual value was {failureStatusCode}.");
+                ex.Message.Should().Be($"Unknown return value when creating delegate record. (Parameter 'errorCodeOrCandidateNumber')\r\nActual value was {failureStatusCode}.");
                 ex.ActualValue.Should().Be(failureStatusCode);
             }
             finally
--- a/DigitalLearningSolutions.Data.Tests/Services/DelegateUploadFileServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/DelegateUploadFileServiceTests.cs
@@ -48,6 +48,10 @@ namespace DigitalLearningSolutions.Data.Tests.Services
             registrationDataService = A.Fake<IRegistrationDataService>(x => x.Strict());
             supervisorDelegateService = A.Fake<ISupervisorDelegateService>();
 
+            A.CallTo(() => userDataService.GetDelegateUserByCandidateNumber(A<string>._, A<int>._))
+                .Returns(UserTestHelper.GetDefaultDelegateUser());
+
+
             delegateUploadFileService = new DelegateUploadFileService(
                 jobGroupsDataService,
                 userDataService,
@@ -649,7 +653,7 @@ namespace DigitalLearningSolutions.Data.Tests.Services
             catch (ArgumentOutOfRangeException ex)
             {
                 // Then
-                ex.Message.Should().Be($"Unknown return value when creating delegate record. (Parameter 'status')\r\nActual value was {failureStatusCode}.");
+                ex.Message.Should().Be($"Unknown return value when creating delegate record. (Parameter 'statusOrCandidateNumber')\r\nActual value was {failureStatusCode}.");
                 ex.ActualValue.Should().Be(failureStatusCode);
             }
             finally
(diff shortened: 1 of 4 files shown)


DigitalLearningSolutions.Data/Services/SupervisorDelegateService.cs
```diff
@@ -20,23 +19,23 @@ public interface ISupervisorDelegateService
     public class SupervisorDelegateService : ISupervisorDelegateService
     {
         private readonly ISupervisorDelegateDataService supervisorDelegateDataService;
-        private readonly IUserDataService userDataService;
 
         public SupervisorDelegateService(
-            ISupervisorDelegateDataService supervisorDelegateDataService,
-            IUserDataService userDataService
```
DanBloxham-sw, line 27: I saw that this userDataService wasn't being used so I removed it.

DigitalLearningSolutions.Data.Tests/Services/DelegateUploadFileServiceTests.cs
```diff
@@ -618,6 +622,104 @@ public void ProcessDelegateTable_calls_register_with_expected_values_when_welcom
...
+        {
+            // Given
+            const string aliasId = "ALIAS";
+            var row = GetSampleDelegateDataRow(candidateNumber: string.Empty, aliasId: aliasId);
+            var table = CreateTableFromData(new[] { row });
+
+            A.CallTo(() => userDataService.GetDelegateUserByAliasId(aliasId, CentreId)).Returns(null);
+            A.CallTo(() => userService.IsDelegateEmailValidForCentre("email@test.com", CentreId)).Returns(true);
+            A.CallTo(() => registrationDataService.RegisterDelegateByCentre(A<DelegateRegistrationModel>._))
+                .Returns(failureStatusCode);
```
livzorn, line 642: Confused by the variable name a bit, does RegisterDelegateByCentre actually return a status code? I see the return variable in the method is called candidateNumber

livzorn (commented):
Looks good, just one question

DigitalLearningSolutions.Data.Tests/Services/DelegateUploadFileServiceTests.cs
DanBloxham-sw: RegisterDelegateByCentre calls a stored procedure and returns what it returns. This can either by a candidate number if a delegate is registered successfully, or a non-zero status code if something goes wrong. These tests are specifically testing those error status codes prevent further processing.

livzorn: Understood, thanks for the clarification! :)

livzorn (approved):

DigitalLearningSolutions.Data/Services/DelegateUploadFileService.cs
```diff
@@ -199,11 +202,33 @@ private void RegisterDelegate(DelegateTableRow delegateRow, DateTime? welcomeEma
...
+                    centreId,
+                    emailAddress
+                ).Select(supervisor => supervisor.ID).ToList();
+
+            if (!pendingSupervisorDelegateIds.Any())
+            {
+                return;
+            }
+
+            var newDelegateRecord = userDataService.GetDelegateUserByCandidateNumber(candidateNumber, centreId)!;
```
SteveJacksonSoft, line 224: I think this would be better placed outside this method.
We've no guarantee that this method will only be called just after the delegate has successfully been registered, so can't safely use the null assertion here. But if we move it outside the method, into the switch statement, we can.
Also the method is kind of doing two things currently: looking up the delegate matching the candidate and updating supervisor-delegate records.

SteveJacksonSoft (changes requested):

DigitalLearningSolutions.Data/Services/DelegateUploadFileService.cs
```diff
@@ -187,8 +187,8 @@ private void UpdateDelegate(DelegateTableRow delegateRow, DelegateUser delegateU
         private void RegisterDelegate(DelegateTableRow delegateRow, DateTime? welcomeEmailDate, int centreId)
         {
             var model = new DelegateRegistrationModel(delegateRow, centreId, welcomeEmailDate);
-            var status = registrationDataService.RegisterDelegateByCentre(model);
-            switch (status)
+            var statusOrCandidateNumber = registrationDataService.RegisterDelegateByCentre(model);
+            switch (statusOrCandidateNumber)
```
SteveJacksonSoft, line 191: Ah, good point - could we just tweak this to errorCodeOrCandidateNumber? That makes it a bit clearer that it's not a candidate number when an error's occurred. Currently it's not so obvious when it's a status and when it's a candidate number. No need for re-review. Unless it demands approval before letting you merge.

DanBloxham-sw: Changed to errorCodeOrCandidateNumber. No need for re-review, so I will merge it once the checks pass.

PR #655, opened 2021-10-01
Heedls 501 remove course

JBsoftwire:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-501

### Description
The 'remove' button on delegate course cards now goes to a page where the removal can be confirmed. If it is, the delegate's current progress on that course is marked as removed.

### Screenshots
image
image

-----
### Developer checks

I have:
- [*] Run the formatter and made sure there are no IDE errors.
- [*] Written tests for the changes (controller, data services, services, view models etc) and manually tested my work with and without JavaScript. Full manual testing guidelines can be found here: https://softwiretech.atlassian.net/wiki/spaces/HEE/pages/6703648740/Testing
- [*] Scanned over my own MR to ensure everything is as expected.


DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/ViewDelegate/ConfirmRemoveFromCourse.cshtml
```diff
@@ -0,0 +1,47 @@
...
+  var errorHasOccurred = !ViewData.ModelState.IsValid;
+  ViewData["Title"] = $"Remove delegate enrolment from {Model.CourseName}";
+  ViewData["Application"] = "Tracking System";
+  ViewData["HeaderPath"] = $"{Configuration["AppRootPath"]}/TrackingSystem/Centre/Dashboard";
+  ViewData["HeaderPathName"] = "Tracking System";
+}
+
+@section NavMenuItems {
+  <partial name="~/Views/TrackingSystem/Shared/_NavMenuItems.cshtml" />
+}
```
AlexJacksonDS, line 16: We should be using the new ApplicationType enumeration for these ViewData values now, and using the AutoNavMenuItems partial.

```
ViewData["Application"] = ApplicationType.TrackingSystem.HeaderExtension;
ViewData["HeaderPath"] = ApplicationType.TrackingSystem.HeaderPath;
ViewData["HeaderPathName"] = ApplicationType.TrackingSystem.HeaderPathName;
```

DigitalLearningSolutions.Data/Services/CourseService.cs
```diff
@@ -52,5 +60,18 @@ public IEnumerable<DelegateCourseDetails> GetAllCoursesForDelegate(int delegateI
                 }
             );
         }
+
+        public bool RemoveDelegateFromCourse(int delegateId, int customisationId, RemovalMethod removalMethod)
```
AlexJacksonDS, line 64: We should probably name this method something that indicates it's usage and the fact that it can return false and not actually mark a progress record as removed.

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/BasicAuthenticatedAccessibilityTests.cs
```diff
@@ -52,6 +52,7 @@ public class BasicAuthenticatedAccessibilityTests : AccessibilityTestsBase,
         [InlineData("/TrackingSystem/Delegates/Groups/Add", "Add new delegate group")]
         [InlineData("/TrackingSystem/Delegates/3/View", "xxxx xxxxxx")]
         [InlineData("/TrackingSystem/Delegates/3/SetPassword", "Set delegate user password")]
+        [InlineData("/TrackingSystem/Delegates/3/View/100/Remove", "Remove delegate enrolment from Entry Level - Win XP, Office 2003/07 OLD - Standard")]
```
AlexJacksonDS, line 55: Formatting? I'm surprised this hasn't been broken across new lines like some of the above cases.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/ViewDelegateController.cs
```diff
@@ -82,7 +88,59 @@ public IActionResult DeactivateDelegate(int delegateId)
...
+            if (delegateUser == null || delegateUser.CentreId != centreId)
+            {
+                return new NotFoundResult();
+            }
+
+            var course = courseDataService.GetCourseNameAndApplication(customisationId);
+            if (course == null)
+            {
+                return new NotFoundResult();
+            }
```
AlexJacksonDS, line 109: Is it worth shuffling these around so we only have one return NotFoundResult?

AlexJacksonDS (changes requested):
A few things to change relating to new set up/naming/formatting.

Could we also have some unit tests for the controller to check that the correct results are being returned? e.g. the Not Found on failed failure etc.

AlexJacksonDS (approved):

DigitalLearningSolutions.Data/Services/CourseService.cs
```diff
@@ -159,5 +168,18 @@ public void UpdateCourseOptions(CourseOptions courseOptions, int customisationId
 
             return new DelegateCourseDetails(info, customPrompts, attemptStats);
         }
+
+        public bool RemoveDelegateFromCourseIfDelegateHasCurrentProgress(int delegateId, int customisationId, RemovalMethod removalMethod)
+        {
+            var currentProgress = progressDataService.GetDelegateProgressForCourse(delegateId, customisationId)
+                .FirstOrDefault(p => p.Completed == null && p.RemovedDate == null);
```
DanBloxham-sw, line 175: The GroupService is the only other place where ProgressDataService.GetDelegateProgressForCourse is currently used, but in there we do not use .FirstOrDefault() despite the fact that there should only ever be one record that matches these requirements. I think the logic was that the database does not actually enforce that, therefore we should err on the cautious side and replace the FirstOrDefault() with Where(). Probably good to be consistent here as well.

DigitalLearningSolutions.Data.Tests/Services/CourseServiceTests.cs
```diff
@@ -244,13 +266,31 @@ public void UpdateLearningPathwayDefaultsForCourse_calls_data_service()
                 .MustHaveHappened();
         }
 
+        [Test]
+        public void RemoveDelegateFromCourse_returns_false_if_no_current_progress()
```
DanBloxham-sw, line 270: Minor, but could we have this test directly after the other RemoveDelegateFromCourse test.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/ViewDelegateController.cs
```diff
@@ -78,5 +82,52 @@ public IActionResult DeactivateDelegate(int delegateId)
...
+        [Route("{customisationId:int}/Remove")]
+        public IActionResult ConfirmRemoveFromCourse(int delegateId, int customisationId)
+        {
+            var centreId = User.GetCentreId();
+            var delegateUser = userDataService.GetDelegateUserCardById(delegateId);
+            var course = courseDataService.GetCourseNameAndApplication(customisationId);
+            if (delegateUser == null || delegateUser.CentreId != centreId || course == null)
+            {
+                return new NotFoundResult();
+            }
```
DanBloxham-sw, line 96: We should be able to get rid of this by using the new service filters:
VerifyAdminUserCanAccessDelegateUser
VerifyAdminUserCanAccessCourse
The first one already looks to be applied at the controller level. The second can probably be applied to the two new controller actions. 

However, you might want to double check it does what you want it to do. Currently, the manual check in this method only determines that the course exists, while the service filter also checks the admin user has access to edit the course. That might not be applicable here if the admin should be able to remove delegates from AllCentre courses that they don't have permission to edit.

Looking at the HLD comments, Stella mentioned that the Admin Category Id should be checked against to make sure the admin user can remove the user from the course. Seeing as how the admin ability is directly restricted by course properties, I think the course service filter should be fine to apply here (it should also check against the admin category id for you). Thoughts?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/ViewDelegateController.cs
```diff
@@ -78,5 +82,52 @@ public IActionResult DeactivateDelegate(int delegateId)
...
+        public IActionResult ExecuteRemoveFromCourse(int delegateId, int customisationId, RemoveFromCourseViewModel model)
+        {
+            if (!ModelState.IsValid)
+            {
+                return View("ConfirmRemoveFromCourse", model);
+            }
+
+            var centreId = User.GetCentreId();
+            var delegateUser = userDataService.GetDelegateUserCardById(delegateId);
+            if (delegateUser == null || delegateUser.CentreId != centreId ||
```
DanBloxham-sw, line 120: Similarly the delegate user checks in this if statement can be removed thanks to the service filter.

Will want to double check whether the course service filter is appropriate too.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/ViewDelegate/ConfirmRemoveFromCourse.cshtml
```diff
@@ -0,0 +1,43 @@
+﻿@inject IConfiguration Configuration
+@using DigitalLearningSolutions.Web.Models.Enums
+@using DigitalLearningSolutions.Web.ViewModels.TrackingSystem.Delegates.ViewDelegate
+@using Microsoft.Extensions.Configuration
+@model RemoveFromCourseViewModel
+
+@{
+  var errorHasOccurred = !ViewData.ModelState.IsValid;
+  ViewData["Title"] = $"Remove delegate enrolment from {Model.CourseName}";
```
DanBloxham-sw, line 9: In other pages, the tab title has "Error: " prefixed to it when we display an error summary. This page should do this as well.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/ViewDelegate/ConfirmRemoveFromCourse.cshtml
```diff
@@ -0,0 +1,43 @@
...
+          <div class="nhsuk-checkboxes">
+            <div class="nhsuk-checkboxes__item">
+              <input type="hidden" name="name" value="@Model.Name" />
+              <input type="hidden" name="courseName" value="@Model.CourseName" />
+              <input class="nhsuk-checkboxes__input" id="confirm" name="confirm" type="checkbox" value="true" aria-describedby="confirm-item-hint">
+              <label class="nhsuk-label nhsuk-checkboxes__label" for="confirm">
+                I am sure that I want to remove @Model.Name from the @Model.CourseName course.
+              </label>
+              <div class="nhsuk-hint nhsuk-checkboxes__hint" id="confirm-item-hint">
+                This action will remove the course from the delegate's Current/Completed courses list and archive their progress.
```
DanBloxham-sw, line 33: Will this actually remove the course from the delegate's Completed course list? We don't edit any completed progress records so I don't think it will. This might need to be confirmed and/or reworded.

DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/Delegates/ViewDelegateControllerTests.cs
```diff
@@ -47,5 +55,167 @@ public void Deactivating_delegate_returns_redirect()
             // then
             result.Should().BeRedirectToActionResult();
         }
+
+        [Test]
+        public void Removal_confirmation_page_displays_for_valid_delegate_and_course()
```
DanBloxham-sw, line 60: Seeing as the logic for some of this can be moved out into the ServiceFilters, the tests can be removed that test that removed logic.

DanBloxham-sw (changes requested):
Few minor comments about some things, and a couple of questions clarifying how this works. 

Could we also have some documentation on the swiki about the RemovalMethodIds as requested in the HLD comments?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/DelegateProgressController.cs
```diff
@@ -33,7 +33,11 @@ public IActionResult Index(int progressId, DelegateProgressAccessRoute accessedV
             var courseDelegatesData =
                 courseService.GetDelegateCourseProgress(progressId, centreId);
 
-            var model = new DelegateProgressViewModel(accessedVia, courseDelegatesData!);
+            var model = new DelegateProgressViewModel(
+                accessedVia,
+                courseDelegatesData!,
+                courseDelegatesData!.DelegateCourseInfo.DelegateId
```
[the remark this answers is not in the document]
JBsoftwire: The compiler complains about it not being there, it doesn't seem like it refers to other lines to check possible null status.

DigitalLearningSolutions.Web.Tests/ViewModels/TrackingSystem/Delegates/DelegateProgressViewModelTests.cs
```diff
@@ -103,10 +109,12 @@ public void ViewModel_sets_Enrolment_method_text_correctly(int enrolmentMethodId
                 EnrolledByAdminId = 1,
                 EnrolledByForename = "Ronnie",
                 EnrolledBySurname = "Dio",
+                DelegateId = 1,
             };
             var viewModel = new DelegateProgressViewModel(
                 DelegateProgressAccessRoute.ViewDelegate,
-                new DelegateCourseDetails(delegateInfo, new List<CustomPromptWithAnswer>(), new AttemptStats(0, 0))
+                new DelegateCourseDetails(delegateInfo, new List<CustomPromptWithAnswer>(), new AttemptStats(0, 0)),
```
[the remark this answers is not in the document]
JBsoftwire: The formatter doesn't mind it, which surprised me at the time, but I've split it manually.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/ViewDelegate/ConfirmRemoveFromCourse.cshtml
JBsoftwire: Good catch. I'll change this wording and we should ask them about this when we demo it.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/ViewDelegateController.cs
JBsoftwire: I think Stella's comments on the HLD say we want to have the check, following her discussion with Kevin. If it's not the desired behaviour they can let us know and we can change it.

JBsoftwire:
Removal method documentation already exists on the https://softwiretech.atlassian.net/wiki/spaces/HEE/pages/6713476239/Course+tables page.

DigitalLearningSolutions.Data.Tests/DataServices/CourseDataServiceTests.cs
```diff
@@ -236,6 +240,32 @@ public void Remove_current_course_should_prevent_a_course_from_being_returned()
...
+        public async Task Remove_current_course_sets_removal_date_and_method_correctly()
+        {
+            using var transaction = new TransactionScope(TransactionScopeAsyncFlowOption.Enabled);
+
+            try
+            {
+                // Given
+                var removedDate = DateTime.UtcNow;
+                const int progressId = 94323;
+                const int candidateId = 1;
```
DanBloxham-sw, line 253: Can safely move this outside the try block.

DanBloxham-sw (approved):
This is looking good. I've put one minor comment about moving some lines, but it won't need re-review from me.

As for the documentation, it looks to be in a sensible place, not sure how I missed them before. I don't think anything further need to be done with them.

PR #658, opened 2021-10-05
HEEDLS-632 Remove unnecessary application type from permission based …

AlexJacksonDS:
…redirection

### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-632

### Description
Removed the ApplicationType == Main or null check that was grouped up with the Admin accessing tracking system check causing pages with the Main application type (e.g. accessing My Account from the homepage) to be redirected if the user was not a centre admin (or higher permissions).
Sanity checked that the rest of the conditions produced by the attribute, e.g. redirecting a delegate to the learning portal still worked as expected.

### Screenshots
N/A

-----
### Developer checks

I have:

### DIFF
--- a/DigitalLearningSolutions.Web/Attributes/ValidateAllowedApplicationTypeAttribute.cs
+++ b/DigitalLearningSolutions.Web/Attributes/ValidateAllowedApplicationTypeAttribute.cs
@@ -49,9 +49,7 @@
             if (!user.HasLearningPortalPermissions() && ApplicationType.LearningPortal.Equals(application) ||
                 !user.HasFrameworksAdminPermissions() && ApplicationType.Frameworks.Equals(application) ||
                 !user.HasSupervisorAdminPermissions() && ApplicationType.Supervisor.Equals(application) ||
-                !user.HasCentreAdminPermissions() && (ApplicationType.TrackingSystem.Equals(application) ||
-                                                      ApplicationType.Main.Equals(application) ||
-                                                      application is null))
+                !user.HasCentreAdminPermissions() && ApplicationType.TrackingSystem.Equals(application))
             {
                 RedirectToHome(context);
             }


DigitalLearningSolutions.Web/Attributes/ValidateAllowedApplicationTypeAttribute.cs
```diff
@@ -49,9 +49,7 @@ public void OnActionExecuting(ActionExecutingContext context)
             if (!user.HasLearningPortalPermissions() && ApplicationType.LearningPortal.Equals(application) ||
                 !user.HasFrameworksAdminPermissions() && ApplicationType.Frameworks.Equals(application) ||
                 !user.HasSupervisorAdminPermissions() && ApplicationType.Supervisor.Equals(application) ||
-                !user.HasCentreAdminPermissions() && (ApplicationType.TrackingSystem.Equals(application) ||
-                                                      ApplicationType.Main.Equals(application) ||
-                                                      application is null))
+                !user.HasCentreAdminPermissions() && ApplicationType.TrackingSystem.Equals(application))
```
DanBloxham-sw, line 52: I can see how this removal would fix the bug, but I'm wondering why it was there in the first place.

In the comments of the ticket that implemented this (https://softwiretech.atlassian.net/browse/HEEDLS-588) it seemed important that "Main" was not a valid application. I then wonder why we had a Main ApplicationType at all. Do we need an extra check in here or the MyAccount controller to redirect to the main version if these current conditions are all false?

DanBloxham-sw (commented):
I might be getting overly confused but it looks like there bug ticket description's expected outcomes are inconsistent with the expected outcomes of the original implementation ticket.

DigitalLearningSolutions.Web/Attributes/ValidateAllowedApplicationTypeAttribute.cs
AlexJacksonDS: I've finally got around to checking this, and Main still behaves as per that ticket. I don't think we need an extra checks, if we have an ApplicationType that we do not handle (e.g. Main) we fail out before this if statement and return a not found. i.e. ApplicationType.Main.Equals(application) must always be false by the time we reach this code.

At the point we've reached this we must have an admin account and we must have one of the ApplicationTypes we want to handle, so we can safely redirect to home if they've tried to access an area they have no access to.

SteveJacksonSoft (approved):

PR #678, opened 2021-10-15
Heedls 541e edit filters test fixes

JBsoftwire:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-541

### Description
Issues raised in testing for this ticket have been addressed, except where later deemed not in need of fixing in comments on that ticket.

### Screenshots
image
-----
### Developer checks

I have:
- [*] Run the formatter and made sure there are no IDE errors.
- [*] Written tests for the changes (controller, data services, services, view models etc) and manually tested my work with and without JavaScript. Full manual testing guidelines can be found here: https://softwiretech.atlassian.net/wiki/spaces/HEE/pages/6703648740/Testing
- [*] Updated my Jira ticket with information about other parts of the system that were touched as part of the MR and have to be sanity tested to ensure nothing’s broken.
- [*] Scanned over my own MR to ensure everything is as expected.


DigitalLearningSolutions.Data/Services/ActivityService.cs
```diff
@@ -217,5 +224,18 @@ private static int GetFirstMonthOfQuarter(int quarter)
...
+            public bool Equals(Course? x, Course? y)
+            {
+                return x?.CustomisationId == y?.CustomisationId;
+            }
+
+            public int GetHashCode(Course obj)
+            {
+                return obj.CustomisationId;
+            }
+        }
```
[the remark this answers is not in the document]
SteveJacksonSoft: Is there a reason not to add these methods as overrides on Course? Is course equality checked anywhere else yet?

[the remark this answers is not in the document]
JBsoftwire: There's no architectural reason and I found examples of both solutions in use. The reason it's done this way is that it works - I couldn't get it to recognise the overrides when I put them directly on the Course class, with no immediately clear reason why.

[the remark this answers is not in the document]
JBsoftwire: Since this method worked, I decided to use it rather than spend time getting to the bottom of the other,

DigitalLearningSolutions.Data/Models/TrackingSystem/DateInformation.cs
```diff
@@ -5,22 +5,66 @@
...
-            return Date.ToString(format);
+            return startDate.ToString(startFormat) + " to " + endDate.ToString(endFormat);
+        }
+
+        private string ConvertQuartersInFormatString(string format, DateTime date)
+        {
+            var quarter = date.Month / 3 + 1;
+
+            return format.Replace("Q", $"'Quarter' {quarter}").Replace("q", $"Q{quarter}");
+        }
```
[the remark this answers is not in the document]
JBsoftwire: The DateRangeLabel one I left as dynamic (is that the term here?) because it's an overload of the other method of the same name, which can't be static. It felt appropriate that they both be called in the same way, but I don't know if that seems like a good principle to you.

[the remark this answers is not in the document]
davidm-m: Hm, I still think it makes sense to make it static if it doesn't rely on any object properties, so I'm happy with how it is now.

davidm-m (approved):
Great stuff!

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Reports/_FilterDisplayCard.cshtml
```diff
@@ -33,7 +33,7 @@
             Date range
           </dt>
           <dd class="nhsuk-summary-list__value">
-            @Model.DateRange
+            <span>@Model.StartDate</span><span class="screen-reader-only">to</span><span> - @Model.EndDate</span>
```
AlexJacksonDS, line 36: Is there a reason for this screen-reader-only class? rather than using the nhsuk-u-visually-hidden class?

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/Centre/Reports/ReportsViewModel.cs
```diff
@@ -28,11 +30,44 @@ IEnumerable<EvaluationResponseBreakdown> evaluationResponseBreakdowns
...
+                        : new ActivityDataRowModel(p, DateHelper.StandardDateFormat, startDate, true)
+                );
+                var last = new[] { activity.Last() }.Select(
+                    p => p.DateInformation.Interval == ReportInterval.Days
+                        ? new ActivityDataRowModel(
+                            p,
+                            DateHelper.GetFormatStringForDateInTable(p.DateInformation.Interval)
+                        )
+                        : new ActivityDataRowModel(p, DateHelper.StandardDateFormat, endDate, false)
+                );
```
AlexJacksonDS, line 60: Why do we need to use new lists of single items for these rather than just having them be converted directly?

DigitalLearningSolutions.Data/Helpers/DateHelper.cs
```diff
@@ -7,6 +7,8 @@
 
     public static class DateHelper
     {
+        public static string StandardDateFormat = "dd/MM/yyyy";
```
AlexJacksonDS, line 10: We now have this standard date format in 2 places again (there is one in a helper in the web project too). Either we need to get rid of that one in favour of this one, or we need that one to reference this one.

DigitalLearningSolutions.Data/DataServices/CourseDataService.cs
```diff
@@ -396,11 +399,32 @@ public IEnumerable<Course> GetCentrallyManagedAndCentreCourses(int centreId, int
...
+                        c.CustomisationName,
+                        c.Active
+                    FROM Candidates AS cn
+                    INNER JOIN Progress AS p ON p.CandidateID = cn.CandidateID
+                    INNER JOIN Customisations AS c ON c.CustomisationID = p.CustomisationId
+                    INNER JOIN dbo.Applications AS ap ON ap.ApplicationID = c.ApplicationID
+                    WHERE cn.CentreID = @centreID
 	                AND (ap.CourseCategoryID = @categoryId OR @categoryId IS NULL)
-                    AND ca.CentreID = @centreId
                     AND ap.ArchivedDate IS NULL",
```
AlexJacksonDS, line 428: Indentation is a bit funny in the WHERE clauses of both these queries. May just be a github thing though. If they are fine in the IDE just leave them.

DigitalLearningSolutions.Data/Models/TrackingSystem/DateInformation.cs
```diff
@@ -5,22 +5,61 @@
```
AlexJacksonDS, line 5: I don't really understand the whole of this file, it seems to be doing 2 things. We have the DateInformation entity, but also static methods that seem to have nothing to do with it.

DigitalLearningSolutions.Data/Models/TrackingSystem/DateInformation.cs
```diff
@@ -5,22 +5,61 @@
...
-            format = format.Replace("Q", $"'Quarter' {quarter}").Replace("q", $"Q{quarter}");
+            return StartDate.ToString(format);
+        }
+
+        public string GetDateRangeLabel(string format, DateTime rangeTerminator, bool startRangeFromTerminator)
+        {
+            var dateFromThisObject = startRangeFromTerminator ? GetFinalDate() : StartDate;
+            var formatForDateFromThisObject = ConvertQuartersInFormatString(format, dateFromThisObject);
+
+            var formatForTerminator = ConvertQuartersInFormatString(format, rangeTerminator);
```
AlexJacksonDS, line 29: I find these variable/parameter names a bit confusing, particularly formatForDateFromThisObject. Isn't this actually a formatted date not a format?

DigitalLearningSolutions.Data/Services/ActivityService.cs
```diff
@@ -118,12 +124,60 @@ public byte[] GetActivityDataFileForCentre(int centreId, ActivityFilterData filt
...
+                        p.DateInformation.GetDateRangeLabel(
+                            DateHelper.StandardDateFormat,
+                            filterData.EndDate ?? DateTime.Now,
+                            true
+                        ),
+                        p.Registrations,
+                        p.Completions,
+                        p.Evaluations
+                    )
+                );
```
AlexJacksonDS, line 167: Same as the view model. Why do we need to use new lists of single items for these rather than just having them be converted directly?

If we're having to do all this messing around with first and last and display strings, would it perhaps make sense to pull all this logic into the GetFilteredActivity method, so we only have this logic in one place.

DigitalLearningSolutions.Web.Tests/ViewModels/TrackingSystem/Centre/Reports/ReportsViewModelTests.cs
```diff
@@ -18,20 +18,28 @@ public void UsageStatsTableViewModel_reverses_data_in_time()
             {
```
AlexJacksonDS, line 18: I removed all the extra PeriodOfActivity declarations in a previous version of this since they weren't needed for that version of the display strings. If we're adding them back, presumably we need them, but does this mean we should have more assertions in each test?
This applies to all tests in this file.

Also can you capitalise all the // given // when // then for consistency with the rest of the solution.

AlexJacksonDS (changes requested):
I think we may need to consider where some of the logic is as it seems we're doing the same thing in a couple of places.

Also a few things to tidy up.

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/Centre/Reports/ReportsViewModel.cs
JBsoftwire: It was the first way that came to mind to deal with creating and concatenating the three components neatly, but now you've put a light on it I think just using prepend/append would be clearer.

DigitalLearningSolutions.Data/DataServices/CourseDataService.cs
JBsoftwire: Seems like just a github thing, though I'd be curious to find out how this thing where GH has a different indentation even happens.

AlexJacksonDS: Looking at them in Notepad++ there is a tab (--->) on the category filter line which github obviously renders differently to VS/Rider:
image

DigitalLearningSolutions.Data/Models/TrackingSystem/DateInformation.cs
JBsoftwire: It's a format string for the .ToString() calls below. The reason it also contains formatted date elements is that C#'s datetime formatting doesn't have any handling for Q/quarters, so it processes them there and then passes the format string on to ToString(). The already-processed quarter elements aren't interpreted by the ToString() date formatter so are passed through unchanged.

DigitalLearningSolutions.Data/Models/TrackingSystem/DateInformation.cs
JBsoftwire: The file contains the DateInformation class and some methods to generate a date or date range label from it. One of the date range label methods doesn't actually require information from the instance of the class (it's for the edge case where there's only one data point, with a date range specified by the start and end dates from the filters) but is still in this class as it is a special case of the other two methods, used in the same place in the code.

DigitalLearningSolutions.Web.Tests/ViewModels/TrackingSystem/Centre/Reports/ReportsViewModelTests.cs
JBsoftwire: If I understand what you're referring to correctly, they're needed because single, start, end and middle periods are all converted to labels differently, so there need to be additional periods wrapping middle periods to test their conversion. That the start and end periods are converted correctly has its own test, which only has one variant because those conversions don't depend on the period.

AlexJacksonDS: Ok that's fine then. Wasn't sure that all cases were being tested.

DigitalLearningSolutions.Data/Services/ActivityService.cs
JBsoftwire: I think moving all the display logic into the GetFilteredActivity method would overload that method, and be an illogical place for it compared to the view model which is the main use case. This logic is only in the service at all because it needs to be done before the file is prepared, which happens in the data project.

It's a little ugly and involves some duplication, but I don't think a completely neat version of this logic is possible, or at least easily achievable at this point.

DigitalLearningSolutions.Data/DataServices/CourseDataService.cs
```diff
@@ -36,7 +36,10 @@ public interface ICourseDataService
 
         CourseDetails? GetCourseDetailsForAdminCategoryId(int customisationId, int centreId, int categoryId);
 
-        IEnumerable<Course> GetCentrallyManagedAndCentreCourses(int centreId, int? categoryId);
+        IEnumerable<Course> GetCoursesAvailableToCentreByCategory(int centreId, int? categoryId);
+
+        IEnumerable<Course> GetCoursesEverUsedAtCentreByCategory(int centreId, int? categoryId);
+
```
AlexJacksonDS, line 43: Extra line here

AlexJacksonDS (approved):
Just an extra new line left to sort. No need for re-review

PR #696, opened 2021-10-26
Heedls 535 all delegates edit details

AlexJacksonDS:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-535

### Description
Add new edit delegate details screen in the tracking system.

Refactored some of the common code from the MyAccount edit details.

### Screenshots
image
image
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -100,21 +100,28 @@ DlsSubApplication dlsSubApplication
...
             var userAdminId = User.GetAdminId();
             var userDelegateId = User.GetCandidateId();
 
             if (userDelegateId.HasValue)
             {
-                ValidateJobGroup(formData);
-                ValidateCustomPrompts(formData);
+                if (!formData.JobGroupId.HasValue)
+                {
+                    ModelState.AddModelError(nameof(EditDetailsFormData.JobGroupId), "Select a job group");
```
davidm-m, line 115: You can achieve this with a custom validator in the View Model (see SummaryViewModel for an example). I think it's better to do validation in the View Model where possible than in the controller, but I'm open to being persuaded otherwise.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/EditDelegateController.cs
```diff
@@ -0,0 +1,112 @@
...
+    using DigitalLearningSolutions.Web.ViewModels.Common;
+    using DigitalLearningSolutions.Web.ViewModels.TrackingSystem.Delegates.EditDelegate;
+    using Microsoft.AspNetCore.Authorization;
+    using Microsoft.AspNetCore.Mvc;
+    using Microsoft.FeatureManagement.Mvc;
+
+    [FeatureGate(FeatureFlags.RefactoredTrackingSystem)]
+    [Authorize(Policy = CustomPolicies.UserCentreAdmin)]
+    [Route("TrackingSystem/Delegates/{delegateId:int}/Edit")]
+    [SetDlsSubApplication(nameof(DlsSubApplication.TrackingSystem))]
```
davidm-m, line 18: Does this not need a selected tab attribute?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/EditDelegateController.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            var customPrompts =
+                centreCustomPromptHelper.GetEditCustomFieldViewModelsForCentre(delegateUser, centreId);
+            var model = new EditDelegateViewModel(delegateUser, jobGroups, customPrompts);
+
+            return View(model);
+        }
+
+        [HttpPost]
+        public IActionResult Index(EditDelegateFormData formData, int delegateId)
```
davidm-m, line 57: We don't check here that the delegate id exists or if the centre ids match - I think we should make a Service Filter that checks for that (if one doesn't already exist) and add it to the controller.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/EditDelegateController.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            var centreId = User.GetCentreId();
+            if (!formData.JobGroupId.HasValue)
+            {
+                ModelState.AddModelError(nameof(EditDetailsFormData.JobGroupId), "Select a job group");
```
davidm-m, line 62: You should be able to do this in the View Model (I'm not 100% on exactly how that should work with the Form Data structure but at the very least a custom Validate method on EditDelegateFormData will do it I think)

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/EditDelegateController.cs
```diff
@@ -0,0 +1,112 @@
...
+            }
+
+            if (!userService.NewEmailAddressIsValid(formData.Email!, null, delegateId, centreId))
+            {
+                ModelState.AddModelError(
+                    nameof(EditDetailsFormData.Email),
+                    "A user with this email address is already registered at this centre"
+                );
+                return ReturnToEditDetailsViewWithErrors(formData, delegateId, centreId);
+            }
```
davidm-m, line 87: Why do we not do this before the previous !ModelState.IsValid check?

DigitalLearningSolutions.Data/Models/User/AccountDetailsData.cs
```diff
@@ -3,30 +3,18 @@
     public class AccountDetailsData
     {
         public AccountDetailsData(
```
davidm-m, line 5: If we do make this class abstract, this constructor should probably be protected.

davidm-m (changes requested):
Found a few things

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
AlexJacksonDS: So this behaves a bit oddly. I have to have the IValidatableObject at the level above the EditDetailsFormData, the type on the method parameter, otherwise it doesn't work. I can have the Validate method ishared in EditDetailsFormData though.
This also means I've had to remove some test assertions since they don't actually validate.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/EditDelegateController.cs
AlexJacksonDS: See this comment chain https://github.com/TechnologyEnhancedLearning/DLSV2/pull/696#discussion_r738471840

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/EditDelegateController.cs
AlexJacksonDS: This was already in progress with 633 when this ticket was being done, merged master and used VerifyAdminUserCanAccessDelegateUser ServiceFilter (same as ViewDelegateController)

davidm-m (approved):
Great stuff, looking good!

PR #697, opened 2021-10-27
Heedls 448 remove old hash

JBsoftwire:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-448

### Description
_Describe what has changed and how that will affect the app. If relevant, add references to the resources you used. Use this as your opportunity to highlight anything odd and give people context around particular decisions._

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests/DelegateUserDataServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests/DelegateUserDataServiceTests.cs
@@ -94,7 +94,7 @@
         public void GetDelegateUsersByEmailAddress_Returns_delegate_user()
         {
             // Given
-            var expectedDelegateUser = UserTestHelper.GetDefaultDelegateUser();
+            var expectedDelegateUser = UserTestHelper.GetDefaultDelegateUser(resetPasswordId: 1);
 
             // When
             var returnedDelegateUsers = userDataService.GetDelegateUsersByEmailAddress("email@test.com");
--- a/DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests/DelegateUserDataServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests/DelegateUserDataServiceTests.cs
@@ -3,6 +3,7 @@
     using System;
     using System.Linq;
     using System.Transactions;
+    using Dapper;
     using DigitalLearningSolutions.Data.Tests.TestHelpers;
     using FluentAssertions;
     using FluentAssertions.Execution;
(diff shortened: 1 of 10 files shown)


DigitalLearningSolutions.Data/DataServices/UserDataService/DelegateUserDataService.cs
```diff
@@ -140,7 +140,8 @@ public List<DelegateUser> GetDelegateUsersByEmailAddress(string emailAddress)
                         cd.Answer5,
                         cd.Answer6,
                         cd.JobGroupID,
-                        jg.JobGroupName
+                        jg.JobGroupName,
+                        cd.ResetPasswordID
```
DanBloxham-sw, line 144: This change leads me to think that the original code for invalidating hashes did not work as expected as it did not recover the delegate's reset password Ids. Are there any unit tests we could implement that would have spotted the issue? I can only think that the unit tests for this method and the admin equivalent should retrieve data that has a non-null reset password id. Thoughts?

DanBloxham-sw (commented):
Looks good. Left a comment about the used code and potentially unit test modifications.

DigitalLearningSolutions.Data/DataServices/UserDataService/DelegateUserDataService.cs
JBsoftwire: Good question, it did look to me like the original code didn't work for delegate hashes. I only discovered this problem when manually testing. I'll take another look at the tests, I recall I tried to find a quick fix and failed, but it's worth fixing.

JBsoftwire:
I haven't run the new script on the remote DB yet so the jenkins build will fail, since Stephen was thinking a different approach might be better.

SQLScripts/20211103_UpdateResetPasswordIdForTests.sql
```diff
@@ -0,0 +1 @@
+UPDATE Candidates SET ResetPasswordID = 1 WHERE CandidateID = 2
```
DanBloxham-sw, line 1: I believe the ResetPasswordID column in the Candidates table has a foreign key constraint on it. Can we guarantee that there will be a ResetPassword record with an ID of 1? or should it perhaps be added along with this as well?

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests/DelegateUserDataServiceTests.cs
```diff
@@ -94,7 +94,7 @@ public void GetAllDelegateUsersByUsername_searches_AliasID()
         public void GetDelegateUsersByEmailAddress_Returns_delegate_user()
         {
             // Given
-            var expectedDelegateUser = UserTestHelper.GetDefaultDelegateUser();
+            var expectedDelegateUser = UserTestHelper.GetDefaultDelegateUser(resetPasswordId: 1);
```
DanBloxham-sw, line 97: Couple of things with this.

Firstly, do you think we should have a similar test for admin users to make sure their reset password Id is properly retrieved?

Secondly, I feel like the default values for the GetDefaultDelegateUser should match the values of the record with the default Id. In this case the resetPasswordId should be 1 by default. This would then prevent any data tests that query the same user from mysteriously failing. But then on the other hand, that _might_ cause other tests to fail that create users based on the default delegate. But I'm not too sure on that. Thoughts?

DanBloxham-sw (commented):
Couple of potentially redundant concerns with the new changes.

DigitalLearningSolutions.Data/DataServices/UserDataService/AdminUserDataService.cs
```diff
@@ -132,7 +132,8 @@ ELSE cc.CategoryName
                         au.IsWorkforceContributor,
                         au.IsLocalWorkforceManager,
                         au.ImportOnly,
-                        au.FailedLoginCount
+                        au.FailedLoginCount,
+                        au.ResetPasswordId
```
JBsoftwire, line 136: On further investigation it turned out the lookup also doesn't work for admins. The actual reset login works fine because it uses its own methods, but I don't think the old password resets were ever getting deleted. It evaded our unit testing because the result is a formally valid output from the data service, and the service-service could assume useful data.

DanBloxham-sw (approved):

PR #699, opened 2021-10-28
Heedls 568 edit supervisor

AlexJacksonDS:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-568

### Description
Added a new Edit Supervisor page linked from the View Delegate course cards and the Delegate Progress page

### Screenshots
image
image
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


AlexJacksonDS:
Updated post show and tell to allow unsetting a supervisor
image

DigitalLearningSolutions.Data.Tests/TestHelpers/ProgressTestHelper.cs
```diff
@@ -26,5 +36,15 @@ public static class ProgressTestHelper
...
         }
+
+        public DateTime? GetSupervisorVerificationRequestedByAspProgressId(int aspProgressId)
+        {
+            return connection.Query<DateTime?>(
+                @"SELECT SupervisorVerificationRequested
+                    FROM aspProgress
+                    WHERE aspProgressId = @aspProgressId",
+                new { aspProgressId }
+            ).Single();
```
JBsoftwire, line 47: I assume this must work since the tests pass, but my understanding of the Single() LINQ operator is that it throws if there's not exactly one result, so this can never return null. Can you tell me what I've got wrong here?

JBsoftwire (changes requested):
Some comments about the use of `!` and a question about `Single()`.

DigitalLearningSolutions.Data.Tests/TestHelpers/ProgressTestHelper.cs
AlexJacksonDS: The column SupervisorVerificationRequested is a nullable datetime, so if the aspProgress record exists but has null it will return null, hence the DateTime? return type.
As you say, this will throw an exception if there is not exactly one result, but since we are looking it up by aspProgressId and using it in a unit test this is the behaviour we would want. i.e. this will only throw an exception if the aspProgress record gets deleted from the database for some reason.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/DelegateProgressController.cs
```diff
@@ -34,5 +43,48 @@ public IActionResult Index(int progressId, DelegateProgressAccessRoute accessedV
...
+            var centreId = User.GetCentreId();
+            var delegateCourseProgress =
+                courseService.GetDelegateCourseProgress(progressId, centreId);
+            var supervisors = userService.GetSupervisorsAtCentre(centreId);
+
+            var model = new EditSupervisorViewModel(
+                progressId,
+                accessedVia,
+                supervisors,
+                delegateCourseProgress!.DelegateCourseInfo
```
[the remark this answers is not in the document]
AlexJacksonDS: Yes, because we are protected by the VerifyAdminUserCanAccessProgress ServiceFilter, which handles the progess == null, return NotFound case. I suppose it is technically less efficient since we're doing double database queries, but that seems to be the pattern we've decided with all these service filters.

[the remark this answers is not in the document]
JBsoftwire: That makes sense, I forgot to check the controller attributes. No problem with this, then.

JBsoftwire (approved):

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/Delegates/DelegateProgress/EditSupervisorViewModel.cs
```diff
@@ -0,0 +1,52 @@
...
+            Supervisors = PopulateSupervisors(formData.SupervisorId, supervisors);
+        }
+
+        public int ProgressId { get; set; }
+
+        public DelegateProgressAccessRoute AccessedVia { get; set; }
+
+        public IEnumerable<SelectListItem> Supervisors { get; set; }
+
+        private IEnumerable<SelectListItem> PopulateSupervisors(int? supervisorId, IEnumerable<AdminUser> supervisors)
```
davidm-m, line 43: You could make this static

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateProgress/EditSupervisor.cshtml
```diff
@@ -0,0 +1,66 @@
...
+  var errorHasOccured = !ViewData.ModelState.IsValid;
+  ViewData["Title"] = errorHasOccured ? "Error: Edit supervisor" : "Edit supervisor";
+  var routeParamsForBackLink = new Dictionary<string, string?>();
+
+  if (Model.AccessedVia.Equals(DelegateProgressAccessRoute.CourseDelegates)) {
+    routeParamsForBackLink.Add("accessedVia", Model.AccessedVia.ToString());
+    routeParamsForBackLink.Add("progressId", Model.ProgressId.ToString());
+  } else {
+    routeParamsForBackLink.Add("delegateId", Model.DelegateId.ToString());
+  }
```
davidm-m, line 17: Feels like since this only depends on the model it could be set inside it, but I can also see that we already do it like this elsewhere so we might have already decided this is better. Any thoughts?

davidm-m (changes requested):
A few very small things

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateProgress/EditSupervisor.cshtml
AlexJacksonDS: I think it's probably better to leave it as is for now? It's the same pattern as pretty much every other page with a cancel/back link that requires a route parameter. Unless we want to start a new code style here, and tell everyone to start putting this on the model instead?

davidm-m: Yeah I'm not sure it's worth it then. Happy to leave it like it is.

davidm-m (approved):

PR #701, opened 2021-10-28
HEEDLS-639 implement new selected tab system in tracking system

livzorn:
### JIRA link
[_HEEDLS-639_](https://softwiretech.atlassian.net/browse/HEEDLS-639)

### Description
Implemented the new SetSelectedTab and SetDlsSubApplication attributes on all of the tracking system controllers and removes the redundant NavMenuItems section and ViewData values from all the views.

Note that there is a bug with submitting forms in the MyAccount tab, it takes you back to the Main application rather than staying in Tracking system, but this is being fixed in HEEDLS-618

-----
### Developer checks

I have:

### DIFF
--- a/DigitalLearningSolutions.Web/Controllers/Support/SupportController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/Support/SupportController.cs
@@ -1,7 +1,9 @@
 ﻿namespace DigitalLearningSolutions.Web.Controllers.Support
 {
     using System.Threading.Tasks;
+    using DigitalLearningSolutions.Data.Enums;
     using DigitalLearningSolutions.Data.Helpers;
+    using DigitalLearningSolutions.Web.Attributes;
     using DigitalLearningSolutions.Web.Helpers;
     using DigitalLearningSolutions.Web.Models.Enums;
     using DigitalLearningSolutions.Web.ViewModels.Support;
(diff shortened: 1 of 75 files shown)


DigitalLearningSolutions.Web/Controllers/Support/SupportController.cs
```diff
@@ -10,6 +12,8 @@
     using Microsoft.Extensions.Configuration;
     using Microsoft.FeatureManagement;
 
+    [SetDlsSubApplication]
```
davidm-m, line 15: Is this doing anything without any arguments?

livzorn: Ah yes, good question. Without any arguments the attribute will automatically try to find a route parameter of type DlsSubApplication, this is for controllers that are used in multiple applications. More info in [this documentation](https://softwiretech.atlassian.net/wiki/spaces/HEE/pages/19822084230/Selected+tab+functionality)

davidm-m: Fantastic, thanks for the link!

davidm-m (approved):

SteveJacksonSoft (approved):
I think `[SetDlsSubApplication]` on its own looks a little strange, and it'd be kinda nice if we could find a name which works better for this case as well as the case where it's set specifically. But I'm not sure it's worth the faff. This is definitely nicer than what we had before!

PR #705, opened 2021-11-02
Heedls 521 tracking system resources

AlexJacksonDS:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-521

### Description
Added new resources page in the Support area for the TrackingSystem and Frameworks sub applications.
Updated the links in the side nav and and Support body to link to this new page.

### Screenshots
image
image
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Web.Tests/Controllers/Support/ResourcesControllerTests.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            featureManager = A.Fake<IFeatureManager>();
+            configuration = A.Fake<IConfiguration>();
+            resourcesService = A.Fake<IResourcesService>();
+            A.CallTo(() => featureManager.IsEnabledAsync(FeatureFlags.RefactoredTrackingSystem))
+                .Returns(true);
+        }
+
+        [Test]
+        public async Task Frameworks_Resources_page_should_be_shown_for_valid_claims()
```
davidm-m, line 31: Should we have a similar test for the tracking system?

DigitalLearningSolutions.Web.Tests/Controllers/Support/ResourcesControllerTests.cs
```diff
@@ -0,0 +1,107 @@
...
+
+            // When
+            var result = await controller.Index("TrackingSystem");
+
+            // Then
+            result.Should().BeRedirectToActionResult().WithControllerName("Home").WithActionName("Index");
+        }
+
+        [Test]
+        public async Task Home_page_should_be_shown_when_accessing_frameworks_Resources_without_appropriate_claims()
```
davidm-m, line 93: Here too, maybe we should test the same situation for the tracking system.

davidm-m (changes requested):
A couple of suggestions

DigitalLearningSolutions.Web/Controllers/Support/ResourcesController.cs
```diff
@@ -0,0 +1,66 @@
...
+            {
+                return NotFound();
+            }
+
+            var trackingSystemSupportEnabled =
+                DlsSubApplication.TrackingSystem.Equals(dlsSubApplication) &&
+                User.HasCentreAdminPermissions() &&
+                await featureManager.IsEnabledAsync(FeatureFlags.RefactoredTrackingSystem);
+            var frameworksSupportEnabled = DlsSubApplication.Frameworks.Equals(dlsSubApplication) &&
+                                           User.HasFrameworksAdminPermissions();
```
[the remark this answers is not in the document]
AlexJacksonDS: As above, https://softwiretech.atlassian.net/browse/HEEDLS-608 should resolve this

DigitalLearningSolutions.Data/Models/Support/Resource.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace DigitalLearningSolutions.Data.Models.Support
+{
+    using System;
+
+    public class Resource
```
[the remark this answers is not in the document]
AlexJacksonDS: I've separated them as the downloads represent more. A list of Downloads represents a list of ResourceCategories each containing a list of Resources. I could combine the 2 and just have the extra Category string property though?

AlexJacksonDS:
Forgot to push the file extension change from show and tell yesterday. Pushed that now.

DigitalLearningSolutions.Web.Tests/Controllers/Support/ResourcesControllerTests.cs
davidm-m: If this were getting released in the meantime I'd say it is worth it, but it isn't, so fine to leave.

DigitalLearningSolutions.Data/Models/Support/Resource.cs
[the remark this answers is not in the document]
davidm-m: I'm just struggling to see what benefit we get from grabbing downloads from the database then immediately turning them into an identical data type except with less information. I think combining the two is sensible unless there's any specific reason you wouldn't want a resource to know what category it's in.

DigitalLearningSolutions.Data/DataServices/DownloadDataService.cs
```diff
@@ -7,7 +7,7 @@
 
     public interface IDownloadDataService
```
davidm-m, line 8: Does this want to be renamed ResourceDataService now?

davidm-m (approved):
One quick question but apart from that looks great!

DigitalLearningSolutions.Web/Helpers/DisplayStringHelper.cs
```diff
@@ -17,7 +17,7 @@ public static string FormatBytesWithLimit(long number, long limit)
             return GenerateBytesDisplayString(number) + Divider + GenerateBytesDisplayString(limit);
         }
 
-        private static string GenerateBytesDisplayString(long byteCount)
+        public static string GenerateBytesDisplayString(long byteCount)
```
SteveJacksonSoft, line 20: What kind of output does this give us if I put in something like `477210`?

SteveJacksonSoft: Just checking the name makes sense in an external context.

DigitalLearningSolutions.Web/Controllers/Support/ResourcesController.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        public async Task<IActionResult> Index(DlsSubApplication dlsSubApplication)
+        {
+            if (!DlsSubApplication.TrackingSystem.Equals(dlsSubApplication) &&
+                !DlsSubApplication.Frameworks.Equals(dlsSubApplication))
+            {
+                return NotFound();
+            }
+
+            var trackingSystemSupportEnabled =
```
SteveJacksonSoft, line 45: I don't think these bool names are quite right. Tracking system support might be enabled even if the user is currently trying to access frameworks/resources. The bools really mean shouldContinueToTrackingSystemResources and shouldContinueToFrameworksResources.

SteveJacksonSoft (changes requested):
Haven't got the whole way through this, but leaving some comments so Jon can crack on with the ValidateAllowedApplication attribute

DigitalLearningSolutions.Web/Helpers/DisplayStringHelper.cs
AlexJacksonDS: 466KB

DigitalLearningSolutions.Web/Controllers/Support/ResourcesController.cs
```diff
@@ -0,0 +1,66 @@
...
+                var model = new ResourcesViewModel(
+                    dlsSubApplication,
+                    SupportPage.Resources,
+                    configuration.GetCurrentSystemBaseUrl(),
+                    resourcesService.GetAllResources()
+                );
+                return View("Index", model);
+            }
+
+            return RedirectToAction("Index", "Home");
```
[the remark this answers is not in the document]
AlexJacksonDS: Is it worth doing all this here? The access denied change will come automatically with the change to using the ValidateAllowed...Attribute.

The 404 is a bit more difficult as we have an extra condition that it can be enabled by each feature flag separately. 404 is the behaviour we get when we use the feature flag attribute normally, but I don't think we can this twice so we probably need something custom to deal with this. This has a few options depending on the behaviour we want:

- Have the same 404 check in the actions for ResourcesController and SupportController (or one of these if we don't want it in both)
- Have some new attribute that will check application + feature flag and return the 404
- Roll this check into the Validate...Attribute. This will potentially be useful for other pages that have the same pattern. E.g. MyAccount can probably currently be accessed at /TrackingSystem/MyAccount since I don't think we have a similar flag check there.

DigitalLearningSolutions.Web/Controllers/Support/ResourcesController.cs
AlexJacksonDS: I just copied them from the SupportController. Do we want to rename them or is it not worth it given they're going to be replaced?

DigitalLearningSolutions.Web/ViewModels/Support/Resources/ResourcesCategoryViewModel.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace DigitalLearningSolutions.Web.ViewModels.Support.Resources
+{
+    using DigitalLearningSolutions.Data.Models.Support;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    public class ResourcesCategoryViewModel
+    {
+        public ResourcesCategoryViewModel(ResourceCategory resourceCategory, string currentSystemBaseUrl)
```
SteveJacksonSoft, line 9: Why not do the standard thing here, take in a category and IEnumerable<ResourceItemViewModel>?
Then in ResourcesService you can convert the list of resources directly to a list of ResourcesCategoryViewModel rather than going through `ResourceCategory`.

SteveJacksonSoft (changes requested):

DigitalLearningSolutions.Web/Helpers/DisplayStringHelper.cs
SteveJacksonSoft: I wonder if a more understandable interface would be `GenerateSizeDisplayString(long sizeInBytes)`. "Bytes" can be displayed / interpreted in a few different ways - signed or unsigned, or as characters after being sent through a character encoding. This name is worth getting right because the logic in the method is very mathsy and hard to get your head around. What do you think?

DigitalLearningSolutions.Web/Controllers/Support/ResourcesController.cs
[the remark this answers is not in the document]
SteveJacksonSoft: Yeah, I think rolling it into the Validate...Attribute makes the most sense. Yeah, let's not implement this here, but leave todos here referencing 608 and comment on the tickets that we'll be testing these cases in 608. Would you mind doing that?

DigitalLearningSolutions.Data/Services/ResourcesService.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            this.resourceDataService = resourceDataService;
+        }
+
+        public IEnumerable<ResourceCategory> GetAllResources()
+        {
+            var resources = resourceDataService.GetAllResources();
+
+            var resourceCategories = resources.GroupBy(d => d.Category).Select(
+                dg => new ResourceCategory(
```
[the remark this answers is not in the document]
AlexJacksonDS: Ah whoops, we were using Download before, so we had d downloads and dg downloadgroups

DigitalLearningSolutions.Web/ViewModels/Support/Resources/ResourcesViewModel.cs
```diff
@@ -0,0 +1,24 @@
...
+        public ResourcesViewModel(
+            DlsSubApplication dlsSubApplication,
+            SupportPage currentPage,
+            string currentSystemBaseUrl,
+            IEnumerable<ResourceCategory> resourceCategories
+        ) : base(dlsSubApplication, currentPage, currentSystemBaseUrl)
+        {
+            CurrentPage = currentPage;
+            DlsSubApplication = dlsSubApplication;
+            Categories = resourceCategories.Select(rc => new ResourcesCategoryViewModel(rc, currentSystemBaseUrl));
```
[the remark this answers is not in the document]
SteveJacksonSoft: Belay, VM not accessible from service

DigitalLearningSolutions.Web/ViewModels/Support/Resources/ResourcesCategoryViewModel.cs
SteveJacksonSoft: Belay, VM not accessible from services.

DigitalLearningSolutions.Data.Tests/DataServices/ResourceDataServiceTests.cs
```diff
@@ -0,0 +1,44 @@
...
+                Tag = "casestudytemplate",
+                UploadDateTime = new DateTime(2015, 05, 05, 09, 22, 36),
+            };
+
+            // When
+            var result = resourceDataService.GetAllResources().ToList();
+
+            // Then
+            result.Should().HaveCount(96);
+            result.Should().ContainEquivalentOf(expectedResource);
```
SteveJacksonSoft, line 41: Nitpick - I'd inline `expectedResource`. I'm more likely to look at the end of a test to find out what the expected output is than the start.

DigitalLearningSolutions.Web/Controllers/Support/ResourcesController.cs
SteveJacksonSoft: Okay, let's leave a todo for 608 here.

DigitalLearningSolutions.Web/Controllers/Support/ResourcesController.cs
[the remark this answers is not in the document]
SteveJacksonSoft: I've added this to the ACs on 608.

SteveJacksonSoft (approved):
Looks good, go ahead and merge.

PR #708, opened 2021-11-03
HEEDLS-587 Display alternate message if too many data points are sele…

AlexJacksonDS:
…cted for the reports graph

### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-587

### Description
Limit the number of allowed data points on the reports activity graph to 17 (this was the max I think reasonably fits at all page widths).
Display a message instead when the graph is not displayed.

### Screenshots
image
image
image
Mobile graph with 17 data points
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Web/Scripts/trackingSystem/reports.ts
+++ b/DigitalLearningSolutions.Web/Scripts/trackingSystem/reports.ts
@@ -9,6 +9,8 @@ const path = getPathForEndpoint('TrackingSystem/Centre/Reports/Data');
 const activityToggleableRowDisplayNone = 'none';
 const activityToggleableRowDisplayTableRow = 'table-row';
 
+const maxNumberOfEntriesForActivityGraph = 17;
+
 interface IActivityDataRowModel {
   period: string;
   completions: number;
@@ -34,40 +36,48 @@ function generateChart(request: XMLHttpRequest) {
     response = JSON.parse(response);
   }
   const data = constructChartistData(response);
+  const numberOfEntries = data.labels?.length;
 
-  const options = {
-    axisY: {
-      scaleMinSpace: 10,
-      onlyInteger: true,
-    },
-    chartPadding: {
-      bottom: 32,
-    },
-  };
+  if (numberOfEntries !== undefined && numberOfEntries <= maxNumberOfEntriesForActivityGraph) {
+    const options = {
+      axisY: {
+        scaleMinSpace: 10,
+        onlyInteger: true,
+      },
+      chartPadding: {
+        bottom: 32,
+      },
+    };
 
-  const chart = new Chartist.Line('.ct-chart', data, options);
+    const chart = new Chartist.Line('.ct-chart', data, options);
 
-  chart.on('draw',
-    // The type here is Chartist.ChartDrawData, but the type specification is missing getNode()
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    (drawnElement: any) => {
-      const { element } = drawnElement;
-      // IE renders text in SVGs with 'text' tags that do not work with most CSS properties
-      // so we set the relevant attributes manually
-      if (element.getNode().tagName === 'text'
-        && element.classes().indexOf('ct-horizontal') >= 0
-        && element.classes().indexOf('ct-label') >= 0) {
-        const xOrigin = Number(element.getNode().getAttribute('x'));
-        const yOrigin = element.getNode().getAttribute('y');
-        const width = Number(element.getNode().getAttribute('width'));
-        // this should match the NHS tablet breakpoint
-        const mediaQuery = window.matchMedia('(min-width: 641px)');
-        const rotation = mediaQuery.matches ? -45 : -60;
-        element.attr({
-          transform: `translate(-${width}) rotate(${rotation} ${xOrigin + width} ${yOrigin})`,
-        });
-      }
-    });
+    chart.on('draw',
+      // The type here is Chartist.ChartDrawData, but the type specification is missing getNode()
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      (drawnElement: any) => {
+        const { element } = drawnElement;
+        // IE renders text in SVGs with 'text' tags that do not work with most CSS properties
+        // so we set the relevant attributes manually
+        if (element.getNode().tagName === 'text'
+          && element.classes().indexOf('ct-horizontal') >= 0
+          && element.classes().indexOf('ct-label') >= 0) {
+          const xOrigin = Number(element.getNode().getAttribute('x'));
+          const yOrigin = element.getNode().getAttribute('y');
+          const width = Number(element.getNode().getAttribute('width'));
+          // this should match the NHS tablet breakpoint
+          const mediaQuery = window.matchMedia('(min-width: 641px)');
+          const rotation = mediaQuery.matches ? -45 : -60;
+          element.attr({
+            transform: `translate(-${width}) rotate(${rotation} ${xOrigin + width} ${yOrigin})`,
+          });
+        }
+      });
+  } else {
+    const chartContainer = document.getElementById('activity-graph');
+    if (chartContainer !== null) {
+      chartContainer.innerHTML = 'Too many data points selected. Please select a smaller date range or a bigger reporting interval to view a supporting graph.';
+    }
+  }
 }
 
 function toggleVisibleActivityRows() {
--- a/DigitalLearningSolutions.Web/Scripts/trackingSystem/reports.ts
+++ b/DigitalLearningSolutions.Web/Scripts/trackingSystem/reports.ts
@@ -1,4 +1,4 @@
-import Chartist from 'chartist';
+import Chartist, { IChartistData } from 'chartist';
 import getPathForEndpoint from '../common';
 
 // These constants are used in _ActivityTable.cshtml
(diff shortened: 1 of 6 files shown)


showkath10 (approved):
Looks good

DigitalLearningSolutions.Web/Scripts/trackingSystem/reports.ts
```diff
@@ -34,40 +36,48 @@ function generateChart(request: XMLHttpRequest) {
...
-        const mediaQuery = window.matchMedia('(min-width: 641px)');
-        const rotation = mediaQuery.matches ? -45 : -60;
-        element.attr({
-          transform: `translate(-${width}) rotate(${rotation} ${xOrigin + width} ${yOrigin})`,
-        });
-      }
-    });
+  const numberOfEntries = data.labels?.length;
+
+  if (numberOfEntries !== undefined && numberOfEntries <= maxNumberOfEntriesForActivityGraph) {
```
SteveJacksonSoft, line 41: Think we could take the chance to improve the code by extracting the first branch of the if-statement into a function of its own. It'd also make it easier to see what this if-statement does.

DigitalLearningSolutions.Web/Scripts/trackingSystem/reports.ts
```diff
@@ -9,6 +9,8 @@ const path = getPathForEndpoint('TrackingSystem/Centre/Reports/Data');
 const activityToggleableRowDisplayNone = 'none';
 const activityToggleableRowDisplayTableRow = 'table-row';
 
+const maxNumberOfEntriesForActivityGraph = 17;
```
SteveJacksonSoft, line 12: 17 feels a little bit restrictive, and limits what you can do with a couple of the intervals - e.g. you can't do daily over a month, or weekly over a half-year. I'd hazard a guess that people will generally be using a tablet or computer to look at reports, so they'd probably find it a bit annoying to be restricted by the size of a phone screen.
E.g. IPad view of 6 months weekly: 
image

I reckon this is a question worth asking HEE about.

I'd be inclined to raise this to 30 or so and just ignore phones, or maybe set up breakpoints to chose the limit based on the width of the element, but responding to screen dimention changes gets fiddly e.g. when you rotate a phone screen.

SteveJacksonSoft (changes requested):

DigitalLearningSolutions.Web/Scripts/trackingSystem/reports.ts
AlexJacksonDS: The problem is typically at the bottom of the desktop range on these pages with a sidebar. 30 gets a bit congested but is probably just about manageable:
image

AlexJacksonDS: Also I could implement a different limit for if the page is loaded on mobile initially. I think it is too much trouble to change responsively if the window is resized, other than perhaps dropping the graph if they go from tablet to mobile with 30 data points. Thoughts?

SteveJacksonSoft: Yeah, sounds reasonable to determine the limit on page load. Extra good if we can make that clear on the alternative text. Dropping the graph if they go from tablet to mobile - that sounds nearly as hard as recalculating on every window resize, no?

AlexJacksonDS: Yeah it's still recalculating on window resize, but at least it isn't recalculating a limit and having to repeatedly call the server to get the data again.

SteveJacksonSoft: I think we should still show a graph on mobile, so still have that 17-point limit or whatever.
You could just get the js to store the data rather than re-requesting it.

AlexJacksonDS:
Error is now in an NHS inset text:
image

DigitalLearningSolutions.Web/Scripts/trackingSystem/reports.ts
```diff
@@ -33,8 +38,25 @@ function generateChart(request: XMLHttpRequest) {
   if (typeof request.response === 'string') {
     response = JSON.parse(response);
   }
-  const data = constructChartistData(response);
+  chartData = constructChartistData(response);
+  drawChartOrDataPointMessage();
```
SteveJacksonSoft, line 42: I think it makes more sense split the functions as 
* Save chart data
* Draw the graph

It's a natural split, but we're doing one once and the other repeatedly.
So I would move this function call out into `request.onload` and rename `generateChart` to `saveChartData`. I'd probably change the param to just take in the response instead of the whole request.

DigitalLearningSolutions.Web/Scripts/trackingSystem/reports.ts
```diff
@@ -68,6 +90,30 @@ function generateChart(request: XMLHttpRequest) {
...
+  }
+  const chartContainer = document.getElementById('activity-graph');
+  if (chartContainer !== null) {
+    chartContainer.hidden = false;
+    chartContainer.style.display = 'block';
+  }
+}
+
+function displayTooManyDataPointsMessage() {
+  const dataPointErrorContainer = document.getElementById('activity-graph-data-error');
```
SteveJacksonSoft, line 107: Can we extract the two `getElementById` calls in this function (and the one above) into two functions of their own? It would also be worth extracting the id strings into constants at the top of the file and commenting that they need to match the elements in the view (as is done with the `_ActivityTable.cshtml` element ids).

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Reports/Index.cshtml
```diff
@@ -33,9 +33,15 @@
...
 
-    <div class="js-only-block">
+    <div id="activity-graph" class="js-only-block">
       <partial name="_ActivityGraph.cshtml" />
     </div>
+    <div id="activity-graph-data-error" class="js-only-block" hidden>
+      <div class="nhsuk-inset-text nhsuk-u-margin-top-0 nhsuk-u-margin-bottom-4">
+        <span class="nhsuk-u-visually-hidden">Information: </span>
+        <p>Too many chart data points to display. Please select a smaller date range or a bigger reporting interval to view data on a chart.</p>
+      </div>
```
SteveJacksonSoft, line 43: I think this would be a good thing to pull out into a partial...

SteveJacksonSoft (changes requested):
The page worked very nicely in the demo! Just some improvements I think we can make to the js.

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Reports/Index.cshtml
```diff
@@ -51,7 +57,8 @@
     <p>Summary of evaluation responses collected from delegates on completion of learning activities.</p>
 
     <ul class="nhsuk-card-group">
-      @foreach (var evaluationSummary in Model.EvaluationSummaryBreakdown) {
+      @foreach (var evaluationSummary in Model.EvaluationSummaryBreakdown)
```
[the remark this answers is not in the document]
AlexJacksonDS: I always have to fight Visual Studio on this. It automatically does this, then the Resharper format undoes it again. I have no idea how to get Visual Studio to respect the chosen style.

DigitalLearningSolutions.Web/Scripts/trackingSystem/reports.ts
AlexJacksonDS: I'm not sure I can change the parameter due to the strange typing that is browser dependent (IE). This method is what handles that.

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Reports/_ActivityGraphDataPointError.cshtml
```diff
@@ -0,0 +1,4 @@
+﻿<div class="nhsuk-inset-text nhsuk-u-margin-top-0 nhsuk-u-margin-bottom-4">
+  <span class="nhsuk-u-visually-hidden">Information: </span>
+  <p>Too many chart data points to display. Please select a smaller date range or a bigger reporting interval to view data on a chart.</p>
```
SteveJacksonSoft, line 3: Sorry, I should have been more specific maybe. The "inset" component would be a good partial / VC to have, but we should pass the text in in a viewmodel so that this partial isn't specific to the graph data point error, and can be used in other places. Can the inset have different levels (like `information`, `warning`, `error` kind of thing)? If so we could make that parameterizable as well.

DigitalLearningSolutions.Web/Scripts/trackingSystem/reports.ts
SteveJacksonSoft: Hmm, that is very strange. Fair, let's leave it then.

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Reports/Index.cshtml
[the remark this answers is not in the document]
SteveJacksonSoft: Hmm, good to know. I'll think on that.

DigitalLearningSolutions.Web/ViewComponents/InsetTextViewComponent.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace DigitalLearningSolutions.Web.ViewComponents
+{
+    using DigitalLearningSolutions.Web.ViewModels.Common.ViewComponents;
+    using Microsoft.AspNetCore.Mvc;
+
+    public class InsetTextViewComponent : ViewComponent
+    {
+        public IViewComponentResult Invoke(
+            string text,
+            string cssClass
```
SteveJacksonSoft, line 10: cssClassString?
Sounds a bit like you're only allowed to pass one in. Feel free to merge after that.

AlexJacksonDS: This is the same parameter naming as every other view component that has a css class parameter

SteveJacksonSoft: Ah. Fair, let's leave it as-is then.

SteveJacksonSoft (approved):

PR #711, opened 2021-11-08
HEEDLS-617 Remove centre references from admin field selection

AlexJacksonDS:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-617

### Description
Changed the admin field selection query to not filter on centre ID as we should be doing this at the course level before we query the admin fields.

Changed the get courses for delegate method to filter out courses not at the centre. This fixes the original bug with View Delegate page which was caused by accessing a delegate that had courses at a different centre attempting to call the admin field with the centre filter, and ending up with no result.

Pages affected by this change: 
- View Delegate (new centre filter on courses)
- Admin Fields management pages (behaviour should be unchanged)

### Screenshots
N/A

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


rhimcg (commented):
Looks good to me!

DigitalLearningSolutions.Data.Tests/Services/CourseServiceTests.cs
```diff
@@ -244,13 +240,32 @@ public void UpdateLearningPathwayDefaultsForCourse_calls_data_service()
...
+            var delegateCourseInfoAtCentre = new DelegateCourseInfo { CustomisationCentreId = centreId, CourseCategoryId = categoryId };
+            var delegateCourseInfoNotAtCentre = new DelegateCourseInfo { CustomisationCentreId = 1000, CourseCategoryId = categoryId };
+            A.CallTo(() => courseDataService.GetDelegateCoursesInfo(delegateId))
+                .Returns(new[] { delegateCourseInfoAtCentre, delegateCourseInfoNotAtCentre });
+
+            // When
+            var result = courseService.GetAllCoursesInCategoryForDelegate(delegateId, centreId, categoryId);
+
+            // Then
+            result.Count().Should().Be(1);
```
DanBloxham-sw, line 259: I think we should also check that the result's returned model should have the correct centre Id.

DigitalLearningSolutions.Data/Services/CourseService.cs
```diff
@@ -66,17 +66,17 @@ public IEnumerable<CourseStatistics> GetCentreSpecificCourseStatistics(int centr
         )
         {
             return courseDataService.GetDelegateCoursesInfo(delegateId)
+                .Where(info => info.CustomisationCentreId == centreId)
                 .Where(info => courseCategoryId == null || info.CourseCategoryId == courseCategoryId)
```
DanBloxham-sw, line 70: Could these Wheres be combined, or do you think it reads better as is?

DigitalLearningSolutions.Data/Services/CourseService.cs
```diff
@@ -66,17 +66,17 @@ public IEnumerable<CourseStatistics> GetCentreSpecificCourseStatistics(int centr
         )
         {
             return courseDataService.GetDelegateCoursesInfo(delegateId)
+                .Where(info => info.CustomisationCentreId == centreId)
```
DanBloxham-sw, line 69: I just want to double check the changes do actually do what we want. 

With this new CustomisationCentreId check, an admin looking at the View Delegate page will not be able to see any AllCentre courses that the delegate is enrolled on that are defined at a different centre. An admin at the centre the AllCentre course is defined at will not be able to see this either, as the delegate is registered at a different centre. 

As such, I don't think any admin will be naturally be able to follow a standard journey in the View Delegate to see/edit the progress details of a delegate for an AllCentre course. Is this the intended design? The service filter VerifyAdminUserCanAccessProgress does allow for AllCentre courses to be viewed and modified, so the user can still reach those pages by altering URLs, and also from the CourseDelegates page (I think).

DanBloxham-sw (changes requested):
I think this is mostly good, just left a minor comment about making sure a test is specific enough to match what the change, and some minor tidying. 

Another comment is clarifying the changes do what we expect, as it feels a little off to me at the moment.

DigitalLearningSolutions.Data/Services/CourseService.cs
AlexJacksonDS: yeah they could be, but I prefer to keep the separate ones in different statements for readability. This would be the following if combined:
`.Where(info => info.CustomisationCentreId == centreId && (courseCategoryId == null || info.CourseCategoryId == courseCategoryId))`

DigitalLearningSolutions.Data/Services/CourseService.cs
DanBloxham-sw: I've just seen 612, and it does indeed seem like the behaviour we want. As such, the course delegates and progress pages will probably need modifying as well. Though it still feels an odd choice to me.

AlexJacksonDS: I think you are correct that they should probably be able to view the All Centres courses that the delegate is enrolled on. It's not like they are managing the courses (as noted when I split VerifyAdminUserCanAccessCourse for HEEDLS-655 on the internal slack channel).

Updated the service method to have this. (This means I really don't think the above point about combining the wheres should happen)

DigitalLearningSolutions.Data/Services/CourseService.cs
DanBloxham-sw: That's fair, I agree, it can be left as it is.

DanBloxham-sw (approved):

PR #602, opened 2021-09-17
HEEDLS-594 Centre administrators - deactivate admin account

showkath10:
https://softwiretech.atlassian.net/browse/HEEDLS-594

Description
Added new view and action methods to be able to deactivate the admin user account.

Screenshots
image
image

PR #723, opened 2021-11-15
HEEDLS-594 Deactivate Admin Account - test fix

showkath10:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-594

### Description
Additional left margin to the user text is removed 
Added error summary block

### Screenshots
image
image
image
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)
- [ ] Updated my Jira ticket with information about other parts of the system that were touched as part of the MR and have to sanity tested to ensure nothing’s broken.


DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/DeactivateAdmin.cshtml
```diff
@@ -2,16 +2,22 @@
...
+    @if (errorHasOccurred) {
+      <vc:error-summary order-of-property-names="@(new[] { nameof(Model.Confirm) })" />
+    }
+
     <h1 class="nhsuk-heading-l">Are you sure you would like to deactivate this admin account?</h1>
   </div>
 </div>
 
-<div class="nhsuk-grid-row nhsuk-u-margin-6">
+
```
wilkinsm, line 19: Extra space here is not required.

wilkinsm (approved):

AlexJacksonDS:
I had a look through the history of this ticket to understand why the strange spacing raised as a bug occurred, and I found that several comments from the [original MR](https://github.com/TechnologyEnhancedLearning/DLSV2/pull/602) are still unresolved here (plus some extra thoughts that I think are incorrect) having got lost in the Closing/New MR change. I can't comment them against the code since it isn't changed here, so I'm just going to raise them all in this comment with screenshots of the relevant lines from VS.

1.  Martin's comment above is not yet resolved

2. DeactivateAdmin.cshtml: We have an extra div nhsuk-form-group that is also contained in the view component. This View component also has populate-with-current-value that is never used. I also think we should perhaps replace this with SingleCheckboxViewComponent, although that does behave slightly differently.
image

3. ConfirmCheckboxViewComponent: I think we should perhaps get rid of this in favour of SingleCheckboxViewComponent, but their behaviour for errors is slightly different. Single checkbox only shows one error, but I'm not sure why ConfirmCheckbox needs to show multiple errors?

4. DeactivateAdminVewModel.cs: FullName should not be nullable, the AdminUser property we are populating into it isn't. UserId isn't used anywhere outside the constructor, remove it.
image

5. [Steve's original comment about spacing](https://github.com/TechnologyEnhancedLearning/DLSV2/pull/602#pullrequestreview-766469579) has been reverted by this MR (It was done incorrectly before). We want vertical spacing, but no horizontal spacing, hence the bug.

6. AdminUserDataServiceTests.cs: Relating to [this comment](https://github.com/TechnologyEnhancedLearning/DLSV2/pull/602#discussion_r723122784), we're still using hard coded 7 (and more of them now). image

7. AdministratorController.cs: There is an unused call to get the admin user by ID here. Remove it.
image

8. AdministratorControllerTests.cs: Relating to [this comment](https://github.com/TechnologyEnhancedLearning/DLSV2/pull/602#discussion_r730838782) for DeactivateAdminUser_deactivates_admin_user_with_confirmation, we're not using NBuilder or just creating a new DeactivateAdminViewModel using its parameterless contructor. We also have these configured calls to GetAdminUserById which aren't needed since that statement is unused (see 6.).
image

I believe all of your other comments from the previous MRs have been resolved @SteveJacksonSoft @davidm-m

AlexJacksonDS (changes requested):
See above comment

SteveJacksonSoft:
For the record...
Showkath and I did have a discussion about ConfirmCheckboxVC. It came along before SingleCheckboxVC, but I prefer the latter because it's more generic. So I agree to getting rid of CCVC.
It feels like unnecessary logic to prune your list of error messages down to a single one in the SingleCheckboxVC, it's simpler for it just to show what it's given, but it's not a big deal because as Alex says, why would you have > 1 validation message for a binary input.

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests/AdminUserDataServiceTests.cs
```diff
@@ -198,13 +198,13 @@ public void UpdateAdminUserFailedLoginCount_updates_user()
...
             try
             {
                 // When
-                userDataService.DeactivateAdmin(7);
-                var updatedAdminUser = userDataService.GetAdminUserById(7)!;
+                userDataService.DeactivateAdmin(adminUser.Id);
+                var updatedAdminUser = userDataService.GetAdminUserById(adminUser.Id)!;
 
                 // Then
                 adminUser.Active.Should().Be(true);
```
AlexJacksonDS, line 210: This assertion isn't needed since adminUser is defined in code

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/DeactivateAdmin.cshtml
```diff
@@ -2,16 +2,20 @@
...
   <div class="nhsuk-grid-column-full">
+    @if (errorHasOccurred) {
+      <vc:error-summary order-of-property-names="@(new[] { nameof(Model.Confirm) })" />
+    }
     <h1 class="nhsuk-heading-l">Are you sure you would like to deactivate this admin account?</h1>
   </div>
 </div>
 
-<div class="nhsuk-grid-row nhsuk-u-margin-6">
+<div class="nhsuk-grid-row">
```
AlexJacksonDS, line 18: We still need some vertical spacing applied here as per comment 5 of my large comment. Can we have some screenshots once you've done that.

AlexJacksonDS (changes requested):
Just a couple of things left to resolve now.

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests/AdminUserDataServiceTests.cs
showkath10: I have removed this assertion

DigitalLearningSolutions.Web/Views/TrackingSystem/Centre/Administrator/DeactivateAdmin.cshtml
showkath10: I have added vertical spacing. Hope this look ok.

Desktop version 
image

Mobile versioin
image

AlexJacksonDS: Not sure it's made much difference, but I guess it's fine

AlexJacksonDS (approved):

PR #727, opened 2021-11-16
HEEDLS-640 Add super admin centres page

wilkinsm:
### JIRA link
[https://softwiretech.atlassian.net/browse/HEEDLS-640](https://softwiretech.atlassian.net/browse/HEEDLS-640)

### Description
Adds a list all centres page to the Super Admin sub-app.

### Screenshots
**Desktop**
image

**Desktop expanded**
image

**Tablet**
image

**Mobile**
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Data.Tests/DataServices/CentresDataServiceTests.cs
```diff
@@ -301,5 +301,24 @@ public void SetCentreAutoRegistered_should_set_AutoRegistered_true()
                 transaction.Dispose();
             }
         }
+
+        [Test]
+        public void GetAllCentreSummaries_returns_all_summary_details_and_reference_data()
+        {
+            using var transaction = new TransactionScope();
```
DanBloxham-sw, line 308: We are using a transaction here, but we are not modifying any records, so it can probably get gotten rid of.

DigitalLearningSolutions.Data.Tests/DataServices/CentresDataServiceTests.cs
```diff
@@ -301,5 +301,24 @@ public void SetCentreAutoRegistered_should_set_AutoRegistered_true()
...
+        {
+            using var transaction = new TransactionScope();
+            try
+            {
+                // When
+                var summaries = centresDataService.GetAllCentreSummaries();
+
+                // Then
+                summaries.Should()
+                    .OnlyContain(c => c.CentreType != null && c.RegionName != null && (c.Active || !c.Active));
```
DanBloxham-sw, line 316: I'm a bit confused as to what this line is testing. Centres are either Active or Inactive so that part is always true. The CentreType and RegionName property check seems off to me as well - this seems to just be checking that data in the database exists. Does that make sense or am I just confused?

I would expect that we would have some kind of count check to make sure that all centres are returned.

DigitalLearningSolutions.Web/Views/SuperAdmin/Centres/_CentreCard.cshtml
```diff
@@ -0,0 +1,90 @@
...
+       role="button"
+       href="#">
+      Manage centre
+    </a>
+    <a class="nhsuk-button nhsuk-u-margin-right-2 nhsuk-u-margin-bottom-2"
+       role="button"
+       href="#">
+      View courses
+    </a>
+    @if (Model.Active)
```
DanBloxham-sw, line 80: I noticed there was a question on the ticket description about a potential "reactivate button". Was there a decision made about this?

DanBloxham-sw (changes requested):
Looking good. Some minor changes can be made, and one comment that might just be me being confused.

DigitalLearningSolutions.Web/Controllers/SuperAdmin/Centres/CentresController.cs
```diff
@@ -0,0 +1,35 @@
+﻿namespace DigitalLearningSolutions.Web.Controllers.SuperAdmin.Admins
```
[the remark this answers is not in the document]
wilkinsm: Yeah spotted that after I re-enabled Resharper.

DigitalLearningSolutions.Data/DataServices/CentresDataService.cs
```diff
@@ -161,6 +162,28 @@ FROM Centres
             return centre;
         }
 
+        public IEnumerable<Centre> GetAllCentreSummaries()
+        {
+            var centres = connection.Query<Centre>(
```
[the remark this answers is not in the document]
wilkinsm: I've seen a mix of approaches here. If this is the prefered I'm happy to go with it so long as we try to stick to one convention.

DigitalLearningSolutions.Data.Tests/DataServices/CentresDataServiceTests.cs
wilkinsm: Indeed, good spot.

DigitalLearningSolutions.Data.Tests/DataServices/CentresDataServiceTests.cs
wilkinsm: I've had some question marks over these data services tests and the reliance on a "static" database for testing.
For now I've changed the test to be very specific. Although I have concerns given the reliance on the test data always being the same.

DigitalLearningSolutions.Web/Views/SuperAdmin/Centres/_CentreCard.cshtml
wilkinsm: It is still outstanding, Carolyn is away right now. I have planned in a more detailed discussion with Carolyn, not just on this but the whole Super Admin section.

DigitalLearningSolutions.Data/DataServices/CentresDataService.cs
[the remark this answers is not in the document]
DanBloxham-sw: Yeah I see what you mean about the mix of approaches. I don't think there is a clear set of coding standards that are overly specific and within easy reach of this project, so there are probably quite a few places where different styles have slipped through. Sticking with this should be adequate.

DigitalLearningSolutions.Data.Tests/DataServices/CentresDataServiceTests.cs
DanBloxham-sw: That's a good point. We have been (mostly) relying on the assumption of a "static" database throughout the data tests, and I believe that the assumption is a pretty safe one to make.

As far as I am aware, there are not any plans to majorly change a bunch of data, and seeing as we use a separate database only for unit tests (that should be identical to the back up file) then it should be pretty reliable. We've only recently moved to a preference of setting up data in a unit test, and the docs are pretty clear about transaction usage. So overall, I think we can rely on the data being consistent. If you have any questions/concerns, feel free to ask.

DigitalLearningSolutions.Web/Views/SuperAdmin/Centres/_CentreCard.cshtml
DanBloxham-sw: Good to know. Seeing as it will be a relatively simple change (if required), and everything else in the review seems good to me, I'm happy to move this on and the change can be tackled during tech lead review (or in another ticket if the decision is made far in the future for some reason)

DanBloxham-sw (approved):

DigitalLearningSolutions.Web/Views/SuperAdmin/Centres/_CentreCard.cshtml
```diff
@@ -0,0 +1,90 @@
...
+        @Model.CentreName (@Model.CentreId)
+      </span>
+    </summary>
+
+    <div class="nhsuk-details__text">
+      <dl class="nhsuk-summary-list">
+
+        <div class="nhsuk-summary-list__row">
+        <dt class="nhsuk-summary-list__key">
+          Centre id
```
SteveJacksonSoft, line 17: ID is capitalised on the delegates lists, so I think we should keep that consistent. Appreciate lower case matches the wireframes. See https://www.dls.nhs.uk/dev-my-learning-portal/TrackingSystem/Delegates/CourseDelegates

DigitalLearningSolutions.Web/Views/SuperAdmin/Centres/_CentreCard.cshtml
```diff
@@ -0,0 +1,90 @@
...
+        @Model.ContactForename @Model.ContactSurname
+      </dd>
+    </div>
+
+    <div class="nhsuk-summary-list__row">
+      <dt class="nhsuk-summary-list__key">
+        Contact email
+      </dt>
+      <dd class="nhsuk-summary-list__value">
+        @Model.ContactEmail
```
SteveJacksonSoft, line 56: We generally use dashes when these strings are null. I guess we should do the same here.

DigitalLearningSolutions.Data/Models/Centre.cs
```diff
@@ -31,8 +31,11 @@ public class Centre
         public int TrainerSpots { get; set; }
         public string? IpPrefix { get; set; }
         public string? ContractType { get; set; }
+        public int CentreTypeId { get; set; }
+        public string CentreType { get; set; }
```
SteveJacksonSoft, line 35: This is breaking c#'s null checks, because `CentresDataService.GetCentreDetailsById()` returns a `Centre?` but doesn't populate these fields.
And there are other nullable fields which don't get populated by the new `GetAllCentreSummaries()`.

It's a tricky one to navigate, because we might end up with 5 or 6 models which have centre properties and which overlap in different ways. It's not ideal to have a separate class for each if you're changing some property names. But maybe it's better than this? The only other thing I can think of is having an interface for each one, which are all implemented by `Centre`. What are your thoughts?

DigitalLearningSolutions.Data.Tests/Services/CentresServiceTests.cs
```diff
@@ -69,5 +71,24 @@ public void GetCentreRankForCentre_returns_null_with_no_data_for_centre()
...
             result.Should().BeNull();
         }
+
+        [Test]
+        public void GetAllCentreSummaries_calls_dataService_and_returns_all_summary_details()
+        {
+            // Given
+            var centres = Builder<Centre>.CreateListOfSize(10)
+                .All()
+                .With((c, i) => c.CentreId = i + 1)
```
SteveJacksonSoft, line 81: Does this not happen automatically?

wilkinsm: It does indeed.

DigitalLearningSolutions.Web/Views/SuperAdmin/Centres/_CentreCard.cshtml
wilkinsm: I've now added the button following confirmation.

DigitalLearningSolutions.Web/Views/SuperAdmin/Centres/_CentreCard.cshtml
wilkinsm: Done.

DigitalLearningSolutions.Web/Views/SuperAdmin/Centres/_CentreCard.cshtml
wilkinsm: Done.

DigitalLearningSolutions.Data/Models/Centre.cs
wilkinsm: I wasn't sure on this one.
I've experienced the mutli-overlapping model approach before and see it get out of hand. Although a strict oversight should keep it in check.
One thing I was thinking about but needs discussing is the use of a base class which holds the core properties and then other variations extend from that. But where do you draw the line on what are the core properties?

DigitalLearningSolutions.Web/Views/SuperAdmin/Centres/_CentreCard.cshtml
SteveJacksonSoft: I think D should be capital as well.

DigitalLearningSolutions.Data/Models/Centre.cs
SteveJacksonSoft: We've decided to use a new model for this instead of expanding Centre.

DigitalLearningSolutions.Web/Views/SuperAdmin/Centres/_CentreCard.cshtml
wilkinsm: Done.

DigitalLearningSolutions.Data/Models/Centre.cs
wilkinsm: Done.

DigitalLearningSolutions.Data/Models/Centres/Centre.cs
```diff
@@ -31,8 +31,11 @@ public class Centre
         public int TrainerSpots { get; set; }
         public string? IpPrefix { get; set; }
         public string? ContractType { get; set; }
+        public int CentreTypeId { get; set; }
```
SteveJacksonSoft, line 34: These three new fields are no longer used right?

SteveJacksonSoft: If that's correct, you can go ahead and merge after they're gone.

wilkinsm: Ah yes, I missed those. I've removed them now.

PR #739, opened 2021-11-19
HEEDLS-673 progress not being updated for long tutorial - release 21.9.1

showkath10:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-673

### Description
Small code change to increase auth token lifetime to 8 hours

### Screenshots

-----
### Developer checks

I have:
- [x ] Run the formatter and made sure there are no IDE errors.
- [ x] Manually tested my work with and without JavaScript. Full manual testing guidelines can be found here: https://softwiretech.atlassian.net/wiki/spaces/HEE/pages/6703648740/Testing
  - [doc_1_here](link_1_here)
- [ x] Scanned over my own MR to ensure everything is as expected.


JBsoftwire (approved):
Looks good.

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -175,6 +175,12 @@ UserAccountSet accounts
                 IsPersistent = rememberMe,
                 IssuedUtc = DateTime.UtcNow
             };
+
+            if (!rememberMe)
+            {
+                authProperties.ExpiresUtc = DateTime.UtcNow.AddHours(8);
+            }
```
SteveJacksonSoft, line 182: Nice, could we just add a comment to explain why we're doing it? Most important to point to the bug ticket, but also just say that tutorial progress doesn't get updated if the auth token has expired by the end of the tutorial.

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -175,6 +175,14 @@ UserAccountSet accounts
                 IsPersistent = rememberMe,
                 IssuedUtc = DateTime.UtcNow
             };
+
+            //HEEDLS-673
+            //Curse progress doesn't get updated if the auth token gets expires by the end of the tutorial so it's set to max 8 hours
```
SteveJacksonSoft, line 180: Nice, just a couple of tweaks would be good. 
* Space at the start of the comment
* Full stop at the end
* Typo in course
* Rather than saying "so it's set to max 8 hours" can we say that some tutorials are longer than the default auth token lifetime.
* Can we give a bit more detail about how HEEDLS-673 relates to this, so it doesn't get mistaken for todo or something else. Just say "see HEEDLS-637 for more context" or something like that.

DigitalLearningSolutions.Web/Controllers/LoginController.cs
```diff
@@ -175,6 +175,16 @@ UserAccountSet accounts
                 IsPersistent = rememberMe,
                 IssuedUtc = DateTime.UtcNow
             };
+
+
+            /* Course progress doesn't get updated if the auth token expires by the end of the tutorials as some
```
SteveJacksonSoft, line 180: "as" is not quite right here. Better to have "... tutorials. Some...". There's no causal relation between those two things. Looks good other than that.

SteveJacksonSoft (approved):
Had another check of this and it seems correct.

PR #741, opened 2021-11-19
HEEDLS-652 Add back end code to add resource to action plan

AlexJacksonDS:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-652

### Description
Add back end code for the Add to action plan functionality

### Screenshots
N/A

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/DataServices/CompetencyLearningResourcesDataServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/DataServices/CompetencyLearningResourcesDataServiceTests.cs
@@ -1,0 +1,85 @@
+﻿namespace DigitalLearningSolutions.Data.Tests.DataServices
+{
+    using System.Transactions;
+    using DigitalLearningSolutions.Data.DataServices;
+    using DigitalLearningSolutions.Data.Models.LearningResources;
+    using DigitalLearningSolutions.Data.Tests.TestHelpers;
+    using FluentAssertions;
+    using NUnit.Framework;
+
+    public class CompetencyLearningResourcesDataServiceTests
+    {
+        private ICompetencyLearningResourcesDataService service = null!;
+        private CompetencyLearningResourcesTestHelper testHelper = null!;
+
+        [SetUp]
+        public void Setup()
+        {
+            var connection = ServiceTestHelper.GetDatabaseConnection();
+            service = new CompetencyLearningResourcesDataService(connection);
+
+            testHelper = new CompetencyLearningResourcesTestHelper(connection);
+        }
+
+        [Test]
+        public void GetCompetencyLearningResourceById_returns_expected_resource()
+        {
+            using var transaction = new TransactionScope();
+            try
+            {
+                // Given
+                var expectedResource = new CompetencyLearningResource
+                {
+                    Id = 1,
+                    CompetencyId = 1,
+                    LearningHubResourceReference = 2,
+                    AdminId = 7,
+                };
+                InsertCompetencyLearningResources();
+
+                // When
+                var result = service.GetCompetencyLearningResourceById(1);
+
+                // Then
+                result.Should().BeEquivalentTo(expectedResource);
+            }
+            finally
+            {
+                transaction.Dispose();
+            }
+        }
+
+        [Test]
+        public void GetCompetencyIdsByLearningHubResourceReference_returns_expected_ids()
+        {
+            using var transaction = new TransactionScope();
+            try
+            {
+                // Given
+                var expectedIds = new[] { 1, 2, 3, 5, 6 };
+                InsertCompetencyLearningResources();
+
+                // When
+                var result = service.GetCompetencyIdsByLearningHubResourceReference(2);
+
+                // Then
+                result.Should().BeEquivalentTo(expectedIds);
+            }
+            finally
+            {
+                transaction.Dispose();
+            }
+        }
+
+        private void InsertCompetencyLearningResources()
+        {
+            var adminId = UserTestHelper.GetDefaultAdminUser().Id;
+            var id1 = testHelper.InsertCompetencyLearningResource(1, 1, 2, adminId);
+            var id2 = testHelper.InsertCompetencyLearningResource(2, 2, 2, adminId);
+            var id3 = testHelper.InsertCompetencyLearningResource(3, 3, 2, adminId);
+            var id4 = testHelper.InsertCompetencyLearningResource(4, 4, 3, adminId);
+            var id5 = testHelper.InsertCompetencyLearningResource(5, 5, 2, adminId);
+            var id6 = testHelper.InsertCompetencyLearningResource(6, 6, 2, adminId);
+        }
+    }
+}
--- a/DigitalLearningSolutions.Data.Tests/DataServices/CompetencyLearningResourcesDataServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/DataServices/CompetencyLearningResourcesDataServiceTests.cs
@@ -32,7 +32,7 @@
                 {
                     Id = 1,
                     CompetencyId = 1,
-                    LearningHubResourceReference = 2,
+                    LearningHubResourceReferenceId = 2,
                     AdminId = 7,
                 };
                 InsertCompetencyLearningResources();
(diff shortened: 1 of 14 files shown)


DigitalLearningSolutions.Data.Tests/DataServices/LearningLogItemsDataServiceTests.cs
```diff
@@ -0,0 +1,131 @@
...
+
+                // Then
+                using (new AssertionScope())
+                {
+                    result.Should().NotBeNull();
+                    result!.DelegateId.Should().Be(delegateId);
+                    result.AddedDate.Should().Be(addedDate);
+                    result.CompetencyLearningResourceId.Should().Be(competencyLearningResourceId);
+                    result.ResourceLink.Should().Be(resourceLink);
+                    result.ResourceName.Should().Be(resourceName);
```
DanBloxham-sw, line 64: Do you think it would be worth checking the other columns are null, just to make sure illogical additions are not made (e.g. CompletedDate not being null etc.)?

DigitalLearningSolutions.Data.Tests/TestHelpers/LearningLogItemsTestHelper.cs
```diff
@@ -0,0 +1,78 @@
...
+                @"SELECT
+                        ID,
+                        LearningLogItemID,
+                        CompetencyID,
+                        AssociatedDate
+                    FROM LearningLogItemCompetencies"
+            );
+        }
+
+        public class LearningLogItemDetails
```
DanBloxham-sw, line 54: We are going to be using LearningLogItems in a bunch of other tickets. Do you reckon its worth moving this class into its own file under Models, to reduce the number of potentially redundant classes when everything gets merged?

DigitalLearningSolutions.Data.Tests/Services/ActionPlanServiceTests.cs
```diff
@@ -0,0 +1,110 @@
...
+            ).MustHaveHappenedOnceExactly();
+            A.CallTo(
+                () => learningLogItemsDataService.InsertCandidateAssessmentLearningLogItem(
+                    selfAssessmentId,
+                    learningLogId
+                )
+            ).MustHaveHappenedOnceExactly();
+            A.CallTo(
+                () => learningLogItemsDataService.InsertLearningLogItemCompetencies(learningLogId, A<int>._, addedDate)
+            ).MustHaveHappened(5, Times.Exactly);
```
DanBloxham-sw, line 107: At the risk of making this test even larger, what do you think of additionally adding a similar set of assertions for this to test the competency Ids are the correct ones, such as (excuse the formatting):

```
 var expectedMatchingCompetencies = new[] { 2, 3, 5, 6, 8 };
            foreach (var competencyId in expectedMatchingCompetencies)
            {
                A.CallTo(
                    () => learningLogItemsDataService.InsertLearningLogItemCompetencies(learningLogId, competencyId, addedDate)
                ).MustHaveHappenedOnceExactly(); 
            }
```
Would probably need expectedMatchingCompetencies to be defined alongside the other collections of competencies somewhere obvious.

DanBloxham-sw (commented):
Looks good on the whole. Left a few comments about potential improvements.

Unfortunately, at time of writing Kevin has not confirmed my guess at the LearningLogItemCompetencies insertion process, but I feel pretty confident with what we have here. With that, and the state of the API client work, feel free to argue against the suggestions in the comments.

DanBloxham-sw:
Kevin has just confirmed that the process for adding things to LearningLogItemCompetencies is exactly how we have implemented it here, so no need to worry about changing it.

DigitalLearningSolutions.Data.Tests/DataServices/CompetencyLearningResourcesDataServiceTests.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        private void InsertCompetencyLearningResources()
+        {
+            var adminId = UserTestHelper.GetDefaultAdminUser().Id;
+            var id1 = testHelper.InsertCompetencyLearningResource(1, 1, 2, adminId);
+            var id2 = testHelper.InsertCompetencyLearningResource(2, 2, 2, adminId);
+            var id3 = testHelper.InsertCompetencyLearningResource(3, 3, 2, adminId);
+            var id4 = testHelper.InsertCompetencyLearningResource(4, 4, 3, adminId);
+            var id5 = testHelper.InsertCompetencyLearningResource(5, 5, 2, adminId);
+            var id6 = testHelper.InsertCompetencyLearningResource(6, 6, 2, adminId);
```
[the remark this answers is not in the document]
AlexJacksonDS: I don't even need to do that, just don't assign them to variables.

DigitalLearningSolutions.Data.Tests/DataServices/CompetencyLearningResourcesDataServiceTests.cs
```diff
@@ -0,0 +1,85 @@
...
+                // Then
+                result.Should().BeEquivalentTo(expectedIds);
+            }
+            finally
+            {
+                transaction.Dispose();
+            }
+        }
+
+        private void InsertCompetencyLearningResources()
```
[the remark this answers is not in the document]
AlexJacksonDS: I don't see any sensible way to do it without identity insert since we won't be able to get the item by Id since the Id will auto-increment.

Given it is against the test database though, I don't think it'll matter as long as we assume we always insert records in transactions like we are doing here (i.e. no permanent test data ever) or we create some permanent test data right now.

Otherwise I think the only options that would be more future proof would be to insert them at the max int value (unlikely to hit this) or delete all data in the transaction (which would likely violate some foreign keys if we did have data)

DigitalLearningSolutions.Data.Tests/TestHelpers/CompetencyLearningResourcesTestHelper.cs
```diff
@@ -0,0 +1,33 @@
...
+            int adminId
+        )
+        {
+            return connection.QuerySingle<int>(
+                @"SET IDENTITY_INSERT dbo.CompetencyLearningResources ON
+                    INSERT INTO CompetencyLearningResources
+                    (ID, CompetencyId, LHResourceReferenceID, AdminId)
+                    OUTPUT Inserted.ID
+                    VALUES (@id, @competencyId, @lhResourceId, @adminId)
+                    SET IDENTITY_INSERT dbo.Progress OFF",
```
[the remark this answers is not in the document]
AlexJacksonDS: Copy and paste error, will fix.

DigitalLearningSolutions.Data.Tests/DataServices/LearningLogItemsDataServiceTests.cs
AlexJacksonDS: Yeah probably.
If I'm adding a new LearningLogItems class for the below, I guess it needs to have properties for all the columns in the database, and we can then check the rest are null with a full select

DigitalLearningSolutions.Data.Tests/TestHelpers/LearningLogItemsTestHelper.cs
AlexJacksonDS: Added this with columns names as per the database. We'll probably want to rename columns as we actually use this properly later.

DigitalLearningSolutions.Data.Tests/DataServices/LearningLogItemsDataServiceTests.cs
```diff
@@ -0,0 +1,147 @@
...
+                // Then
+                using (new AssertionScope())
+                {
+                    result.Should().NotBeNull();
+                    result!.LoggedById.Should().Be(delegateId);
+                    result.LoggedDate.Should().Be(addedDate);
+                    result.LinkedCompetencyLearningResourceId.Should().Be(competencyLearningResourceId);
+                    result.ExternalUri.Should().Be(resourceLink);
+                    result.Activity.Should().Be(resourceName);
+                    result.ActivityTypeId.Should().NotBe(0);
```
DanBloxham-sw, line 65: I feel like the ActivityTypeId should be 8 explicitly, as it comes from a reference table, but I suppose we can't guarantee that given the migration. Do you think this is worth changing? An alternative might be making the query used JOIN onto the ActivityTypes table, and recovering the TypeLabel and making sure that is "Learning Hub Resource"

DanBloxham-sw (commented):
I think this is good. I've left one comment about making a test more specific, not sure if it's worth it.

Also unsure if I should approve this given it doesn't use the real API client yet, but what I've seen so far is worthy of approval.

DigitalLearningSolutions.Data.Tests/DataServices/LearningLogItemsDataServiceTests.cs
AlexJacksonDS: Changed this to get the TypeLabel as I think it is safer than assuming an ID

DanBloxham-sw (commented):
Yeah I think this is now good sans API client.

I'll not "approve" it yet, as I will wait until we're fully happy to merge with the API client, but I'm happy for this to be moved on.

DanBloxham-sw (approved):
Actually, I'm just approving this properly, so it can be merged as is, then a second MR can be made to incorporate the API client methods.

livzorn (approved):
Looks good!

PR #742, opened 2021-11-22
HEEDLS-604-Add-Support-tickets-page

wilkinsm:
### JIRA link
[_HEEDLS-XXXX_](https://softwiretech.atlassian.net/browse/HEEDLS-604)

### Description
Create a new page for support tickets.
This holds an iframe which show the "old site" support tickets page.
Update the support side nav so that "support tickets" link takes you to the new page rather than redirect to the old site.

### Screenshots
**Desktop**
image

**Tablet**
image

**Mobile**
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DanBloxham-sw (changes requested):
Looking good, few minor comments on formatting, and one about potentially making the support pages more inline with the rest of the system.

DigitalLearningSolutions.Web/Views/Support/SupportTickets.cshtml
```diff
@@ -0,0 +1,40 @@
...
+  ViewData["HeaderPath"] = Model.DlsSubApplication.HeaderPath;
+  ViewData["HeaderPathName"] = Model.DlsSubApplication.HeaderPathName;
+}
+
+@section NavMenuItems {
+  @if (Model.DlsSubApplication.Equals(DlsSubApplication.TrackingSystem)) {
+    <partial name="~/Views/TrackingSystem/Shared/_NavMenuItems.cshtml" />
+  } else if (Model.DlsSubApplication.Equals(DlsSubApplication.Frameworks)) {
+    <partial name="~/Views/Frameworks/Shared/_NavMenuItems.cshtml" />
+  }
```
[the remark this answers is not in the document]
DanBloxham-sw: Sorry, actually looks like this is being dealt with in 608, so never mind about this.

DigitalLearningSolutions.Web/ViewModels/Support/SupportViewModel.cs
```diff
@@ -4,7 +4,7 @@
 
     public class SupportViewModel
     {
-        private readonly string currentSystemBaseUrl;
+        protected readonly string currentSystemBaseUrl;
```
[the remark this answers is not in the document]
wilkinsm: Done.

DigitalLearningSolutions.Web/ViewModels/Support/SupportTicketsViewModel.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace DigitalLearningSolutions.Web.ViewModels.Support
+{
+    using DigitalLearningSolutions.Web.Models.Enums;
+
+    public class SupportTicketsViewModel : SupportViewModel
+    {
+        public SupportTicketsViewModel(DlsSubApplication dlsSubApplication, SupportPage currentPage, string currentSystemBaseUrl) : base(dlsSubApplication, currentPage, currentSystemBaseUrl)
```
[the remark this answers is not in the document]
wilkinsm: Done. Regarding the formatter I have noticed something which I've described below.

DigitalLearningSolutions.Web/Views/Support/SupportTickets.cshtml
```diff
@@ -0,0 +1,40 @@
...
+
+<div class="nhsuk-grid-row">
+  <div class="nhsuk-grid-column-one-quarter">
+    <nav class="side-nav-menu" aria-label="Side navigation bar">
+      <partial name="Shared/_SupportSideNavMenu.cshtml" model="@Model" />
+    </nav>
+  </div>
+
+  <div class="nhsuk-grid-column-three-quarters">
+    <h1 id="page-heading" class="nhsuk-heading-xl">Support tickets</h1>
```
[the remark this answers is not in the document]
wilkinsm: Nice. Done.

DigitalLearningSolutions.Web/Views/Support/Shared/_SupportSideNavMenu.cshtml
```diff
@@ -15,11 +15,14 @@
                      link-text="Support"
                      is-current-page="@(Model.CurrentPage == SupportPage.Support)" />
 
-  @if (Model.CurrentPage == SupportPage.HelpDocumentation) {
+  @if (Model.CurrentPage == SupportPage.HelpDocumentation)
+  {
```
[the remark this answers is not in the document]
wilkinsm: Done.
So I've noticed that after running the formatter which puts the open braces on the same line as the IF,
if you then make a trivial change after this and save, in VS at least, "auto-format on save" changes it back. But only sometimes. Something I need to watch out for. It's wierd because I'm sure that auto-format on change is a ReSharper thing.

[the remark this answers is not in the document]
DanBloxham-sw: Interesting. Yeah, something to keep an eye out on. If we see it happening again, might be worth changing the .editorconfig  file with a fix if we can find a reliable one.

DanBloxham-sw (approved):
Looks good. 

On the off chance that 608 gets merged before this, it would be good to pull over whatever changes it makes to this page as well. If this gets merged first, it will be worth mentioning on the other ticket/pull request about this new page that the changes can be enacted on.

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/BasicAuthenticatedAccessibilityTests.cs
```diff
@@ -80,6 +80,7 @@ public class BasicAuthenticatedAccessibilityTests : AccessibilityTestsBase,
         [InlineData("/NotificationPreferences/Edit/DelegateUser", "Update notification preferences")]
         [InlineData("/ChangePassword", "Change password")]
         [InlineData("/TrackingSystem/Support", "Support")]
+        [InlineData("/TrackingSystem/Support/Tickets", "Support tickets")]
```
AlexJacksonDS, line 83: I wouldn't be surprised if this would fail if there were an instance of the old site running to load the old site into the page. We don't have these tests for the other pages loading in the old site in iframes (Pricing, Find a Centre, LearningContent in the main area of the site).
I'm not sure if it is worth keeping this or not?

DigitalLearningSolutions.Web/Styles/index.scss
```diff
@@ -128,6 +128,10 @@ h1#page-heading {
 .responsive-iframe {
   width: 100%;
   height: 100%;
+
+  &.responsive-iframe-min-height {
+    min-height: 550px;
+  }
```
AlexJacksonDS, line 134: Is there a reason we need this here? If we're using the rest of these iframe classes we already have a min height?

DigitalLearningSolutions.Web/Controllers/Support/SupportTicketsController.cs
```diff
@@ -0,0 +1,58 @@
...
+        private readonly IFeatureManager featureManager;
+
+        public SupportTicketsController(IFeatureManager featureManager, IConfiguration configuration)
+        {
+            this.featureManager = featureManager;
+            this.configuration = configuration;
+        }
+
+        [Route("/{dlsSubApplication}/Support/Tickets")]
+        [Authorize(Policy = CustomPolicies.UserCentreAdminOrFrameworksAdmin)]
```
AlexJacksonDS, line 29: We typically have these on the Controller if possible, is there a reason not to do so here?

DigitalLearningSolutions.Web/Controllers/Support/SupportTicketsController.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                var model = new SupportTicketsViewModel(
+                    dlsSubApplication,
+                    SupportPage.SupportTickets,
+                    configuration.GetCurrentSystemBaseUrl()
+                );
+                return View("SupportTickets", model);
+            }
+
+            return RedirectToAction("Index", "Home");
```
AlexJacksonDS, line 55: I suspect this will get merged before 608, so can you put a todo here to resolve these by implementing the new attribute style there. I've done the same in ResourcesController for 521 (just merged). We have all the same issues as the other ticket. Incorrect redirect when unauthorized or features aren't enabled, not great bool names (that will be removed) etc.

DigitalLearningSolutions.Web/Views/Support/SupportTickets.cshtml
```diff
@@ -0,0 +1,40 @@
...
+@using DigitalLearningSolutions.Web.ViewModels.Support
+@model SupportTicketsViewModel
+
+<link rel="stylesheet" href="@Url.Content("~/css/support/support.css")" asp-append-version="true">
+
+@{
+  ViewData["Title"] = "Support tickets";
+  ViewData["Application"] = Model.DlsSubApplication.HeaderExtension;
+  ViewData["HeaderPath"] = Model.DlsSubApplication.HeaderPath;
+  ViewData["HeaderPathName"] = Model.DlsSubApplication.HeaderPathName;
```
AlexJacksonDS, line 11: We shouldn't need these, you have the SetDlsSubApplication attribute on the controller

DigitalLearningSolutions.Web/Views/Support/SupportTickets.cshtml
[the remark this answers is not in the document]
AlexJacksonDS: Either this needs dealing with here or a todo adding (and a note on the 608 ticket). I see no reason why it can't be dealt with here, it was possible for the ResourcesController/View from 521. Saves a bit of work on 608.

DigitalLearningSolutions.Web/Startup.cs
```diff
@@ -121,6 +121,7 @@ public void ConfigureServices(IServiceCollection services)
                         options.ViewLocationFormats.Add("/Views/TrackingSystem/Delegates/{1}/{0}.cshtml");
                         options.ViewLocationFormats.Add("/Views/TrackingSystem/CourseSetup/{1}/{0}.cshtml");
                         options.ViewLocationFormats.Add("/Views/SuperAdmin/{1}/{0}.cshtml");
+                        options.ViewLocationFormats.Add("/Views/Support/{0}.cshtml");
```
AlexJacksonDS, line 124: Just need to confirm this won't conflict at all with the slightly different location I added for Resources.

DigitalLearningSolutions.Web/Views/Support/Shared/_SupportSideNavMenu.cshtml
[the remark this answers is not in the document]
AlexJacksonDS: Yeah I've been having similar issues with VS formatting them differently to Resharper, but I don't use "auto-format on save".
I suspect unless you're using this one in the Resharper settings, it's using a default VS format.
image

AlexJacksonDS (changes requested):
Some stuff to tidy up. This has a lot of the same issues as 521 (i.e. all the not great stuff in the controller will be replaced by 608), so we just need to make sure that it does get done depending on which gets merged first (likely this).

DigitalLearningSolutions.Web/Views/Support/Shared/_SupportSideNavMenu.cshtml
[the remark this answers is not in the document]
wilkinsm: I don't have this option turned on so it will be some VS default going on that I've not noticed before or maybe is new.

DigitalLearningSolutions.Web/Views/Support/SupportTickets.cshtml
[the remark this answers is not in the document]
wilkinsm: Ah yes, done.

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/BasicAuthenticatedAccessibilityTests.cs
wilkinsm: Originally I had the old site running when I did this. I've since turned that off and re-tested. Under thos condition the test still ran fine.
But I know what you mean I could introducing a flaky test so I have dropped this.

DigitalLearningSolutions.Web/Styles/index.scss
wilkinsm: Yes, despite the wrapper having the min-height, in all the browsers I tested the Iframe would not stretch to fill that space and would only fill around half of that.
I tried in a Chrome, Firefox, and Edge - they all appeared to behave the same.
Weighing it up, is it worth spending time trying to figure this out or just add this extra bit of re-usable css?
Could you run this locally and see if you can re-create?

DigitalLearningSolutions.Web/Controllers/Support/SupportTicketsController.cs
wilkinsm: Good spot, its hang over from the main support controller. I've moved them up.

DigitalLearningSolutions.Web/Controllers/Support/SupportTicketsController.cs
wilkinsm: I've added some todos, same as the ones in 521.

DigitalLearningSolutions.Web/Views/Support/SupportTickets.cshtml
wilkinsm: Done. I've removed them from the original support view also.

DigitalLearningSolutions.Web/Startup.cs
wilkinsm: I've removed this in favour of yours.

DigitalLearningSolutions.Web/Styles/index.scss
AlexJacksonDS: Your new support tickets page doesn't behave the same as the Pricing (and others) pages for me in Chrome at least:
image

But I also see what you mean about when this isn't there? (removed the new class from the wrapper div)
image

I'm pretty sure the issue without the class is because we're stuck in a 2 column layout so it's only allowed to reach the bottom of the left column rather than the page. Ideally we would have the same behaviour as the other pages, where the column is allowed to the bottom of the page, but if that proves too difficult, I think what we have now is acceptable. Maybe time box for an hour to see if you can force the column full height without destroying the layout and if not we just go with what we have?

wilkinsm: Attempting a few things doesn't seems to work. 
If I remove the left column and basically match the pricing markup then it works as expected.
So there is something about the 2 column layout that is causing the issue.

AlexJacksonDS: I guess we'll have to leave it for now then, I wouldn't be surprised if this comes back in test though.

AlexJacksonDS (approved):

DigitalLearningSolutions.Web/Styles/index.scss
AlexJacksonDS: I think it might come back (possibly in UAT) as a "Why do we have to scroll this small box when there is all the space on the page"

wilkinsm: Ah ok I see. I will merge this in and we can then re-think if this end result is not acceptable.

PR #743, opened 2021-11-23
DLSV2-415 Prevent SelfAssessment result duplicate entries

daniel-manta-hee:
### JIRA link
https://hee-dls.atlassian.net/browse/DLSV2-415

### Description
1. If a SelfAssessmentResult exists for the same candidate, assessment, result, competency and question, then update it with the new comment and dateTime. 
2. If it doesn't exist, insert a new one. 
3. If there are already not one but multiple results for the same candidate, assessment, result, competency and question, then update only the most recent.

### Screenshots
image

-----
### Developer checks

For a given candidate, assessment, competency and question, I have:
- [ ] Changed the result and checked that is creating new record in table SelfAssessmentResults
- [ ] Changed the comment and checked that is updating most recent version in table SelfAssessmentResults
- [ ] Submitted multiple questions in the same form for checking combinations of the above cases

### DIFF
--- a/DigitalLearningSolutions.Data/Services/SelfAssessmentService.cs
+++ b/DigitalLearningSolutions.Data/Services/SelfAssessmentService.cs
@@ -345,15 +345,29 @@ LEFT OUTER JOIN CandidateAssessmentOptionalCompetencies AS CAOC
                             AND CAQ.AssessmentQuestionID = @assessmentQuestionId
                     )
                     BEGIN
-                        INSERT INTO SelfAssessmentResults
-                          ([CandidateID]
-                          ,[SelfAssessmentID]
-                          ,[CompetencyID]
-                          ,[AssessmentQuestionID]
-                          ,[Result]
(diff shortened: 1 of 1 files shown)


kevwhitt-hee (changes requested):
Looking at the change to the service method, it checks whether any record in SelfAssessmentResults matches on the self assessment ID, candidate ID, assessment question ID and result and if so, returns the most recent one. 
This will usually work but isn't quite what we need. Instead, we need to load their most recent SelfAssessmentResults for the SelfAssessmentID, CandidateID and AssessmentQuestionID and then check whether the Result returned matches the Result that the user has posted in order to decide whether to update it or insert a new one. This will require two separate queries.

kevwhitt-hee (commented):
This looks right. I will do some in-application testing to make sure the change is having the desired effect before approving.

kevwhitt-hee (approved):

PR #744, opened 2021-11-23
Heedls 608 validate sub application attribute has whitelist

JBsoftwire:
### JIRA link
[HEEDLS-608](https://softwiretech.atlassian.net/browse/HEEDLS-608)

### Description
The ValidateAllowedDlsSubApplication attribute now checks the subapplication against a list of valid applications (if provided) 

### Screenshots
N/A

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


JBsoftwire:
This currently doesn't deal with the Resources Controller, because that branch hasn't been merged yet. Once one of these tickets is, the other can be adapted to incorporate it straightforwardly.

DigitalLearningSolutions.Web/ServiceFilter/ValidateAllowedDlsSubApplication.cs
```diff
@@ -1,26 +1,37 @@
...
     {
         private readonly string applicationArgumentName;
+        private readonly IFeatureManager featureManager;
+        private readonly IEnumerable<DlsSubApplication> validApplications;
 
-        public ValidateAllowedDlsSubApplicationAttribute(string applicationArgumentName = "dlsSubApplication")
+        public ValidateAllowedDlsSubApplication(
+            string[] validApplicationNames,
+            IFeatureManager featureManager,
+            string applicationArgumentName = "dlsSubApplication"
```
DanBloxham-sw, line 24: As mentioned in the Knowledge Sharing Session, this can probably be replaced by using the same methodology as in SetDlsSubApplication.

DanBloxham-sw (changes requested):
Few minor things plus what was discussed in the Knowledge Sharing

DigitalLearningSolutions.Web/Helpers/ControllerHelper.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace DigitalLearningSolutions.Web.Helpers
+{
+    public static class ControllerHelper
+    {
+        public static string GetControllerAspName(Type controllerType)
+        {
+            var controllerName = controllerType.Name;
+
+            return controllerName.Substring(0, controllerName.LastIndexOf("Controller", StringComparison.Ordinal));
```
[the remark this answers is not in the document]
JBsoftwire: Do you know if this would also prevent the normal ASP functions identifying it from the name prefix?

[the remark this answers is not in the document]
DanBloxham-sw: Good point, I think they probably would. There'd almost certainly be noticeable knock on effects so I don't think the change here would necessary.

DigitalLearningSolutions.Web.Tests/TestHelpers/ContextHelper.cs
```diff
@@ -22,5 +32,40 @@ public static ActionExecutingContext GetDefaultActionExecutingContext(Controller
                 controller
             );
         }
+
+        public static HttpContext WithMockUser(
```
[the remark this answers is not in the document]
JBsoftwire: That leaves us with no tests in this file, but the controller is very simple at this point so I think that's acceptable.

DigitalLearningSolutions.Web.Tests/Controllers/Support/SupportControllerTests.cs
```diff
@@ -7,28 +7,23 @@
...
         public void Setup()
         {
-            featureManager = A.Fake<IFeatureManager>();
             configuration = A.Fake<IConfiguration>();
-            A.CallTo(() => featureManager.IsEnabledAsync(FeatureFlags.RefactoredTrackingSystem))
-                .Returns(true);
         }
 
         [Test]
         public async Task Frameworks_Support_page_should_be_shown_for_valid_claims()
```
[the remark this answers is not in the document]
JBsoftwire: That leaves us with no tests in this file, but the controller is very simple at this point so I think that's acceptable.

DanBloxham-sw (commented):
I think this is looking good. I've put a single comment on something that I think can be simplified, but feel free to argue back.

DigitalLearningSolutions.Web/ServiceFilter/ValidateAllowedDlsSubApplication.cs
```diff
@@ -40,15 +38,20 @@ public async void OnActionExecuting(ActionExecutingContext context)
...
 
-            var application = (DlsSubApplication?)
-                (context.ActionArguments.ContainsKey(applicationArgumentName)
-                    ? context.ActionArguments[applicationArgumentName]
+            var application = dlsSubApplicationParameterName == null
+                ? null
+                : (DlsSubApplication?)
+                (context.ActionArguments.ContainsKey(dlsSubApplicationParameterName)
+                    ? context.ActionArguments[dlsSubApplicationParameterName]
                     : null);
```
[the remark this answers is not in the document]
JBsoftwire: I don't think it's safe to assume there will always be a sub application, as for example this attribute is on the MyAccount controller which can be accessed with or without a sub application in the route.

The ActionArguments implements a dictionary, which I understand from https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2?view=net-6.0 to throw if it has no key, so the pre-check is needed.

However, the declaration is definitely messy and I'll see if I can find a way to express the logic more clearly.

DanBloxham-sw (approved):
Yeah this looks clearer and now good to me.

DigitalLearningSolutions.Web/ServiceFilter/ValidateAllowedDlsSubApplication.cs
```diff
@@ -0,0 +1,134 @@
...
+            if (!string.IsNullOrEmpty(dlsSubApplicationParameterName) &&
+                HasModelBindingError(context, dlsSubApplicationParameterName!))
+            {
+                SetNotFoundResult(context);
+                return;
+            }
+
+            var application = dlsSubApplicationParameterName == null
+                ? null
+                : GetDlsSubApplicationFromContext(context, dlsSubApplicationParameterName);
```
SteveJacksonSoft, line 52: If you just go ahead and do this when there's a random string (or empty string) passed in as dlsSubApplication, what happens?

SteveJacksonSoft: And what if the url I go to is <dls>/MyAccount - how does it pick up the `Main` application?

DigitalLearningSolutions.Web/ServiceFilter/ValidateAllowedDlsSubApplication.cs
```diff
@@ -0,0 +1,134 @@
...
+            IFeatureManager featureManager
+        )
+        {
+            validApplications = validApplicationNames.Select(Enumeration.FromName<DlsSubApplication>);
+            this.featureManager = featureManager;
+        }
+
+        public void OnActionExecuted(ActionExecutedContext context) { }
+
+        public async void OnActionExecuting(ActionExecutingContext context)
```
SteveJacksonSoft, line 31: It would be nice if we could pull a couple of methods out of this.
I think some stages which could possibly be pulled out are
* Redirect user if application param not recognised
* Parse application param into DlsSubApplication
* Validate whether this page exists on this application
* Validate whether user has permission to access this application

Can we make some improvements in this direction?

DigitalLearningSolutions.Web/Helpers/ControllerHelper.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace DigitalLearningSolutions.Web.Helpers
+{
+    using System;
+
+    public static class ControllerHelper
+    {
+        public static string GetControllerAspName(Type controllerType)
```
SteveJacksonSoft, line 7: Is `AspName` a thing? If not, the risk with calling it `AspName` is that readers will assume it is a thing they don't understand. To make it slightly less intimidating, perhaps `GetControllerNameForAspHttpMethods`?

DigitalLearningSolutions.Web/ServiceFilter/ValidateAllowedDlsSubApplication.cs
```diff
@@ -0,0 +1,134 @@
...
+        {
+            var descriptor = ((ControllerBase)context.Controller).ControllerContext.ActionDescriptor;
+            var routeValues = new Dictionary<string, object>
+            {
+                [applicationArgumentName] = DlsSubApplication.LearningPortal,
+            };
+            context.Result = new RedirectToActionResult(descriptor.ActionName, descriptor.ControllerName, routeValues);
+        }
+
+        private void RedirectToAccessDenied(ActionExecutingContext context)
```
SteveJacksonSoft, line 109: This method does the same kind of thing as the one below. Can we have similar names for them? I prefer `Set...Result` because `RedirectTo...` looks synchronous.

SteveJacksonSoft (changes requested):

DigitalLearningSolutions.Web/ServiceFilter/ValidateAllowedDlsSubApplication.cs
```diff
@@ -0,0 +1,134 @@
...
+    using Microsoft.AspNetCore.Mvc.ModelBinding;
+    using Microsoft.FeatureManagement;
+
+    public class ValidateAllowedDlsSubApplication : IActionFilter
+    {
+        private readonly IFeatureManager featureManager;
+        private readonly IEnumerable<DlsSubApplication> validApplications;
+
+        public ValidateAllowedDlsSubApplication(
+            string[] validApplicationNames,
```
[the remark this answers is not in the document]
JBsoftwire: The compiler gives me the compile-time constant objection if I try to set the default to even an empty string array. However verifying this just now I may have found a slightly ugly workaround.

DigitalLearningSolutions.Web/ServiceFilter/ValidateAllowedDlsSubApplication.cs
JBsoftwire: When I go to /MyAccount raw, the context contains a 'dlsSubApplicaition' ActionArgument with value 'main'. I admit I wasn't aware this is how it worked before - I thought it just had a null application if there was no route value. I'm not sure where the configuration responsible for this behaviour is.

If the string in the subapp argument is empty or unparseable, it throws.

DigitalLearningSolutions.Web/ServiceFilter/ValidateAllowedDlsSubApplication.cs
JBsoftwire: I've reorganised it into something that looks clearer, but I'm not sure whether two out variables is best practice, my intuition dislikes them.

DigitalLearningSolutions.Web/ServiceFilter/ValidateAllowedDlsSubApplication.cs
```diff
@@ -33,42 +37,83 @@ public async void OnActionExecuting(ActionExecutingContext context)
...
             }
 
-            var application = dlsSubApplicationParameterName == null
-                ? null
-                : GetDlsSubApplicationFromContext(context, dlsSubApplicationParameterName);
+            application = GetDlsSubApplicationFromContext(context, dlsSubApplicationParameterName!);
+            return true;
+        }
 
+        private async Task<bool> ApplicationIsInaccessibleByPage(
```
SteveJacksonSoft, line 84: This name makes the method seem pure, but it actually sets the response.
Can we change the name to show that?

DigitalLearningSolutions.Web/ServiceFilter/ValidateAllowedDlsSubApplication.cs
```diff
@@ -33,42 +37,83 @@ public async void OnActionExecuting(ActionExecutingContext context)
...
+
+            if (await ApplicationIsInaccessibleByPage(context, application))
+            {
+                return;
+            }
+
+            ValidateUserHasPermissionForApplication(context, user, application, applicationParameterName!);
+        }
+
+        private bool TryGetDlsSubApplication(
```
SteveJacksonSoft, line 57: We can return a tuple of the param name and the application instead of a bool to say whether they've been successfully parsed - the values of param name and application give us more info than the bool anyway.

DigitalLearningSolutions.Web/ServiceFilter/ValidateAllowedDlsSubApplication.cs
```diff
@@ -33,42 +37,83 @@ public async void OnActionExecuting(ActionExecutingContext context)
...
                 SetNotFoundResult(context);
-                return;
+                application = null;
+                return false;
             }
 
-            var application = dlsSubApplicationParameterName == null
-                ? null
-                : GetDlsSubApplicationFromContext(context, dlsSubApplicationParameterName);
+            application = GetDlsSubApplicationFromContext(context, dlsSubApplicationParameterName!);
```
SteveJacksonSoft, line 80: Does this definitely give us a non-null value for application? I think it would be nice to rule out the null-application case when we call this method.
E.g. if the application param on the action is nullable or optional.

DigitalLearningSolutions.Web/ServiceFilter/ValidateAllowedDlsSubApplication.cs
SteveJacksonSoft: The top level definitely looks nicer, but I think you're right, we can improve that method with two out variables.

DigitalLearningSolutions.Web/ServiceFilter/ValidateAllowedDlsSubApplication.cs
JBsoftwire: There's a check that the actual arguments actually contains the key the actionDescription claims they do, but I don't know if there's a way to stop the value attached to that key being null. Maybe the model binder would have pitched a fit at that point, however.

DigitalLearningSolutions.Web/ServiceFilter/ValidateAllowedDlsSubApplication.cs
JBsoftwire: Strictly speaking they don't contain the same information. It's possible for a successful parse to return a null application, and this will then pass the permissions-based validation (you can't lack permission to access a null application) but fail the whitelist-based validation (because null isn't one of the applications permitted to access the page). This is distinct from the parse failing, which will cause the validation to end after either setting or not setting the Not Found Result.

SteveJacksonSoft: But either way we return a NotFound right? If the application we parse out is null, I would always expect it to return a NotFound. So we don't really care if it's a successful parse in that case. Do you disagree?

DigitalLearningSolutions.Web/ServiceFilter/ValidateAllowedDlsSubApplication.cs
SteveJacksonSoft: Okay, well I think it's fine to return a nullable application if we're not sure, but I think we should return a NotFound response if it's null after we call `TryGetDlsSubApplication()`, because I don't think this attribute should ever be used by an endpoint that isn't expecting some non-null dlsSubApplication. There are only a few endpoints that aren't part of an sub-application (if any), and it's very unlikely any of them will need to appear with other sub-application nav bars .

DigitalLearningSolutions.Web/ServiceFilter/ValidateAllowedDlsSubApplication.cs
```diff
@@ -0,0 +1,182 @@
...
+        private readonly IFeatureManager featureManager;
+        private readonly IEnumerable<DlsSubApplication> validApplications;
+
+        public ValidateAllowedDlsSubApplication(
+            IFeatureManager featureManager,
+            string[]? validApplicationNames = null
+        )
+        {
+            validApplications = validApplicationNames == null
+                ? new DlsSubApplication[] { }
```
SteveJacksonSoft, line 28: Think it would be slightly nicer to use `Enumeration.GetAll<DlsSubApplication>()` in the default case. Then ValidApplications has a more consistent meaning, and you don't need the `.Any()` in `ApplicationIsInaccessibleByPage()`

DigitalLearningSolutions.Web/ServiceFilter/ValidateAllowedDlsSubApplication.cs
JBsoftwire: We don't necessarily return a NotFound if the parse is successful. If the endpoint has a restricted list of valid applications it will return NotFound (because 'null' is not on that list), but if it doesn't have such a list it will just wave the action through.

JBsoftwire: If we want to return a Not Found either way I can implement that.

SteveJacksonSoft (approved):
Looks nice, merge at will!

PR #755, opened 2021-11-25
Heedls 595 course delegates unlock course

showkath10:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-595

### Description
1. Added new action method in DelegateProgressConroller to unlock the course progress
2. When "Unlock progress" is clicked, it will update PLLocked to 0 and refresh the page
3. The "Unlock progress" button is available in different two pages - Course course card available in Course delegates page and View Delegate> Manage delegate 

### Screenshots
image
image

-----
### Developer checks

I have:
- [] Updated/added documentation in Swiki and/or Readme. Links (if any) are below:
  - [doc_1_here](link_1_here)
- [] Updated my Jira ticket with information about other parts of the system that were touched as part of the MR and have to be sanity tested to ensure nothing’s broken.

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/Delegates/DelegateProgressControllerTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/Delegates/DelegateProgressControllerTests.cs
@@ -14,6 +14,7 @@
     {
         private ICourseService courseService = null!;
         private DelegateProgressController delegateProgressController = null!;
+        private INotificationService notificationService = null;
         private IProgressService progressService = null!;
         private IUserService userService = null!;
 
@@ -30,13 +31,32 @@
             }
         }
 
+        private static IEnumerable<TestCaseData> UnlockCourseProgressData
+        {
+            get
+            {
+                yield return new TestCaseData(DelegateProgressAccessRoute.CourseDelegates, "CourseDelegates", "Index")
+                    .SetName("UnlockCourseProgress_redirects_to_course_delegates_progress");
+                yield return
+                    new TestCaseData(DelegateProgressAccessRoute.ViewDelegate, "ViewDelegate", "Index").SetName(
+                        "UnlockCourseProgress_redirects_to_view_delegate"
+                    );
+            }
+        }
+
         [SetUp]
         public void Setup()
         {
             courseService = A.Fake<ICourseService>();
             userService = A.Fake<IUserService>();
             progressService = A.Fake<IProgressService>();
-            delegateProgressController = new DelegateProgressController(courseService, userService, progressService)
+            notificationService = A.Fake<INotificationService>();
+            delegateProgressController = new DelegateProgressController(
+                    courseService,
+                    userService,
+                    progressService,
+                    notificationService
+                )
                 .WithDefaultContext()
                 .WithMockUser(true);
         }
@@ -62,6 +82,41 @@
             var result = delegateProgressController.EditSupervisor(formData, progressId, accessedVia);
 
             // Then
+            result.Should().BeRedirectToActionResult().WithControllerName(expectedController)
+                .WithActionName(expectedAction);
+        }
+
+        [Test]
+        [TestCaseSource(
+            typeof(DelegateProgressControllerTests),
+            nameof(UnlockCourseProgressData)
+        )]
+        public void UnlockCourseProgress_redirects_to_correct_action_and_unlocks_progress_and_sends_notification(
+            DelegateProgressAccessRoute accessedVia,
+            string expectedController,
+            string expectedAction
+        )
+        {
+            // Given
+            const int progressId = 1;
+            const int delegateId = 1;
+            const int customisationId = 1;
+
+            A.CallTo(() => progressService.UnlockProgress(progressId)).DoesNothing();
+            A.CallTo(() => notificationService.SendUnlockRequest(progressId)).DoesNothing();
+
+            // When
+            var result = delegateProgressController.UnlockProgress(
+                progressId,
+                customisationId,
+                delegateId,
+                accessedVia
+            );
+
+            // Then
+            A.CallTo(() => progressService.UnlockProgress(progressId)).MustHaveHappened();
+            A.CallTo(() => notificationService.SendUnlockRequest(progressId)).MustHaveHappened();
+
             result.Should().BeRedirectToActionResult().WithControllerName(expectedController)
                 .WithActionName(expectedAction);
         }
--- a/DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/DelegateProgressController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/DelegateProgressController.cs
@@ -22,18 +22,21 @@
     public class DelegateProgressController : Controller
     {
         private readonly ICourseService courseService;
+        private readonly INotificationService notificationService;
         private readonly IProgressService progressService;
         private readonly IUserService userService;
 
         public DelegateProgressController(
             ICourseService courseService,
             IUserService userService,
-            IProgressService progressService
+            IProgressService progressService,
+            INotificationService notificationService
         )
         {
             this.courseService = courseService;
             this.userService = userService;
             this.progressService = progressService;
+            this.notificationService = notificationService;
         }
 
         public IActionResult Index(int progressId, DelegateProgressAccessRoute accessedVia)
@@ -92,18 +95,21 @@
             return RedirectToAction("Index", "ViewDelegate", new { formData.DelegateId });
         }
 
-        [HttpPost]
-        [Route("/UnlockProgress/{delegateId:int}")]
+        [HttpGet]
+        [Route("UnlockProgress")]
         public IActionResult UnlockProgress(
             int progressId,
+            int customisationId,
             int delegateId,
-            DelegateProgressAccessRoute accessedVia)
+            DelegateProgressAccessRoute accessedVia
+        )
         {
-            progressService.UnlockProgress(progressId, delegateId);
+            progressService.UnlockProgress(progressId);
+            notificationService.SendUnlockRequest(progressId);
 
             if (accessedVia.Equals(DelegateProgressAccessRoute.CourseDelegates))
             {
-                return RedirectToAction("Index", new { progressId, accessedVia });
+                return RedirectToAction("Index", "CourseDelegates", new { customisationId });
             }
 
             return RedirectToAction("Index", "ViewDelegate", new { delegateId });
--- a/DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/DelegateProgressController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/DelegateProgressController.cs
@@ -91,5 +91,22 @@
 
             return RedirectToAction("Index", "ViewDelegate", new { formData.DelegateId });
         }
+
+        [HttpPost]
+        [Route("/UnlockProgress/{delegateId:int}")]
+        public IActionResult UnlockProgress(
+            int progressId,
+            int delegateId,
+            DelegateProgressAccessRoute accessedVia)
+        {
+            progressService.UnlockProgress(progressId, delegateId);
+
+            if (accessedVia.Equals(DelegateProgressAccessRoute.CourseDelegates))
+            {
+                return RedirectToAction("Index", new { progressId, accessedVia });
+            }
+
+            return RedirectToAction("Index", "ViewDelegate", new { delegateId });
+        }
     }
 }
--- a/DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/CourseDelegates/_SearchableCourseDelegateCard.cshtml
+++ b/DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/CourseDelegates/_SearchableCourseDelegateCard.cshtml
@@ -64,6 +64,14 @@
             @Model.PassRate%
           </dd>
         </div>
+        <div class="nhsuk-summary-list__row">
+          <dt class="nhsuk-summary-list__key">
+            Progress locked
+          </dt>
+          <dd class="nhsuk-summary-list__value">
+            @(Model.Locked? "Yes": "No")
+          </dd>
+        </div>
       </dl>
 
       <a class="nhsuk-button expander-card__button"
@@ -74,17 +82,17 @@
          asp-route-accessedVia="@DelegateProgressAccessRoute.CourseDelegates">
         View progress
       </a>
-      @*@if (Model.Locked) {*@
+      @if (Model.Locked) {
         <a class="nhsuk-button nhsuk-button--secondary expander-card__button"
            role="button"
            asp-controller="DelegateProgress"
            asp-action="UnlockProgress"
            asp-route-progressId="@Model.ProgressId"
-           asp-route-delegateId="@Model.DelegateId"
+           asp-route-customisationId="@Model.CustomisationId"
            asp-route-accessedVia="@DelegateProgressAccessRoute.CourseDelegates">
           Unlock progress
         </a>
-      @*}*@
+      }
       @if (Model.RemovedDate == null) {
         <a class="nhsuk-button delete-button expander-card__button" role="button">
           Remove from course
--- a/DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/CourseDelegates/_SearchableCourseDelegateCard.cshtml
+++ b/DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/CourseDelegates/_SearchableCourseDelegateCard.cshtml
@@ -74,11 +74,17 @@
          asp-route-accessedVia="@DelegateProgressAccessRoute.CourseDelegates">
         View progress
       </a>
-      @if (Model.Locked) {
-        <a class="nhsuk-button nhsuk-button--secondary expander-card__button" role="button">
+      @*@if (Model.Locked) {*@
+        <a class="nhsuk-button nhsuk-button--secondary expander-card__button"
+           role="button"
+           asp-controller="DelegateProgress"
+           asp-action="UnlockProgress"
+           asp-route-progressId="@Model.ProgressId"
+           asp-route-delegateId="@Model.DelegateId"
+           asp-route-accessedVia="@DelegateProgressAccessRoute.CourseDelegates">
           Unlock progress
         </a>
-      }
+      @*}*@
       @if (Model.RemovedDate == null) {
         <a class="nhsuk-button delete-button expander-card__button" role="button">
           Remove from course
--- a/DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/ViewDelegate/_DelegateCourseInfoCard.cshtml
+++ b/DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/ViewDelegate/_DelegateCourseInfoCard.cshtml
@@ -105,6 +105,15 @@
           <dd class="nhsuk-summary-list__actions"></dd>
         </div>
 
+        <div class="nhsuk-summary-list__row">
+          <dt class="nhsuk-summary-list__key">
+            Progress locked
+          </dt>
+          <dd class="nhsuk-summary-list__value">
+            @(Model.IsProgressLocked ? "Yes" : "No")
+          </dd>
+        </div>
+
         @if (Model.IsAssessed) {
           <div class="nhsuk-summary-list__row details-list-with-button__row">
             <dt class="nhsuk-summary-list__key">
@@ -150,13 +159,24 @@
         View progress
       </a>
       <a class="nhsuk-button nhsuk-button--secondary" role="button">See learning log</a>
-      <a
-        class="nhsuk-button delete-button"
-        role="button"
-        asp-controller="ViewDelegate"
-        asp-action="ConfirmRemoveFromCourse"
-        asp-route-delegateId="@Model.DelegateId"
-        asp-route-customisationId="@Model.CustomisationId">
+
+      @if (Model.IsProgressLocked) {
+        <a class="nhsuk-button nhsuk-button--secondary"
+           role="button"
+           asp-controller="DelegateProgress"
+           asp-route-progressId="@Model.ProgressId"
+           asp-route-delegateId="@Model.DelegateId"
+           asp-action="UnlockProgress"
+           asp-route-accessedVia="@DelegateProgressAccessRoute.ViewDelegate">
+          Unlock progress
+        </a>
+      }
+      <a class="nhsuk-button delete-button"
+         role="button"
+         asp-controller="ViewDelegate"
+         asp-action="ConfirmRemoveFromCourse"
+         asp-route-delegateId="@Model.DelegateId"
+         asp-route-customisationId="@Model.CustomisationId">
         Remove
       </a>
     </div>
--- a/DigitalLearningSolutions.Data.Tests/DataServices/CourseDataServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/DataServices/CourseDataServiceTests.cs
@@ -53,7 +53,8 @@ namespace DigitalLearningSolutions.Data.Tests.DataServices
             "xxxx",
             "xxxxxx",
             "",
-            101
+            101,
+            false
         );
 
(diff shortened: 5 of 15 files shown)


DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/Delegates/DelegateProgressControllerTests.cs
```diff
@@ -14,6 +14,7 @@ public class DelegateProgressControllerTests
     {
         private ICourseService courseService = null!;
         private DelegateProgressController delegateProgressController = null!;
+        private INotificationService notificationService = null;
```
wilkinsm, line 17: Missing null-forgiving operator.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/CourseDelegates/_SearchableCourseDelegateCard.cshtml
```diff
@@ -64,6 +64,14 @@
             @Model.PassRate%
           </dd>
         </div>
+        <div class="nhsuk-summary-list__row">
```
wilkinsm, line 67: In the wireframe this is a label rather than a table row. Assuming this is how they want it, please could you update the Jira ticket to reflect this change so as not to be flagged as an issue during testing.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/ViewDelegate/_DelegateCourseInfoCard.cshtml
```diff
@@ -105,6 +105,15 @@
           <dd class="nhsuk-summary-list__actions"></dd>
         </div>
 
+        <div class="nhsuk-summary-list__row">
```
wilkinsm, line 108: In the wireframe this is a label rather than a table row. Assuming this is how they want it, please could you update the Jira ticket to reflect this change so as not to be flagged as an issue during testing.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/DelegateProgressController.cs
```diff
@@ -91,5 +94,25 @@ DelegateProgressAccessRoute accessedVia
...
+        [Route("UnlockProgress")]
+        public IActionResult UnlockProgress(
+            int progressId,
+            int customisationId,
+            int delegateId,
+            DelegateProgressAccessRoute accessedVia
+        )
+        {
+            progressService.UnlockProgress(progressId);
+            notificationService.SendUnlockRequest(progressId);
```
wilkinsm, line 108: There is a requirement to ensure the link sent out in the notification email directs the user to the this refactored site rather than the old site if the TrackingSystem feature flag is on (i.e. TrackingSystem feature is enabled).
I don't see any changes made related to this.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/ViewDelegate/_DelegateCourseInfoCard.cshtml
```diff
@@ -150,13 +159,24 @@
...
       <a class="nhsuk-button nhsuk-button--secondary" role="button">See learning log</a>
-      <a
-        class="nhsuk-button delete-button"
-        role="button"
-        asp-controller="ViewDelegate"
-        asp-action="ConfirmRemoveFromCourse"
-        asp-route-delegateId="@Model.DelegateId"
-        asp-route-customisationId="@Model.CustomisationId">
+
+      @if (Model.IsProgressLocked) {
```
wilkinsm, line 163: Please could you note down this additional button in the Jira ticket requirements so that if can be checked during testing.

wilkinsm (changes requested):
Generally looking good.
One minor refactor, but then there is a missing requirement regarding the email. And some notes required on the Jira ticket.

wilkinsm:
Can we have a screenshot of the other unlock button?

DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/Delegates/DelegateProgressControllerTests.cs
showkath10: Sorry, added null-forgiving operator as suggested

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/CourseDelegates/_SearchableCourseDelegateCard.cshtml
showkath10: Updated JIra as suggested

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/ViewDelegate/_DelegateCourseInfoCard.cshtml
showkath10: Updated Jira as suggested

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/DelegateProgressController.cs
showkath10: The url is read from the database. This value is read from ConfigName column which is in config table 
SELECT * FROM Config WHERE ConfigName = 'TrackingSystemBaseURL'
We need to update the url here. I have updated the ticket

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/ViewDelegate/_DelegateCourseInfoCard.cshtml
showkath10: Updated jira thanks

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/DelegateProgressController.cs
wilkinsm: So the requirement here would be to check the feature manager to see if the tracking system is enabled. 
If tracking is enabled then use the new site base url in the email.
If tracking is disabled then use the old site base url in the email as before.
This should not require a DB change.
We either use the old value from the DB/AppSettings.
Or the new url from AppSettings.
The DigitalLearningSolutions.Data.Helpers.ConfigHelper will be of use here.

showkath10: Added a check as suggested. This will set url based on tracking system 
Please ensure you have config settings in the database (in config table)

image

DigitalLearningSolutions.Data.Tests/Services/UnlockServiceTests.cs
```diff
@@ -1,71 +1,74 @@
...
+    public class UnlockServiceTests
+    {
+        private IConfigService configService;
+        private IEmailService emailService;
+        private INotificationDataService notificationDataService;
+        private NotificationService notificationService;
+        private IFeatureManager featureManager;
+
+        [SetUp]
+        public void Setup()
```
wilkinsm, line 19: Can we have a test that checks the feature manager is implemented and working.
The `// Then`  section can be something along the lines of:

```
A.CallTo(() =>
			emailService.SendEmail(A<Email>.That.Matches(e => e.Body.TextBody.Contains(<Expected URL>)))
		)
		.MustHaveHappened();
		```

File: DigitalLearningSolutions.Data/Services/NotificationService.cs [hunk 7]
```diff
@@ -1,61 +1,81 @@
...
+            var unlockData = notificationDataService.GetUnlockData(progressId);
+            if (unlockData == null)
+            {
+                throw new NotificationDataException(
+                    $"No record found when trying to fetch Unlock Data. Progress ID: {progressId}"
+                );
+            }
+
+            unlockData.ContactForename = unlockData.ContactForename == "" ? "Colleague" : unlockData.ContactForename;
+            var trackingSystemBaseUrl = configService.GetConfigValue(ConfigService.TrackingSystemBaseUrl) ??
```
wilkinsm, line 45: Looking at lines 45 to 58, can we look at simplyfying this? Example could be:

```
var refactoredTrackingSystemEnabled = await featureManager.IsEnabledAsync("RefactoredTrackingSystem");
var baseUrlConfigOption = refactoredTrackingSystemEnabled ? ConfigService.AppBaseUrl : ConfigService.TrackingSystemBaseUrl;
var trackingSystemBaseUrl = configService.GetConfigValue(baseUrlConfigOption) ??
    throw new ConfigValueMissingException(
      configService.GetConfigValueMissingExceptionMessage(baseUrlConfigOption)
    ); 
var unlockUrl = new UriBuilder(trackingSystemBaseUrl);
```

DigitalLearningSolutions.Data/Services/NotificationService.cs
```diff
@@ -1,61 +1,81 @@
...
+                                        throw new ConfigValueMissingException(
+                                            configService.GetConfigValueMissingExceptionMessage("TrackingSystemBaseUrl")
+                                        );
+
+            var v2AppBaseUrl = configService.GetConfigValue(ConfigService.AppBaseUrl) ??
+                               throw new ConfigValueMissingException(
+                                   configService.GetConfigValueMissingExceptionMessage("AppBaseUrl")
+                               );
+
+            var trackingSystemSupportEnabled = featureManager.IsEnabledAsync("RefactoredTrackingSystem").Result;
```
wilkinsm, line 55: Could we look to make the `SendUnlockRequest` method `async` and then we can `await` this line.

DigitalLearningSolutions.Data.Tests/Services/UnlockServiceTests.cs
wilkinsm: We should additionally check that a call to the feature manager happened.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/DelegateProgressController.cs
wilkinsm: @showkath10 Can you verify that this value exists in all the required databases.
And/or write a migration script that ensures it is populated correctly.

DigitalLearningSolutions.Data/Services/NotificationService.cs
showkath10: I asked if we need to make it async in show and tell. It was told it is not necessary so I left it as existed before

DigitalLearningSolutions.Data/Services/NotificationService.cs
showkath10: I refactored it as suggested but SendUnlockRequest is not async since it was told it is not necessary in show and tell

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/DelegateProgressController.cs
rhimcg: I've added this migration

DigitalLearningSolutions.Data.Tests/Services/UnlockServiceTests.cs
rhimcg: added a test that its calling the feature manager and that its setting the url properly in the email

DigitalLearningSolutions.Data.Migrations/202201060948_AddV2AppBaseUrl.cs
```diff
@@ -0,0 +1,23 @@
+namespace DigitalLearningSolutions.Data.Migrations
```
wilkinsm, line 1: Looking at this I have a feeling that for any version of this app that has not run this migration but already has a config by name of `V2AppBaseUrl` may end up with 2 rows with the same name given `ConfigName` does not have any constriants on it.
My DB does not have this config value so this will work fine but I think Showkaths did so running this migration would lead to 2 rows with the same name.

DigitalLearningSolutions.Data.Tests/Services/UnlockServiceTests.cs
```diff
@@ -1,71 +1,124 @@
...
+
+            // Then
+            A.CallTo(() =>
+                    emailService.SendEmail(A<Email>._)
+                )
+                .MustHaveHappened();
+        }
+
+        [Test]
+        public void Trying_to_sent_unlock_makes_request_to_feature_manager_to_get_correct_url()
```
wilkinsm, line 75: NitPick - Test name `Trying_to_sent...` should read `Trying_to_send...`

DigitalLearningSolutions.Data.Tests/Services/UnlockServiceTests.cs
```diff
@@ -1,71 +1,124 @@
...
+
+            //When
+            notificationService.SendUnlockRequest(1);
+
+            //Then
+            A.CallTo(() => featureManager.IsEnabledAsync(A<string>._)).MustHaveHappened();
+        }
+
+        [Test]
+        public void trying_to_send_unlock_request_send_email_with_correct_old_url()
```
wilkinsm, line 89: NitPick - should be uppercase `T` at start of test name.

DigitalLearningSolutions.Data.Tests/Services/UnlockServiceTests.cs
```diff
@@ -1,71 +1,124 @@
...
+            //Then
+            A.CallTo(() =>
+                    emailService.SendEmail(A<Email>.That.Matches(e => e.Body.TextBody.Contains("https://old-tracking-system.com/")))
+                )
+                .MustHaveHappened();
+
+        }
+
+        [Test]
+        public void trying_to_send_unlock_request_send_email_with_correct_new_url()
```
wilkinsm, line 107: NitPick - should be uppercase `T` at start of test name.

wilkinsm (changes requested):
Thanks for sorting these Rhianna, just a few minor points.

DigitalLearningSolutions.Data.Migrations/202201060948_AddV2AppBaseUrl.cs
rhimcg: cool I've changed it so it so it checks there isn't one with the same config value before it add, the down still removes anything with that configName

wilkinsm: This has got me thinking on whether we need a `down`. Given the original concern that the value may already exist, if that is true, then removing this value may inadvertently break anything that was reliant on this value prior to this point.

wilkinsm: Nice one thanks.

wilkinsm (approved):
Nice one, good to go to TLR.

DigitalLearningSolutions.Data/Services/NotificationService.cs
```diff
@@ -1,61 +1,74 @@
...
+To review and unlock their progress, visit the this url: ${unlockUrl.Uri}.",
+                HtmlBody = $@"<body style= 'font-family: Calibri; font-size: small;'>
+                                    <p>Dear {unlockData?.ContactForename}</p>
+                                    <p>Digital Learning Solutions Delegate, {unlockData?.DelegateName}, has requested that you unlock their progress for the course {unlockData?.CourseName}</p>
+                                    <p>They have reached the maximum number of assessment attempt allowed without passing.</p><p>To review and unlock their progress, <a href='{unlockUrl.Uri}'>click here</a>.</p>
+                                </body>",
+            };
+
+            emailService.SendEmail(
+                new Email(emailSubjectLine, builder, unlockData.ContactEmail, unlockData.DelegateEmail)
```
AlexJacksonDS, line 70: There's several instances of null access in this chunk of code, the first of which will show as unnecessary, and these cause the access unlockData.ContactEmail to be marked as deference of a null reference.

All of these can be removed since we check unlockData == null at the top of the method.

DigitalLearningSolutions.Data/Services/NotificationService.cs
```diff
@@ -1,61 +1,74 @@
...
+            var unlockData = notificationDataService.GetUnlockData(progressId);
+            if (unlockData == null)
+            {
+                throw new NotificationDataException(
+                    $"No record found when trying to fetch Unlock Data. Progress ID: {progressId}"
+                );
+            }
+
+            unlockData.ContactForename = unlockData.ContactForename == "" ? "Colleague" : unlockData.ContactForename;
+            var refactoredTrackingSystemEnabled = featureManager.IsEnabledAsync("RefactoredTrackingSystem").Result;
```
AlexJacksonDS, line 45: We should really be awaiting the result here rather than calling `.Result`. This will cause some modifications to make all the calling methods async/await up the chain

DigitalLearningSolutions.Data/Services/NotificationService.cs
```diff
@@ -1,61 +1,74 @@
...
+            var baseUrlConfigOption = refactoredTrackingSystemEnabled ? ConfigService.AppBaseUrl : ConfigService.TrackingSystemBaseUrl;
+            var trackingSystemBaseUrl = configService.GetConfigValue(baseUrlConfigOption) ??
+                                        throw new ConfigValueMissingException(
+                                            configService.GetConfigValueMissingExceptionMessage(baseUrlConfigOption)
+                                        );
```
AlexJacksonDS, line 50: I don't understand why we are using this over Configuration.GetCurrentSystemBaseUrl and Configuration.GetAppRootPath in ConfigHelper which pulls the base url out of the appSetting.json

This will mean that we have to build the url more carefully, as the values won't have the /tracking/ part in them when we get them from the database. That this is currently incorrect with feature flag on based on the migration anyway, it'll generate https://hee-dls-test.softwire.com/coursedelegate?CustomisationID=xxxx instead of e.g. https://hee-dls-test.softwire.com/TrackingSystem/Delegates/CourseDelegates?CustomisationId=19262

DigitalLearningSolutions.Data.Migrations/202201060948_AddV2AppBaseUrl.cs
AlexJacksonDS: This migration is not useful as it will not allow the application to work on all environments, i.e. the value should not be test on any environment other than test (and maybe dev in certain cases). I've commented about it on the place where we're accessing this value as well.

I also don't think it will create the correct url in its current setup, the old one from ConfigService.TrackingSystemBaseUrl works because the url stored in the database is https://www.dls.nhs.uk/tracking/

DigitalLearningSolutions.Data.Tests/Services/UnlockServiceTests.cs
```diff
@@ -1,71 +1,122 @@
...
+            A.CallTo(() =>
+                    emailService.SendEmail(A<Email>._)
+                )
+                .MustHaveHappened();
+        }
+
+        [Test]
+        public void Trying_to_send_unlock_makes_request_to_feature_manager_to_get_correct_url()
+        {
+            //setup
```
AlexJacksonDS, line 77: // Given (and in the next 3 tests)

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/DelegateProgressController.cs
```diff
@@ -178,6 +181,24 @@ DelegateProgressAccessRoute accessedVia
...
```
AlexJacksonDS, line 194: I don't think this is the correct behaviour. The ticket says nothing about send an email when we're actually doing the unlock, the unlock request just needs to direct the admin it is sent to to the correct page, where they can click the button for this action.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/CourseDelegates/_SearchableCourseDelegateCard.cshtml
```diff
@@ -68,6 +68,14 @@
             @Model.PassRate%
           </dd>
         </div>
+        <div class="nhsuk-summary-list__row">
+          <dt class="nhsuk-summary-list__key">
+            Progress locked
+          </dt>
+          <dd class="nhsuk-summary-list__value">
+            @(Model.Locked? "Yes": "No")
+          </dd>
```
AlexJacksonDS, line 77: There is a partial _SummaryBooleanField that we can use instead of this

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/ViewDelegate/_DelegateCourseInfoCard.cshtml
```diff
@@ -113,6 +113,15 @@
           <dd class="nhsuk-summary-list__actions"></dd>
         </div>
 
+        <div class="nhsuk-summary-list__row">
+          <dt class="nhsuk-summary-list__key">
+            Progress locked
+          </dt>
+          <dd class="nhsuk-summary-list__value">
+            @(Model.IsProgressLocked ? "Yes" : "No")
+          </dd>
```
AlexJacksonDS, line 122: Same here, use the partial

AlexJacksonDS (changes requested):
I think the implementation is pretty significantly incorrect in several places for this ticket, the work done doesn't match the stated behaviour in ticket, and the emails it is attempting to send out will never work as intended with the current implementaion.

DigitalLearningSolutions.Data/Services/NotificationService.cs
rhimcg: Should I get rid of the migration and just use the CurrentSystemBaseUrl from appsettings?

AlexJacksonDS: CurrentSystemBaseUrl is the old site base url, we could keep that one as coming from the database, since it is an existing value and would mean we don't have to change the url calculation quite as much, but the new site url should not come from the migration, use the AppRootPath from config.

So yes, remove the migration (I guess manually delete the record on the test mbdbx101_test db since the migration will have run with the jenkins build).

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/DelegateProgressController.cs
rhimcg: ah I see so an admin should be able to unlock but a non admin will have to request an unlock? But this ticket is just for an admin being able to unlock? and changing the link in the email sent to an admin when a non admin requests an unlock is that right?

AlexJacksonDS: Not exactly, they're from different parts of the system. The delegate accesses Learning Portal, then requests a course to be unlocked. This sends an email to an admin with the request (NotificationService.SendUnlockRequest), which links to the new page created here (or the old site equivalent) where they can then unlock the progress.

DigitalLearningSolutions.Data/Services/NotificationService.cs
```diff
@@ -1,61 +1,94 @@
...
+                throw new ConfigValueMissingException(
+                    configService.GetConfigValueMissingExceptionMessage(
+                        refactoredTrackingSystemEnabled ? "AppRootPath" : "CurrentSystemBaseUrl"
+                    )
+                );
+            }
+
+            var baseUrl = refactoredTrackingSystemEnabled
+                ? $"{configuration.GetAppRootPath()}/TrackingSystem/Delegates/CourseDelegates"
+                : $"{configuration.GetCurrentSystemBaseUrl()}/Tracking/";
```
AlexJacksonDS, line 68: Feels a bit strange that we're getting these values twice/in 2 different ways

AlexJacksonDS, line 68: I don't think this will create the correct old site url, there is no coursedelegates in the path any more. With the feature off, the code needs to do the same as it did previously.

AlexJacksonDS (changes requested):
I think the logic around generating the urls still isn't quite right

DigitalLearningSolutions.Data/Services/NotificationService.cs
rhimcg: I've replaces it with just baseUrl here

PR #756, opened 2021-11-25
HEEDLS-672 Learning Hub API client

livzorn:
### JIRA link
[_HEEDLS-672_](https://softwiretech.atlassian.net/browse/HEEDLS-672)
[_HEEDLS-672_](https://softwiretech.atlassian.net/browse/HEEDLS-651)

### Description
Created the Learning hub Api client and associated view models. Configured user secrets to store the API key, and each developer will have to add it to their user secrets file locally. 

### Screenshots
None

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Data/Helpers/ExternalApis/LearningHubApiClient.cs
```diff
@@ -0,0 +1,153 @@
...
+
+            var response = await client.GetStringAsync(requestUrl);
+            var result = JsonConvert.DeserializeObject<ResourceSearchResult>(response);
+            return result;
+        }
+
+        public async Task<ResourceReferenceWithReferenceDetails> GetResourceByReferenceId(int resourceReferenceId)
+        {
+            const string resourceEndpoint = "/Resource/";
+            var requestUrl = learningHubApiBaseUrl + resourceEndpoint + resourceReferenceId;
```
AlexJacksonDS, line 63: I think we should just string interpolate this into `$"Resource/{resourceReferenceId}"`

DigitalLearningSolutions.Data/Helpers/ExternalApis/LearningHubApiClient.cs
```diff
@@ -0,0 +1,153 @@
...
+            const string bulkEndpoint = "/Resource/Bulk";
+            var endpointPath = learningHubApiBaseUrl + bulkEndpoint;
+            var requestUrl = GetUrlWithResourceReferenceIdQueryParams(endpointPath, resourceReferenceIds);
+
+            var response = await client.GetStringAsync(requestUrl);
+            var result = JsonConvert.DeserializeObject<BulkResourceReferences>(response);
+            return result;
+        }
+
+        private static string GetUrlWithSearchQueryParams(
```
AlexJacksonDS, line 83: I'm pretty sure there must be a better way to do all this query string construction without using a bunch of string concatenation and interpolation.

DigitalLearningSolutions.Data/Helpers/ExternalApis/LearningHubApiClient.cs
```diff
@@ -0,0 +1,153 @@
...
+
+            return queryString;
+        }
+
+        private static string GetQueryString(string key, string? value)
+        {
+            return string.IsNullOrEmpty(value) ? "" : $"{key}={value}&";
+        }
+
+        private static string GetUrlWithoutLastCharacter(string url)
```
AlexJacksonDS, line 148: Particularly without needing to use this method...

DigitalLearningSolutions.Data/DigitalLearningSolutions.Data.csproj
```diff
@@ -1,8 +1,9 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.1</TargetFramework>
     <Nullable>enable</Nullable>
+    <UserSecretsId>cc50a716-0c33-41f2-aae9-55493a9f6437</UserSecretsId>
```
AlexJacksonDS, line 6: Is the fact that this is a specific Guid going to be a problem? No idea how the secrets stuff works

AlexJacksonDS (changes requested):
I think there are some improvements that could be made to tidy up the url handling.

DigitalLearningSolutions.Data/Models/LearningHubApiClient/ResourceReferenceWithReferenceDetails.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace DigitalLearningSolutions.Data.Models.LearningHubApiClient
+{
+    public class ResourceReferenceWithReferenceDetails
```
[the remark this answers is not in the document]
livzorn: oops, thanks for catching that. added it

DigitalLearningSolutions.Data/Helpers/ExternalApis/LearningHubApiClient.cs
```diff
@@ -0,0 +1,153 @@
...
+    public class LearningHubApiClient : ILearningHubApiClient
+    {
+        private readonly HttpClient client;
+        private readonly string learningHubApiBaseUrl;
+
+        public LearningHubApiClient(HttpClient httpClient, IConfiguration config)
+        {
+            string learningHubApiKey = config.GetLearningHubApiKey();
+            learningHubApiBaseUrl =
+                "https://uks-learninghubnhsuk-openapi-dev.azurewebsites.net";
```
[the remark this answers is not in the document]
livzorn: Added the url to the appsettings files, is that the right place for it? It's not a user secret is it?

DigitalLearningSolutions.Data/Helpers/ExternalApis/LearningHubApiClient.cs
```diff
@@ -0,0 +1,153 @@
...
+
+        Task<BulkResourceReferences> GetBulkResourcesByReferenceIds(
+            IEnumerable<int>? resourceReferenceIds = null
+        );
+    }
+
+    public class LearningHubApiClient : ILearningHubApiClient
+    {
+        private readonly HttpClient client;
+        private readonly string learningHubApiBaseUrl;
```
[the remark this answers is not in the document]
livzorn: Removed and refactored the formation of the urls

DigitalLearningSolutions.Data/Helpers/ExternalApis/LearningHubApiClient.cs
livzorn: done

DigitalLearningSolutions.Data/Helpers/ExternalApis/LearningHubApiClient.cs
livzorn: I initially used QueryHelpers.AddQueryString to add the query strings, but this method only accepts string values, not StringValues values, so it was impossible to add the query strings dealing with enumerables. Steve suggested I create the query strings manually, but if you can think of a better way, please enlighten me. It does feel a bit icky

DigitalLearningSolutions.Data/DigitalLearningSolutions.Data.csproj
livzorn: I don't think so, the docs say that this Guid is unique to the project rather than a user. But not totally sure
https://docs.microsoft.com/en-us/aspnet/core/security/app-secrets?view=aspnetcore-6.0&tabs=windows#secret-manager

AlexJacksonDS: Yeah I think it'll be fine looking at the documentation. It creates a folder in appdata/.../UserSecrets with the Guid as the name, so we should be fine to all have the same folder name.

DigitalLearningSolutions.Data/Helpers/ExternalApis/LearningHubApiClient.cs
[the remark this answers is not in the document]
AlexJacksonDS: I would assume this is fine, might be best to just check with Steve before we merge @SteveJacksonSoft

DigitalLearningSolutions.Data/Helpers/ExternalApis/LearningHubApiClient.cs
```diff
@@ -0,0 +1,142 @@
...
+            return result;
+        }
+
+        public async Task<BulkResourceReferences> GetBulkResourcesByReferenceIds(
+            IEnumerable<int>? resourceReferenceIds = null
+        )
+        {
+            var queryParams = GetBulkResourcesQueryParams(resourceReferenceIds);
+
+            var response = await client.GetStringAsync($"/Resource/Bulk{queryParams}");
```
AlexJacksonDS, line 69: Not super keen on the fact we've got a different setup here to the Search endpoint. I think the ? should probably be included in this string.
Also this doesn't seem to return anything if we don't enter any parameters, is that the intended behaviour?

AlexJacksonDS (changes requested):
Might be worth having a chat about/pairing on some ideas to sort out the query strings. I think it certainly should be possible to simplify some of the steps by being a bit smarter about whether we are appending & to the strings or not and making better use of the fact we've got IEnumerables we can use to join/split etc.

DigitalLearningSolutions.Data/Helpers/ExternalApis/LearningHubApiClient.cs
livzorn: I didn't include the ? here because it is possible to make the call with an empty array of referenceIds, in which case there is no ? appended onto the url. If you don't enter parameters it should return something like this:
`{
  "resourceReferences": [],
  "unmatchedResourceReferenceIds": []
}`
Is that what you're getting?

livzorn: I'll do a bit more refactoring to make it smarter, and we can have a chat or pair on it tomorrow :)

AlexJacksonDS: Yeah, but I get the same result if I move the ? here and just don't have anything after it because the private method returns string.Empty.

AlexJacksonDS: I'm only commenting on the behaviour because that seems strange. Why have a nullable parameter on the API if it returns nothing useful without it.

livzorn: Unless...are we assuming that we'll always call this Bulk method will at least one id? Then I could use the same set up as the search endpoint. I was just going off what [swagger ](https://uks-learninghubnhsuk-openapi-dev.azurewebsites.net/swagger/index.html) allows you to enter, but we'll never call this without parameters right?

DigitalLearningSolutions.Data/Models/LearningHubApiClient/ResourceMetadata.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace DigitalLearningSolutions.Data.Models.LearningHubApiClient
+{
+    public class ResourceMetadata
+    {
+        public int ResourceId { get; set; }
+
+        public string? Title { get; set; }
```
SteveJacksonSoft, line 7: Some of these properties aren't nullable in the Open API response.
Check out the view models in the [Open API repo](https://dev.azure.com/e-LfH/LearningHub/_git/LearningHub.Nhs.OpenApi?path=/LearningHub.Nhs.OpenApi.Models/ViewModels&version=GBCI&_a=contents) (make sure you're on branch `CI` rather than `Main`).

DigitalLearningSolutions.Data/Helpers/ExternalApis/LearningHubApiClient.cs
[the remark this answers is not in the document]
SteveJacksonSoft: This should go in config.
Generally it's best to have URLs of external dependencies in config because they might change, and also in this case different envs will point to different envs of Learning Hub Open API.
Can we include LearningHub**Open**Api in the name of the config value? LH have two APIs, and anyone on the LH team will think of the other one when you say "LearningHubApi"?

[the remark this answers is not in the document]
SteveJacksonSoft: But it's not secret. Thanks for asking though.

DigitalLearningSolutions.Data/Helpers/ExternalApis/LearningHubApiClient.cs
livzorn: refactored so we don't use this anymore

DigitalLearningSolutions.Data/Helpers/ExternalApis/LearningHubApiClient.cs
```diff
@@ -0,0 +1,142 @@
...
+        {
+            var textQueryString = GetQueryString("text", text);
+            var offSetQueryString = GetQueryString("offset", offset.ToString());
+            var limitQueryString = GetQueryString("limit", limit.ToString());
+            var resourceTypesQueryString = GetMultipleQueryStrings("resourceTypes", resourceTypes);
+
+            var queryParams =
+                $"{textQueryString}{offSetQueryString}{limitQueryString}{resourceTypesQueryString}";
+
+            queryParams = GetStringWithoutLastCharacter(queryParams);
```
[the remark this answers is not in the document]
livzorn: changed to not return the &

DigitalLearningSolutions.Data/Helpers/ExternalApis/LearningHubApiClient.cs
```diff
@@ -0,0 +1,142 @@
...
+            }
+
+            var queryParams = "?";
+
+            foreach (var resourceId in resourceIds)
+            {
+                queryParams += "resourceReferenceIds=" + resourceId + "&";
+            }
+
+            queryParams = GetStringWithoutLastCharacter(queryParams);
```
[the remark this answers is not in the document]
livzorn: a method JoinQueryStrings which does this

DigitalLearningSolutions.Data/Models/LearningHubApiClient/ResourceMetadata.cs
livzorn: Should the view models match the Open API repo view models exactly? The ticket describes these fields as nullable

DigitalLearningSolutions.Data/Helpers/ExternalApis/LearningHubApiClient.cs
SteveJacksonSoft: Since it's an IEnumerable, yeah, we should discourage people from calling this method with `null`. They can always call it with an empty list, which is preferable as a "nothing" case.

DigitalLearningSolutions.Data/Models/LearningHubApiClient/ResourceMetadata.cs
SteveJacksonSoft: Ah, the ticket might have been made before the final version of the view models.

In terms of nullability, yeah, we want to be as restrictive as possible here - the data we get back will always match the Open API repo view models, so it's unhelpful to tell DLS devs "the Title might be null" when it actually won't.

AlexJacksonDS (approved):

DigitalLearningSolutions.Data/DigitalLearningSolutions.Data.csproj
SteveJacksonSoft: We don't need this - the secrets are linked to the Web project, and won't be picked up as config using this key.

DigitalLearningSolutions.Data/Helpers/ExternalApis/LearningHubApiClient.cs
```diff
@@ -0,0 +1,106 @@
+﻿namespace DigitalLearningSolutions.Data.Helpers.ExternalApis
```
SteveJacksonSoft, line 1: I wouldn't put it in Helpers - I'd normally think of Helpers as static methods that help you out.
DigitalLearningSolutions.Data/ApiClients makes more sense to me.

DigitalLearningSolutions.Data/Helpers/ExternalApis/LearningHubApiClient.cs
```diff
@@ -0,0 +1,106 @@
...
+
+        public async Task<ResourceReferenceWithResourceDetails> GetResourceByReferenceId(int resourceReferenceId)
+        {
+            var response = await client.GetStringAsync($"/Resource/{resourceReferenceId}");
+            var result = JsonConvert.DeserializeObject<ResourceReferenceWithResourceDetails>(response);
+            return result;
+        }
+
+        public async Task<BulkResourceReference> GetBulkResourcesByReferenceIds(
+            IEnumerable<int> resourceReferenceIds
```
DanBloxham-sw, line 65: Given that an empty collection of ints could be passed in, and we know that we will get empty lists backs, we can probably bypass the unnecessary call and just return the model with the empty lists (or maybe null if you think that might be better)

DanBloxham-sw (changes requested):
Looking good overall, just a couple of minor changes.

DigitalLearningSolutions.Data/Helpers/ExternalApis/LearningHubApiClient.cs
AlexJacksonDS: We should keep the return type correct (so no null), but I'm not sure whether we should short circuit the API call or not. I kind of feel like the API client should always call the API, and perhaps any calling method should be deciding whether it should call the API or not.

DanBloxham-sw: Yeah, that sounds reasonable, it can stay as it is.

DanBloxham-sw (approved):

DigitalLearningSolutions.Data/Models/LearningHubApiClient/BulkResourceReference.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace DigitalLearningSolutions.Data.Models.LearningHubApiClient
+{
+    using System.Collections.Generic;
+
+    public class BulkResourceReference
```
SteveJacksonSoft, line 5: I'd be inclined to keep this plural - you have lists of resource references

AlexJacksonDS: This matches the naming of the API BulkResourceReferenceViewModel.

PR #759, opened 2021-11-25
Heedls 409 tracker endpoint getobjectivearraycc

JBsoftwire:
### JIRA link
HEEDLS-409

### Description
Added second tracker endpoint action with relevant tests.

### Screenshots
N/A

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/DataServices/TutorialContentDataServiceTests/ObjectiveTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/DataServices/TutorialContentDataServiceTests/ObjectiveTests.cs
@@ -65,6 +65,8 @@
                     result.Last().Possible.Should().Be(3);
                     result.Last().WrongCount.Should().Be(0);
                     result.Last().MyScore.Should().Be(0);
+
+                    result.Select(x => x.TutorialId).Should().NotContain(1138);
                 }
             }
         }
@@ -75,11 +77,12 @@
         {
             using (new TransactionScope())
             {
-                connection.Execute("UPDATE Tutorials SET OriginalTutorialID = 1 WHERE TutorialID = 179");
+                connection.Execute("UPDATE Tutorials SET OriginalTutorialID = 1 WHERE TutorialID = 1137");
+                connection.Execute("UPDATE CustomisationTutorials SET DiagStatus = 0 WHERE CusTutID = 704039");
 
                 // When
                 var result = tutorialContentDataService
-                    .GetNonArchivedCcObjectivesBySectionAndCustomisationId(83, 2793, true)
+                    .GetNonArchivedCcObjectivesBySectionAndCustomisationId(248, 22062, true)
                     .ToList();
 
                 // Then
@@ -87,16 +90,14 @@
                 {
                     result.Count.Should().Be(4);
                     result.First().TutorialId.Should().Be(1);
-                    result.First().TutorialName.Should().Be("Open and arrange windows");
-                    result.First().Possible.Should().Be(6);
-                    result.First().WrongCount.Should().Be(0);
-                    result.First().MyScore.Should().Be(0);
 
-                    result.Last().TutorialId.Should().Be(182);
-                    result.Last().TutorialName.Should().Be("Change the zoom");
-                    result.Last().Possible.Should().Be(2);
-                    result.Last().WrongCount.Should().Be(0);
-                    result.Last().MyScore.Should().Be(0);
+                    result.Last().TutorialId.Should().Be(1257);
+
+                    var postLearnResult = result.Single(x => x.TutorialId == 1138);
+                    postLearnResult.TutorialName.Should().Be("Managing files");
+                    postLearnResult.Possible.Should().Be(1);
+                    postLearnResult.WrongCount.Should().Be(0);
+                    postLearnResult.MyScore.Should().Be(0);
                 }
             }
         }
--- a/DigitalLearningSolutions.Data.Tests/DataServices/TutorialContentDataServiceTests/ObjectiveTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/DataServices/TutorialContentDataServiceTests/ObjectiveTests.cs
@@ -36,5 +36,66 @@
                 }
             }
         }
+
+        [Test]
+        public void GetNonArchivedCcObjectivesBySectionAndCustomisationId_returns_objectives_correctly()
+        {
+            using (new TransactionScope())
+            {
+                connection.Execute("UPDATE Tutorials SET OriginalTutorialID = 1 WHERE TutorialID = 1137");
+                connection.Execute("UPDATE CustomisationTutorials SET DiagStatus = 0 WHERE CusTutID = 704039");
+
+                // When
+                var result = tutorialContentDataService.GetNonArchivedCcObjectivesBySectionAndCustomisationId(248, 22062, false)
+                    .ToList();
+
+                // Then
+                using (new AssertionScope())
+                {
+                    result.Count.Should().Be(3);
+                    result.First().TutorialId.Should().Be(1);
+                    result.First().TutorialName.Should().Be("Manufacturers, operating systems and applications");
+                    result.First().Possible.Should().Be(4);
+                    result.First().WrongCount.Should().Be(0);
+                    result.First().MyScore.Should().Be(0);
+
+                    result.Last().TutorialId.Should().Be(1257);
+                    result.Last().TutorialName.Should().Be("Sharing information");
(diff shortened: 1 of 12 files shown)


DigitalLearningSolutions.Data.Tests/Services/TrackerActionServiceTests.cs
```diff
@@ -82,5 +82,81 @@ public void GetObjectiveArray_returns_empty_object_json_if_no_results_found()
...
+        {
+            // given
+            A.CallTo(
+                    () => dataService.GetNonArchivedCcObjectivesBySectionAndCustomisationId(
+                        A<int>._,
+                        A<int>._,
+                        A<bool>._
+                    )
+                )
+                .Returns(new[] { new CcObjective(1, "name", 9) });
```
DanBloxham-sw, line 153: I feel like this set up is unnecessary as we shouldn't be calling this if any of the params are null. Would probably be better to make the assertion that this is never called in the Then section.

DigitalLearningSolutions.Data.Tests/DataServices/TutorialContentDataServiceTests/ObjectiveTests.cs
```diff
@@ -36,5 +36,69 @@ public void GetNonArchivedObjectivesBySectionAndCustomisationId_returns_objectiv
...
+            GetNonArchivedCcObjectivesBySectionAndCustomisationId_returns_objectives_correctly_with_postlearning()
+        {
+            using (new TransactionScope())
+            {
+                connection.Execute("UPDATE Tutorials SET OriginalTutorialID = 1 WHERE TutorialID = 179");
+
+                // When
+                var result = tutorialContentDataService
+                    .GetNonArchivedCcObjectivesBySectionAndCustomisationId(83, 2793, true)
+                    .ToList();
```
DanBloxham-sw, line 83: Do you think it might be worth changing these two GetNonArchivedCcObjectivesBySectionAndCustomisationId to be called with the same customisationId and sectionId? Might be a bit of a pain to set up, but that way it is easier to see that the changing of the bool does causes differences in the return.

DigitalLearningSolutions.Data.Tests/DataServices/TutorialContentDataServiceTests/ObjectiveTests.cs
```diff
@@ -36,5 +36,69 @@ public void GetNonArchivedObjectivesBySectionAndCustomisationId_returns_objectiv
...
+
+                // When
+                var result = tutorialContentDataService
+                    .GetNonArchivedCcObjectivesBySectionAndCustomisationId(248, 22062, false)
+                    .ToList();
+
+                // Then
+                using (new AssertionScope())
+                {
+                    result.Count.Should().Be(3);
```
DanBloxham-sw, line 56: Because the PostLearning param is false, ideally the test would exclude some data that does not match the specific requirements. Do you think it is worth adding another assertion to check that the results do not include specific tutorial Ids that should be ignored?

DanBloxham-sw (changes requested):
Some minor changes.

DigitalLearningSolutions.Data.Tests/Services/TrackerActionServiceTests.cs
```diff
@@ -82,5 +82,81 @@ public void GetObjectiveArray_returns_empty_object_json_if_no_results_found()
...
+
+            // then
+            result.Should().Be(null);
+        }
+
+        [Test]
+        [TestCase(null, null, null)]
+        [TestCase(null, 1, true)]
+        [TestCase(1, null, true)]
+        [TestCase(1, 1, null)]
```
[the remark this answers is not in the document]
JBsoftwire: Probably we don't, I just retained it since I did every combination for the other objective method, but that was only three total. Doing them individually should be enough.

DanBloxham-sw (approved):

DigitalLearningSolutions.Data.Tests/Services/TrackerServiceTests.cs
```diff
@@ -65,7 +65,7 @@ public void ProcessQuery_with_GetObjectiveArray_action_passes_query_params()
         }
 
         [Test]
-        public void ProcessQuery_with_GetObjectiveArray_action_correctly_serialises_contentful_response()
+        public void ProcessQuery_with_valid_action_correctly_serialises_contentful_response()
```
AlexJacksonDS, line 68: I'm not sure about this renaming, why are we taking away the reference to the action that is being used here?

DigitalLearningSolutions.Data.Tests/Services/TrackerServiceTests.cs
```diff
@@ -90,7 +90,7 @@ public void ProcessQuery_with_GetObjectiveArray_action_correctly_serialises_cont
         }
 
         [Test]
-        public void ProcessQuery_with_GetObjectiveArray_action_correctly_serialises_null_response()
+        public void ProcessQuery_with_valid_action_correctly_serialises_null_response()
```
AlexJacksonDS, line 93: Same with this renaming

DigitalLearningSolutions.Data/DataServices/TutorialContentDataService.cs
```diff
@@ -31,6 +31,7 @@ bool learningEnabled
         );
 
         IEnumerable<Objective> GetNonArchivedObjectivesBySectionAndCustomisationId(int sectionId, int customisationId);
+        IEnumerable<CcObjective> GetNonArchivedCcObjectivesBySectionAndCustomisationId(int sectionId, int customisationId, bool isPostLearning);
```
AlexJacksonDS, line 34: I'd expect the formatter to put this over multiple lines

DigitalLearningSolutions.Data/Models/Tracker/CcObjective.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace DigitalLearningSolutions.Data.Models.Tracker
+{
+    using System.Collections.Generic;
+
+    public class CcObjective : ObjectiveBase
+    {
+        public CcObjective(int tutorialId, string tutorialName, int possible) : base(
+            tutorialId,
+            possible
+        )
```
AlexJacksonDS, line 10: This formatting also looks a bit odd

DigitalLearningSolutions.Data/Models/Tracker/Objective.cs
```diff
@@ -2,19 +2,16 @@
...
     using System.Collections.Generic;
 
-    public class Objective
+    public class Objective : ObjectiveBase
     {
-        public Objective(int tutorialId, IEnumerable<int> interactions, int possible)
+        public Objective(int tutorialId, IEnumerable<int> interactions, int possible) : base(
+            tutorialId,
+            possible
+        )
```
AlexJacksonDS, line 10: Same for the formatting here

DigitalLearningSolutions.Data.Tests/DataServices/TutorialContentDataServiceTests/ObjectiveTests.cs
```diff
@@ -36,5 +36,70 @@ public void GetNonArchivedObjectivesBySectionAndCustomisationId_returns_objectiv
...
             }
         }
+
+        [Test]
+        public void GetNonArchivedCcObjectivesBySectionAndCustomisationId_returns_objectives_correctly()
+        {
+            using (new TransactionScope())
+            {
+                connection.Execute("UPDATE Tutorials SET OriginalTutorialID = 1 WHERE TutorialID = 1137");
+                connection.Execute("UPDATE CustomisationTutorials SET DiagStatus = 0 WHERE CusTutID = 704039");
```
AlexJacksonDS, line 46: These should probable have a // Given before them both here and in the next test below

AlexJacksonDS (changes requested):
I think overall this looks fine, just some cases where the formatter may need to be run and a question about test naming.

DigitalLearningSolutions.Data.Tests/Services/TrackerServiceTests.cs
JBsoftwire: This is a test for the serialisation of the response, rather than the success of the specific action, which is tested in `ProcessQuery_with_GetObjectiveArray_action_passes_query_params` and its Cc sibling.

AlexJacksonDS: OK fine, as long as it's covered.

AlexJacksonDS (approved):

PR #773, opened 2021-11-29
Heedls 680 sso security helper

JBsoftwire:
### JIRA link
[_HEEDLS-XXXX_](https://softwiretech.atlassian.net/browse/HEEDLS-680)

### Description
Added a helper class to handle generating and verifying hashes for Learning Hub SSO. The implemented version varies from the example by not being static, allowing the use of an injected ClockService and so more robust testing.

### Screenshots
N/A

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Data/Helpers/LearningHubSsoSecurityHelper.cs
```diff
@@ -0,0 +1,62 @@
+﻿namespace DigitalLearningSolutions.Data.Helpers
+{
+    using System;
+    using System.Security.Cryptography;
+    using System.Text;
+    using DigitalLearningSolutions.Data.Services;
+
+    public class LearningHubSsoSecurityHelper
```
wilkinsm, line 8: To fully utilise dependency injection this will need an interface so this helper can be injected/mocked also.

DigitalLearningSolutions.Data/Helpers/LearningHubSsoSecurityHelper.cs
```diff
@@ -0,0 +1,62 @@
+﻿namespace DigitalLearningSolutions.Data.Helpers
+{
+    using System;
+    using System.Security.Cryptography;
+    using System.Text;
+    using DigitalLearningSolutions.Data.Services;
+
+    public class LearningHubSsoSecurityHelper
+    {
+        private static readonly DateTime UnixEpoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
```
wilkinsm, line 10: It might be worth noting on the ticket or clarifying with LH that times should be based off this epoc.
The documentation is not explicit on this but we can imply it based on their examples. 
I think it's worth ticking this box and getting clarification.

DigitalLearningSolutions.Data/Helpers/LearningHubSsoSecurityHelper.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            this.clockService = clockService;
+        }
+
+        public string GenerateHash(string state, string secretKey)
+        {
+            var secondsSinceEpoch = GetSecondsSinceEpoch();
+            var encoder = new UTF8Encoding();
+            var salt = encoder.GetBytes(secretKey);
+            var pwd = encoder.GetBytes(state + secondsSinceEpoch);
```
wilkinsm, line 24: Could the `pwd` var be renamed. Maybe `hashInput`? Feel free to come up with something better than that.

DigitalLearningSolutions.Data/Helpers/LearningHubSsoSecurityHelper.cs
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        private static string GetHash(byte[] password, byte[] salt)
+        {
+            using var byteResult = new Rfc2898DeriveBytes(password, salt, 10000, HashAlgorithmName.SHA512);
+            var hash = Convert.ToBase64String(byteResult.GetBytes(32));
+            return hash;
+        }
+
+        private long GetSecondsSinceEpoch()
```
wilkinsm, line 57: I'm wondering whether this has potential to be used elsewhere in the future outside of this helper. That particular future is unknown so just food for thought right now and possibly nothing to do here in regards to moving/refactoring.

DigitalLearningSolutions.Data/Helpers/LearningHubSsoSecurityHelper.cs
```diff
@@ -0,0 +1,62 @@
...
+                if (hash == GetHash(pwd, salt))
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
+        private static string GetHash(byte[] password, byte[] salt)
```
wilkinsm, line 50: Could the `password` param be renamed to follow suit with which ever name is chosen in the other method?

DigitalLearningSolutions.Data/Helpers/LearningHubSsoSecurityHelper.cs
```diff
@@ -0,0 +1,62 @@
...
+        private static readonly DateTime UnixEpoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
+
+        private readonly IClockService clockService;
+
+        public LearningHubSsoSecurityHelper(IClockService clockService)
+        {
+            this.clockService = clockService;
+        }
+
+        public string GenerateHash(string state, string secretKey)
```
wilkinsm, line 19: The secret value will eventually come from config. It probably makes sense to keep this as a parameter and to be more "helper-like" assuming the service using this helper will itself utilise config to obtain the value.

DigitalLearningSolutions.Data/Helpers/LearningHubSsoSecurityHelper.cs
```diff
@@ -0,0 +1,62 @@
...
+            var pwd = encoder.GetBytes(state + secondsSinceEpoch);
+            return GetHash(pwd, salt);
+        }
+
+        public bool VerifyHash(string state, string secretKey, string hash)
+        {
+            var secondsSinceEpoch = GetSecondsSinceEpoch();
+            var encoder = new UTF8Encoding();
+            var salt = encoder.GetBytes(secretKey);
+            var toleranceInSec = 60;
```
wilkinsm, line 33: The documentation is explicit on this tolerance so this is probably ok. But I'm wondering whether it is worth this living in config - but then actually within this context that would mean a parameter? Just in ase they change the tolerance level.

DigitalLearningSolutions.Data/Helpers/LearningHubSsoSecurityHelper.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            // loop optimisation, iterates 0,-1,-2,-3 .. -60,1,2,3 .. 60
+            for (var counter = 0; counter <= toleranceInSec * 2; counter++)
+            {
+                var step = counter > toleranceInSec ? counter - toleranceInSec : -1 * counter;
+                var pwd = encoder.GetBytes(state + (secondsSinceEpoch + step));
```
wilkinsm, line 39: Can the `pwd` var name be changed to match the other name change?

DigitalLearningSolutions.Data/Helpers/LearningHubSsoSecurityHelper.cs
```diff
@@ -0,0 +1,62 @@
...
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
+        private static string GetHash(byte[] password, byte[] salt)
+        {
+            using var byteResult = new Rfc2898DeriveBytes(password, salt, 10000, HashAlgorithmName.SHA512);
```
wilkinsm, line 52: Similar to tolerance, is the hash rounds `10000` ? It's implied based on the example but not explicitly documented. Worth clarifying.

DigitalLearningSolutions.Data/Helpers/LearningHubSsoSecurityHelper.cs
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        public bool VerifyHash(string state, string secretKey, string hash)
+        {
+            var secondsSinceEpoch = GetSecondsSinceEpoch();
+            var encoder = new UTF8Encoding();
+            var salt = encoder.GetBytes(secretKey);
+            var toleranceInSec = 60;
+
+            // loop optimisation, iterates 0,-1,-2,-3 .. -60,1,2,3 .. 60
```
wilkinsm, line 35: I don't think this comment is needed. 
If more clarity was ever needed on what is going on here, maybe some refactor is required or pulling this sub-section into another method.

DigitalLearningSolutions.Data/Helpers/LearningHubSsoSecurityHelper.cs
```diff
@@ -0,0 +1,62 @@
...
+                }
+            }
+
+            return false;
+        }
+
+        private static string GetHash(byte[] password, byte[] salt)
+        {
+            using var byteResult = new Rfc2898DeriveBytes(password, salt, 10000, HashAlgorithmName.SHA512);
+            var hash = Convert.ToBase64String(byteResult.GetBytes(32));
```
wilkinsm, line 53: The returned byte length here - this is another point of clarification. It's in the example so we can imply this is the expected length to use but might be worth getting that clarified.

DigitalLearningSolutions.Data.Tests/Helpers/LearningHubSsoSecurityHelperTests.cs
```diff
@@ -0,0 +1,160 @@
...
+            var stateString = "stateString";
+            var clockService = new BinaryClockService(now, now);
+            var helper = new LearningHubSsoSecurityHelper(clockService);
+
+            // when
+            var hash1 = helper.GenerateHash(stateString, Secret);
+            var hash2 = helper.GenerateHash(stateString, Secret);
+
+            // then
+            hash1.Should().BeEquivalentTo(hash2);
```
wilkinsm, line 27: I think case matters here and so `Should().Be<string>)` would be more appropriate.
`BeEquivalentTo` is case insensitive.

DigitalLearningSolutions.Data.Tests/Helpers/LearningHubSsoSecurityHelperTests.cs
```diff
@@ -0,0 +1,160 @@
...
+            var stateString = "stateString";
+            var clockService = new BinaryClockService(now, now.AddSeconds(1));
+            var helper = new LearningHubSsoSecurityHelper(clockService);
+
+            // when
+            var hash1 = helper.GenerateHash(stateString, Secret);
+            var hash2 = helper.GenerateHash(stateString, Secret);
+
+            // then
+            hash1.Should().NotBeEquivalentTo(hash2);
```
wilkinsm, line 44: I think case matters here and so `Should().NotBe<string>)` would be more appropriate.
`NotBeEquivalentTo` is case insensitive.

DigitalLearningSolutions.Data.Tests/Helpers/LearningHubSsoSecurityHelperTests.cs
```diff
@@ -0,0 +1,160 @@
...
+            var stateString = "stateString";
+            var clockService = new BinaryClockService(now, now.AddMilliseconds(999));
+            var helper = new LearningHubSsoSecurityHelper(clockService);
+
+            // when
+            var hash1 = helper.GenerateHash(stateString, Secret);
+            var hash2 = helper.GenerateHash(stateString, Secret);
+
+            // then
+            hash1.Should().BeEquivalentTo(hash2);
```
wilkinsm, line 61: I think case matters here and so `Should().Be<string>)` would be more appropriate.
`BeEquivalentTo` is case insensitive.

DigitalLearningSolutions.Data.Tests/Helpers/LearningHubSsoSecurityHelperTests.cs
```diff
@@ -0,0 +1,160 @@
...
+            var differentStateString = "stateStrinh";
+            var clockService = new BinaryClockService(now, now.AddSeconds(1));
+            var helper = new LearningHubSsoSecurityHelper(clockService);
+
+            // when
+            var hash1 = helper.GenerateHash(stateString, Secret);
+            var hash2 = helper.GenerateHash(differentStateString, Secret);
+
+            // then
+            hash1.Should().NotBeEquivalentTo(hash2);
```
wilkinsm, line 79: I think case matters here and so `Should().NotBe<string>)` would be more appropriate.
`NotBeEquivalentTo` is case insensitive.

DigitalLearningSolutions.Data.Tests/Helpers/LearningHubSsoSecurityHelperTests.cs
```diff
@@ -0,0 +1,160 @@
...
+            var stateString = "stateString";
+            var clockService = new BinaryClockService(now, now);
+            var helper = new LearningHubSsoSecurityHelper(clockService);
+
+            // when
+            var hash1 = helper.GenerateHash(stateString, Secret);
+            var hash2 = helper.GenerateHash(stateString, "open sesame");
+
+            // then
+            hash1.Should().NotBeEquivalentTo(hash2);
```
wilkinsm, line 96: I think case matters here and so `Should().NotBe<string>)` would be more appropriate.
`NotBeEquivalentTo` is case insensitive.

DigitalLearningSolutions.Data.Tests/Helpers/LearningHubSsoSecurityHelperTests.cs
```diff
@@ -0,0 +1,160 @@
...
+            // when
+            var hash1 = helper.GenerateHash(stateString, Secret);
+            var hash2 = helper.GenerateHash(stateString, Secret);
+
+            // then
+            hash1.Should().BeEquivalentTo(hash2);
+        }
+
+        [Test]
+        public void GenerateHash_varies_hash_with_clock_time()
```
wilkinsm, line 31: Not sure this name fully represents the test.
What about `GenerateHash_returns_different_hashes_for_different_timestamps`

DigitalLearningSolutions.Data.Tests/Helpers/LearningHubSsoSecurityHelperTests.cs
```diff
@@ -0,0 +1,160 @@
...
+    using NUnit.Framework;
+
+    public class LearningHubSsoSecurityHelperTests
+    {
+        private const string Secret = "where the wild rose blooms";
+
+        [Test]
+        public void GenerateHash_is_consistent()
+        {
+            // given
```
wilkinsm, line 16: Nit-picking Capitalisation on region **G**iven

DigitalLearningSolutions.Data.Tests/Helpers/LearningHubSsoSecurityHelperTests.cs
```diff
@@ -0,0 +1,160 @@
...
+            // when
+            var hash1 = helper.GenerateHash(stateString, Secret);
+            var hash2 = helper.GenerateHash(stateString, Secret);
+
+            // then
+            hash1.Should().NotBeEquivalentTo(hash2);
+        }
+
+        [Test]
+        public void GenerateHash_varies_hash_with_clock_time_by_second()
```
wilkinsm, line 48: Not sure this name fully represents the test.
What about `GenerateHash_returns_hashes_for_timestamps_accurate_to_the_second`

DigitalLearningSolutions.Data.Tests/Helpers/LearningHubSsoSecurityHelperTests.cs
```diff
@@ -0,0 +1,160 @@
...
+            // when
+            var hash1 = helper.GenerateHash(stateString, Secret);
+            var hash2 = helper.GenerateHash(stateString, "open sesame");
+
+            // then
+            hash1.Should().NotBeEquivalentTo(hash2);
+        }
+
+        [Test]
+        public void VerifyHash_verifies_hash_created_within_60_seconds([Range(-60, 60, 10)] int delay)
```
wilkinsm, line 100: Could we tweak the name slightly: `VerifyHash_returns_true_for_hashed_created_within_60_seconds` ?

DigitalLearningSolutions.Data.Tests/Helpers/LearningHubSsoSecurityHelperTests.cs
```diff
@@ -0,0 +1,160 @@
...
+            // when
+            var hash1 = helper.GenerateHash(stateString, Secret);
+            var result = helper.VerifyHash(stateString, Secret, hash1);
+
+            // then
+            result.Should().BeTrue();
+        }
+
+        [Test]
+        public void VerifyHash_does_not_verify_hash_created_outside_60_seconds([Values(-61, 61)] int delay)
```
wilkinsm, line 117: Could we tweak the name slightly: `VerifyHash_returns_false_for_hashes_created_outside_60_second_tolerance` ?

wilkinsm (changes requested):
Nice one Jonathan. Just a few minor things really. Most of my comments are about clarifications on values from LH.

DigitalLearningSolutions.Data/Helpers/LearningHubSsoSecurityHelper.cs
```diff
@@ -0,0 +1,68 @@
...
+            this.clockService = clockService;
+        }
+
+        public string GenerateHash(string state, string secretKey)
+        {
+            var secondsSinceEpoch = GetSecondsSinceEpoch();
+            var encoder = new UTF8Encoding();
+            var salt = encoder.GetBytes(secretKey);
+            var timedState = encoder.GetBytes(state + secondsSinceEpoch);
+            return GetHash(timedState, salt);
```
wilkinsm, line 32: It does not say in the documentation but I imagine the output needs to be url-encoded.
Is that something for this helper class to do or the service that uses this helper to do?

JBsoftwire: I think that should be in the service calling the helper, on grounds that the helper should only perform its specific task of creating or verifying hashes. If those hashes need to be sent somewhere else or further processed in some way then that's the job of the rest of the system.

DigitalLearningSolutions.Data/Helpers/LearningHubSsoSecurityHelper.cs
SteveJacksonSoft: I'm down with putting it in config aye.

DigitalLearningSolutions.Data/Helpers/LearningHubSsoSecurityHelper.cs
SteveJacksonSoft: I'll confirm with Dave in my email. Putting it in config makes sense to me.

DigitalLearningSolutions.Data/Helpers/LearningHubSsoSecurityHelper.cs
SteveJacksonSoft: I'll clarify this with Dave.

DigitalLearningSolutions.Data/Helpers/LearningHubSsoSecurityHelper.cs
wilkinsm: Sounds reasonable to me. Url encoding will probably be another helper if we even need one.

DigitalLearningSolutions.Web/appsettings.json
```diff
@@ -12,5 +12,10 @@
     "RefactoredTrackingSystem": false,
     "RefactoredSuperAdminInterface": false
   },
-  "LearningHubOpenAPIBaseUrl": "https://uks-learninghubnhsuk-openapi-dev.azurewebsites.net"
+  "LearningHubOpenAPIBaseUrl": "https://uks-learninghubnhsuk-openapi-dev.azurewebsites.net",
+  "LearningHubSSO": {
+    "ToleranceInSeconds": 60,
+    "HashRounds": 10000,
```
wilkinsm, line 18: This is named `HashIterations` in the helper.

DigitalLearningSolutions.Data/Helpers/LearningHubSsoSecurityHelper.cs
wilkinsm: Now that config is being injected directly into this helper service, the secret key might as well be taken from there, rather than a parameter, to be consistent.

DigitalLearningSolutions.Web/Startup.cs
```diff
@@ -314,10 +317,10 @@ public void Configure(IApplicationBuilder app, IMigrationRunner migrationRunner,
 
         private Task RedirectToLogin(RedirectContext<CookieAuthenticationOptions> context)
         {
-            var applicationPath = new Uri(config.GetAppRootPath()).AbsolutePath.TrimEnd('/');
+            var applicationPath = new Uri(ConfigHelper.GetAppRootPath(config)).AbsolutePath.TrimEnd('/');
```
wilkinsm, line 320: I can see why this has had to change but this is also highlighting an issue in that we now have conflicts.
There is a ticket (HEEDLS-609) which is looking to address these issues I think.
Could you add a TODO here and reference the above ticket number please?

DigitalLearningSolutions.Data/Helpers/LearningHubSsoSecurityHelper.cs
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        public bool VerifyHash(string state, string secretKey, string hash)
+        {
+            var secondsSinceEpoch = GetSecondsSinceEpoch();
+            var encoder = new UTF8Encoding();
+            var salt = encoder.GetBytes(secretKey);
+            var toleranceInSec = config.GetSection("LearningHubSSO").GetValue<int>("ToleranceInSeconds");
```
wilkinsm, line 43: Although we are probably only using these config values here we should stick to the convention that is in place.
If we had the fullness of time I wuld actually opt for the `IOptions` approach for a strongly typed and easily `mocked` implementation over the standard `IConfiguration` approach but I think that is going to have to wait.
In the mean time can we look to implement some extension methods in the ConfigHelper and use those to match other areas like the `ResourcesController` ?

DigitalLearningSolutions.Data/Helpers/ConfigHelper.cs
```diff
@@ -34,5 +39,20 @@ public static bool IsSignpostingUsed(this IConfiguration config)
         {
             return bool.Parse(config[UseSignposting]);
         }
+
+        public static int GetLearningHubSsoHashTolerance(this IConfiguration config)
+        {
+            return int.Parse(config[string.Join(':', LearningHubSsoSectionKey, LearningHubSsoToleranceKey)]);
```
wilkinsm, line 45: For these string concatenations the common approach now is to use string interpolation, e.g.
`return int.Parse(config[$"{LearningHubSsoSectionKey}:{LearningHubSsoToleranceKey}"]);`
Could these be changes to use the string interpolation approach?

DigitalLearningSolutions.Data/Helpers/ConfigHelper.cs
```diff
@@ -14,6 +14,7 @@ public static class ConfigHelper
         private const string LearningHubSsoToleranceKey = "ToleranceInSeconds";
         private const string LearningHubSsoIterationsKey = "HashIterations";
         private const string LearningHubSsoByteLengthKey = "ByteLength";
+        private const string LearningHubSsoSecretKey = "SecretKey";
```
wilkinsm, line 17: Can this also be placed into the `appSettings.json` file with a "fake" value set?

wilkinsm (approved):

DigitalLearningSolutions.Web/appsettings.json
```diff
@@ -13,5 +13,11 @@
     "RefactoredSuperAdminInterface": false,
     "UseSignposting": true
   },
-  "LearningHubOpenAPIBaseUrl": "https://uks-learninghubnhsuk-openapi-dev.azurewebsites.net"
+  "LearningHubOpenAPIBaseUrl": "https://uks-learninghubnhsuk-openapi-dev.azurewebsites.net",
+  "LearningHubSSO": {
+    "ToleranceInSeconds": 60,
+    "HashIterations": 10000,
+    "ByteLength": 32,
+    "SecretKey": "test value"
```
SteveJacksonSoft, line 21: MapsAPIKey's value is blank in appsettings; I think we should do the same thing here to stay consistent.

DigitalLearningSolutions.Web/Startup.cs
```diff
@@ -315,10 +318,11 @@ public void Configure(IApplicationBuilder app, IMigrationRunner migrationRunner,
 
         private Task RedirectToLogin(RedirectContext<CookieAuthenticationOptions> context)
         {
-            var applicationPath = new Uri(config.GetAppRootPath()).AbsolutePath.TrimEnd('/');
+            // TODO HEEDLS-609 update this
+            var applicationPath = new Uri(ConfigHelper.GetAppRootPath(config)).AbsolutePath.TrimEnd('/');
             var url = HttpUtility.UrlEncode(applicationPath + context.Request.Path);
             var queryString = HttpUtility.UrlEncode(context.Request.QueryString.Value);
-            context.HttpContext.Response.Redirect(config.GetAppRootPath() + $"/Login?returnUrl={url}{queryString}");
+            context.HttpContext.Response.Redirect(ConfigHelper.GetAppRootPath(config) + $"/Login?returnUrl={url}{queryString}");
```
SteveJacksonSoft, line 325: I'd say just port this helper method over to the Data project in this commit. I'm guessing VS can move static methods for you with its refactoring tool? Rider can.

DigitalLearningSolutions.Data/Helpers/LearningHubSsoSecurityHelper.cs
```diff
@@ -0,0 +1,78 @@
+﻿namespace DigitalLearningSolutions.Data.Helpers
+{
+    using System;
+    using System.Security.Cryptography;
+    using System.Text;
+    using DigitalLearningSolutions.Data.Services;
+    using Microsoft.Extensions.Configuration;
+
+    public interface ILearningHubSsoSecurityHelper
```
SteveJacksonSoft, line 9: Ah, okay, so this escaped me when I made the ticket.

This doesn't fit the Helper pattern we've been using - they're all static classes.
This is really a service, in that you're registering it with DI and are using instances of it.

Can you move it to the services folder and tweak the name accordingly?

DigitalLearningSolutions.Data.Tests/Helpers/LearningHubSsoSecurityHelperTests.cs
```diff
@@ -0,0 +1,185 @@
...
+        public void VerifyHash_returns_true_for_hashed_created_within_tolerance_time([Range(-TestTolerance, TestTolerance, 1)] int delay)
+        {
+            // Given
+            var now = DateTime.UtcNow;
+            var stateString = "stateString";
+            var clockService = new BinaryClockService(now, now.AddSeconds(delay));
+            var helper = new LearningHubSsoSecurityHelper(clockService, Config);
+
+            // When
+            var hash1 = helper.GenerateHash(stateString);
```
SteveJacksonSoft, line 134: Nitpick that in this test and the one below there's only one hash, so you can just call it `hash` instead of `hash1`.

SteveJacksonSoft (changes requested):
Nice tests!

DigitalLearningSolutions.Web/appsettings.json
wilkinsm: Yes good point. We can move all of these into `secrets` for local dev work. Then the actual values will need plumbing in in the pipeline. Probably good to add/update keeper with the values required to make it work in test/UAT.
Maybe the zero-to-hero wiki should be updated to reflect the need to update local `secrets` from these keeper values too?

SteveJacksonSoft: Thanks, updated the z2h. Not sure how secrets are set on the environments, I'll ask Kevin about it.

DigitalLearningSolutions.Web/Startup.cs
JBsoftwire: The method is already in the data project it turns out, I just need to point the code at the right one (this whole original refactor was done automatically and I didn't notice it at first, an example of the IDE being too smart).

SteveJacksonSoft: We'll leave this as it was.

DigitalLearningSolutions.Data/Helpers/ConfigHelper.cs
```diff
@@ -4,6 +4,7 @@
 
     public static class ConfigHelper
     {
+        public const string DefaultConnectionStringName = "DefaultConnection";
```
SteveJacksonSoft, line 7: What's this for?

SteveJacksonSoft (approved):

PR #784, opened 2021-11-30
DLSV2-422 Only show RAG colour of self-assessment result when the result has been verified if supervisor verification of results is switched on for the self assessment

daniel-manta-hee:
### JIRA link
https://hee-dls.atlassian.net/browse/DLSV2-422

### Description
Passing SupervisorSelfAssessmentReview to shared partial view _AssessmentQuestionResponse.cshtml from controllers returning views ReviewSelfAssessment, ReviewCompetencySelfAssessment, SelfAssessmentOverview and VerificationPickSupervisor. Then display SupervisorSelfAssessmentReview Rag coloured LevelRAG (levelDescriptor.LevelLabel) on that view if SupervisorSelfAssessmentReview is true, SelfAssessmentResultSupervisorVerifications is Verified and SignedOff. Otherwise display coloured Result/MaxValue.

-----
### Developer checks
Review verified and signed off self assessment for non-slider questions (AssessmentQuestionInputTypeID <> 2). 

- Check that LevelRAG label is visible when SupervisorSelfAssessmentReview is true on table SelfAssessments.
- Check that if any of the above conditions are not met, coloured Result/MaxValue is shown.

### DIFF
--- a/DigitalLearningSolutions.Data/Models/SelfAssessments/CurrentSelfAssessment.cs
+++ b/DigitalLearningSolutions.Data/Models/SelfAssessments/CurrentSelfAssessment.cs
@@ -9,6 +9,7 @@
         public DateTime? SubmittedDate { get; set; }
         public bool IsSupervised { get; set; }
         public bool IsSupervisorResultsReviewed { get; set; }
+        public bool SupervisorSelfAssessmentReview { get; set; }
         public string? Vocabulary { get; set; }
         public string? VerificationRoleName { get; set; }
         public string? SignOffRoleName { get; set; }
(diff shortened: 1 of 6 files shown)


kevwhitt-hee (changes requested):
GitHub is reporting a merge conflict with Supervisor.cs. Please could you review and fix?
The razor markup for the _AssessmentQuestionResponse only applies the new logic to one of it's two cases:
image
The second case (highlighted) also needs the new logic applying to it. This screenshot shows an unverified score against an assessment question without level descriptors (the second case) still showing up green:
image

PR #798, opened 2021-12-06
HEEDLS-698 Refactor the Filtered Results actions to use the feature flag

AlexJacksonDS:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-698

### Description
Basic feature flagged page for the recommended learning results, behind the signposting feature flag. With the feature flag off, the behaviour should be unchanged apart from the route.

### Screenshots
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Web/Controllers/LearningPortalController/FilteredResults.cs
```diff
@@ -1,126 +0,0 @@
...
-             {
-                 SelfAssessment = assessment,
-                 CompetencyPlayLists = await filteredApiHelperService.GetPlayListsPoll<IEnumerable<PlayList>>(filteredToken, "playlist.FetchCompetencyPlaylists"),
-                 RecommendedPlayLists = await filteredApiHelperService.GetPlayListsPoll<IEnumerable<PlayList>>(filteredToken, "playlist.FetchNexRexPlaylists"),
-                 FavouritePlayList = favouritePlayList
-            };
-            return View("SelfAssessments/FilteredMgp/FilteredResults", model);
-        }
-        [Route("/LearningPortal/SelfAssessment/{selfAssessmentId:int}/Filtered/Dashboard")]
-        public async Task<IActionResult> FilteredRecommendations(int selfAssessmentId)
```
DanBloxham-sw, line 64: There are a few places that still attempt to link to these old actions that will probably be broken if accessing the relevant pages now with signposting disabled. These are mostly in back links.
Searching for "FilteredRecommendations" should bring up a few locations to bring back to.

DigitalLearningSolutions.Web.Tests/Controllers/LearningPortal/RecommendedLearningControllerTests.cs
```diff
@@ -0,0 +1,87 @@
...
+                .WithMockUser(true, delegateId: DelegateId);
+        }
+
+        [Test]
+        public async Task
+            SelfAssessmentResults_redirect_to_expected_action_does_not_call_filtered_api_when_using_signposting()
+        {
+            // Given
+
+            A.CallTo(() => configuration[ConfigHelper.UseSignposting]).Returns("true");
```
DanBloxham-sw, line 48: Stray empty line

DigitalLearningSolutions.Web.Tests/Controllers/LearningPortal/RecommendedLearningControllerTests.cs
```diff
@@ -0,0 +1,87 @@
...
+                    selfAssessmentService,
+                    configuration
+                )
+                .WithDefaultContext()
+                .WithMockUser(true, delegateId: DelegateId);
+        }
+
+        [Test]
+        public async Task
+            SelfAssessmentResults_redirect_to_expected_action_does_not_call_filtered_api_when_using_signposting()
```
DanBloxham-sw, line 44: I'm debating whether there is a point to introducing tests to make sure everything works as expected when signposting is not used. Given that there weren't any before, and the old code will probably end up unused, I think it's fine to not implement any new tests for that. Thoughts?

DanBloxham-sw (changes requested):
Mostly minor things and a few questions.

DigitalLearningSolutions.Data/Services/SelfAssessmentService.cs
```diff
@@ -289,6 +291,13 @@ public void LogAssetLaunch(int candidateId, int selfAssessmentId, LearningAsset
             selfAssessmentDataService.LogAssetLaunch(candidateId, selfAssessmentId, learningAsset);
         }
 
+        public bool CanDelegateAccessSelfAssessment(int delegateId, int selfAssessmentId)
+        {
+            var candidateAssessments = selfAssessmentDataService.GetCandidateAssessments(delegateId, selfAssessmentId);
+
+            return candidateAssessments.Any(ca => ca.CompletedDate == null && ca.RemovedDate == null);
```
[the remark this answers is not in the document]
SteveJacksonSoft: We shouldn't. Delegates should be able to do the self-assessment again e.g. after they've done some training in an area they were struggling with.

DigitalLearningSolutions.Web/Controllers/LearningPortalController/RecommendedLearningController.cs
```diff
@@ -0,0 +1,183 @@
...
+            ISelfAssessmentService selfAssessmentService,
+            IConfiguration configuration
+        )
+        {
+            this.filteredApiHelperService = filteredApiHelperService;
+            this.selfAssessmentService = selfAssessmentService;
+            this.configuration = configuration;
+        }
+
+        [Route("/LearningPortal/SelfAssessment/{selfAssessmentId:int}/Results")]
```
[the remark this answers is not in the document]
AlexJacksonDS: There's an action SetCompleteAsset that doesn't have a route that we'd probably need to verify I haven't broken by adding a route to it.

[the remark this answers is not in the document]
DanBloxham-sw: Wouldn't the route have already broken with this moving into a new controller?

DigitalLearningSolutions.Web.Tests/Controllers/LearningPortal/RecommendedLearningControllerTests.cs
AlexJacksonDS: Yeah that was roughly my thinking in not having them there. I'd also have to do a lot more figuring out what we need setting up all the faked API calls.
I think we should probably be covered by testing the system with the feature flag off, but we could always come back and add them if we want later if that testing all passes.

DigitalLearningSolutions.Web/Controllers/LearningPortalController/RecommendedLearningController.cs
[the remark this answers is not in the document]
AlexJacksonDS: It should work fine at the default value (as it did before). Adding the Route to the controller would break it unless we also added some route to the action as well.

DigitalLearningSolutions.Data/Services/SelfAssessmentService.cs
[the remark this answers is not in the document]
AlexJacksonDS: I've based the service filter that this is used in on the old behaviour of returning access denied when the GetSelfAssessmentForCandidateById returns null, the where clause of this query has both of these values as NULL.

DigitalLearningSolutions.Web/Views/LearningPortal/RecommendedLearning/RecommendedLearning.cshtml
```diff
@@ -0,0 +1,10 @@
+﻿@{
+  Layout = "../SelfAssessments/_Layout";
+  ViewData["Title"] = "Recommended learning";
```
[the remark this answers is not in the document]
AlexJacksonDS: Added a basic model just containing the whole self assessment record for now, same as the old page.

DigitalLearningSolutions.Data.Tests/Services/SelfAssessmentServiceTests.cs
```diff
@@ -16,5 +22,57 @@ public void Setup()
...
+
+            // When
+            var result = selfAssessmentService.CanDelegateAccessSelfAssessment(1, 1);
+
+            // Then
+            result.Should().BeTrue();
+        }
+
+        [Test]
+        public void CanDelegateAccessSelfAssessment_returns_false_with_only_completed_or_removed_assessments()
```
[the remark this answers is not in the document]
AlexJacksonDS: Added extra test for now, but may end up removing some depending on the other comment chain

DigitalLearningSolutions.Web/Controllers/LearningPortalController/RecommendedLearningController.cs
```diff
@@ -0,0 +1,183 @@
...
+                    null
+                ),
+            };
+            return View("../LearningPortal/SelfAssessments/FilteredMgp/FilteredResults", model);
+        }
+
+        private string GetCandidateNumber()
+        {
+            return User.GetCustomClaim(CustomClaimTypes.LearnCandidateNumber) ?? "";
+        }
```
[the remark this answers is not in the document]
AlexJacksonDS: Moved to a known not null helper

DigitalLearningSolutions.Web/Controllers/LearningPortalController/RecommendedLearningController.cs
```diff
@@ -0,0 +1,183 @@
...
+            selfAssessmentService.UpdateLastAccessed(selfAssessmentId, candidateId);
+
+            return configuration.IsSignpostingUsed()
+                ? View("RecommendedLearning")
+                : await ReturnFilteredResultsView(selfAssessmentId, candidateId);
+        }
+
+        [Route("/LearningPortal/SelfAssessment/{selfAssessmentId:int}/Filtered/PlayList/{playListId}")]
+        public async Task<IActionResult> FilteredCompetencyPlaylist(int selfAssessmentId, string playListId)
+        {
```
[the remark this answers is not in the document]
AlexJacksonDS: Added a simple config check and return not found to all the other actions. I don't think it's really worth adding an attribute instead

DigitalLearningSolutions.Web/Controllers/LearningPortalController/FilteredResults.cs
AlexJacksonDS: I think I've caught all these now.

DigitalLearningSolutions.Web/Controllers/LearningPortalController/RecommendedLearningController.cs
```diff
@@ -0,0 +1,211 @@
...
+
+            return configuration.IsSignpostingUsed()
+                ? ReturnSignpostingRecommendedLearningView(selfAssessmentId, candidateId)
+                : await ReturnFilteredResultsView(selfAssessmentId, candidateId);
+        }
+
+        [Route("/LearningPortal/SelfAssessment/{selfAssessmentId:int}/Filtered/PlayList/{playListId}")]
+        public async Task<IActionResult> FilteredCompetencyPlaylist(int selfAssessmentId, string playListId)
+        {
+            if (configuration.IsSignpostingUsed())
```
SteveJacksonSoft, line 75: Is there a reason not to use the FeatureGate attribute here? That feels more straightforward.

SteveJacksonSoft: Belay, just realised this is an upside-down feature gate.

DigitalLearningSolutions.Web/Controllers/LearningPortalController/RecommendedLearningController.cs
```diff
@@ -0,0 +1,211 @@
...
+            selfAssessmentService.SetUpdatedFlag(selfAssessmentId, candidateId, false);
+
+            if (!configuration.IsSignpostingUsed())
+            {
+                var filteredToken = await GetFilteredToken();
+                var profile = selfAssessmentService.GetFilteredProfileForCandidateById(selfAssessmentId, candidateId);
+                var goals = selfAssessmentService.GetFilteredGoalsForCandidateId(selfAssessmentId, candidateId)
+                    .ToList();
+                var response = await filteredApiHelperService.UpdateProfileAndGoals(filteredToken, profile, goals);
+            }
```
SteveJacksonSoft, line 54: I'm a bit confused why this doesn't take the user straight to the Filtered/Results page if Signposting is turned off, rather than going through the new RecommendedLearning endpoint first.
I would have expected this endpoint to direct the user either to Signposting or to Filtered depending on which was enabled.

DigitalLearningSolutions.Web/Views/LearningPortal/RecommendedLearning/RecommendedLearning.cshtml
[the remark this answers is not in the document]
DanBloxham-sw: Not sure if this is ready for re-review, but just quickly checked this out. Looks like the layout is not quite working correctly with this. The breadcrumbs try to lead to /RecommendedLearning/Current with the controller change, which causes an error page to spring up instead. The additional breadcrumbs from the previous page aren't carried forward either.

DigitalLearningSolutions.Web/Controllers/LearningPortalController/FilteredResults.cs
DanBloxham-sw: Looks like there are still a couple on Playlist.cshtml.

DigitalLearningSolutions.Web/Views/LearningPortal/RecommendedLearning/RecommendedLearning.cshtml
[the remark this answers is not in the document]
AlexJacksonDS: Fixed the link with no controller specified and added the rest of the breadcrumbs to the new page.

image

DigitalLearningSolutions.Web/Controllers/LearningPortalController/RecommendedLearningController.cs
AlexJacksonDS: I've tried to reduce the duplicated code between the 2 actions here, although this may have made it more confusing. We are either submitting and displaying, or just displaying depending on whether the self assessment has unprocessed updates.

Previously both actions contained the contents of the RecommendedLearning action (I've renamed this to remove the Filtered from the name) and this SelfAssessmentResults action contained the extra stuff that still remains. Both actions had the same bookmark etc.

We need to preserve the 2 actions logic from the view since we still care about marking the self assessment as submitted only when it can be but we could potentially change the redirect pattern here if there is a better way to do this.

DigitalLearningSolutions.Data/Services/SelfAssessmentService.cs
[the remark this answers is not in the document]
SteveJacksonSoft: Stick with what the Filtered code was doing regarding completed assessments (so belay my last comment).
Only admins can mark self-assessments as completed currently, and Kevin has some more thinking to do about what should happen in that scenario (if it should ever happen).

DigitalLearningSolutions.Web/Controllers/LearningPortalController/RecommendedLearningController.cs
SteveJacksonSoft: @AlexJacksonDS Okay, I think this is generally alright. Couple of things in here that I'm wondering about
* We're setting the url of the RecommendedLearning endpoint as a bookmark on the user's self assessment, and this code change modifies the url (removes "Filtered") - could this cause problems?
* Can the bit inside the if-statement here be pulled out into a method `UpdateFilteredProfileAndGoalsForDelegate(candidateId)`?

AlexJacksonDS: I'm not 100% sure on the bookmarks, I guess it might. Maybe we need to have an endpoint that still has the old route and sets the bookmark accordingly, then change that to redirect to the new one?

SteveJacksonSoft: Talked to Kevin about this, and we'd have to run a script on prod to update the bookmarks if we changed the url for this endpoint. We agreed it would be better just to keep it as it is, even though "Filtered" is in the current url.

SteveJacksonSoft: (The bookmark is the url the user gets taken to when they launch the self-assessment.) Given that the endpoint it currently bookmarks updates the submission time, I think it's fine to stick with bookmarking the url of this endpoint. The extra bonus in doing that is that you redirect them to an endpoint with a non-filtered-related url.

AlexJacksonDS: Does [this logic](https://github.com/TechnologyEnhancedLearning/DLSV2/blob/HEEDLS-698-controller-refactor-temp/DigitalLearningSolutions.Web/Controllers/LearningPortalController/RecommendedLearningController.cs) look any better? I've re-added the Filtered/Dashboard action, which will redirect to the new one when signposting is on.

The /Results endpoint should be fine to change, it was never bookmarked, so we can't break the continue button on the first page of the self assessment which is the only place that the bookmark is used.

SteveJacksonSoft: @AlexJacksonDS Think the pattern looks about right. Thoughts:
* We should move the `if (configuration.IsSignpostingUsed())` to the top of `RecommendedLearning()` to mirror what happens in `FilteredDashboard`. 
    * On a related note, we can just redirect from RecommendedLearning() to FilteredDashboard() in this case.

DanBloxham-sw (commented):
I think this is good. Couple of minor questions.

DigitalLearningSolutions.Web/Views/LearningPortal/RecommendedLearning/RecommendedLearning.cshtml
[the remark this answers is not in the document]
AlexJacksonDS: Just because I've copied the breadcrumbs from the old filtered results view, which they're set up the same as here in. I can fix it to include the name, but I'd prefer to resolve it properly in the next subtask where I'm actually making the page/view model etc. properly.

DigitalLearningSolutions.Web/Controllers/LearningPortalController/RecommendedLearningController.cs
```diff
@@ -0,0 +1,236 @@
...
+            {
+                await UpdateFilteredProfileAndGoalsForDelegate(selfAssessmentId, candidateId);
+                return RedirectToAction("FilteredDashboard", new { selfAssessmentId });
+            }
+
+            return RedirectToAction("RecommendedLearning", new { selfAssessmentId });
+        }
+
+        [Route("/LearningPortal/SelfAssessment/{selfAssessmentId:int}/RecommendedLearning")]
+        public async Task<IActionResult> RecommendedLearning(int selfAssessmentId)
```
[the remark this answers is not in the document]
AlexJacksonDS: It will once I implement the back end stuff that calls the API, might as well just leave it now as it doesn't affect how the page works.

DigitalLearningSolutions.Web/Views/LearningPortal/RecommendedLearning/RecommendedLearning.cshtml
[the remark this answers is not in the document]
DanBloxham-sw: Fair enough,

DigitalLearningSolutions.Web/Controllers/LearningPortalController/RecommendedLearningController.cs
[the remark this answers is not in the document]
DanBloxham-sw: Good point.

DanBloxham-sw (approved):
I think this is good now.

DigitalLearningSolutions.Web/Views/LearningPortal/SelfAssessments/SelfAssessmentOverview.cshtml
```diff
@@ -1,103 +1,104 @@
...
+            @competencyGroup.Key
+          </span>
+        </summary>
+        <div class="nhsuk-details__text">
+          @if (Model.SelfAssessment.LinearNavigation) {
+            <partial name="SelfAssessments/_MeanScores"
+                     view-data="@(new ViewDataDictionary(ViewData) { { "competencyGroup", competencyGroup } })" />
+          }
+
+          <partial name="SelfAssessments/_OverviewTable" view-data="@( new ViewDataDictionary(ViewData) { { "linearNavigation", Model.SelfAssessment.LinearNavigation }, { "selfAssessment", Model.SelfAssessment } })" model="competencyGroup" />
```
livzorn, line 57: Some more long lines to break if possible

DigitalLearningSolutions.Web/Views/LearningPortal/SelfAssessments/FilteredMgp/FilteredResults.cshtml
```diff
@@ -55,7 +55,7 @@
         }
         else
         {
-          <p>Check back soon for recommended learning learning resources. <a class="trigger-loader" asp-action="FilteredRecommendations" asp-route-selfAssessmentId="@(ViewContext.RouteData.Values["selfAssessmentId"])">Reload resources.</a></p>
+          <p>Check back soon for recommended learning learning resources. <a class="trigger-loader" asp-controller="RecommendedLearning" asp-action="RecommendedLearning" asp-route-selfAssessmentId="@(ViewContext.RouteData.Values["selfAssessmentId"])">Reload resources.</a></p>
```
livzorn, line 58: long line

DigitalLearningSolutions.Web/Views/LearningPortal/SelfAssessments/SelfAssessmentOverview.cshtml
```diff
@@ -1,103 +1,104 @@
...
+       asp-action=@(Model.SelfAssessment.UnprocessedUpdates ? "SelfAssessmentResults" : "RecommendedLearning")>
+      Submit results
+    </a>
+  } else {
+    <a class="nhsuk-button finish-review-button trigger-loader"
+       asp-controller="RecommendedLearning"
+       asp-route-selfAssessmentId="@Model.SelfAssessment.Id"
+       asp-action=@(Model.SelfAssessment.UnprocessedUpdates ? "SelfAssessmentResults" : "RecommendedLearning")>
+      Submit self assessment
+    </a>
```
livzorn, line 87: role=button?

livzorn (changes requested):
Just some formatting things, breaking long lines and adding role=button (not sure if we need role=button but I've seen it mentioned in failed testing so just wanted to raise it)

DigitalLearningSolutions.Web/Views/LearningPortal/SelfAssessments/SelfAssessmentOverview.cshtml
```diff
@@ -1,103 +1,104 @@
...
+    <partial name="../../Supervisor/Shared/_SupervisorSignOffSummary.cshtml" model="Model.SupervisorSignOffs" />
+    @if (Model.AllQuestionsVerifiedOrNotRequired()) {
+      @if (!Model.SupervisorSignOffs.Any()) {
+        <p class="nhsuk-body-l">You have not yet requested @Model.SelfAssessment.SignOffRoleName sign-off for this self assessment.</p>
+      }
+      <a class="nhsuk-button nhsuk-button--secondary nhsuk-u-margin-bottom-2" asp-action="RequestSignOff" asp-route-vocabulary="@Model.SelfAssessment.Vocabulary" asp-route-selfAssessmentId="@Model.SelfAssessment.Id">Request @Model.SelfAssessment.SignOffRoleName sign-off</a>
+    } else {
+      <p class="nhsuk-body-l">All required @Model.SelfAssessment.Vocabulary.ToLower() self-assessments must be completed and verified, before requesting @Model.SelfAssessment.SignOffRoleName sign off of the @Model.SelfAssessment.Name.</p>
+    }
+
```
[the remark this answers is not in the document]
AlexJacksonDS: Done

DigitalLearningSolutions.Web/Views/LearningPortal/SelfAssessments/SelfAssessmentOverview.cshtml
AlexJacksonDS: Broken these down within reason, there are still some pretty long lines.

DigitalLearningSolutions.Web/Views/LearningPortal/SelfAssessments/SelfAssessmentOverview.cshtml
AlexJacksonDS: I've added them to all links with class nhsuk-button in this view

DigitalLearningSolutions.Web/Views/LearningPortal/SelfAssessments/FilteredMgp/AssetComplete.cshtml
```diff
@@ -7,14 +7,14 @@
...
+  <li class="nhsuk-breadcrumb__item"><a class="nhsuk-breadcrumb__link" asp-controller="RecommendedLearning" asp-action="RecommendedLearning" asp-route-selfAssessmentId="@(ViewContext.RouteData.Values["selfAssessmentId"])">Recommended Learning Resources</a></li>
+  <li class="nhsuk-breadcrumb__item"><a class="nhsuk-breadcrumb__link" asp-controller="RecommendedLearning" asp-action="FilteredLearningAsset" asp-route-selfAssessmentId="@(ViewContext.RouteData.Values["selfAssessmentId"])" asp-route-assetId="@Model.Id">Learning Resource</a></li>
 <li class="nhsuk-breadcrumb__item">Complete</li>
 }
 <link rel="stylesheet" href="@Url.Content("~/css/learningPortal/selfAssessment.css")" asp-append-version="true">
 
 <div class="nhsuk-back-link">
-  <a class="nhsuk-back-link__link trigger-loader" asp-action="FilteredLearningAsset" asp-route-selfAssessmentId="@(ViewContext.RouteData.Values["selfAssessmentId"])" asp-route-assetId="@Model.Id">
+  <a class="nhsuk-back-link__link trigger-loader" asp-controller="RecommendedLearning" asp-action="FilteredLearningAsset" asp-route-selfAssessmentId="@(ViewContext.RouteData.Values["selfAssessmentId"])" asp-route-assetId="@Model.Id">
```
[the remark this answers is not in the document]
AlexJacksonDS: I've actively avoided running the formatter/breaking down the really long lines in these views in the FilteredMgp folder (except where it was really dumb) since these are the views for the old page that I'm replacing in 648 and I wasn't sure if it was really worth spending the time to do this. This applies to the button role too. I can go back and change them all if we want that though.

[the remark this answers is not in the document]
livzorn: Ah I see, now worries then!

livzorn (approved):

DigitalLearningSolutions.Web/Controllers/LearningPortalController/RecommendedLearningController.cs
```diff
@@ -0,0 +1,236 @@
...
+
+            return ReturnSignpostingRecommendedLearningView(selfAssessmentId, candidateId);
+        }
+
+        [Route("/LearningPortal/SelfAssessment/{selfAssessmentId:int}/Filtered/Dashboard")]
+        public async Task<IActionResult> FilteredDashboard(int selfAssessmentId)
+        {
+            if (configuration.IsSignpostingUsed())
+            {
+                return RedirectToActionPermanent("RecommendedLearning", new { selfAssessmentId });
```
SteveJacksonSoft, line 77: I fear the `Permanent` could cause trouble if we want to revert to Filtered temporarily for some reason. I think it's better to use a temporary redirect because users shouldn't end up going to this endpoint twice after we turn signposting on, unless they're doing it manually (they'll only go here because of the leftover bookmark).

SteveJacksonSoft (approved):

PR #803, opened 2021-12-08
Heedls 686 prn add to edit details for delegates

showkath10:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-686

### Description
Added new radio buttons and professional registration box which will only be required when user selects Yes 
Professional registration has validation (must be between 5 and 20 characters allowed, only alphanumeric and hyphens allowed)
Professional registration also added to details card (this is not mentioned in the ticket)

### Screenshots
image
image
image
image
image
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -131,6 +133,24 @@ private IActionResult EditDetailsPostSave(EditDetailsFormData formData, DlsSubAp
...
+                        {
+                            continue;
+                        }
+                    }
+
+                    ModelState.TryAddModelError(
+                        key,
+                        result.ErrorMessage
+                    );
+                }
```
davidm-m, line 152: Why is this necessary? As far as I can tell we haven't needed anything like this before.

DigitalLearningSolutions.Web/Views/MyAccount/_EditRegistrationNumber.cshtml
```diff
@@ -0,0 +1,63 @@
...
+
+<div class="nhsuk-form-group @(errorHasOccurred? "nhsuk-form-group--error": "")">
+  <fieldset class="nhsuk-fieldset" aria-describedby="professional-registration-hint">
+    <div class="nhsuk-hint" id="professional-registration-hint">
+      Do you have a Professional Registration Number?
+    </div>
+
+  @if (errorHasOccurred) {
+    <span class="nhsuk-error-message" id="no-selection-error">
+      <span class="nhsuk-u-visually-hidden">Error:</span> @(!optionYesSelected? "Select an option":"")
```
davidm-m, line 20: We should be getting the error message from the error itself ideally, as this currently is if something unexpected happens we could display a message to users that just says "Error:"

DigitalLearningSolutions.Web/Views/MyAccount/_EditRegistrationNumber.cshtml
```diff
@@ -0,0 +1,63 @@
+﻿@using DigitalLearningSolutions.Data.Enums
+@using DigitalLearningSolutions.Web.ViewModels.MyAccount
+@model EditDetailsViewModel
+
+@{
+  var errorHasOccurred = !ViewData.ModelState.IsValid && @Model.ProfessionalRegNumberSelectionAnswer == YesNoSelectionEnum.None;
```
davidm-m, line 6: Not sure I like this, it relies too much on how the validation works elsewhere. Could we do something like check specifically where there's an error on the `ProfessionalRegNumberSelectionAnswer` or `ProfessionalRegistrationNumber` properties?

DigitalLearningSolutions.Web/ViewModels/MyAccount/EditDetailsFormData.cs
```diff
@@ -90,5 +105,70 @@ protected EditDetailsFormData(EditDetailsFormData formData)
         public string? Answer5 { get; set; }
 
         public string? Answer6 { get; set; }
+
+        public string? ProfessionalRegistrationNumber { get; set; }
+
+        public YesNoSelectionEnum ProfessionalRegNumberSelectionAnswer { get; set; }
```
davidm-m, line 111: I'm not a big fan of the abbreviation to Reg here, but agree it would be a very long property name without it. Would something like `HasProfessionalRegistrationNumber` work instead?

DigitalLearningSolutions.Web/Views/MyAccount/_EditRegistrationNumber.cshtml
```diff
@@ -0,0 +1,63 @@
...
+        <vc:text-input asp-for="@nameof(Model.ProfessionalRegistrationNumber)"
+                       label="Professional Registration Number"
+                       populate-with-current-value="true"
+                       type="text"
+                       spell-check="false"
+                       hint-text=""
+                       autocomplete="off"
+                       css-class="nhsuk-u-width-one-half" />
+      </div>
+      <input type="hidden" asp-for="ProfessionalRegistrationNumber" />
```
davidm-m, line 60: Is this needed? What happens if we don't have it?

davidm-m (changes requested):
A few questions and things I think could be tidied up

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
showkath10: Since EditDetailsFormData has other data annotations, they take priority over IValidatableObject so when user tries to save the page without selecting the professional registration number and when there are other entries are not entered like first name, password etc. -- only list of data annotations validations will be shown

IValidatableObject validations will only be shown when all other validations are passed. To resolve this, we are calling validate method implicitly so all the validations are shown. There is check to avoid duplicate validation being added to ModelState.

If we remove this code, professional registration validation will only be shown when all other existing validations are passed

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests/DelegateUserDataServiceTests.cs
```diff
@@ -3,6 +3,7 @@
     using System;
     using System.Linq;
     using System.Transactions;
+    using DigitalLearningSolutions.Data.Enums;
```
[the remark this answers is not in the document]
showkath10: Removed this

DigitalLearningSolutions.Web/Views/MyAccount/_EditRegistrationNumber.cshtml
showkath10: sorry, yes, I removed it as we don't need this here

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
davidm-m: Aha, I see the problem. Sounds like this could come up elsewhere then, is there any way to make this reusable? Maybe a helper function like `public static void ManuallyValidate(this IValidatableObject model, ModelStateDictionary ModelState)` that runs this exact code?

showkath10: I think, yes, we can do. I will make changes. Thanks

DigitalLearningSolutions.Web/ViewModels/MyAccount/EditDetailsFormData.cs
showkath10: Sorry, I have renamed it as suggested

DigitalLearningSolutions.Web/Views/MyAccount/_EditRegistrationNumber.cshtml
showkath10: This is not needed so I removed it

DigitalLearningSolutions.Web/Views/MyAccount/_EditRegistrationNumber.cshtml
showkath10: I have updated the code. We only show validation in case the professional registration number is selected on postback, when model state is invalid. If we don't add this check, the validation may appear when page loads.

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
showkath10: I have moved it to helper class

DigitalLearningSolutions.Web/Views/MyAccount/_EditRegistrationNumber.cshtml
```diff
@@ -3,27 +3,28 @@
...
       Do you have a Professional Registration Number?
     </div>
 
-  @if (errorHasOccurred) {
+  @if (errorHasOccurred && optionNoneSelected) {
     <span class="nhsuk-error-message" id="no-selection-error">
-      <span class="nhsuk-u-visually-hidden">Error:</span> @(!optionYesSelected? "Select an option":"")
+       @(!optionYesSelected? "Select an option":"")
     </span>
   }
```
davidm-m, line 23: We definitely still want the visually hidden `Error: ` so that a screen reader announces this as being an error to a user. What we don't want is to have the logic of whether and error has occurred and what the error message should be to be set in the view - that's the responsibility of the model.
We should be displaying an error here only if there is an error on the `HasProfessionalRegistrationNumber` property and then we should be displaying the error message from that error, not hard-coding a string in the view. You can check e.g. `TextInputViewComponent.cs` to see how to get the error messages for a specific property.

davidm-m (changes requested):
Still needs some tweaks to how we display error messages on the radio buttons, but the other changes are good.

DigitalLearningSolutions.Web/Views/MyAccount/_EditRegistrationNumber.cshtml
showkath10: I have done this change

DigitalLearningSolutions.Web/Views/MyAccount/_EditRegistrationNumber.cshtml
```diff
@@ -4,23 +4,23 @@
 
 @{
   var errorHasOccurred = !ViewData.ModelState.IsValid;
```
davidm-m, line 6: I think `errorHasOccurred` should only be true on this view if there is an error on the `HasProfessionalRegistrationNumber` property

DigitalLearningSolutions.Web/Views/MyAccount/_EditRegistrationNumber.cshtml
```diff
@@ -4,23 +4,23 @@
...
     <div class="nhsuk-hint" id="professional-registration-hint">
       Do you have a Professional Registration Number?
     </div>
 
-  @if (errorHasOccurred && optionNoneSelected) {
-    <span class="nhsuk-error-message" id="no-selection-error">
-       @(!optionYesSelected? "Select an option":"")
-    </span>
-  }
+    @if (errorHasOccurred) {
```
davidm-m, line 19: This currently means this error section will show up if there's an error anywhere on the ViewModel, but if we change errorHasOccurred like I've described above we should get around that.

DigitalLearningSolutions.Web/Views/MyAccount/_EditRegistrationNumber.cshtml
```diff
@@ -4,23 +4,23 @@
...
+      <span class="nhsuk-error-message" id="no-selection-error">
+        <span class="nhsuk-u-visually-hidden">Error:</span> @(optionNoneSelected ? ViewData.ModelState[nameof(Model.HasProfessionalRegistrationNumber)].Errors[0].ErrorMessage : "")
```
davidm-m, line 21: This is still leaving us too open to future issues if things change down the line. We shouldn't be checking `optionNoneSelected` here, the conditions should be:

- If there is an error on the `HasProfessionalRegistrationNumber` property (i.e. if `errorHasOccurred` is true, which we are checking in the `if` above), display the error message from the ViewModel
- If there is no error on the `HasProfessionalRegistrationNumber` property, add nothing here

That way if we ever want to tweak what causes an error here, we only have to change the validation in the ViewModel and the view will just pick it up.

davidm-m (changes requested):
Still some changes to be made to how we display errors on the view

DigitalLearningSolutions.Web/Views/MyAccount/_EditRegistrationNumber.cshtml
showkath10: I have updated this to use HasProfessionalRegistrationNumber for validation

DigitalLearningSolutions.Web/Views/MyAccount/_EditRegistrationNumber.cshtml
showkath10: This error section will now only be shown when there is an error on HasProfessionalRegistrationNumber property

DigitalLearningSolutions.Web/Views/MyAccount/_EditRegistrationNumber.cshtml
showkath10: Yes, I agree. I have now changed the logic how we set model state. So this validation works only based on HasProfessionalRegistrationNumber property

davidm-m: We're still checking the value of `optionNoneSelected` here - since we're setting `errorHasOccurred` to only be true if there is definitely an error on `HasProfessionalRegistrationNumber` I don't think we need that, can we get rid of it?

showkath10: I have removed this completely and all look ok now.

DigitalLearningSolutions.Web/Helpers/CentreCustomPromptHelper.cs
```diff
@@ -129,7 +132,21 @@ public List<CustomFieldViewModel> GetCustomFieldViewModelsForCentre(int centreId
...
+                formData.Answer2,
+                formData.Answer3,
+                formData.Answer4,
+                formData.Answer5,
+                formData.Answer6,
+                modelState
+            );
+        }
+
+        public void ValidateCustomPrompts(MyAccountEditDetailsFormData formData, int centreId, ModelStateDictionary modelState)
```
davidm-m, line 149: Can these two methods be consolidated as one that takes an EditDetailsFormData argument?

DigitalLearningSolutions.Web/ViewModels/MyAccount/MyAccountEditDetailsFormData.cs
```diff
@@ -33,6 +35,17 @@ public class MyAccountEditDetailsFormData : EditDetailsFormData, IValidatableObj
             Answer4 = delegateUser?.Answer4;
             Answer5 = delegateUser?.Answer5;
             Answer6 = delegateUser?.Answer6;
+
+            if(IsDelegateUser)
```
davidm-m, line 39: Missing a space

DigitalLearningSolutions.Web/Views/MyAccount/EditDetails.cshtml
```diff
@@ -15,7 +15,7 @@
 <div class="nhsuk-grid-row">
   <div class="nhsuk-grid-column-full">
     @if (errorHasOccurred) {
-      <vc:error-summary order-of-property-names="@(new[] { nameof(Model.FirstName), nameof(Model.LastName), nameof(Model.Email), nameof(Model.Password) })" />
+      <vc:error-summary order-of-property-names="@(new[] { nameof(Model.FirstName), nameof(Model.LastName), nameof(Model.Email), nameof(Model.Password), nameof(Model.ProfessionalRegistrationNumber), nameof(Model.HasProfessionalRegistrationNumber) })" />
```
davidm-m, line 18: I think the new properties are the wrong way round; we want `HasProfessionalRegistrationNumber` before `ProfessionalRegistrationNumber`

DigitalLearningSolutions.Web/Views/MyAccount/_EditRegistrationNumber.cshtml
```diff
@@ -0,0 +1,67 @@
...
+
+@{
+  var errorHasOccurred = false;
+  if (ViewData.ModelState.TryGetValue(nameof(Model.HasProfessionalRegistrationNumber), out var modelState)) {
+    errorHasOccurred = modelState.ValidationState != ModelValidationState.Valid;
+  }
+
+  var optionYesSelected = Model.HasProfessionalRegistrationNumber == YesNoSelectionEnum.Yes;
+  var optionNoSelected = Model.HasProfessionalRegistrationNumber == YesNoSelectionEnum.No;
+  var professionalRegConditionalId = "professional-reg-conditional-Id";
```
davidm-m, line 15: Can you switch this to a const?

davidm-m (changes requested):
A few really small things to fix but overall looks great.

DigitalLearningSolutions.Web/Views/MyAccount/_EditRegistrationNumber.cshtml
wilkinsm: Done.

DigitalLearningSolutions.Web/Views/MyAccount/EditDetails.cshtml
wilkinsm: Good spot, done.

DigitalLearningSolutions.Web/ViewModels/MyAccount/MyAccountEditDetailsFormData.cs
wilkinsm: Done.

DigitalLearningSolutions.Web/Helpers/CentreCustomPromptHelper.cs
wilkinsm: Ah yes good spot, done.

davidm-m (approved):
Fantastic stuff

DigitalLearningSolutions.Web/ViewModels/MyAccount/MyAccountEditDetailsFormData.cs
```diff
@@ -62,5 +77,84 @@ protected MyAccountEditDetailsFormData(MyAccountEditDetailsFormData formData)
         public IFormFile? ProfileImageFile { get; set; }
 
         public bool IsDelegateUser { get; set; }
+
+        [RunValidateMethod(nameof(ValidateProfessionalRegistrationNumber))]
+        public string? ProfessionalRegistrationNumber { get; set; }
+
+        [RunValidateMethod(nameof(ValidateHasProfessionalRegistrationNumber))]
+        public YesNoSelectionEnum HasProfessionalRegistrationNumber { get; set; }
```
SteveJacksonSoft, line 85: Do you think it's better to use `YesNoSelectionEnum` than `bool?`?
I think the general use case for radio buttons will be "some list of values or null", so having a nullable type for this kind of data makes sense to me (if c# casts '' to null nicely).

DigitalLearningSolutions.Web/ViewModels/MyAccount/MyAccountEditDetailsFormData.cs
```diff
@@ -33,6 +35,17 @@ public class MyAccountEditDetailsFormData : EditDetailsFormData, IValidatableObj
...
+            if (IsDelegateUser)
+            {
+                ProfessionalRegistrationNumber = delegateUser?.ProfessionalRegistrationNumber!;
+                if (delegateUser!.HasBeenPromptedForPrn)
+                {
+                    HasProfessionalRegistrationNumber = !string.IsNullOrEmpty(ProfessionalRegistrationNumber)
+                        ? YesNoSelectionEnum.Yes
+                        : YesNoSelectionEnum.No;
+                }
+            }
```
SteveJacksonSoft, line 48: It feels like this belongs in the Controller / Service really, because it's logic rather than just housing data.

DigitalLearningSolutions.Web/ViewModels/MyAccount/MyAccountEditDetailsFormData.cs
```diff
@@ -62,5 +77,84 @@ protected MyAccountEditDetailsFormData(MyAccountEditDetailsFormData formData)
...
 
         public bool IsDelegateUser { get; set; }
+
+        [RunValidateMethod(nameof(ValidateProfessionalRegistrationNumber))]
+        public string? ProfessionalRegistrationNumber { get; set; }
+
+        [RunValidateMethod(nameof(ValidateHasProfessionalRegistrationNumber))]
+        public YesNoSelectionEnum HasProfessionalRegistrationNumber { get; set; }
+
+        public IEnumerable<ValidationResult> ValidateHasProfessionalRegistrationNumber()
```
SteveJacksonSoft, line 87: Similar to the other validate method, this will be wanted by the registration journey as well. I think it's right to make it an attribute of its own.

DigitalLearningSolutions.Web/Attributes/RunValidateMethodAttribute.cs
```diff
@@ -0,0 +1,38 @@
+﻿namespace DigitalLearningSolutions.Web.Attributes
+{
+    using System.Collections.Generic;
+    using System.ComponentModel.DataAnnotations;
+    using System.Linq;
+
+    public class RunValidateMethodAttribute : ValidationAttribute
```
SteveJacksonSoft, line 7: This is a nice idea, but I'm worried it could be misused. I haven't yet seen a case where writing a validation method on a form data model is better than making a globally-usable attribute for it. Do you still think there are benefits to using this pattern?

SteveJacksonSoft (changes requested):

DigitalLearningSolutions.Web/Views/MyAccount/_MyDetailsCard.cshtml
```diff
@@ -42,12 +42,14 @@
           }
         }
 
-        <div class="nhsuk-summary-list__row details-list-with-button__row">
-          <dt class="nhsuk-summary-list__key">
-            Professional registration number
-          </dt>
-          <partial name="_SummaryFieldValue" model="@Model.ProfessionalRegistrationNumber" />
-        </div>
+        @if (!string.IsNullOrEmpty(Model.ProfessionalRegistrationNumber)) {
```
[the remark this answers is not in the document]
wilkinsm: Ah, I didn't spot that one.
Also, looking at the view model, this value is always populated with something.
In saying that though I think something does need to be done here.
If the user is an admin only user then I don't think this field needs to show.
Other delegate only fields here are wrapped:
```
@if (Model.DelegateNumber != null) {
  ....
}
```

DigitalLearningSolutions.Web/Attributes/RunValidateMethodAttribute.cs
wilkinsm: It may have some uses elsewhere but that is an unknown right now. 
There is a bit of a detachment here from the up-coming `HEEDLS-687` which was not considered during this refactor.
Given the actual target methods will become attributes this can go.

DigitalLearningSolutions.Web/ViewModels/MyAccount/MyAccountEditDetailsFormData.cs
wilkinsm: So this is part of the issue I raised on slack regarding the whole approach to the radio buttons - this hand rolled solution here although simpler steps away from what is done elsewhere.

DigitalLearningSolutions.Web/ViewModels/MyAccount/MyAccountEditDetailsFormData.cs
```diff
@@ -62,5 +77,84 @@ protected MyAccountEditDetailsFormData(MyAccountEditDetailsFormData formData)
...
+                    )
+                );
+
+                return validationResults;
+            }
+
+            return validationResults;
+        }
+
+        public IEnumerable<ValidationResult> ValidateProfessionalRegistrationNumber()
```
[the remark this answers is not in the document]
wilkinsm: I originally had these in seperate attributes but there is a requirement to only run this validation for delegates.
This is something we can do within the attribute but it felt less transparent.
Given we have common need, now thinking about `HEEDLS-687`, it makes more sense to have them as attributes.

DigitalLearningSolutions.Web/Views/MyAccount/_EditRegistrationNumber.cshtml
```diff
@@ -0,0 +1,67 @@
...
+      Do you have a Professional Registration Number?
+    </div>
+
+    @if (errorHasOccurred) {
+      <span class="nhsuk-error-message" id="no-selection-error">
+        <span class="nhsuk-u-visually-hidden">Error:</span> @ViewData.ModelState[nameof(Model.HasProfessionalRegistrationNumber)].Errors[0].ErrorMessage
+      </span>
+    }
+
+    <div class="nhsuk-radios nhsuk-radios--conditional">
```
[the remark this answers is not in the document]
wilkinsm: This is related to the point I raised in slack regarding the radio implementation.
We have a view component used for radios in `EditAdminRoles.cshtml`. 
Whether that implementation is overly complicated for this scenario or not is a question. But as you point out the dependent expanders add complexity. 
There is certainly scope to create a component.

DigitalLearningSolutions.Web/Helpers/CentreCustomPromptHelper.cs
```diff
@@ -1,13 +1,16 @@
...
     using DigitalLearningSolutions.Data.Models.User;
     using DigitalLearningSolutions.Data.Services;
     using DigitalLearningSolutions.Web.ViewModels.Common;
+    using DigitalLearningSolutions.Web.ViewModels.TrackingSystem.Delegates.EditDelegate;
     using Microsoft.AspNetCore.Mvc.ModelBinding;
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using DigitalLearningSolutions.Web.ViewModels.MyAccount;
+    using EditDetailsFormData = DigitalLearningSolutions.Web.ViewModels.Common.EditDetailsFormData;
```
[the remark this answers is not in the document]
wilkinsm: Done.

DigitalLearningSolutions.Data/Models/User/MyAccountDetailsData.cs
```diff
@@ -9,19 +9,24 @@ public class MyAccountDetailsData : AccountDetailsData
...
+            ProfessionRegistrationNumber = professionalRegNumber;
+            HasBeenPromptedForPrn = hasBeenPromptedForPrn;
             ProfileImage = profileImage;
         }
 
         public int? AdminId { get; set; }
         public int? DelegateId { get; set; }
         public string Password { get; set; }
         public byte[]? ProfileImage { get; set; }
+        public string? ProfessionRegistrationNumber { get; set; }
```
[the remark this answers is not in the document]
wilkinsm: Done.

DigitalLearningSolutions.Web/Views/MyAccount/_EditRegistrationNumber.cshtml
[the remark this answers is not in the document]
wilkinsm: Some of the challenges here which may over complicate matters include:
- The disconnect between the radios and associated expandables and the desire to ensure clarity on the relationship between the two
- A solution to encompasses the expanded content looks overly complex, especially given the expandable content varies.

DigitalLearningSolutions.Web/ViewModels/MyAccount/MyAccountEditDetailsFormData.cs
wilkinsm: Specifically on the bool option. I've gone for that. In this context it works fine.

DigitalLearningSolutions.Web/ViewModels/MyAccount/MyAccountEditDetailsFormData.cs
wilkinsm: I've simplified this and moved the `logic` bits out to a helper. Could argue that it could go in a service.

DigitalLearningSolutions.Web/Attributes/RunValidateMethodAttribute.cs
wilkinsm: This has gone.

DigitalLearningSolutions.Web/ViewModels/MyAccount/MyAccountEditDetailsFormData.cs
wilkinsm: See other comment.

DigitalLearningSolutions.Web/ViewModels/MyAccount/MyAccountEditDetailsFormData.cs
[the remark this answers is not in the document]
wilkinsm: So in the interest of reusabiliy and without over complicating matters this has been moved to a helper (rather than an attribute) and is called where ever required. In this case it's called in the controller.
Registration model and my account models are different. The logic here is dependant on other factors (other model property values). I could of gone for an interface or pushed the concerning properties into a class of their own.

DigitalLearningSolutions.Web/Views/MyAccount/_EditRegistrationNumber.cshtml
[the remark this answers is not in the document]
SteveJacksonSoft: I had a sense it would be hard to commonise.
You could pull this out into a PRN-specific partial to share with registration if it seems convenient?

DigitalLearningSolutions.Web/ViewModels/MyAccount/MyAccountEditDetailsFormData.cs
SteveJacksonSoft: Ah, sorry, I completely missed that this is effectively in the viewModel constructor, in which case I think it's fine here... but going in the helper works too.

DigitalLearningSolutions.Web/Views/MyAccount/_MyDetailsCard.cshtml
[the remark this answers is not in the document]
SteveJacksonSoft: Yeah, do you want to add it in then, and just comment on your ticket that that should be checked to?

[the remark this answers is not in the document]
SteveJacksonSoft: Good catch!

DigitalLearningSolutions.Web.Tests/Helpers/ProfessionalRegistrationNumberHelperTests.cs
```diff
@@ -0,0 +1,137 @@
...
+            // Then
+            state.IsValid.Should().BeTrue();
+        }
+
+        [Test]
+        public void ValidateProfessionalRegistrationNumber_sets_error_when_hasPrn_is_not_set()
+        {
+            // Given
+            var state = new ModelStateDictionary();
+            var expectedError = "Select your professional registration number status.";
```
SteveJacksonSoft, line 93: I've realised my problem with putting expected output in `Given` is that all of the test data and mock setups go there. So when I see a string literal here I always assume it's input data. Alright if we put this in `Then`?

SteveJacksonSoft: Also c.f. above test where expected output of `true` has to go in `Then` because it's a fluent assertion method.

SteveJacksonSoft (commented):
Couple of things I'd tweak just. Nice test cases!

DigitalLearningSolutions.Web/Views/MyAccount/_EditRegistrationNumber.cshtml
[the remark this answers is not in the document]
wilkinsm: Something needs to be done. I took a look at this but the way it was originally put together means it is slightly more complex.
The view takes a specific model for editing the account details so as it stands does not work as a shared view given registration is a different model. This needs more work.

DigitalLearningSolutions.Web/Views/MyAccount/_MyDetailsCard.cshtml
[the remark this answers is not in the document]
wilkinsm: Yep done.

DigitalLearningSolutions.Web/Helpers/ProfessionalRegistrationNumberHelper.cs
```diff
@@ -0,0 +1,50 @@
+﻿namespace DigitalLearningSolutions.Web.Helpers
+{
+    using System.Text.RegularExpressions;
+    using Microsoft.AspNetCore.Mvc.ModelBinding;
+
+    public class ProfessionalRegistrationNumberHelper
+    {
+        public static bool? SetHasProfessionalRegistrationNumber(bool? hasBeenPromptedForPrn, string? prn)
```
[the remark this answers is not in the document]
wilkinsm: Yeah thats fine, done.

DigitalLearningSolutions.Web.Tests/Helpers/ProfessionalRegistrationNumberHelperTests.cs
wilkinsm: Done.

DigitalLearningSolutions.Web/Helpers/AccountDetailsDataHelper.cs
```diff
@@ -20,6 +20,10 @@ int centreId
                 formData.FirstName!,
                 formData.LastName!,
                 formData.Email!,
+                formData.HasProfessionalRegistrationNumber == true
```
wilkinsm, line 23: @SteveJacksonSoft note this additional change here.
This is to prevent saving a PRN when it is not required or invalid.
Example:
Click `yes` then add an invalid PRN.
Then click `no`.
Saving will set that invalid prn on the account.
This change prevents that, if the PRN question was answered with a `no`.

DigitalLearningSolutions.Web.Tests/Helpers/ProfessionalRegistrationNumberHelperTests.cs
```diff
@@ -100,6 +102,7 @@ public void ValidateProfessionalRegistrationNumber_sets_error_when_hasPrn_is_not
             );
 
             // Then
+            const string expectedError = "Select your professional registration number status.";
```
SteveJacksonSoft, line 105: Is there any reason not to inline this? No need for re-review, do what you think is best.

DigitalLearningSolutions.Web/Helpers/AccountDetailsDataHelper.cs
SteveJacksonSoft: Oho, good catch! And thanks for the clarification.

DigitalLearningSolutions.Web/Views/MyAccount/_EditRegistrationNumber.cshtml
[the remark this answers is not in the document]
SteveJacksonSoft: Sure, let's look at it again in the Registration ticket. Would an interface `IIncludesPrnForm`, implemented by the two form data classes, be the solution?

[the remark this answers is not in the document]
wilkinsm: @SteveJacksonSoft that was my initial though. It should be very straight forward. It's kind of a workaround but the lesser of the two evils as it will avoid more complex refactor.

DigitalLearningSolutions.Web.Tests/Helpers/ProfessionalRegistrationNumberHelperTests.cs
wilkinsm: There is a habit of explicitly declaring test input and output where it appears to matter more for clarity on test scope.
But to be fair these are quite simplistic tests so we can inline this.
Sorted this now.

PR #804, opened 2021-12-08
Adds brackets around parameterised IN lists

kevwhitt-hee:
### JIRA link
N/A

### Description
Fix to a bug with the My Account / Change Password update password query. This was previously throwing an error because the query uses parameterised lists of admin and candidate IDs but these were missing brackets in the SQL query text.


SteveJacksonSoft:
@kevwhitt-hee I found the cause of the bug - it was a type handler we'd added. Here's a PR to remove it and replace it with a better solution which doesn't break random things: https://github.com/TechnologyEnhancedLearning/DLSV2/pull/806

kevwhitt-hee:
Superseded by #806

PR #809, opened 2021-12-09
HEEDLS-703 - Migration to add question comparison columns to Competen…

DanBloxham-sw:
…cyResourceAssessmentQuestionParameters

### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-703

### Description
New migration adding two new columns to the ComptencyResourceAssessmentQuestionParameters that will be used in the Recommendation Score generation for learning resources.

### Screenshots
N/A

-----
### Developer checks

I have:
  - https://softwiretech.atlassian.net/wiki/spaces/HEE/pages/19875758087/Recommendation+Score

### DIFF
--- a/DigitalLearningSolutions.Data.Migrations/202112091009_AddColumnsToCompetencyResourceAssessmentQuestionParameters.cs
+++ b/DigitalLearningSolutions.Data.Migrations/202112091009_AddColumnsToCompetencyResourceAssessmentQuestionParameters.cs
@@ -20,7 +20,7 @@
                 )
                 .OnTable("CompetencyResourceAssessmentQuestionParameters");
             Delete.Column("RelevanceAssessmentQuestionID").FromTable("CompetencyResourceAssessmentQuestionParameters");
-            Delete.Column("CompareToRoleRequirements ").FromTable("CompetencyResourceAssessmentQuestionParameters");
+            Delete.Column("CompareToRoleRequirements").FromTable("CompetencyResourceAssessmentQuestionParameters");
         }
     }
 }
--- a/DigitalLearningSolutions.Data.Migrations/202112091009_AddColumnsToCompetencyResourceAssessmentQuestionParameters.cs
+++ b/DigitalLearningSolutions.Data.Migrations/202112091009_AddColumnsToCompetencyResourceAssessmentQuestionParameters.cs
@@ -1,0 +1,26 @@
+﻿namespace DigitalLearningSolutions.Data.Migrations
+{
+    using FluentMigrator;
+
+    [Migration(202112091009)]
+    public class AddColumnsToCompetencyResourceAssessmentQuestionParameters : Migration
+    {
+        public override void Up()
+        {
+            Alter.Table("CompetencyResourceAssessmentQuestionParameters")
+                .AddColumn("RelevanceAssessmentQuestionID").AsInt32().Nullable().WithDefaultValue(null)
+                .ForeignKey("AssessmentQuestions", "ID")
+                .AddColumn("CompareToRoleRequirements").AsBoolean().NotNullable().WithDefaultValue(false);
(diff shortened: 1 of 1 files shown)


DigitalLearningSolutions.Data.Migrations/202112091009_AddColumnsToCompetencyResourceAssessmentQuestionParameters.cs
```diff
@@ -0,0 +1,26 @@
...
+        }
+
+        public override void Down()
+        {
+            Delete.ForeignKey(
+                    "FK_CompetencyResourceAssessmentQuestionParameters_RelevanceAssessmentQuestionID_AssessmentQuestions_ID"
+                )
+                .OnTable("CompetencyResourceAssessmentQuestionParameters");
+            Delete.Column("RelevanceAssessmentQuestionID").FromTable("CompetencyResourceAssessmentQuestionParameters");
+            Delete.Column("CompareToRoleRequirements ").FromTable("CompetencyResourceAssessmentQuestionParameters");
```
livzorn, line 23: Looks like an extra space after CompareToRoleRequirements

livzorn (changes requested):
Just one extra space, not sure if that affects anything?

DigitalLearningSolutions.Data.Migrations/202112091009_AddColumnsToCompetencyResourceAssessmentQuestionParameters.cs
DanBloxham-sw: Good spot. I tested the Down migration while it had the space, and the column was removed fine, so it looks like the Migrator is clever enough to trim these. I've removed the space now.

livzorn (approved):

AlexJacksonDS (commented):
In theory this looks fine, I'm not particularly keen on this column name though "RelevanceAssessmentQuestionID". Do we actually need it to be called this, or should it really be called "RelevantAssessmentQuestionID"?

DanBloxham-sw:
> In theory this looks fine, I'm not particularly keen on this column name though "RelevanceAssessmentQuestionID". Do we actually need it to be called this, or should it really be called "RelevantAssessmentQuestionID"?

I agree that it is not the nicest name, but I think it is the most appropriate. These columns are used in the context of generating the recommendation score for a particular learning resource. In that, we recommend items that are more useful to the user, which is partially determined by their answers in self assessment questions concerning related competencies. 

These questions should follow along a pattern of "How confident are you in {dealing with x}" followed by "How well trained do you think you need to be in {dealing with x}". These two types of questions are effectively "Confidence" and "Relevance" questions respectively - i.e., how confident are you in X, and how relevant do you think it actually is to you and your job. 

So, in a case where we look at these questions, we need to know which question is which, so we can look at the user's answers and alter the recommendations score from them. This table already has an "`AssessmentQuestionID`" that points to the Confidence question. The new "`RelevanceAssessmentQuestionID`" points to the Relevance question. So a name like "`RelevantAssessmentQuestionID`" might be subtly confusing - what makes a question the relevant one? I think `RelevanceAssessmentQuestionID` is more "off" sounding in natural English, which hopefully keys anyone in that it is more specific than just a relevant question.

While it would be good to rename the old `AssessmentQuestionID` as "`ConfidenceAssessmentQuestionID`", it is not guaranteed to be a Confidence question - the managers of self assessments can set them up differently, and choose different ways to assess delegates. But, the new `RelevanceAssessmentQuestionID` column should only be populated when confidence/relevance questions are used, and in that case the `AssessmentQuestionID` is going to be a confidence question.

Hopefully that makes sense, but I'm happy to clear anything up further if it doesn't.

AlexJacksonDS (approved):

PR #815, opened 2021-12-10
HEEDLS-671 - disable resource retrieval when signposting disabled

DanBloxham-sw:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-671

### Description
In 671 I mistakenly removed a Signposting config check from a method that it should not have been removed from. This MR adds the check back.

### Screenshots
N/A

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Services/ActionPlanServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/ActionPlanServiceTests.cs
@@ -145,6 +145,24 @@
         }
 
         [Test]
+        public async Task GetIncompleteActionPlanResources_returns_empty_list_if_signposting_is_disabled()
+        {
+            // Given
+            const int delegateId = 1;
+            A.CallTo(() => config[ConfigHelper.UseSignposting]).Returns("false");
+
+            // When
+            var result = await actionPlanService.GetIncompleteActionPlanResources(delegateId);
+
+            // Then
+            result.Should().BeEmpty();
+            A.CallTo(() => learningLogItemsDataService.GetLearningLogItems(A<int>._))
+                .MustNotHaveHappened();
+            A.CallTo(() => learningHubApiClient.GetBulkResourcesByReferenceIds(A<IEnumerable<int>>._))
+                .MustNotHaveHappened();
+        }
+
+        [Test]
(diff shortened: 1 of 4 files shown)


DigitalLearningSolutions.Data/Services/ActionPlanService.cs
```diff
@@ -102,6 +102,11 @@ public void AddResourceToActionPlan(int competencyLearningResourceId, int delega
 
         public async Task<IEnumerable<ActionPlanResource>> GetIncompleteActionPlanResources(int delegateId)
         {
+            if (!config.IsSignpostingUsed())
+            {
+                return new List<ActionPlanResource>();
+            }
```
SteveJacksonSoft, line 108: Is there a strong preference to put the gate in here rather than where this method is used? It feels like this method shouldn't be being called if signposting isn't enabled.

I think it's worth updating the name of the method if you're putting the gate in here. I generally wouldn't expect a service method to conk out if a config flag is set unless it explicitly says so.

DigitalLearningSolutions.Data.Tests/Services/ActionPlanServiceTests.cs
```diff
@@ -145,7 +145,8 @@ public void AddResourceToActionPlan_calls_expected_insert_data_service_methods()
         }
 
         [Test]
-        public async Task GetIncompleteActionPlanResources_returns_empty_list_if_no_incomplete_learning_log_items_found()
+        public async Task
+            GetIncompleteActionPlanResources_returns_empty_list_if_no_incomplete_learning_log_items_found()
         {
             // Given
             const int delegateId = 1;
```
AlexJacksonDS, line 152: If the responsibility is the caller now, do we need the line below this mocking the config?

AlexJacksonDS (changes requested):
Just a comment about mocked config values

DigitalLearningSolutions.Data.Tests/Services/ActionPlanServiceTests.cs
DanBloxham-sw: Good spot, removed it.

AlexJacksonDS (approved):

livzorn (approved):
Looks good to me :)

PR #817, opened 2021-12-10
Heedls 592 exclude old competency self assessments

JBsoftwire:
### JIRA link
[_HEEDLS-592_](https://softwiretech.atlassian.net/browse/HEEDLS-592)

### Description
The endpoints in the tracking system that deliver courses to the pages listed on the ticket (excluding Export, which I understand to not yet be implemented) now exclude courses with a DefaultContentTypeID of 4. The 'AdminUserCanAccessCourse' attribute also excludes such courses for any individual course endpoint.

### Screenshots
N/A

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Data/DataServices/ActivityDataService.cs
```diff
@@ -47,14 +47,16 @@ public ActivityDataService(IDbConnection connection)
                         Registered,
                         Completed,
                         Evaluated
-                    FROM tActivityLog
+                    FROM tActivityLog AS al
+                    JOIN Applications AS ap ON ap.ApplicationID = al.ApplicationID
```
wilkinsm, line 51: When a relational table is used for the sole purpose of filter then an `exists/not exists` where clause would be the preffered convention I beleive the, rather that using a `JOIN`.

wilkinsm (changes requested):
Nice one Jonathan, just a minor comment.

wilkinsm (approved):

wilkinsm (changes requested):
Thanks for making the updates.
If you can resolve the conflicts that are now detected then we should be good to move to TL review.

wilkinsm (approved):
Nice one Jonathan.

DigitalLearningSolutions.Data/DataServices/ActivityDataService.cs
```diff
@@ -47,14 +47,19 @@ public ActivityDataService(IDbConnection connection)
...
+                        AND CentreID = @centreId
+                        AND (@jobGroupId IS NULL OR JobGroupID = @jobGroupId)
+                        AND (@customisationId IS NULL OR al.CustomisationID = @customisationId)
+                        AND (@courseCategoryId IS NULL OR al.CourseCategoryId = @courseCategoryId)
+                        AND (Registered = 1 OR Completed = 1 OR Evaluated = 1))
+                        AND EXISTS (
+                            SELECT ap.ApplicationID
+                            FROM Applications ap
+                            WHERE ap.ApplicationID = al.ApplicationID
+                            AND ap.DefaultContentTypeID <> 4)",
```
SteveJacksonSoft, line 62: How easy would it be to add a test for this?

DigitalLearningSolutions.Data/DataServices/CourseDataService.cs
```diff
@@ -533,7 +541,8 @@ FROM CentreApplications
                             )
                             THEN CAST(1 AS BIT)
                             ELSE CAST(0 AS BIT)
-                        END AS CentreHasApplication
+                        END AS CentreHasApplication,
+                        a.DefaultContentTypeID AS DefaultContentTypeId
```
SteveJacksonSoft, line 545: Why fetch the DefaultContentTypeID here rather than just discount courses where it's 4?

DigitalLearningSolutions.Data.Tests/DataServices/GroupsDataServiceTests.cs
```diff
@@ -90,10 +90,6 @@ public void GetGroupCoursesForCentre_returns_expected_courses()
             {
                 1,
                 2,
-                21,
-                22,
-                23,
-                24,
```
SteveJacksonSoft, line 96: It would be good to have a couple of tests that courses with DefaultContentTypeID = 4 are discounted. I guess these are such courses?
I wouldn't worry about adding a test for every method you've changed, but even if it's just for the groups data service and activity data service, and maybe one for course data service.

DigitalLearningSolutions.Data/Services/CourseService.cs
```diff
@@ -134,7 +136,8 @@ public IEnumerable<CourseStatistics> GetCentreSpecificCourseStatistics(int centr
         {
```
SteveJacksonSoft, line 136: We could simplify this method a bit, by calling VerifyAdminUserCanViewCourse (which is a weaker condition) then checking the rest of the conditions (I think that's just checking the course isn't an all-centres course, but I haven't made sure of that). That would be worth doing I think.

SteveJacksonSoft (changes requested):

DigitalLearningSolutions.Data.Tests/DataServices/GroupsDataServiceTests.cs
JBsoftwire: Those are the courses that vanished when filtering out ID 4, yeah. I didn't add a new test for each method but did update several tests where failures were introduced. I'll have a look at what ones haven't been covered.

JBsoftwire: With a change I've done to activity data service tests I think all of it, groups and course are covered.

DigitalLearningSolutions.Data/Services/CourseService.cs
```diff
@@ -134,22 +132,17 @@ public IEnumerable<CourseStatistics> GetCentreSpecificCourseStatistics(int centr
...
-            {
-                return false;
-            }
+            var courseValidationDetails = courseDataService.GetCourseValidationDetails(customisationId, centreId);
 
-            if (categoryId != null && courseValidationDetails.CourseCategoryId != categoryId)
+            if (courseValidationDetails!.AllCentres)
             {
-                return false;
+                return null;
```
JBsoftwire, line 145: This should return false, not null.

DigitalLearningSolutions.Data.Tests/Services/CourseServiceTests.cs
```diff
@@ -418,29 +418,6 @@ public void
         }
 
         [Test]
-        public void VerifyAdminUserCanAccessCourse_should_return_null_with_content_type_id_4()
```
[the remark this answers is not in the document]
JBsoftwire: The course service no longer validates on content type ID, because the filtering happens in the query itself.

DigitalLearningSolutions.Data.Tests/DataServices/ActivityDataServiceTests.cs
```diff
@@ -37,7 +45,7 @@ public void GetFilteredActivity_gets_expected_activity()
             // then
             using (new AssertionScope())
             {
-                result.Count().Should().Be(13);
+                result.Count().Should().Be(9);
```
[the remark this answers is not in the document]
JBsoftwire: That information isn't returned by the query and I don't think there's a way to check against it that's much of an improvement on the implicit meaning of taking an exact count of results.

DigitalLearningSolutions.Data.Tests/Services/CourseServiceTests.cs
```diff
@@ -183,7 +183,7 @@ public void
             {
                 CentreId = 2,
                 CourseCategoryId = 2,
-                AllCentres = true,
+                AllCentres = false,
```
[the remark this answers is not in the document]
SteveJacksonSoft: Oh, sorry, I get it now. We are changing the logic. The new logic is only equivalent to the original if we assume that customisations will either be `allCentres` or will have a `centreId` matching the user's (and not both). But actually, the centre that made the customisation should be able to manage it, even if it is `allCentres`.

So I need to check something with Kevin (I'll do that on Monday), but I _think_ the fix will be to change the `all centres course => false` line to `all centres course && course.centreId != user.centreId => false`.

The case we're concerned about is where the user's centre matches the centre on an All Centres customisation. Presumably the user should be able to manage that course. I'll get back to you on Monday.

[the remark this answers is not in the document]
SteveJacksonSoft: Confirmation from Kevin around `AllCentres` courses:
* admins at the centre specified by Customisation.CentreId should be able to view and manage the course.
* admins at other centres should be able to view but **not** manage the course.

DigitalLearningSolutions.Data.Tests/Services/CourseServiceTests.cs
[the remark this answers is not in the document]
SteveJacksonSoft: Gotcha, ta

DigitalLearningSolutions.Data.Tests/DataServices/ActivityDataServiceTests.cs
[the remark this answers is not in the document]
SteveJacksonSoft: Ah okay. Fair.

DigitalLearningSolutions.Data.Tests/Services/CourseServiceTests.cs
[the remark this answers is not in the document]
SteveJacksonSoft: Almost there I think. Just one case that concerns me.
Can you write a test that AdminUserCanViewCourse returns true if all of the following are satisfied?
* The course is at their centre
* The course matches their categoryId
* The course is an all-centres course
* courseValidationDetails.CentreHasApplication = false.

At the moment I think it will fail, but it should pass.

DigitalLearningSolutions.Data/Services/CourseService.cs
```diff
@@ -157,22 +157,16 @@ public IEnumerable<CourseStatistics> GetCentreSpecificCourseStatistics(int centr
...
-            if (courseValidationDetails.AllCentres && !courseValidationDetails.CentreHasApplication)
-            {
-                return false;
-            }
+            var centreIdMatches = courseValidationDetails.CentreId == centreId;
+            var allCentresCourseWithApplication =
+                courseValidationDetails.AllCentres && courseValidationDetails.CentreHasApplication;
 
-            return true;
+            return centreIdMatches || allCentresCourseWithApplication;
```
SteveJacksonSoft, line 169: So much neater. But could we have a test?

SteveJacksonSoft (commented):
So much neater.

DigitalLearningSolutions.Data.Tests/Services/CourseServiceTests.cs
```diff
@@ -526,22 +549,6 @@ public void RemoveDelegateFromCourse_does_nothing_if_no_current_progress()
                 .MustNotHaveHappened();
         }
 
-        [Test]
-        public void VerifyAdminUserCanAccessCourse_should_return_null_when_course_does_not_exist()
```
SteveJacksonSoft: This was effectively a duplicate test right?

SteveJacksonSoft (commented):
Just one question about the test you deleted. If you deleted the test because it was a duplicate, then go ahead and merge.

DigitalLearningSolutions.Data.Tests/Services/CourseServiceTests.cs
JBsoftwire: Yeah, the body of the test was identical.

PR #838, opened 2021-12-29
Heedls 413 tracker endpoint store diagnostic json

livzorn:
### JIRA link
[_HEEDLS-413_](https://softwiretech.atlassian.net/browse/HEEDLS-413)

### Description
Created a new tracker endpoint for updating the diagnostic score on an aspProgress record.

### Screenshots
N/A

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Data/Enums/TrackerEndpointResponse.cs
```diff
@@ -0,0 +1,29 @@
...
+    using DigitalLearningSolutions.Data.Models.Tracker;
+
+    public class TrackerEndpointResponse : Enumeration, ITrackerEndpointDataModel
+    {
+        public static readonly TrackerEndpointResponse Success =
+            new TrackerEndpointResponse(1, nameof(Success));
+
+        public static readonly TrackerEndpointResponse UnexpectedException =
+            new TrackerEndpointResponse(-1, nameof(UnexpectedException));
+
```
livzorn, line 12: I renamed TrackerEndpointErrorResponse to TrackerEndpointResponse so it could include the Success response that this action returns and the switch statement will still work. Not sure if there's a better way to do that

DigitalLearningSolutions.Data.Tests/Services/TrackerActionServiceTests.cs
```diff
@@ -136,5 +142,63 @@ public void GetObjectiveArrayCc_returns_empty_object_json_if_no_results_found()
...
+                    425,
+                    4
+                )
+            ).MustHaveHappenedOnceExactly();
+        }
+
+        [Test]
+        [TestCase(null, "[{'tutorialid':425,'myscore':4},{'tutorialid':424,'myscore':3}]")]
+        [TestCase(1, null)]
+        public void StoreDiagnosticJson_returns_StoreDiagnosticScoreException_if_parameter_missing(
```
AlexJacksonDS, line 185: Can we also have a test for the exception case where it goes through the catch block with some error?

DigitalLearningSolutions.Data/Services/TrackerActionService.cs
```diff
@@ -51,5 +67,34 @@ public TrackerActionService(ITutorialContentDataService tutorialContentDataServi
...
         }
+
+        public TrackerEndpointResponse StoreDiagnosticJson(int? progressId, string? diagnosticOutcome)
+        {
+            if (!progressId.HasValue || string.IsNullOrEmpty(diagnosticOutcome))
+            {
+                return TrackerEndpointResponse.StoreDiagnosticScoreException;
+            }
+
+            var json = JsonConvert.DeserializeObject<IEnumerable<JObject>>(diagnosticOutcome);
```
AlexJacksonDS, line 78: What happens if the json string is something invalid here?

AlexJacksonDS (changes requested):
Just a couple of questions about the error behaviour

DigitalLearningSolutions.Data.Tests/Services/TrackerActionServiceTests.cs
livzorn: added one!

DigitalLearningSolutions.Data/Services/TrackerActionService.cs
livzorn: It returns TrackerEndpointResponse.UnexpectedException, which is [#Result:-1]. Do you think it should be something else?

AlexJacksonDS: It depends on the interpretation of the final sentence of the ticket: If there is an exception then log it and return "[#Result:-25]" else return "[#Result:1]". I scoped the ticket originally so I'll go and have a look what should happen.

AlexJacksonDS (approved):

DigitalLearningSolutions.Data/Services/TrackerActionService.cs
```diff
@@ -51,5 +67,34 @@ public TrackerActionService(ITutorialContentDataService tutorialContentDataServi
...
+
+            try
+            {
+                foreach (var obj in json)
+                {
+                    var tutorialId = (int)obj["tutorialid"];
```
wilkinsm, line 84: Given we know what the data structure should be, if a model was defined for this such as `DiagnosticOutcome` containing the two relevant properties then the code could become:
```
var diagnosticOutcomes = JsonConvert.DeserializeObject<IEnumerable<DiagnosticOutcome>>(diagnosticOutcome);
try
{
  foreach (var diagnosticOutcome in diagnosticOutcomes)
  {
    progressService.UpdateDiagnosticScore(progressId.Value, diagnosticOutcome .TutorialId, diagnosticOutcome.MyScore);
    ....
```
We can then reduce the need for casting and using magic strings here to identify properties of the object.

DigitalLearningSolutions.Data/Services/TrackerActionService.cs
```diff
@@ -1,23 +1,39 @@
...
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Linq;
 
     public interface ITrackerActionService
     {
         TrackerObjectiveArray? GetObjectiveArray(int? customisationId, int? sectionId);
 
         TrackerObjectiveArrayCc? GetObjectiveArrayCc(int? customisationId, int? sectionId, bool? isPostLearning);
+
+        public TrackerEndpointResponse StoreDiagnosticJson(int? progressId, string? diagnosticOutcome);
```
wilkinsm, line 19: `public` access modifier not required here.

DigitalLearningSolutions.Data/Services/ProgressService.cs
```diff
@@ -12,6 +12,8 @@ public interface IProgressService
         void UpdateCompleteByDate(int progressId, DateTime? completeByDate);
 
         void UpdateCompletionDate(int progressId, DateTime? completionDate);
+
+        public void UpdateDiagnosticScore(int progressId, int tutorialId, int myScore);
```
wilkinsm, line 16: `public` access modifier not required here.

DigitalLearningSolutions.Data/Models/Tracker/TrackerEndpointQueryParams.cs
```diff
@@ -6,5 +6,9 @@ public class TrackerEndpointQueryParams
         public int? CustomisationId { get; set; }
         public int? SectionId { get; set; }
         public string? IsPostLearning { get; set; }
+
```
wilkinsm, line 9: Could we match the spacings up here to be the same as above.
The convention is usually no spacing line between properties.

DigitalLearningSolutions.Data/Services/TrackerActionService.cs
livzorn: oops removed :)

DigitalLearningSolutions.Data/Services/ProgressService.cs
livzorn: removed

DigitalLearningSolutions.Data/Models/Tracker/TrackerEndpointQueryParams.cs
livzorn: yep, done :)

DigitalLearningSolutions.Data/Services/TrackerActionService.cs
livzorn: Gave this a go. This means we'll treat the keys in the json as case insensitive, is that okay? I know @SteveJacksonSoft mentioned we want to keep tracker endpoints as close to the old behavior as possible.

I'm getting some issues when testing - sometimes entering an unexpected key still gets parsed to TutorialId or MyScore, sometimes I get Result:-25, and sometimes it just stalls out. Let me know if I missed something if or we should go back to the magic string way.

wilkinsm: **Regarding case sensitivity:**
In `.net core` the built in JSON deserialisation is case sensitive.
As we are using `JSON.NET` library, to achieve case sensitivity then you would need to add a custom converter.
@SteveJacksonSoft is their a requirement to be this strict? Out of the box the original lower-case keys would be accepted and therfore be compatible - it just additionally allows the keys to be camel/upper/etc case too.

**Regarding unexpected keys:**
In order to handle this you will need to change the default option by passing in `JsonSerializerSettings` as the second parameter to `DeserializeObject`, e.g.:
```
 var diagnosticOutcomes = JsonConvert.DeserializeObject<IEnumerable<DiagnosticOutcome>>(
  diagnosticOutcome,
  new JsonSerializerSettings { MissingMemberHandling = MissingMemberHandling.Error }
);
```
This will cause an `exception` which will be caught higher up and ensure a `Result:-25` response.
If implementing the above it will be useful to re-instante the test cases that focus on `unexpected keys` that were removed.

livzorn: Thanks, I've added the MissingMemberHandling to handle the unexpected key, and also added required attributes on the properties of DiagnosticOutcome so that an error will occur if one of the expected keys is missing

SteveJacksonSoft: I think it's fine if we're case-insensitive with the json in DiagnosticOutcome..

DigitalLearningSolutions.Data/Services/TrackerActionService.cs
```diff
@@ -74,22 +74,41 @@ public TrackerEndpointResponse StoreDiagnosticJson(int? progressId, string? diag
...
-            var diagnosticOutcomes = JsonConvert.DeserializeObject<IEnumerable<DiagnosticOutcome>>(diagnosticOutcome);
             try
             {
-                foreach (var diagOutcome in diagnosticOutcomes)
+                var diagnosticOutcomes = JsonConvert.DeserializeObject<IEnumerable<DiagnosticOutcome>>(
+                    diagnosticOutcome,
+                    new JsonSerializerSettings
+                    {
+                        MissingMemberHandling = MissingMemberHandling.Error,
+                        NullValueHandling = NullValueHandling.Include,
```
wilkinsm, line 84: `NullValueHandling.Include` and `DefaultValueHandling.Include` are already the default values for these settings so these can be dropped from here.

DigitalLearningSolutions.Data/Services/TrackerActionService.cs
```diff
@@ -74,22 +74,41 @@ public TrackerEndpointResponse StoreDiagnosticJson(int? progressId, string? diag
...
+                        DefaultValueHandling = DefaultValueHandling.Include,
+                    }
+                );
+
+                try
```
wilkinsm, line 89: I'm not certain we need the double `try-catch` as the calling method in the `TrackerService` will catch the json exception. 

But I get the reasoning though - to ensure that consistent output from this method.

DigitalLearningSolutions.Data.Tests/Services/TrackerActionServiceTests.cs
```diff
@@ -181,11 +181,17 @@ public void StoreDiagnosticJson_returns_success_response_if_successful()
 
         [Test]
```
wilkinsm, line 182: `[Test]` is not strictly required when you also have `[TestCase(...)]` defined.

DigitalLearningSolutions.Data.Tests/Services/TrackerActionServiceTests.cs
```diff
@@ -136,5 +142,69 @@ public void GetObjectiveArrayCc_returns_empty_object_json_if_no_results_found()
...
+                () => progressService.UpdateDiagnosticScore(
+                    1,
+                    425,
+                    4
+                )
+            ).MustHaveHappenedOnceExactly();
+        }
+
+        [Test]
+        [TestCase(null, "[{'tutorialid':425,'myscore':4},{'tutorialid':424,'myscore':3}]")]
```
wilkinsm, line 183: Line 183 and 184 appear to be duplicates.

DigitalLearningSolutions.Data.Tests/Services/TrackerActionServiceTests.cs
```diff
@@ -136,5 +142,69 @@ public void GetObjectiveArrayCc_returns_empty_object_json_if_no_results_found()
...
+        [TestCase(1, null)]
```
wilkinsm, line 184: This is a duplicate from below - looks like a merge conflict.

DigitalLearningSolutions.Data/Services/TrackerActionService.cs
wilkinsm: thanks @SteveJacksonSoft

DigitalLearningSolutions.Data.Tests/Services/TrackerActionServiceTests.cs
livzorn: removed :)

DigitalLearningSolutions.Data.Tests/Services/TrackerActionServiceTests.cs
livzorn: ah cool, removed

DigitalLearningSolutions.Data/Services/TrackerActionService.cs
livzorn: Yeah I wasn't sure if it mattered that the logged message in the catch statement wouldn't really reflect an issue going wrong in the deserialization. I've put it all in one try catch now.

DigitalLearningSolutions.Data/Services/TrackerActionService.cs
livzorn: oops, was playing around with some other settings and forgot to remove them

DigitalLearningSolutions.Data/Services/TrackerActionService.cs
wilkinsm: Yeah that makes sense. Given this try-catch does capture the exception then we can understand from logs why a failure may of occured.

DigitalLearningSolutions.Data/DataServices/ProgressDataService.cs
```diff
@@ -185,6 +192,32 @@ public void SetCompletionDate(int progressId, DateTime? completionDate)
...
+                new { progressId, tutorialId, myScore }
+            );
+
+            if (numberOfAffectedRows < 1)
+            {
+                logger.LogWarning(
+                    "Not updating diagnostic score as db update failed. " +
+                    $"Progress id: {progressId}, tutorial Id: {tutorialId}, myScore: {myScore}"
+                );
+                throw new Exception();
```
wilkinsm, line 217: I missed this. I think it would be good to include a message in the exception. From a logging perspective it will be a bit clearer as to why we have a failure.

DigitalLearningSolutions.Data/Services/TrackerActionService.cs
```diff
@@ -51,5 +66,47 @@ public TrackerActionService(ITutorialContentDataService tutorialContentDataServi
...
+                    {
+                        MissingMemberHandling = MissingMemberHandling.Error,
+                    }
+                );
+
+                foreach (var diagOutcome in diagnosticOutcomes)
+                {
+                    if (diagOutcome.TutorialId == 0)
+                    {
+                        throw new Exception();
```
wilkinsm, line 91: I think it would be good to include a message in the exception. From a logging perspective it will be a bit clearer as to why we have a failure.

wilkinsm (approved):
Nice one Olivia, looks ok to merge.

PR #852, opened 2022-01-11
HEEDLS-712 - Modify the PK on CompetencyAssessmentQuestionRoleRequire…

DanBloxham-sw:
…ments to be a compound key

### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-712

### Description
The CompetencyAssessmentQuestionRoleRequirements table should only have one record for each SelfAssessmentID/CompetencyID combination, but the table did not actually enforce this. As advised by Kevin, I have removed the previous PK on the ID column, and the column itself, and replaced it with a compound PK on SelfAssessmentID and CompetencyID.

I have also made code changes where the ID was used and confirmed the method in question functions appropriately.

### Screenshots
N/A

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data.Migrations/202201111021_ChangeCompetencyAssessmentQuestionRoleRequirementsPrimaryKey.cs
+++ b/DigitalLearningSolutions.Data.Migrations/202201111021_ChangeCompetencyAssessmentQuestionRoleRequirementsPrimaryKey.cs
@@ -18,8 +18,9 @@
         {
             Delete.PrimaryKey("PK_CompetencyAssessmentQuestionRoleRequirements")
                 .FromTable("CompetencyAssessmentQuestionRoleRequirements");
-            Alter.Table("CompetencyAssessmentQuestionRoleRequirements").AddColumn("ID").AsInt32().NotNullable()
-                .PrimaryKey().Identity();
+            Alter.Table("CompetencyAssessmentQuestionRoleRequirements").AddColumn("ID").AsInt32().NotNullable().Identity();
+            Create.PrimaryKey("PK_CompetencyAssessmentQuestionRoleRequirements")
+                .OnTable("CompetencyAssessmentQuestionRoleRequirements").Column("ID");
         }
     }
 }
--- a/DigitalLearningSolutions.Data.Migrations/202201111021_ChangeCompetencyAssessmentQuestionRoleRequirementsPrimaryKey.cs
+++ b/DigitalLearningSolutions.Data.Migrations/202201111021_ChangeCompetencyAssessmentQuestionRoleRequirementsPrimaryKey.cs
@@ -1,0 +1,25 @@
+﻿namespace DigitalLearningSolutions.Data.Migrations
+{
+    using FluentMigrator;
+
+    [Migration(202201111021)]
+    public class ChangeCompetencyAssessmentQuestionRoleRequirementsPrimaryKey: Migration
+    {
+        public override void Up()
+        {
+            Delete.PrimaryKey("PK_CompetencyAssessmentQuestionRoleRequirements")
+                .FromTable("CompetencyAssessmentQuestionRoleRequirements");
+            Delete.Column("ID").FromTable("CompetencyAssessmentQuestionRoleRequirements");
+            Create.PrimaryKey("PK_CompetencyAssessmentQuestionRoleRequirements")
+                .OnTable("CompetencyAssessmentQuestionRoleRequirements").Columns("SelfAssessmentID", "CompetencyID");
+        }
+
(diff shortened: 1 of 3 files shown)


AlexJacksonDS (approved):

DigitalLearningSolutions.Data.Migrations/202201111021_ChangeCompetencyAssessmentQuestionRoleRequirementsPrimaryKey.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace DigitalLearningSolutions.Data.Migrations
+{
+    using FluentMigrator;
+
+    [Migration(202201111021)]
+    public class ChangeCompetencyAssessmentQuestionRoleRequirementsPrimaryKey: Migration
```
DanBloxham-sw, line 6: This was the original migration made when we wanted to replace the ID column with a compound PK on the two necessary columns. As it was pushed into an MR, and thus run on Jenkins, prior to the decision to just use a unique constraint instead, this migration has been left, and the other migration tackles it.

DigitalLearningSolutions.Data.Migrations/202201120821_ChangeUniqueConstraintsOnCompetencyAssessmentQuestionRoleRequirements.cs
```diff
@@ -0,0 +1,30 @@
...
+    [Migration(202201120821)]
+    public class ChangeUniqueConstraintsOnCompetencyAssessmentQuestionRoleRequirements: Migration
+    {
+        public override void Up()
+        {
+            Delete.PrimaryKey("PK_CompetencyAssessmentQuestionRoleRequirements")
+                .FromTable("CompetencyAssessmentQuestionRoleRequirements");
+            Alter.Table("CompetencyAssessmentQuestionRoleRequirements").AddColumn("ID").AsInt32().NotNullable().Identity();
+            Create.PrimaryKey("PK_CompetencyAssessmentQuestionRoleRequirements")
+                .OnTable("CompetencyAssessmentQuestionRoleRequirements").Column("ID");
```
DanBloxham-sw, line 14: You might notice that this is not identical to the Down() method in the other migration. I only noticed when creating this migration that adding a column with `.PrimaryKey()` when altering a table (rather than creating) doesn't actually make the column a primary key. This seems to be due to a subtlety in FluentMigrator that should probably be more obvious but isn't, see here for details: https://github.com/fluentmigrator/fluentmigrator/issues/96

DigitalLearningSolutions.Data.Migrations/202201111021_ChangeCompetencyAssessmentQuestionRoleRequirementsPrimaryKey.cs
```diff
@@ -0,0 +1,25 @@
...
+            Create.PrimaryKey("PK_CompetencyAssessmentQuestionRoleRequirements")
+                .OnTable("CompetencyAssessmentQuestionRoleRequirements").Columns("SelfAssessmentID", "CompetencyID");
+        }
+
+        public override void Down()
+        {
+            Delete.PrimaryKey("PK_CompetencyAssessmentQuestionRoleRequirements")
+                .FromTable("CompetencyAssessmentQuestionRoleRequirements");
+            Alter.Table("CompetencyAssessmentQuestionRoleRequirements").AddColumn("ID").AsInt32().NotNullable()
+                .PrimaryKey().Identity();
```
SteveJacksonSoft, line 22: Given your comment on the next migration, should this Down() have `Create.PrimaryKey...`?

DigitalLearningSolutions.Data.Migrations/202201111021_ChangeCompetencyAssessmentQuestionRoleRequirementsPrimaryKey.cs
SteveJacksonSoft: Ah, missed this comment!

DigitalLearningSolutions.Data.Migrations/202201111021_ChangeCompetencyAssessmentQuestionRoleRequirementsPrimaryKey.cs
DanBloxham-sw: Good point. When I made the other migration, I thought it would be good to leave this one as is for some reason. But now I think it would break anytime someone wanted to migrate down beyond this. I've updated it to include the correct primary key addition.

PR #853, opened 2022-01-11
DLSV2-471 Migration to add OriginalRating field

kevwhitt-hee:
### JIRA link
[DLSV2-471](https://hee-dls.atlassian.net/browse/DLSV2-471)

### Description
Migration to add OriginalRating field to LearningResourceReferences table. This will be populated when creating new LearningResourceReferences records and used as a fallback value if the Learning Bub API is unavailable.

### DIFF
--- a/DigitalLearningSolutions.Data.Migrations/202201111458_LearningResourceReferencesOriginalRatingField.cs
+++ b/DigitalLearningSolutions.Data.Migrations/202201111458_LearningResourceReferencesOriginalRatingField.cs
@@ -11,7 +11,7 @@
         }
         public override void Down()
         {
-            Delete.Column("OriginalRating").FromTable("OriginalRating");
+            Delete.Column("OriginalRating").FromTable("LearningResourceReferences");
         }
     }
 }
--- a/DigitalLearningSolutions.Data.Migrations/202201111458_LearningResourceReferencesOriginalRatingField.cs
+++ b/DigitalLearningSolutions.Data.Migrations/202201111458_LearningResourceReferencesOriginalRatingField.cs
@@ -7,7 +7,7 @@
         public override void Up()
         {
             Alter.Table("LearningResourceReferences")
-                .AddColumn("OriginalRating").AsDecimal(2, 1).Nullable();
+                .AddColumn("OriginalRating").AsDecimal(2, 1).NotNullable().WithDefaultValue(0);
         }
         public override void Down()
         {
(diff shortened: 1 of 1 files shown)


DigitalLearningSolutions.Data.Migrations/202201111458_LearningResourceReferencesOriginalRatingField.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace DigitalLearningSolutions.Data.Migrations
+{
+    using FluentMigrator;
+    [Migration(202201111458)]
+    public class LearningResourceReferencesOriginalRatingField : Migration
+    {
+        public override void Up()
+        {
+            Alter.Table("LearningResourceReferences")
+                .AddColumn("OriginalRating").AsDecimal(2, 1).Nullable();
```
SteveJacksonSoft, line 10: The open API will always return a non-null value for this, so I'd expect this column to be non-null.
I missed this for the previous migration, but that's also true for the other columns.

Are there reasons you would prefer the columns to be nullable?

kevwhitt-hee: Laziness really. It is simply to avoid having to delete existing records in all of the related tables in dev database (because we are actively working with that data). Will nullable fields cause problems?

SteveJacksonSoft: I think with decimals having a non-null column would make a significant difference to the code quality, as it would remove unexplained assumptions that rating will be non-null. You could use `WithDefaultValue()` to set the default value as 0 in the migration, which would save you deleting entries, and would also get you safer test data.
I'm a bit less concerned about string fields because we're going to treat whitespace strings as if they were null anyway.

kevwhitt-hee: Good idea. Updated.

DigitalLearningSolutions.Data.Migrations/202201111458_LearningResourceReferencesOriginalRatingField.cs
```diff
@@ -0,0 +1,17 @@
...
+    public class LearningResourceReferencesOriginalRatingField : Migration
+    {
+        public override void Up()
+        {
+            Alter.Table("LearningResourceReferences")
+                .AddColumn("OriginalRating").AsDecimal(2, 1).NotNullable().WithDefaultValue(0);
+        }
+        public override void Down()
+        {
+            Delete.Column("OriginalRating").FromTable("OriginalRating");
```
SteveJacksonSoft, line 14: Ah, and the table's called LearningResourceReferences. Sorry I didn't spot that earlier.

kevwhitt-hee: Oh dear. Rush job, sorry. Fixed, now.

SteveJacksonSoft (approved):
Looks good now!

PR #855, opened 2022-01-12
Heedls 711 filter results outside range

AlexJacksonDS:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-711

### Description
Filtering of resources where the delegate answer falls outside the range of results specified in CompetencyResourceAssessmentQuestionParameters. Refactored and added new unit tests to accommodate these changes.

### Screenshots
No visual changes

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Extensions/EnumerableExtensionsTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Extensions/EnumerableExtensionsTests.cs
@@ -12,12 +12,12 @@
         {
             // Given
             var list = new List<string?> { "1", "2", null, "3", null, "4" };
-            var expectedResult = new List<string?> { "1", "2", "3", "4" };
 
             // When
             var result = list.WhereNotNull();
 
             // Then
+            var expectedResult = new List<string> { "1", "2", "3", "4" };
             result.Should().BeEquivalentTo(expectedResult);
         }
     }
--- a/DigitalLearningSolutions.Data.Tests/Extensions/EnumerableExtensionsTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Extensions/EnumerableExtensionsTests.cs
@@ -17,8 +17,7 @@
             var result = list.WhereNotNull();
 
             // Then
-            var expectedResult = new List<string> { "1", "2", "3", "4" };
-            result.Should().BeEquivalentTo(expectedResult);
+            result.Should().BeEquivalentTo(new List<string> { "1", "2", "3", "4" });
         }
     }
 }
--- a/DigitalLearningSolutions.Data.Tests/Extensions/EnumerableExtensionsTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Extensions/EnumerableExtensionsTests.cs
@@ -1,0 +1,24 @@
+﻿namespace DigitalLearningSolutions.Data.Tests.Extensions
+{
+    using System.Collections.Generic;
+    using DigitalLearningSolutions.Data.Extensions;
+    using FluentAssertions;
+    using NUnit.Framework;
+
+    public class EnumerableExtensionsTests
+    {
+        [Test]
+        public void WhereNotNull_filters_items_correctly()
+        {
+            // Given
+            var list = new List<string?> { "1", "2", null, "3", null, "4" };
+            var expectedResult = new List<string?> { "1", "2", "3", "4" };
+
+            // When
+            var result = list.WhereNotNull();
+
+            // Then
+            result.Should().BeEquivalentTo(expectedResult);
+        }
+    }
+}
--- a/DigitalLearningSolutions.Data.Tests/Services/RecommendedLearningServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/RecommendedLearningServiceTests.cs
@@ -20,11 +20,13 @@
     {
         private const int SelfAssessmentId = 1;
         private const int CompetencyId = 2;
+        private const int SecondCompetencyId = 3;
         private const int LearningResourceReferenceId = 3;
         private const int LearningHubResourceReferenceId = 4;
         private const int DelegateId = 5;
         private const int LearningLogId = 6;
         private const int CompetencyLearningResourceId = 1;
+        private const int SecondCompetencyLearningResourceId = 2;
         private const int CompetencyAssessmentQuestionId = 1;
         private const int RelevanceAssessmentQuestionId = 2;
         private const string ResourceName = "Resource";
@@ -62,7 +64,7 @@
             // Given
             GivenResourceForSelfAssessmentIsReturnedByLearningHubApi();
             GivenQuestionParametersAreReturned(true, true, 1, 10);
-            GivenSelfAssessmentHasResults(5, 5);
+            GivenSelfAssessmentHasResultsForFirstCompetency(5, 5);
 
             A.CallTo(() => learningLogItemsDataService.GetLearningLogItems(DelegateId))
                 .Returns(new List<LearningLogItem>());
@@ -86,7 +88,7 @@
             // Given
             GivenResourceForSelfAssessmentIsReturnedByLearningHubApi();
             GivenQuestionParametersAreReturned(true, true, 1, 10);
-            GivenSelfAssessmentHasResults(5, 5);
+            GivenSelfAssessmentHasResultsForFirstCompetency(5, 5);
 
             var learningLogItems = Builder<LearningLogItem>.CreateListOfSize(5).All()
                 .With(i => i.LearningHubResourceReferenceId = LearningHubResourceReferenceId + 1)
@@ -113,7 +115,7 @@
             // Given
             GivenResourceForSelfAssessmentIsReturnedByLearningHubApi();
             GivenQuestionParametersAreReturned(true, true, 1, 10);
-            GivenSelfAssessmentHasResults(5, 5);
+            GivenSelfAssessmentHasResultsForFirstCompetency(5, 5);
 
             var learningLogItem = Builder<LearningLogItem>.CreateNew()
                 .With(i => i.LearningHubResourceReferenceId = LearningHubResourceReferenceId)
@@ -142,7 +144,7 @@
             // Given
             GivenResourceForSelfAssessmentIsReturnedByLearningHubApi();
             GivenQuestionParametersAreReturned(true, true, 1, 10);
-            GivenSelfAssessmentHasResults(5, 5);
+            GivenSelfAssessmentHasResultsForFirstCompetency(5, 5);
 
             var learningLogItem = Builder<LearningLogItem>.CreateNew()
                 .With(i => i.LearningHubResourceReferenceId = LearningHubResourceReferenceId)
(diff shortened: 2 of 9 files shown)


DigitalLearningSolutions.Data/Services/RecommendedLearningService.cs
```diff
@@ -67,40 +67,111 @@ int delegateId
...
 
+                var latestConfidenceResult = delegateResults
+                    .Where(
+                        dr => dr.AssessmentQuestionId ==
+                              competencyResourceAssessmentQuestionParameter.AssessmentQuestionId
+                    )
+                    .OrderByDescending(dr => dr.DateTime).FirstOrDefault();
+
+                if (competencyResourceAssessmentQuestionParameter.MinResultMatch <= latestConfidenceResult?.Result &&
+                    latestConfidenceResult?.Result <= competencyResourceAssessmentQuestionParameter.MaxResultMatch)
```
DanBloxham-sw, line 158: The second null conditional operator `?` on latestConfidenceResult is unnecessary as it should be tackled by the first.

DigitalLearningSolutions.Data/Services/RecommendedLearningService.cs
```diff
@@ -67,40 +67,111 @@ int delegateId
...
+                    .GetSelfAssessmentResultsForDelegateSelfAssessmentCompetency(
+                        delegateId,
+                        selfAssessmentId,
+                        competencyLearningResource.CompetencyId
+                    ).ToList();
+
+                var competencyResourceAssessmentQuestionParameter =
+                    competencyResourceAssessmentQuestionParameters.Single(
+                        qp => qp.CompetencyLearningResourceId == competencyLearningResource.Id
+                    );
```
DanBloxham-sw, line 148: The database enforces a one-to-one-or-zero relationship between CompetencyResourceAssessmentQuestionParameters and CompetencyLearningResource so the `.Single()` here should be `SingleOrDefault()`.
In the case of null being returned, then no record exists, and thus the linked competency does not have a range specified. As such, we can return true, as we have at least one linked competency with no range specified.

Also minor, but the record (if found) will technically contain multiple "parameters" so we might need a better name for the singular record variable. I can't think of anything better than maybe "CompetencyParametersForAssessmentQuestion", so feel free to improvise.

DigitalLearningSolutions.Data.Tests/Services/RecommendedLearningServiceTests.cs
```diff
@@ -491,7 +527,33 @@ private void GivenResourceForSelfAssessmentIsReturnedByLearningHubApi(decimal ra
...
+                .With(clr => clr.Id = 2)
+                .And(clr => clr.CompetencyId = 3)
+                .Build();
 
+            A.CallTo(
+                () => competencyLearningResourcesDataService.GetCompetencyLearningResourcesByCompetencyId(CompetencyId)
+            ).Returns(competencyLearningResources);
+        }
+
+        private void GivenLearningUpApiReturnsResources(decimal rating)
```
DanBloxham-sw, line 555: Typo of _LearningHub_ -> _LearningUp_

DigitalLearningSolutions.Data.Tests/Services/RecommendedLearningServiceTests.cs
```diff
@@ -491,7 +527,33 @@ private void GivenResourceForSelfAssessmentIsReturnedByLearningHubApi(decimal ra
...
+                .All()
+                .With(clr => clr.LearningResourceReferenceId = LearningResourceReferenceId)
+                .And(clr => clr.LearningHubResourceReferenceId = LearningHubResourceReferenceId)
+                .And(clr => clr.AdminId = 7)
+                .TheFirst(1)
+                .With(clr => clr.Id = CompetencyLearningResourceId)
+                .And(clr => clr.CompetencyId = CompetencyId)
+                .TheRest()
+                .With(clr => clr.Id = 2)
+                .And(clr => clr.CompetencyId = 3)
```
DanBloxham-sw, line 547: Seeing how we have the first Clr.Id and clr.CompetencyId defined as consts, it would be good to define the second ones as consts as well to make the tests that use it a bit clearer.

DigitalLearningSolutions.Data.Tests/Services/RecommendedLearningServiceTests.cs
```diff
@@ -474,7 +434,83 @@ public async Task
             ).MustNotHaveHappened();
         }
 
+        [Test]
+        public async Task
+            GetRecommendedLearningForSelfAssessment_does_return_resources_with_some_unanswered_optional_competencies()
```
DanBloxham-sw, line 439: Which part of this test defines the question as optional?

DigitalLearningSolutions.Data.Tests/Services/RecommendedLearningServiceTests.cs
```diff
@@ -307,10 +295,10 @@ public async Task
...
-        [TestCase(0, 0)]
-        [TestCase(1, 4)]
-        [TestCase(2.4, 9.6)]
-        [TestCase(5, 20)]
+        [TestCase(0, 175)]
+        [TestCase(1, 179)]
+        [TestCase(2.4, 184.6)]
+        [TestCase(5, 195)]
         public async Task
             GetRecommendedLearningForSelfAssessment_returns_correct_recommendation_score_for_resource_with_learning_hub_ratings_only(
```
DanBloxham-sw, line 303: This test title mentions "...correct score for {factor} only" but the expected scores seem to be higher than just the individual factor's contribution, so might be worth renaming.

DigitalLearningSolutions.Data.Tests/Services/RecommendedLearningServiceTests.cs
```diff
@@ -235,32 +227,19 @@ public async Task GetRecommendedLearningForSelfAssessment_calls_learning_hub_api
...
         [Test]
-        [TestCase(1, true, 100)]
-        [TestCase(1, false, 30)]
-        [TestCase(0, true, 0)]
+        [TestCase(true, 175)]
+        [TestCase(false, 105)]
         public async Task
-            GetRecommendedLearningForSelfAssessment_returns_correct_recommendation_score_for_resource_with_essential_question_parameters_only(
-                int numberOfQuestionParameters,
+            GetRecommendedLearningForSelfAssessment_returns_correct_recommendation_score_for_resource_with_essential_question_parameters(
```
DanBloxham-sw, line 233: The test case where there is no CompetencyResourceAssessmentQuestionParameters record seems to have vanished (probably tied to the other comment about the `.Single()` vs `.SingleOrDefault()`)

DanBloxham-sw (changes requested):
Looks just about how I was expecting, though I've left a few comments to tackle.

DigitalLearningSolutions.Data/Services/RecommendedLearningService.cs
AlexJacksonDS: Changed the Single and added the null behaviour.
I'm not sure I fully agree on the name change but have changed it to competencyResourceAssessmentQuestionParametersForClr.
Plural nouns for a single table record isn't really great as it leaves you unable to make it singular to signify a single record

DigitalLearningSolutions.Data.Tests/Services/RecommendedLearningServiceTests.cs
AlexJacksonDS: There is no definition that is is optional, we only care that it has no answers. (it should not be possible to reach here with required competencies with no answers)
I've updated the name of the results setup to GivenSelfAssessmentHasResultsForFirstCompetency

DigitalLearningSolutions.Data.Tests/Services/RecommendedLearningServiceTests.cs
AlexJacksonDS: This was a consequence of me not realising the relationship was 1-1-0 (The Single comment above)
We now have 2 basic situations, either we can only calc the score from the LH rating, or we can have the LH rating + attempt both other values (req adj may be 0).
Restored these back to just the LH calculation, dropping the QuestionParameters setup.

DigitalLearningSolutions.Data.Tests/Services/RecommendedLearningServiceTests.cs
AlexJacksonDS: See also here https://github.com/TechnologyEnhancedLearning/DLSV2/pull/855#discussion_r783277920, the no record case doesn't really need other tests

DigitalLearningSolutions.Data.Tests/Services/RecommendedLearningServiceTests.cs
DanBloxham-sw: Ah I see, the case where no QuestionParameters are set up is handled here. I think given that that case is specifically considered in the logic for determining what to hide, it might be good to specify it in the title here then. Just will make it easier to see that it has indeed be tested from a quickly glance. 
Maybe something like "...for_resource_only_with_learning_hub_rating_and_no_question_parameters"?

DanBloxham-sw (approved):
Nice, think this covers everything now.

DigitalLearningSolutions.Data/Services/RecommendedLearningService.cs
```diff
@@ -67,40 +67,119 @@ int delegateId
...
+                    delegateId,
+                    resourceReferences[rr.RefId],
+                    delegateLearningLogItems,
+                    rr,
+                    competencyLearningResources
+                )
             );
 
-            return recommendedResources;
+            return recommendedResources.Where(r => r != null).Select(r => r!);
```
SteveJacksonSoft, line 80: Just a thought, this would be a great extension method if you pulled it out, `WhereNotNull()` or something.

livzorn (approved):
Looks good!

DigitalLearningSolutions.Data.Tests/Extensions/EnumerableExtensionsTests.cs
```diff
@@ -0,0 +1,24 @@
...
+    using NUnit.Framework;
+
+    public class EnumerableExtensionsTests
+    {
+        [Test]
+        public void WhereNotNull_filters_items_correctly()
+        {
+            // Given
+            var list = new List<string?> { "1", "2", null, "3", null, "4" };
+            var expectedResult = new List<string?> { "1", "2", "3", "4" };
```
SteveJacksonSoft, line 15: Does the nullability of the type make a difference? I would expect a `List<string>`.
Also, I've a bit of a preference for inlining expected output in the `Then` section - that's where I'd habitually look for the expected output, because often most of the literals in the `Given` section are input data or setting up mocks.

AlexJacksonDS: Ah whoops, this was a copy and paste error, it returns `IEnumerable<T>` from input `IEnumerable<T?>`

DigitalLearningSolutions.Data.Tests/Extensions/EnumerableExtensionsTests.cs
```diff
@@ -12,12 +12,12 @@ public void WhereNotNull_filters_items_correctly()
...
             // Given
             var list = new List<string?> { "1", "2", null, "3", null, "4" };
-            var expectedResult = new List<string?> { "1", "2", "3", "4" };
 
             // When
             var result = list.WhereNotNull();
 
             // Then
+            var expectedResult = new List<string> { "1", "2", "3", "4" };
             result.Should().BeEquivalentTo(expectedResult);
```
SteveJacksonSoft, line 21: May as well inline `expectedResult`, but no need for re-review from me.

PR #857, opened 2022-01-13
Heedls 691 search all delegates candidate number

wilkinsm:
### JIRA link
[https://softwiretech.atlassian.net/browse/HEEDLS-691](https://softwiretech.atlassian.net/browse/HEEDLS-691)

### Description
Add the ability to search all delegates by their candidate number as well as their name.
This adds the ability to search across multiple properties on a model.
A slight refactor of the first version of this concept which affects support FAQs.

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/ViewModels/Support/SearchableFaqTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/ViewModels/Support/SearchableFaqTests.cs
@@ -17,8 +17,7 @@
             var searchableFaq = new SearchableFaq(faq);
 
             // Then
-            const string expectedSearchableFaqAnswer = " p Helpful content  p ";
-            searchableFaq.SearchableFaqAnswer.Should().Be(expectedSearchableFaqAnswer);
+            searchableFaq.SearchableFaqAnswer.Should().Be(" p Helpful content  p ");
         }
     }
 }
--- a/DigitalLearningSolutions.Web.Tests/ViewModels/Support/SearchableFaqTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/ViewModels/Support/SearchableFaqTests.cs
@@ -5,21 +5,20 @@
     using FluentAssertions;
     using NUnit.Framework;
 
-    public class FaqViewModelTests
+    public class SearchableFaqTests
     {
         [Test]
-        public void FaqViewModelTests_returns_model_with_cleaned_searchName()
+        public void SearchableFaq_returns_model_with_cleaned_SearchableFaqAnswer()
         {
             // Given
             var faq = FaqTestHelper.GetDefaultFaq();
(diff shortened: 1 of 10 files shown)


DigitalLearningSolutions.Web/Helpers/GenericSearchHelper.cs
```diff
@@ -73,7 +73,7 @@ public class GenericSearchHelper
             var results = Process.ExtractAll(
                 (T)query,
                 items,
-                item => item.SearchableName.ToLower(),
+                item => string.Join(" ", item.SearchableContent.Where(s => s != null)).ToLower(),
```
JBsoftwire, line 76: Can this produce a situation where the search 'finds' a string that's actually composited from multiple separate content elements? I don't think it can because I think the search is on single words only not phrases-including-spaces, but I have no experience with the search system.

JBsoftwire (approved):
Just one question about the search, but I don't think it's actually an issue.

DigitalLearningSolutions.Web/Helpers/GenericSearchHelper.cs
wilkinsm: Exactly how search behaves depends on the "scorer" used. 
In this context partial matches can be found as that is how it is configured.
If you look back at the tests in `AllDelegatesSearchTests`. The test cases, although basic, pass in multiple values as a search string and will find a match based on partial matches across multiple properties of the model.
This is expected and wanted behaviour.
So for example we have:
```
var delegate = new DelegateUser 
{
FirstName = "Emily",
LastName = "Harper",
CandidateNumber = "EH14"
};
```

We pass in the search string of `Haper H14` and it will find this delegate.

DigitalLearningSolutions.Web.Tests/ViewModels/Support/SearchableFaqTests.cs
```diff
@@ -5,21 +5,20 @@
...
+    public class SearchableFaqTests
     {
         [Test]
-        public void FaqViewModelTests_returns_model_with_cleaned_searchName()
+        public void SearchableFaq_returns_model_with_cleaned_SearchableFaqAnswer()
         {
             // Given
             var faq = FaqTestHelper.GetDefaultFaq();
-            var expectedSearchableName = "A common question?  p Helpful content  p ";
+            var expectedSearchableFaqAnswer = " p Helpful content  p ";
```
SteveJacksonSoft, line 15: Sorry, this is one of those "it was just a refactor!" cases, but this test could be improved by specifying the answer in line 14 and inlining the expected answer in `Then`. No need for re-review.

wilkinsm: done.

PR #863, opened 2022-01-17
Heedls 644 super admin FAQs basic page

rhimcg:
### JIRA link
[_HEEDLS-644_](https://softwiretech.atlassian.net/jira/software/c/projects/HEEDLS/boards/671?modal=detail&selectedIssue=HEEDLS-644&assignee=6065c334b30f0d00704264d4)

### Description
Add the basic page for super admin to view the 10 most recent FAQs

### Screenshots
image
image
image
image
image
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)
- [] Scanned over my own MR to ensure everything is as expected.


DigitalLearningSolutions.Data.Tests/Services/FaqsServiceTests.cs
```diff
@@ -154,5 +155,38 @@ public void GetPublishedFaqsForTargetGroup_returns_empty_list_when_data_service_
                     .MustHaveHappenedOnceExactly();
             }
         }
+
+        [Test]
+        public void GetAllFaqs_calls_data_service_method_and_returns_expected_results()
+        {
+            //Given
+            var expectedFaqs = Builder<Faq>.CreateListOfSize(10)
```
wilkinsm, line 163: Given that this service method is a wrapper service only we would not expect the data from the data service to be altered in any way.
With that it would be worth checking that these `expectedFaqs` equal `results`.

DigitalLearningSolutions.Data/DataServices/FaqsDataService.cs
```diff
@@ -51,5 +53,10 @@ public IEnumerable<Faq> GetPublishedFaqsForTargetGroup(int targetGroup)
                 new { targetGroup }
             );
         }
+
+        public IEnumerable<Faq> GetAllFaqs()
```
wilkinsm, line 57: I originally missed a trick here when I wrote the `GetPublishedFaqsForTargetGroup` method in the DataServices.

I would be tempted to line up with the coding conventions that are now outlined in the swiki: "DataServices to be as generic and reusable as possible. Let the services do the logic".

In this case it would mean:
- Dropping the method `GetPublishedFaqsForTargetGroup`
- Then in the service layer the corresponding method would change to use your new method of `GetAllFaqs` and do the `Publish` and `TargetGroup` filtering.

Some may argue this results in an application taking more than it needs from the DB, which is a valid point. 
But in this instance the `FAQ` dataset is relatively small so the overhead is rather negligable.

DigitalLearningSolutions.Web.Tests/Controllers/SuperAdmin/FaqControllerTests.cs
```diff
@@ -0,0 +1,93 @@
...
+    using FakeItEasy;
+    using FluentAssertions;
+    using FluentAssertions.AspNetCore.Mvc;
+    using NUnit.Framework;
+
+    public class FaqControllerTests
+    {
+        private readonly IEnumerable<Faq> faqs = new List<Faq>
+        {
+            FaqTestHelper.GetDefaultFaq(1, weighting: 40),
```
wilkinsm, line 19: Here I would be tempted to define these in the setup and use soemthing more like `Builder<Faq>.CreateListOfSize(10).Build();` rather than the `FaqTestHelper`.

DigitalLearningSolutions.Web.Tests/Controllers/SuperAdmin/FaqControllerTests.cs
```diff
@@ -0,0 +1,93 @@
...
+            faqService = A.Fake<IFaqsService>();
+            A.CallTo(() => faqService.GetAllFaqs())
+                .Returns(faqs);
+        }
+
+        [Test]
+        public void Faqs_page_should_return_expected_Faqs_view_page()
+        {
+            //Given
+            var controller = new FaqsController(faqService);
```
wilkinsm, line 47: The `controller` could be a private property on the class and initialised on the setup. It would mean you don't have to write this for every test.

DigitalLearningSolutions.Web.Tests/Controllers/SuperAdmin/FaqControllerTests.cs
```diff
@@ -0,0 +1,93 @@
...
+            A.CallTo(() => faqService.GetAllFaqs())
+                .Returns(faqs.Take(2));
+
+            //When
+            var results = controller.Index();
+
+            //Then
+            results.Should().BeViewResult().WithViewName("SuperAdminFaqs")
```
wilkinsm, line 68: I'm not certain on the purpose of this test.
But given the controller orders by `CreatedDate` then we can determine which faq to expect `first` and compare that to an expected model. We can then reduce these two blocks down to one and write:
```
results.Should().BeViewResult().WithViewName("SuperAdminFaqs")
  .Model.Should().BeOfType<FaqsPageViewModel>()
  .Subject.Faqs.First().Should().Match<SearchableFaqViewModel>(f => f.FaqId == 10 && f.Weighting == 10);
```
Additionally you could drop the faqService setup you have on line 61.

DigitalLearningSolutions.Web/ViewModels/SuperAdmin/Faqs/FaqModel.cs
```diff
@@ -0,0 +1,31 @@
+namespace DigitalLearningSolutions.Web.ViewModels.SuperAdmin.Faqs
```
wilkinsm, line 1: This model doesn't look to be used.

DigitalLearningSolutions.Web/ViewModels/SuperAdmin/Faqs/FaqsPageViewModel.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace DigitalLearningSolutions.Web.ViewModels.SuperAdmin.Faqs
+{
+    using System.Collections.Generic;
+    using System.Linq;
+    using DigitalLearningSolutions.Web.Models.Enums;
+
+    public class FaqsPageViewModel
+    {
+        public FaqsPageViewModel(IEnumerable<SearchableFaqModel> faqs, SuperAdminSystemPage currentPage)
```
wilkinsm, line 9: Given I don't think the `CurrentPage` value will change can this be set directly on the property rather than passed into the model?

DigitalLearningSolutions.Web/ViewModels/SuperAdmin/Faqs/SearchableFaqModel.cs
```diff
@@ -0,0 +1,40 @@
+﻿namespace DigitalLearningSolutions.Web.ViewModels.SuperAdmin.Faqs
```
wilkinsm, line 1: This looks to be a duplicate of `SearchableFaq` used in support. 
Can we drop one and share the model across both?

DigitalLearningSolutions.Web/ViewModels/SuperAdmin/Faqs/SearchableFaqViewModel.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        public SearchableFaqViewModel(SearchableFaqModel faq)
+        {
+            FaqId = faq.FaqId;
+            AHtml = faq.AHtml;
+            CreatedDate = faq.CreatedDate.ToShortDateString();
+            Published = faq.Published ? "Published" : "Unpublished";
+            QAnchor = faq.QAnchor;
+            QText = faq.QText;
+            TargetGroup = Enumeration.GetAll<DlsSubApplication>()
```
wilkinsm, line 18: I'm not sure this method of getting a target group will work across all FAQ target groups. There is not a corresponding `DlsSubApplication` for every `TargetGroup` exactly.

There is also no reference data in the DB.

It may be that we go with a simple `enum` containing the 4 TargetGroups with a simple description on each that can be pulled out. Or maybe use the more comprehensive `enumeration` class if that is easier.

DigitalLearningSolutions.Web/Views/SuperAdmin/Faqs/SuperAdminFaqs.cshtml
```diff
@@ -0,0 +1,48 @@
+@inject IConfiguration Configuration
+@using DigitalLearningSolutions.Web.ViewModels.SuperAdmin.Faqs
+@using Microsoft.Extensions.Configuration
+@model FaqsPageViewModel
+
+<link rel="stylesheet" href="@Url.Content("~/css/superAdmin/faqs.css")" asp-append-version="true">
+
+@section scripts {
+  <script src="@Url.Content("~/js/support/faqs.js")" asp-append-version="true"></script>
```
wilkinsm, line 9: Note that this `JS` file has a check built in that will mean it will not work for `SuperAdmin` as it stands. 
The url defined in there includes `/Support/` which will not match the url here.

DigitalLearningSolutions.Web/Views/SuperAdmin/Faqs/_FaqCard.cshtml
```diff
@@ -0,0 +1,60 @@
...
+    </summary>
+
+    <div class="nhsuk-details__text">
+      <div class="nhsuk-u-margin-bottom-4">
+        <span class="searchable-element-content">@Html.Raw(Model.AHtml)</span>
+
+        <div class="nhsuk-grid-row">
+          <div class="nhsuk-grid-column-one-quarter">
+            <div class="nhsuk-u-font-weight-bold">
+              <p>Created</p>
```
wilkinsm, line 31: In the attached screenshot these appear to be spaced quite far apart.

DigitalLearningSolutions.Web/Views/SuperAdmin/Faqs/_FaqCard.cshtml
```diff
@@ -0,0 +1,60 @@
...
+          </div>
+          <div class="nhsuk-grid-column-one-quarter">
+            <p>@Model.CreatedDate</p>
+            <p>@Model.Weighting</p>
+          </div>
+        </div>
+      </div>
+
+      <div>
+        <a class="nhsuk-button nhsuk-button--danger"
```
wilkinsm, line 43: This I think should be on the same line as the others. The wireframe is out of date but it is mentioned in the acceptance criteria.

DigitalLearningSolutions.Web/Views/SuperAdmin/Shared/_SideNavMenu.cshtml
```diff
@@ -0,0 +1,34 @@
+@using DigitalLearningSolutions.Web.Models.Enums
```
wilkinsm, line 1: Nice one @rhimcg thanks for sorting this here!

DigitalLearningSolutions.Web/Views/SuperAdmin/Shared/_SideNavMenu.cshtml
```diff
@@ -0,0 +1,34 @@
...
+                     asp-controller="Faqs"
+                     link-text="Configuration"
+                     is-current-page="@(Model.CurrentPage == SuperAdminSystemPage.Configuration)" />
+
+  <vc:side-menu-link asp-action="Index"
+                     asp-controller="Faqs"
+                     link-text="Manage Lists"
+                     is-current-page="@(Model.CurrentPage == SuperAdminSystemPage.ManageLists)" />
+
+  <vc:side-menu-link asp-action="Index"
```
wilkinsm, line 15: Ah just spotted, the wireframe is out of date. This item can be removed as `Landing Content` is not being refactored here.

wilkinsm (changes requested):
Nice one @rhimcg just a few tweaks to look at.

DigitalLearningSolutions.Data/DataServices/FaqsDataService.cs
rhimcg: cool moved the filtering to the service and the tests of the filtering to the service tests

DigitalLearningSolutions.Web/ViewModels/SuperAdmin/Faqs/SearchableFaqModel.cs
rhimcg: had a go but it didnt seem to live them not being in the same namespace?

DigitalLearningSolutions.Web/ViewModels/SuperAdmin/Faqs/SearchableFaqViewModel.cs
rhimcg: Ok cool so shall I create a new enum with just the data we need from the first 4 sub applications? Shall I also make it nullable so if it gives us something else we say null or maybe 'n/a'

rhimcg (commented):
image
changed the tags

DigitalLearningSolutions.Web/ViewModels/SuperAdmin/Faqs/SearchableFaqViewModel.cs
rhimcg: Oh I've just had a closer look 2 of the sub applications have target group ids but 1,3 dont seem related to a sub application

DigitalLearningSolutions.Web/ViewModels/SuperAdmin/Faqs/SearchableFaqModel.cs
wilkinsm: In terms of sharing I would look for a common place for the class to live like `Web.ViewModels.Shared.Faqs`.
A simple alternative could be to inherit instead. If you want to give that a try. 
This class here would end up looking like this instead:
```
namespace DigitalLearningSolutions.Web.ViewModels.SuperAdmin.Faqs
{
    using DigitalLearningSolutions.Data.Models.Support;
    using BaseSearchableFaqModel = Support.Faqs.SearchableFaq;

    public class SearchableFaqModel : BaseSearchableFaqModel
    {
        public SearchableFaqModel() : base() { }

        public SearchableFaqModel(Faq model) : base(model) { }
    }
}
```

wilkinsm (changes requested):
Thanks for sorting the changes and the refactor.
Theres a few more bits I've noticed a few more suggestion to look through.

DigitalLearningSolutions.Web/ViewModels/SuperAdmin/Faqs/SearchableFaqModel.cs
rhimcg: ahh cool thanks I've put them in a shared directory

DigitalLearningSolutions.Web/Views/SuperAdmin/Faqs/_FaqCard.cshtml
rhimcg: I think it looks similar to the wireframe and seemed ok in show and tell! I have realised this would stack weirdly on phones so I've made some changes to the row/columns

DigitalLearningSolutions.Web/ViewModels/SuperAdmin/Faqs/SearchableFaqModel.cs
wilkinsm: Nice one

DigitalLearningSolutions.Web/Views/SuperAdmin/Faqs/_FaqCard.cshtml
wilkinsm: Ok thanks for looking into it.

wilkinsm (approved):
Looks good thanks @rhimcg

DigitalLearningSolutions.Web/Views/SuperAdmin/Faqs/SuperAdminFaqs.cshtml
```diff
@@ -0,0 +1,44 @@
...
+
+@{
+  ViewData["Title"] = "FAQs";
+}
+
+<div class="nhsuk-grid-row">
+  <div class="nhsuk-grid-column-one-quarter">
+    <nav class="side-nav-menu" aria-label="Side navigation bar">
+      <partial name="~/Views/SuperAdmin/Shared/_SideNavMenu.cshtml" model="Model" />
+    </nav>
```
DanBloxham-sw, line 16: Subtle thing for side nav menus - these were originally taken from the NHS style guide pages, and like those pages, the side nav menu should be shifted to underneath the main content when the screen size is small enough. To do this, we just throw the partial underneath the main content as well, but with slightly different html attributes. This allows us to hide whichever one we don't need (taken care of with the classes)

See the TrackingSystem/Delegates/AllDelegates/Index.cshtml file for an example.

DigitalLearningSolutions.Web/Views/SuperAdmin/Faqs/SuperAdminFaqs.cshtml
```diff
@@ -0,0 +1,44 @@
...
+  </div>
+
+  <div class="nhsuk-grid-column-three-quarters">
+    <div class="nhsuk-grid-row">
+      <div class="nhsuk-grid-column-one-third">
+        <h1 id="page-heading" class="nhsuk-heading-xl">FAQs</h1>
```
DanBloxham-sw, line 22: We don't need this id on the h1. It was originally used on a couple of pages, and has since persisted through copy and pastes onto new ones, but it is completed unnecessary.

DigitalLearningSolutions.Web/Views/SuperAdmin/Faqs/SuperAdminFaqs.cshtml
```diff
@@ -0,0 +1,44 @@
...
+      </div>
+      <div class="nhsuk-grid-column-two-thirds heading-button-group">
+        <a class="nhsuk-button nhsuk-button--primary heading-button" role="button">
+          Add
+        </a>
+
+      </div>
```
DanBloxham-sw, line 29: Stray empty line can be removed.

DigitalLearningSolutions.Web/Views/SuperAdmin/Faqs/_FaqCard.cshtml
```diff
@@ -0,0 +1,61 @@
+<div class="searchable-element nhsuk-expander-group word-break" id="@Model.FaqId-card">
+
+  <details class="nhsuk-details nhsuk-expander">
+    <div class="nhsuk-details__text">
+      <div class="nhsuk-grid-row">
+        <div class="nhsuk-grid-column-one-quarter">
+          <div class="nhsuk-tag tag card-tag--grey">
+            @Model.TargetGroup
```
DanBloxham-sw, line 8: Again, we are using Models here without specifying the model type. In this case, I think it would definitely be useful to specify the view model.

DigitalLearningSolutions.Web/Views/SuperAdmin/Faqs/_FaqCard.cshtml
```diff
@@ -0,0 +1,61 @@
+<div class="searchable-element nhsuk-expander-group word-break" id="@Model.FaqId-card">
+
+  <details class="nhsuk-details nhsuk-expander">
+    <div class="nhsuk-details__text">
+      <div class="nhsuk-grid-row">
+        <div class="nhsuk-grid-column-one-quarter">
+          <div class="nhsuk-tag tag card-tag--grey">
```
DanBloxham-sw, line 7: Think we can replace the class here (and for the other tag) with
`nhsuk-tag nhsuk-tag--grey`
which just makes the tags look a bit nice. Think there was some feedback ages ago that tags that look like they currently do are too similar to buttons.

Though these will eventually be replaced with filterable tags anyway, so up to you if you think there is no point.

DigitalLearningSolutions.Web/Views/SuperAdmin/Faqs/_FaqCard.cshtml
```diff
@@ -0,0 +1,61 @@
...
+          </div>
+        </div>
+        <div class="nhsuk-grid-row">
+          <div class="nhsuk-grid-column-one-quarter">
+            <p class="nhsuk-u-font-weight-bold">Weighting</p>
+          </div>
+          <div class="nhsuk-grid-column-one-quarter">
+            <p>@Model.Weighting</p>
+          </div>
+        </div>
```
DanBloxham-sw, line 42: Think we normally put these in a summary list (without actions): https://service-manual.nhs.uk/design-system/components/summary-list

DigitalLearningSolutions.Web/Views/SuperAdmin/Faqs/_FaqCard.cshtml
```diff
@@ -0,0 +1,61 @@
...
+      </div>
+
+      <div>
+        <a class="nhsuk-button nhsuk-button"
+           role="button">
+          Edit
+        </a>
```
DanBloxham-sw, line 49: This button seems to have "nhsuk-button" class twice. 

Additionally, you might want to fiddle with the margins on these. When using smaller screen sizes, the default nhsuk-buttons add a huge margin underneath, which can be overwritten with the `nhsuk-u-margin-bottom-2` or something similar, depending on whatever size looks good.

DigitalLearningSolutions.Web/ViewModels/SuperAdmin/Faqs/FaqsPageViewModel.cs
```diff
@@ -0,0 +1,19 @@
...
+    public class FaqsPageViewModel
+    {
+        public FaqsPageViewModel(IEnumerable<SearchableFaq> faqs)
+        {
+            Faqs = faqs.Select(f => new SearchableFaqViewModel(f));
+            CurrentPage = SuperAdminSystemPage.FAQs;
+        }
+
+        public IEnumerable<SearchableFaqViewModel> Faqs { get; set; }
+        public SuperAdminSystemPage CurrentPage { get; set; }
```
DanBloxham-sw, line 17: This can be simplified further with
`        public SuperAdminSystemPage CurrentPage => SuperAdminSystemPage.FAQs;`
so we don't even need to set it in the constructor.

DigitalLearningSolutions.Web/Views/SuperAdmin/Faqs/SuperAdminFaqs.cshtml
```diff
@@ -0,0 +1,44 @@
+@inject IConfiguration Configuration
+@using DigitalLearningSolutions.Web.ViewModels.SuperAdmin.Faqs
+@using Microsoft.Extensions.Configuration
```
DanBloxham-sw, line 3: Don't think we are using any config values here, so the set up for it can be removed.

DanBloxham-sw (changes requested):
Looking mostly good. I've left a number of overly pedantic and minor comments to tackle, mostly about styling.

Might also be good to change the nav bar so that Support leads to this page now. Just so its easier to reach.

I also noticed that a bunch of new files you've made here (and files you have modified) have their line endings set to LF rather than CRLF which we use. I left a comment the first time I noticed it, but it seemed overly prevalent so I've mentioned it here as well. All the files should be CRLF endings. Might be worth having a look through the "configuring text editor" section in the readme for some brief pointers: https://github.com/TechnologyEnhancedLearning/DLSV2#configuring-text-editor

DigitalLearningSolutions.Web/Views/SuperAdmin/Faqs/_FaqCard.cshtml
rhimcg: image

rhimcg: cheers this does look better

DanBloxham-sw (changes requested):
Is this ready for re-review? There are still a bunch of comments from last time that have not been tackled (incidentally, all of them are under the "hidden conversations" collapsed section). 

Also added a couple of tiny things to clean up on new changes.

DigitalLearningSolutions.Web/Views/SuperAdmin/Shared/_SideNavMenu.cshtml
```diff
@@ -0,0 +1,29 @@
+@using DigitalLearningSolutions.Web.Models.Enums
+@using DigitalLearningSolutions.Web.ViewComponents
+<h2 class="side-nav__heading">System</h2>
+<ul class="nhsuk-list side-nav__list">
+  <vc:side-menu-link asp-action="Index"
+                     asp-controller="Faqs"
+                     link-text="Configuration"
+                     is-current-page="@(Model.CurrentPage == SuperAdminSystemPage.Configuration)" />
```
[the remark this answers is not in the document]
rhimcg: I've declared the FaqsPageViewModel, seems like we should have that if that's how its done elsewhere

DigitalLearningSolutions.Web/Views/SuperAdmin/Faqs/SuperAdminFaqs.cshtml
```diff
@@ -0,0 +1,42 @@
...
+    </nav>
+  </div>
+
+  <div class="nhsuk-grid-column-three-quarters">
+    <div class="nhsuk-grid-row">
+      <div class="nhsuk-grid-column-one-third">
+        <h1 id="page-heading" class="nhsuk-heading-xl">FAQs</h1>
+      </div>
+      <div class="nhsuk-grid-column-two-thirds heading-button-group">
+        <a class="nhsuk-button nhsuk-button--primary heading-button" role="button">
```
DanBloxham-sw, line 23: Just remembered to put this comment down- there is no "nhsuk-button--primary" class so it can be removed safely. The primary button styling is all contained with nhsuk-button.

DigitalLearningSolutions.Web.Tests/Controllers/SuperAdmin/FaqControllerTests.cs
```diff
@@ -0,0 +1,59 @@
...
+    using DigitalLearningSolutions.Data.Services;
+    using DigitalLearningSolutions.Web.Controllers.SuperAdmin;
+    using DigitalLearningSolutions.Web.ViewModels.SuperAdmin.Faqs;
+    using FakeItEasy;
+    using FizzWare.NBuilder;
+    using FluentAssertions;
+    using FluentAssertions.AspNetCore.Mvc;
+    using NUnit.Framework;
+
+    public class FaqControllerTests
```
DanBloxham-sw, line 16: This file should be CRLF line endings.

DigitalLearningSolutions.Web/Controllers/SuperAdmin/FaqsController.cs
```diff
@@ -0,0 +1,39 @@
...
+    using Microsoft.AspNetCore.Authorization;
+    using Microsoft.AspNetCore.Mvc;
+    using Microsoft.FeatureManagement.Mvc;
+    using DigitalLearningSolutions.Web.ViewModels.SuperAdmin.Faqs;
+
+    [FeatureGate(FeatureFlags.RefactoredSuperAdminInterface)]
+    [Authorize(Policy = CustomPolicies.UserSuperAdmin)]
+    [Route("SuperAdmin/Faqs")]
+    [SetDlsSubApplication(nameof(DlsSubApplication.SuperAdmin))]
+    public class FaqsController : Controller
```
DanBloxham-sw, line 18: This file should be CRLF line endings.

DigitalLearningSolutions.Web/Models/Enums/SuperAdminSystemPage.cs
```diff
@@ -0,0 +1,11 @@
+namespace DigitalLearningSolutions.Web.Models.Enums
+{
+    public enum SuperAdminSystemPage
```
DanBloxham-sw, line 3: This file should be CRLF line endings.

DigitalLearningSolutions.Web/Models/Enums/FaqsTargetGroup.cs
```diff
@@ -0,0 +1,22 @@
+namespace DigitalLearningSolutions.Web.Models.Enums
+{
+    using DigitalLearningSolutions.Data.Enums;
+
+    public class FaqsTargetGroup : Enumeration
```
DanBloxham-sw, line 5: This file should be CRLF line endings.

DigitalLearningSolutions.Web/Styles/superAdmin/faqs.scss
```diff
@@ -0,0 +1,2 @@
+@import "../shared/headingButtons";
+@import "../shared/searchableElements/tags";
```
DanBloxham-sw, line 2: This file should be CRLF line endings.

DigitalLearningSolutions.Web/ViewModels/Common/Faqs/SearchableFaq.cs
```diff
@@ -1,43 +1,43 @@
...
-    }
-}
+namespace DigitalLearningSolutions.Web.ViewModels.Common.Faqs
+{
+    using System;
+    using DigitalLearningSolutions.Data.Models;
+    using DigitalLearningSolutions.Data.Models.Support;
+    using DigitalLearningSolutions.Web.Helpers;
+
+    public class SearchableFaq : BaseSearchableItem
```
DanBloxham-sw, line 8: This file should be CRLF line endings.

DigitalLearningSolutions.Web/ViewModels/SuperAdmin/Faqs/FaqsPageViewModel.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace DigitalLearningSolutions.Web.ViewModels.SuperAdmin.Faqs
+{
+    using System.Collections.Generic;
+    using System.Linq;
+    using DigitalLearningSolutions.Web.Models.Enums;
+    using DigitalLearningSolutions.Web.ViewModels.Common.Faqs;
+
+    public class FaqsPageViewModel
```
DanBloxham-sw, line 8: This file should be CRLF line endings.

DigitalLearningSolutions.Web/ViewModels/SuperAdmin/Faqs/SearchableFaqViewModel.cs
```diff
@@ -0,0 +1,30 @@
+﻿namespace DigitalLearningSolutions.Web.ViewModels.SuperAdmin.Faqs
+{
+    using System.Linq;
+    using DigitalLearningSolutions.Data.Enums;
+    using DigitalLearningSolutions.Web.Models.Enums;
+    using DigitalLearningSolutions.Web.ViewModels.Common.Faqs;
+    using DigitalLearningSolutions.Web.ViewModels.Common.SearchablePage;
+
+    public class SearchableFaqViewModel : BaseFilterableViewModel
```
DanBloxham-sw, line 9: This file should be CRLF line endings.

DigitalLearningSolutions.Web/ViewModels/Support/Faqs/SearchableFaqViewModel.cs
```diff
@@ -1,18 +1,19 @@
...
-        public DlsSubApplication DlsSubApplication { get; set; }
-    }
-}
+﻿namespace DigitalLearningSolutions.Web.ViewModels.Support.Faqs
+{
+    using DigitalLearningSolutions.Web.Models.Enums;
+    using DigitalLearningSolutions.Web.ViewModels.Common.Faqs;
+    using DigitalLearningSolutions.Web.ViewModels.Common.SearchablePage;
+
+    public class SearchableFaqViewModel : BaseFilterableViewModel
```
DanBloxham-sw, line 7: This file should be CRLF line endings.

DigitalLearningSolutions.Web/Views/SuperAdmin/Faqs/SuperAdminFaqs.cshtml
```diff
@@ -0,0 +1,44 @@
+@using DigitalLearningSolutions.Web.ViewModels.SuperAdmin.Faqs
+@model FaqsPageViewModel
+
+<link rel="stylesheet" href="@Url.Content("~/css/superAdmin/faqs.css")" asp-append-version="true">
+
+@{
+  ViewData["Title"] = "FAQs";
```
DanBloxham-sw, line 7: This file should be CRLF line endings.

DigitalLearningSolutions.Web/Views/SuperAdmin/Faqs/_FaqCard.cshtml
```diff
@@ -0,0 +1,68 @@
+@using DigitalLearningSolutions.Web.ViewModels.SuperAdmin.Faqs
+@model SearchableFaqViewModel
+
```
DanBloxham-sw, line 3: This file should be CRLF line endings.

DigitalLearningSolutions.Web/Views/SuperAdmin/Shared/_SideNavMenu.cshtml
```diff
@@ -0,0 +1,31 @@
+@using DigitalLearningSolutions.Web.Models.Enums
+@using DigitalLearningSolutions.Web.ViewModels.SuperAdmin.Faqs
+@model FaqsPageViewModel
+
+<h2 class="side-nav__heading">System</h2>
```
DanBloxham-sw, line 5: This file should be CRLF line endings.

DanBloxham-sw (changes requested):
Couple of things from before. I've also gone through and left a comment on all the files that are still LF line endings when they should have CRLF. The link in my previous review comment might be helpful if you haven't gone through it.

DigitalLearningSolutions.Web.Tests/Controllers/SuperAdmin/FaqControllerTests.cs
DanBloxham-sw: Odd, it could be getting set git when committing/pushing, but if you've just opened this file, without modification, and it reads it as CRLF, then it looks like there is a discrepancy between how our IDEs are looking at it. If I open this, it is specifically LF (I'm using Visual Studio for reference). Running the formatter, it sets it to CRLF, so it looks like all the lines have changed.

Are you able to have a look in a different editor? Something simple like Notepad++ should be able to open the files and show line endings easily enough. I suspect something has gone wrong in your editor somewhere, and if Notepad++ shows them as CRLF, then we can confirm that.

rhimcg: Ok thanks, Notepadd++ and rubymine are both showing CRLF

DanBloxham-sw: Well colour me confused. If you have git auto-CRLF set to false, then I have no idea where this could be coming from, or why our editors are showing discrepancies. 

They definitely seem to have changed, as ViewModels/Support/Faqs/SearchableFaqViewModel.cs has the entire file highlighted as changed in the git differences, despite just the imports changing. On my end, it looks like the line endings have changed from before and after. Out of curiosity, what do your editors show the line endings are for that file, specifically the version currently on master?

rhimcg: Very weird, I'm getting CRLF for everything locally including support/faqs/searchableFaqViewModel.cs on master

DanBloxham-sw: I would expect pretty much everything on master to be CRLF as that is what we have tried to go for as a convention, though there are some places where LF has slipped through. But the fact that you get CRLF before, then it seems to want to change it to what it also sees as CRLF is extremely odd. I'll post a message in hee-internal to see if anyone has any ideas.

DigitalLearningSolutions.Web/Views/SuperAdmin/Shared/_SideNavMenu.cshtml
rhimcg: I think these are all sorted now, went through everything on this branch and changed any LF to CRLF. Sorry for giving you the run around on this!

DanBloxham-sw: Nice, looks like that is all sorted now. No worries, glad this could get sorted.

DanBloxham-sw (approved):
Looks good to me. Looks like there are some minor merge conflicts with the UnlockServiceTests, but seeing as the only changes here are the spacing in the comments, it shouldn't be too hard to solve them.

I've approved this, it shouldn't need re-review once those conflicts are solved and thus can be merged after, just as long as Jenkins builds fine.

PR #867, opened 2022-01-19
HEEDLS-707 - Implemented basic handling of failed learning hub open A…

DanBloxham-sw:
…PI calls

### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-707

### Description
Implemented basic handling of fail API call (based on the learningHubApiClient throwing exceptions). Added warning to pages listing bulk resources when fall back data is used.

### Screenshots
Current page desktop:

current with fallback resource desktop

Current page tablet:

Current page mobile:

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Web/Views/LearningPortal/Shared/_TagRow.cshtml
```diff
@@ -1,69 +1,73 @@
...
+    @if (!string.IsNullOrWhiteSpace(resource.ResourceType)) {
+      <div class="card-filter-tag">
+        <strong class="nhsuk-tag nhsuk-tag--grey">@resource.ResourceType</strong>
+      </div>
+    }
+    @if (!string.IsNullOrWhiteSpace(resource.CatalogueName)) {
+      <div class="card-filter-tag">
+        <strong class="nhsuk-tag nhsuk-tag--grey">@resource.CatalogueName</strong>
+      </div>
+    }
```
DanBloxham-sw, line 46: Changed this file from LF to CRLF, hence the apparent changes everywhere. Only real changes were wrapping these tags in if statements.

DigitalLearningSolutions.Data/Services/LearningHubResourceService.cs
```diff
@@ -0,0 +1,103 @@
...
+                var fallBackResources =
+                    learningResourceReferenceDataService.GetResourceReferenceDetailsByReferenceIds(resourceReferenceIds)
+                        .ToList();
+
+                var bulkResourceReferences = new BulkResourceReferences
+                {
+                    ResourceReferences = fallBackResources,
+                    UnmatchedResourceReferenceIds = resourceReferenceIds.Where(
+                        refId => fallBackResources.All(resource => resource.RefId != refId)
+                    ).ToList(),
```
SteveJacksonSoft, line 97: Maybe slightly clearer to use [Except()](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.except?view=net-6.0), but there's not much in it.

DigitalLearningSolutions.Data/Models/LearningResources/BulkLearningResourceReferences.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace DigitalLearningSolutions.Data.Models.LearningResources
+{
+    using DigitalLearningSolutions.Data.Models.External.LearningHubApiClient;
+
+    public class BulkLearningResourceReferences
```
SteveJacksonSoft, line 5: Now I'm seeing this in action, I'm starting to think is better just to use a tuple - probably shouldn't have expressed a preference on your HLD.
Less code, fewer similar-but-different model names, easier to see what a method's actually returning (as long as tuple components are named obvs). Whatcha think?

SteveJacksonSoft: Also applies to LearningResourceReferenceWithResourceDetails

DigitalLearningSolutions.Web/Controllers/LearningPortalController/Current.cs
```diff
@@ -168,6 +170,12 @@ public IActionResult RequestUnlock(int progressId)
         public async Task<IActionResult> MarkActionPlanResourceAsComplete(int learningLogItemId)
         {
             var actionPlanResource = await actionPlanService.GetActionPlanResource(learningLogItemId);
+
+            if (actionPlanResource == null)
```
SteveJacksonSoft, line 174: Nice, are these NotFound cases mentioned on the ticket as stuff which should be tested?

DigitalLearningSolutions.Web/Controllers/SignpostingController.cs
```diff
@@ -51,11 +50,17 @@ public async Task<IActionResult> LaunchLearningResource(int resourceReferenceId)
                 return RedirectToAction("ViewResource", "Signposting", new { resourceReferenceId });
             }
 
-            // TODO: HEEDLS-707 - handle case where resource reference Id does not match any resource
-            var resource = await learningHubApiClient.GetResourceByReferenceId(resourceReferenceId);
+            var resource = await learningHubResourceService.GetResourceByReferenceId(resourceReferenceId);
+
+            if (resource == null)
```
SteveJacksonSoft, line 55: Nice, again, does this need to be added to the ticket?

DigitalLearningSolutions.Web/Views/Shared/Signposting/_FallbackDataUsedWarning.cshtml
```diff
@@ -0,0 +1,4 @@
+﻿<div class="nhsuk-inset-text nhsuk-u-margin-bottom-2 nhsuk-u-margin-top-2">
+  <span class="nhsuk-u-visually-hidden">Information: </span>
+  <p>Some resource details may be out of date because we are currently unable to retrieve information from the Learning Hub.</p>
+</div>
```
SteveJacksonSoft, line 4: We have an inset text component - can we use that here?

DigitalLearningSolutions.Data/Services/LearningHubResourceService.cs
```diff
@@ -0,0 +1,103 @@
...
+            this.logger = logger;
+        }
+
+        public async Task<LearningResourceReferenceWithResourceDetails?> GetResourceByReferenceId(
+            int resourceReferenceId
+        )
+        {
+            try
+            {
+                // TODO HEEDLS-747: tackle different response codes rather than catching all exceptions
```
SteveJacksonSoft, line 45: Just what I was about to say. I totally agree, but think it's worth doing a bit more of the work in this ticket. It's a bit scary to be relying on response-parsing exceptions to tell us whether our request has been successful. Instead, can we explicitly throw an exception (some type name saying "we got a bad response" - maybe `LearningHubResponseException`) if we don't get a 200 back? Then we know what we're catching here.

DigitalLearningSolutions.Data.Tests/Services/LearningHubResourceServiceTests.cs
```diff
@@ -0,0 +1,210 @@
...
+                A.CallTo(
+                        () => learningResourceReferenceDataService.GetResourceReferenceDetailsByReferenceIds(
+                            A<IEnumerable<int>>.That.IsSameSequenceAs(resourceReferenceIdsToRetrieve)
+                        )
+                    )
+                    .MustHaveHappenedOnceExactly();
+            }
+        }
+
+        private ResourceReferenceWithResourceDetails GenerateGenericResource(int resourceReferenceId)
```
SteveJacksonSoft, line 188: Any reason not to use NBuilder instead?

SteveJacksonSoft (changes requested):

livzorn (approved):
Looks really good! Left one tiny suggestion, no need for re-review :)

DigitalLearningSolutions.Data/ApiClients/LearningHubApiClient.cs
```diff
@@ -103,5 +108,19 @@ private static string GetQueryString(string key, string? value)
...
+        private async Task<string?> GetAsyncAndProcessResult(string requestUri)
+        {
+            var response = await client.GetAsync(requestUri);
+            if (response.IsSuccessStatusCode)
+            {
+                return await response.Content.ReadAsStringAsync();
+            }
+
+            var message =
+                $"Learning Hub Open Api call failed. Status code {response.StatusCode.GetHashCode()} ({response.StatusCode}) when trying {requestUri}";
```
SteveJacksonSoft, line 121: Not sure you want the hash of the status code.

DigitalLearningSolutions.Data/ApiClients/LearningHubApiClient.cs
```diff
@@ -103,5 +108,19 @@ private static string GetQueryString(string key, string? value)
         {
             return string.IsNullOrEmpty(value) ? "" : $"{key}={value}";
         }
+
+        private async Task<string?> GetAsyncAndProcessResult(string requestUri)
```
SteveJacksonSoft, line 112: Name's fine, though it doesn't really give any more information than `GetStringAsync` except that it "does something extra ㊙️ " - I wouldn't have a problem with you calling it `GetStringAsync`.

DigitalLearningSolutions.Data/ApiClients/LearningHubApiClient.cs
SteveJacksonSoft: Might also be worth including the response message in case the API sends an error message - https://stackoverflow.com/a/17082328

DigitalLearningSolutions.Data/Services/LearningHubResourceService.cs
```diff
@@ -45,16 +42,15 @@ int resourceReferenceId
...
-                // TODO HEEDLS-747: tackle different response codes rather than catching all exceptions
                 var upToDateResourceDetails = await learningHubApiClient.GetResourceByReferenceId(resourceReferenceId);
                 return (upToDateResourceDetails, false);
             }
-            catch (HttpRequestException e)
+            catch (LearningHubResponseException e)
             {
-                logger.LogWarning(
-                    $"Call to Learning Hub Open API failed when trying to call /Resource/{resourceReferenceId} endpoint. " +
-                    $"Using fall-back data instead. Exception: {e.Message}"
```
SteveJacksonSoft: I think it could still be useful for debugging if we mentioned that we're using fallback data here (and in bulk get)

DigitalLearningSolutions.Data.Tests/Services/LearningHubResourceServiceTests.cs
SteveJacksonSoft: Ah, well I meant instead of having this method. If you just invoke `Builder<ResourceReferenceWithResourceDetails>.CreateNew.Build()` I'd have thought that would suffice in most cases, rather than having this method which specifies all the properties.

DigitalLearningSolutions.Data/Services/LearningHubResourceService.cs
DanBloxham-sw: I've added the fallback data warning to the logging being done in the API client. I think that is suitable enough, but it does cover the Search endpoint. If you think that we should restrict it to a per-method basis in the LearningHubResourceService, then I can easily change it around.

DigitalLearningSolutions.Data/ApiClients/LearningHubApiClient.cs
DanBloxham-sw: I've refactored the display of these so we grab the value and text in a better fashion.

Are we expecting the API to send its own custom messages for these? Looks like the HttpError class doesn't exist in asp core 3.1 and up, so we'd probably need to make our own class (ideally matching what the API sends back as closely as possible) and use JsonConvert.DeserializeObject as elsewhere to get the message specifically.

DigitalLearningSolutions.Data/Services/LearningHubResourceService.cs
SteveJacksonSoft: I think the problem with doing it there is that it makes the `GetStringAsync` method context-specific. It shouldn't really know what's going to happen if a request fails - that's the responsibility of LearningHubResourceService.

DigitalLearningSolutions.Data/ApiClients/LearningHubApiClient.cs
SteveJacksonSoft: Ah fair. ReasonPhrase looks good.

SteveJacksonSoft (approved):

PR #868, opened 2022-01-19
Heedls 658 add course to group page 3

AlexJacksonDS:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-658

### Description
Added the Save and confirmation page of the  Add Course to group journey. I've refactored out the GroupCoursesController from the DelegateGroupsController in a similar fashion to the refactor of the GroupDelegatesController in HEEDLS-514 (Add delegate to group, in review)

This may conflict slightly with that MR so might need to resolve some stuff later.

### Screenshots
image
image
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Data.Tests/Services/GroupServiceTests/GroupsServiceAddCourseTests.cs
```diff
@@ -0,0 +1,597 @@
...
+                GenericNewProgressId,
+                GenericRelatedTutorialId
+            );
+            SetUpAddCourseEnrolProcessFakes(groupCourse);
+
+            // When
+            groupsService.AddCourseToGroup(
+                reusableGroupCourse.GroupId,
+                reusableGroupCourse.CustomisationId,
+                8,
```
DanBloxham-sw, line 284: Probably want to set CompleteWithinMonths to 0 here just to be consistent, even though it changes nothing.

DigitalLearningSolutions.Data.Tests/Services/GroupServiceTests/GroupsServiceAddCourseTests.cs
```diff
@@ -0,0 +1,597 @@
...
+                GenericNewProgressId,
+                GenericRelatedTutorialId
+            );
+            SetUpAddCourseEnrolProcessFakes(groupCourse);
+
+            // When
+            groupsService.AddCourseToGroup(
+                groupCourse.GroupId,
+                groupCourse.CustomisationId,
+                8,
```
DanBloxham-sw, line 333: Similarly, probably want this similarly set to 3 months to be consistent.

DigitalLearningSolutions.Data.Tests/Services/GroupServiceTests/GroupsServiceAddCourseTests.cs
```diff
@@ -0,0 +1,597 @@
...
+    using DigitalLearningSolutions.Data.Models.Email;
+    using DigitalLearningSolutions.Data.Tests.TestHelpers;
+    using FakeItEasy;
+    using FluentAssertions.Execution;
+    using NUnit.Framework;
+
+    public partial class GroupsServiceTests
+    {
+        [Test]
+        public void AddCourseToGroup_adds_new_group_customisations_record()
```
DanBloxham-sw, line 14: Because so much of this file is effectively copied and pasted tests from the GroupServiceEnrolDelegateTests, do you think we can be cheeky and get rid of most of them? If not, we are basically running the same tests twice. 

Now that we have multiple methods calling the Enrolling logic, and given it is somewhat substantial, might it be worth moving it into its own service, so we only need to test things once?

If keeping these tests, I've noticed there are a couple of tests in GroupServiceEnrolDelegateTests that don't seem to have an equivalent here. Namely:

- EnrolDelegateOnGroupCourses_update_uses_course_supervisor_id_if_course_supervisor_id_is_not_null
- EnrolDelegateOnGroupCourses_updates_existing_progress_record

Is there a specific reason those haven't been brought over?

DigitalLearningSolutions.Data.Tests/Services/GroupServiceTests/GroupsServiceAddCourseTests.cs
```diff
@@ -0,0 +1,597 @@
...
+            using (new AssertionScope())
+            {
+                NewDelegateProgressRecordMustNotHaveBeenAdded();
+                A.CallTo(
+                    () => progressDataService.UpdateProgressSupervisorAndCompleteByDate(
+                        reusableProgressRecord.ProgressId,
+                        adminId,
+                        A<DateTime?>._
+                    )
+                ).MustHaveHappened();
```
DanBloxham-sw, line 395: Is this actually testing that the supervisor id remains the same? Given the title, I would have expected the other supervisor Id to have been present in the call, but then it would seem the new and old supervisor id would be the same?

DigitalLearningSolutions.Data.Tests/Services/GroupServiceTests/GroupsServiceAddCourseTests.cs
```diff
@@ -0,0 +1,597 @@
...
+                A.CallTo(
+                    () => progressDataService.CreateNewDelegateProgress(
+                        reusableGroupDelegate.DelegateId,
+                        reusableGroupCourse.CustomisationId,
+                        reusableGroupCourse.CurrentVersion,
+                        testDate,
+                        3,
+                        adminId,
+                        testDate.AddMonths(12),
+                        supervisorId
```
DanBloxham-sw, line 256: Similar to a comment below in the file, I think the supervisor Id and test set up might need double checking. I don't think it's obvious what is supposed to be happened in this test.

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/Delegates/GroupCourses/AddCourseToGroupCoursesViewModel.cs
```diff
@@ -1,11 +1,11 @@
...
-
-namespace DigitalLearningSolutions.Web.ViewModels.TrackingSystem.Delegates.DelegateGroups
+﻿namespace DigitalLearningSolutions.Web.ViewModels.TrackingSystem.Delegates.GroupCourses
 {
+    using System.Collections.Generic;
     using System.Linq;
+    using DigitalLearningSolutions.Data.Models.Courses;
     using DigitalLearningSolutions.Web.Helpers;
     using DigitalLearningSolutions.Web.ViewModels.Common.SearchablePage;
+    using DigitalLearningSolutions.Web.ViewModels.TrackingSystem.Delegates.DelegateGroups;
```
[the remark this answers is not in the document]
AlexJacksonDS: Was just from moving stuff between folders as I split the controller

DigitalLearningSolutions.Data/Services/GroupsService.cs
```diff
@@ -355,6 +324,116 @@ public void UpdateGroupName(int groupId, int centreId, string groupName)
...
+            }
+        }
+
+        private void EnrolDelegateOnGroupCourse(
+            int delegateUserId,
+            string? delegateUserEmailAddress,
+            string delegateUserFullName,
+            int? addedByAdminId,
+            GroupCourse groupCourse,
+            bool isAddCourseToGroup
```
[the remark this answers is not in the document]
AlexJacksonDS: I'm using it in bool conditions in various places, I think it's better to have a bool parameter than have to do various string == const string logic.

DigitalLearningSolutions.Data.Tests/Services/GroupServiceTests/GroupsServiceAddCourseTests.cs
AlexJacksonDS: Yes, we're passing in supervisorId 12 into the method, the group course has that supervisorId, but we expect the call to have the supervisor parameter = progressRecord.SupervisorId = adminId = 1 rather than 12.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateGroups/GroupDelegates.cshtml
```diff
@@ -10,13 +10,13 @@
 <link rel="stylesheet" href="@Url.Content("~/css/trackingSystem/delegates/groupDelegates.css")" asp-append-version="true">
 
 @section NavBreadcrumbs {
-  <partial name="_DelegateGroupBreadcrumbs" />
+  <partial name="~/Views/TrackingSystem/Delegates/Shared/_DelegateGroupBreadcrumbs.cshtml" />
```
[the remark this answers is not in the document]
AlexJacksonDS: I've tried doing this, but it causes the pages to fail to load despite Visual Studio recognising the path correctly.

DigitalLearningSolutions.Data.Tests/Services/GroupServiceTests/GroupsServiceAddCourseTests.cs
DanBloxham-sw: Okay, then yes this looks correct. I think I've gotten confused with the use of the adminId const here. 

I think we can make this test slightly more secure by changing the parameters around a bit. Looking at this from the test's perspective of "I don't know the details what's going on, but I am checking the expected results (calls)" then you could conclude that the progress records supervisor id is replaced with the adminId from the addedToAdminId parameter in the AddCourseToGroup call.

- Change the value of the addedToAdminId parameter to be different to the progress records' supervisor id
- Change the expected value of the supervisor id in UpdateProgressSupervisorAndCompleteByDate to be reusableProgressRecord.SupervisorId, just so that it is obvious that it came from the record.

Is that sensible?

DanBloxham-sw (changes requested):
Nicely done, few small follow ups.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateGroups/GroupDelegates.cshtml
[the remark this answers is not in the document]
DanBloxham-sw: Fair enough, think we've had trouble with this before.

DigitalLearningSolutions.Data/Services/GroupsService.cs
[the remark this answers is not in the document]
DanBloxham-sw: Good point, I'd forgotten about that, somehow. Best to keep it as a bool then.

DigitalLearningSolutions.Data/Services/GroupsService.cs
```diff
@@ -366,8 +445,16 @@ private bool GroupLabelMatchesAnswer(string groupLabel, string? answer, string?
...
 
-        private static bool ProgressShouldBeUpdatedOnEnrolment(Progress progress)
+        private static bool ProgressShouldBeUpdatedOnEnrolment(
+            Progress progress,
+            bool isAddCourseToGroup
+        )
         {
+            if (isAddCourseToGroup)
+            {
+                return progress.Completed == null && progress.RemovedDate == null && !progress.SystemRefreshed;
```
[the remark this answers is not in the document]
AlexJacksonDS: I re-checked the old stored procedure since it didn't seem right, and the condition is wrong. I just forgot to update the ticket.

DanBloxham-sw (approved):

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/GroupCoursesController.cs
```diff
@@ -0,0 +1,160 @@
...
+                groupCourse!.GroupCustomisationId,
+                groupCourse.CourseName,
+                groupName!
+            );
+
+            return View(model);
+        }
+
+        [HttpPost("{groupCustomisationId:int}/Remove")]
+        [ServiceFilter(typeof(VerifyAdminUserCanAccessGroup))]
```
wilkinsm, line 75: Can we define the `VerifyAdminUserCanAccessGroup` filter once on the controller rather than on each method?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/GroupCoursesController.cs
```diff
@@ -0,0 +1,160 @@
...
+            this.groupsService = groupsService;
+        }
+
+        [Route("{page:int=1}")]
+        public IActionResult Index(int groupId, int page = 1)
+        {
+            var centreId = User.GetCentreId();
+            var groupName = groupsService.GetGroupName(groupId, centreId);
+
+            if (groupName == null)
```
wilkinsm, line 42: If decorating the controller with `VerifyAdminUserCanAccessGroup` it should be same to drop this.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/GroupCoursesController.cs
```diff
@@ -0,0 +1,160 @@
...
+            groupsService.RemoveGroupCourseAndRelatedProgress(
+                groupCustomisationId,
+                groupId,
+                model.DeleteStartedEnrolments
+            );
+
+            return RedirectToAction(nameof(Index), new { groupId });
+        }
+
+        [HttpGet]
```
wilkinsm, line 93: HttpMethod and Route attributes can be combined. Same for the others too.

DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/Delegates/GroupCoursesControllerTests.cs
```diff
@@ -0,0 +1,168 @@
...
+        {
+            // Given
+            A.CallTo(() => groupsService.GetGroupName(1, 2)).Returns("Group");
+            A.CallTo(() => groupsService.GetUsableGroupCoursesForCentre(1, 2)).Returns(new List<GroupCourse>());
+
+            // When
+            var result = groupCoursesController.Index(1);
+
+            // Then
+            result.Should().BeViewResult().WithDefaultViewName();
```
wilkinsm, line 67: Should we wrap these in assertion scopes?

DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/Delegates/GroupCoursesControllerTests.cs
```diff
@@ -0,0 +1,168 @@
...
+            groupCoursesController.ModelState.IsValid.Should().BeFalse();
+            A.CallTo(() => groupsService.RemoveGroupCourseAndRelatedProgress(A<int>._, A<int>._, A<bool>._))
+                .MustNotHaveHappened();
+        }
+
+        [Test]
+        public void RemoveGroupCourse_with_valid_model_calls_remove_service_and_redirects()
+        {
+            // Given
+            var groupId = 1;
```
wilkinsm, line 96: These two can be `const`'s.

wilkinsm (changes requested):
Nice one Alex, just a new suggestions to look at really.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/GroupCoursesController.cs
AlexJacksonDS: Yeah I think so. The Index (renamed) method that I split off didn't have it but instead has a null check within it that would do the same behaviour. I think this can be replaced with the filter which would allow us to have it on the controller.

wilkinsm (approved):
Thanks Alex, looks good to merge.

PR #869, opened 2022-01-19
DLSV2-443 Default max and min to zero if none selected in set trigger…

daniel-manta-hee:
… values

### JIRA link
https://hee-dls.atlassian.net/browse/DLSV2-443

### Description
Default max and min to zero if none selected in set trigger values.

### Screenshots
image

-----
### Developer checks


kevwhitt-hee (approved):
All good.

PR #871, opened 2022-01-20
Develop/features/dlsv2 466 remove resource link signposting

daniel-manta-hee:
### JIRA link
https://hee-dls.atlassian.net/browse/DLSV2-466

### Description
Implement "Remove resource link" in Competency Resource Signposting. When removing a resource link we need to check whether it has active user data against it. If it has, we need to mark it as removed. If it hasn’t, we should simply delete it. Exclude Removed resources from the Competency Resource signposting page.

### Screenshots
image

### Relevant tables
image

### Scripts

```sql
DECLARE @competencyLearningResourceId INT = 1071
DECLARE @adminId INT = 1

IF EXISTS(
	  SELECT * FROM CompetencyLearningResources AS clr
	  WHERE clr.ID = @competencyLearningResourceId
		AND NOT EXISTS (SELECT * FROM LearningLogItems AS lli WHERE lli.LearningResourceReferenceID = clr.LearningResourceReferenceID)
		AND NOT EXISTS (SELECT * FROM CompetencyResourceAssessmentQuestionParameters AS p WHERE p.CompetencyLearningResourceID = clr.ID)
	  )
	Print('HARD DELETE CompetencyLearningResource (DELETE FROM CompetencyLearningResources...)')
ELSE 
	Print('SOFT DELETE. Try to mark record as deleted (UPDATE CompetencyLearningResources SET...)')
```
-----
### Developer checks
- Check that a CompetencyLearningResource that has dependencies on LearningLogItems is marked as deleted.
- Check that a CompetencyLearningResource that has dependencies on CompetencyResourceAssessmentQuestionParameters is marked as deleted.
- Check that a CompetencyLearningResource that has no dependencies on CompetencyResourceAssessmentQuestionParameters or LearningLogItems is hard deleted.


kevwhitt-hee (approved):
Perfect.

PR #876, opened 2022-01-20
HEEDLS-650 - Adding searching and pagination to the Recommended Learn…

DanBloxham-sw:
…ing page

### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-650

### Description
Added searching and pagination to the Recommended Learning page.

### Screenshots
Desktop:

recommended learning desktop

Desktop (No JS):

recommended learning desktop NO JS

Tablet:

Mobile:

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Web/Controllers/LearningPortalController/RecommendedLearningController.cs
```diff
@@ -59,20 +60,39 @@ public async Task<IActionResult> SelfAssessmentResults(int selfAssessmentId)
...
+            var destUrl = "/LearningPortal/SelfAssessment/" + selfAssessmentId + "/RecommendedLearning";
             selfAssessmentService.SetBookmark(selfAssessmentId, candidateId, destUrl);
             selfAssessmentService.UpdateLastAccessed(selfAssessmentId, candidateId);
 
-            return await ReturnSignpostingRecommendedLearningView(selfAssessmentId, candidateId);
+            return await ReturnSignpostingRecommendedLearningView(selfAssessmentId, candidateId, page, searchString);
+        }
+
+        [Route("/LearningPortal/SelfAssessment/{selfAssessmentId:int}/AllRecommendedLearningItems")]
+        public async Task<IActionResult> AllRecommendedLearningItems(int selfAssessmentId)
```
AlexJacksonDS, line 84: Should we perhaps be using the FeatureGate attribute to prevent accessing this endpoint when Signposting is not enabled?

DigitalLearningSolutions.Web/ViewModels/LearningPortal/RecommendedLearning/AllRecommendedLearningItemsViewModel.cs
```diff
@@ -0,0 +1,17 @@
...
+    using System.Collections.Generic;
+    using System.Linq;
+    using DigitalLearningSolutions.Data.Models.LearningResources;
+
+    public class AllRecommendedLearningItemsViewModel
+    {
+        public AllRecommendedLearningItemsViewModel(IEnumerable<RecommendedResource> resources, int selfAssessmentId)
+        {
+            RecommendedResources =
+                resources.OrderByDescending(r => r.RecommendationScore).Select(r => new SearchableRecommendedResourceViewModel(r, selfAssessmentId));
```
AlexJacksonDS, line 12: Do we need the order by here? Or is this to make the javascript ordering work as expected?

AlexJacksonDS (changes requested):
A couple of questions, but looks mostly fine

DigitalLearningSolutions.Web/Controllers/LearningPortalController/RecommendedLearningController.cs
DanBloxham-sw: Good idea, I've applied the FeatureGate to that action which made it a bit neater as well.

DigitalLearningSolutions.Web/ViewModels/LearningPortal/RecommendedLearning/AllRecommendedLearningItemsViewModel.cs
DanBloxham-sw: Without it, the JS ordering doesn't work properly. But that is only because I have no properly set up the JS ordering as we would expect.

We need to order based on the Recommendation Score, but we don't currently pass that through to card (only the wording for the tag). I thought that, rather than adding some hidden values to each card and setting up the SortOptions, I'd just order them in this view model first. The JS manipulation then preserves the order as it is because it doesn't have anything to sort on.

It's only a small reduction in the code necessary, and not a big time sink to implement the usual way, so I'm happy to implement that if you think it'd be better than using this workaround.

AlexJacksonDS: Fair enough, I don't really thing it matters either way, so lets just leave it for now

AlexJacksonDS (approved):

DigitalLearningSolutions.Web/Controllers/LearningPortalController/SelfAssessment.cs
```diff
@@ -38,6 +38,7 @@ private IEnumerable<Competency> PopulateCompetencyLevelDescriptors(List<Competen
             return competencies;
         }
 
+        [NoCaching]
```
davidm-m, line 41: Just for my benefit, why do we need NoCaching?

davidm-m (approved):
All looks good to me, I'm just interested to know why we don't want caching on certain endpoints.

DigitalLearningSolutions.Web/Controllers/LearningPortalController/SelfAssessment.cs
DanBloxham-sw: I was going to link the thread but realised it was in the signposting channel you don't have access to.

This tackles the drive-through bug fix in the ticket description. In short, when going through the self assessment journey, clicking some links causes the current page content to be temporarily replaced with a loading-spinner while waiting for the response for the next page. If you reach the next page, and then hit the back button, the previous page is recovered from the browser cache to display, rather than making a call to the action. Because the cache has the latest version of the page, it has the loading spinner replace all the content on that page, making it useless until you refresh the page. NoCaching just cuts out the need to refresh, as it forces the browser to always make a call to the action.

PR #882, opened 2022-01-24
HEEDLS-749 - Exclude removed competency learning resources

DanBloxham-sw:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-749

### Description
Prevented Removed CompetencyLearningResources from being used in the resource recommendation algorithm. Also prevented them from being extracted when adding a resource to learning log items.

### Screenshots
N/A - all backend

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data/DataServices/CompetencyLearningResourcesDataService.cs
+++ b/DigitalLearningSolutions.Data/DataServices/CompetencyLearningResourcesDataService.cs
@@ -10,7 +10,7 @@
     {
         IEnumerable<int> GetCompetencyIdsByLearningResourceReferenceId(int learningResourceReferenceId);
 
-        IEnumerable<CompetencyLearningResource> GetCompetencyLearningResourcesByCompetencyId(int competencyId);
+        IEnumerable<CompetencyLearningResource> GetActiveCompetencyLearningResourcesByCompetencyId(int competencyId);
 
         IEnumerable<CompetencyResourceAssessmentQuestionParameter> GetCompetencyResourceAssessmentQuestionParameters(IEnumerable<int> competencyLearningResourceIds);
         int AddCompetencyLearningResource(int resourceRefID, string originalResourceName, int competencyID, int adminId);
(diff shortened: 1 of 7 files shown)


DigitalLearningSolutions.Data/DataServices/CompetencyLearningResourcesDataService.cs
```diff
@@ -31,12 +31,12 @@ public IEnumerable<int> GetCompetencyIdsByLearningResourceReferenceId(int learni
                 @"SELECT
                         CompetencyID
                     FROM CompetencyLearningResources
-                    WHERE LearningResourceReferenceID = @learningResourceReferenceId",
+                    WHERE LearningResourceReferenceID = @learningResourceReferenceId AND RemovedDate IS NULL",
```
DanBloxham-sw, line 34: This method is used when adding resources to the action plan. Normally, we add a new item to LearningLogItems for each competency within the self assessment that has that resource listed. This change means that only competencies that have active competency learning resources will be linked to.

Not 100% sure if we need this change, but it made sense to me. Any thoughts/confirmation @SteveJacksonSoft  ?

AlexJacksonDS (approved):
Looks good to me, assuming you don't need to make any further changes for the action plan stage.

davidm-m (approved):
Wait on Steve to confirm the correct behaviour of `GetCompetencyIdsByLearningResourceReferenceId` before merging, but the code looks good to me.

livzorn (approved):
Looks good to me :)

DigitalLearningSolutions.Data/DataServices/CompetencyLearningResourcesDataService.cs
DanBloxham-sw: Sorry, I got confused on the table names when writing the above comment. I meant to say:

- We only add a single record to LearningLogItems, regardless of the number of linked competencies
- We create a new record in LearningLogItemCompetencies for each linked competency. 

So if a user adds an item to their action plan, only one copy of that resource will appear in the Current activities page. We have logic in place that should prevent a user from adding the item multiple times (except in cases where the resource has been completed/removed). The multiple records in LearningLogItemCompetencies are for supposedly for future features in a future signposting phase that I am not 100% sure of the details on (I think it might be for displaying tags of each linked competency on the resource cards).

SteveJacksonSoft: Ah cool. That sounds right, since a removed CompetencyLearningResource means that the resource is no longer linked to that competency. I would propose a name change to make it more conceptual, something like `GetCompetencyIdsLinkedToResource`.

SteveJacksonSoft (approved):

PR #888, opened 2022-01-26
DLSV2-477 Learning Hub api to populate EditCompetencyLearningResources view

daniel-manta-hee:
### JIRA link
https://hee-dls.atlassian.net/browse/DLSV2-477

### Description
When listing competency resources, also call learning hub api to ensure it has up-to-date information in the database about the resource. If the api is unavailable, use database information. 

Other related data like e.g. parameters are always stored in database, so api is used to override some resource fields likely to change as name or description.

### Screenshots
Switch application, go to Frameworks, Framework Structure and click on "Manage resource signposting".

#### Competency resource links with and without parameter
image

#### Learning Hub api is unavailable 
image

#### Unmatched resource warning
image

-----
### Developer checks
- Checked that Learning Hub api "is unavailable" warning message is in place when no internet connection.
- Checked that data retrieved in Swagger overrides database data for resources.
- Checked that unmatched resource is in place: add a record in LearningResourceReferences for ResourceRefID number 66666 that is unmatched in Swagger when calling bulk method
- Checked that Add Resource button is disabled when Learning Hub api is unavailable.


kevwhitt-hee (commented):
All good.

PR #893, opened 2022-01-27
HEEDLS-744 Add warnings when Learning Hub API is down

livzorn:
### JIRA link
[_HEEDLS-744_](https://softwiretech.atlassian.net/browse/HEEDLS-744)

### Description
Added warnings to the following pages:
- Remove from action plan
- Set complete-by date
- Mark as complete
- Resource pre-launch warning page

I replaced the partial that was being used for the warning with an inset text view component, and created an Enumeration class to store the text options. The pages I've added the warnings to display "The resource name may be out of date...", as the resource name is the only detail of the resource displayed, while the pages that already had warnings still display "Some resource details may be out of date..."

I also adjusted some spacing and font size for consistency among these pages with and without the warnings.

### Screenshots
Pages I added the warnings to:
localhost_5001_LearningPortal_Current_ActionPlan_1102_MarkAsComplete (1)
localhost_5001_LearningPortal_Current_ActionPlan_1102_CompleteBy
localhost_5001_LearningPortal_Current_ActionPlan_1102_Remove (1)
localhost_5001_Signposting_LaunchLearningResource_104

Tablet/Mobile example:
localhost_5001_LearningPortal_Current_ActionPlan_1102_MarkAsComplete (2)
localhost_5001_LearningPortal_Current_ActionPlan_1102_MarkAsComplete (3)

I slightly adjusted the spacing on these warnings added in an earlier ticket:
localhost_5001_LearningPortal_Completed
localhost_5001_LearningPortal_Current

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Web/ViewModels/LearningPortal/Current/MarkActionPlanResourceAsCompleteFormData.cs
```diff
@@ -20,6 +20,7 @@ protected MarkActionPlanResourceAsCompleteFormData(MarkActionPlanResourceAsCompl
         public int? Day { get; set; }
         public int? Month { get; set; }
         public int? Year { get; set; }
+        public bool ResourceSourcedFromFallbackData { get; set; }
```
DanBloxham-sw, line 23: Think this should also be populated in the formData constructor. So the message is still displayed if the user gets shunted back to the page after validation failure. Will likely need some hidden input to include it in the form data as well.

Think we will need something similar on the Edit Complete By Date page.

DanBloxham-sw (changes requested):
Looks very good, one minor comment about message consistency across pages that require validation.

Just as a heads up as well, the changes in 747 slightly modify how the bool sourcedFromFallbackData is returned. But I imagine that this ticket will get merged before that, so you'll likely not have to worry about any conflicts.

DigitalLearningSolutions.Web/ViewModels/LearningPortal/Current/MarkActionPlanResourceAsCompleteFormData.cs
livzorn: You're right, made those changes so the warning still shows if an error occurs

DanBloxham-sw (approved):

DigitalLearningSolutions.Data/Services/ActionPlanService.cs
```diff
@@ -138,7 +140,9 @@ public async Task AddResourceToActionPlan(int learningResourceReferenceId, int d
             return await MapLearningLogItemsToActionPlanResources(completedLearningLogItems);
         }
 
-        public async Task<ActionPlanResource?> GetActionPlanResource(int learningLogItemId)
+        public async Task<(ActionPlanResource? actionPlanResource, bool sourcedFromFallbackData)> GetActionPlanResource(
```
SteveJacksonSoft, line 143: The return value of LearningLogItemsDataService.GetLearningLogItem() is being changed in [747](https://github.com/TechnologyEnhancedLearning/DLSV2/pull/896/files) - `sourcedFromFallbackData` is being inverted and named `apiIsAccessible`. Let's do the same with this return value (the main motivation for this is that we're applying special significance to the return values `(null, false)` and `(null, true)`. In one case the resource is deleted, in the other the API is down and we have no fallback data - that's not very clear if the bool is called `sourcedFromFallbackData`.

DigitalLearningSolutions.Data/Services/ActionPlanService.cs
```diff
@@ -148,8 +152,8 @@ public async Task AddResourceToActionPlan(int learningResourceReferenceId, int d
                 );
 
             return response.resource != null
-                ? new ActionPlanResource(learningLogItem, response.resource)
-                : null;
+                ? (new ActionPlanResource(learningLogItem, response.resource), response.sourcedFromFallbackData)
+                : (null, false);
```
SteveJacksonSoft, line 156: Let's not hardcode `false` here - let's just pass the value through from GetLearningLogItems() (see above comment for motivation).

DigitalLearningSolutions.Web/Models/Enums/FallbackDataUsedWarningText.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace DigitalLearningSolutions.Web.Models.Enums
+{
+    using DigitalLearningSolutions.Data.Enums;
+
+    public class FallbackDataUsedWarningText : Enumeration
```
SteveJacksonSoft, line 5: Nice that you've pulled these messages out. Is there a particular reason you've made it an enumeration? It feels like it's just a static helper job to me. When I see an enumeration I generally expect it to be a list of possible values of something (e.g. subapplications of DLS or types of user or states of a video player - paused, FF, playing etc.). Enumeration can definitely work here, but it feels a bit heavyweight.

DigitalLearningSolutions.Data/Services/ActionPlanService.cs
SteveJacksonSoft: Should be changed anywhere else where it's applicable too.

DigitalLearningSolutions.Data/Services/ActionPlanService.cs
SteveJacksonSoft: Should be changed anywhere else where it's applicable too.

SteveJacksonSoft (changes requested):

AlexJacksonDS (commented):
I think this looks OK other than the things Steve has already mentioned and the merge conflicts from 744. Will see what it looks like when you've resolved those

DigitalLearningSolutions.Data/Services/ActionPlanService.cs
livzorn: Ah right, I've passed apiIsAccessible through rather than hard coding

DigitalLearningSolutions.Web/Models/Enums/FallbackDataUsedWarningText.cs
livzorn: Oh nice, I didn't think of a static helper. I've moved these warning texts to a static helper class, along with the other Learning hub-related warning texts I could find

SteveJacksonSoft (approved):
Looking lovely!

DigitalLearningSolutions.Data/Services/ActionPlanService.cs
SteveJacksonSoft: Hats off

AlexJacksonDS (commented):
Just one nitpick thing about usings in the cshtml files

DigitalLearningSolutions.Web/Views/LearningPortal/Completed/Completed.cshtml
```diff
@@ -1,4 +1,6 @@
-﻿@using DigitalLearningSolutions.Web.ViewModels.LearningPortal.Completed
+﻿@using DigitalLearningSolutions.Web.Helpers
+@using DigitalLearningSolutions.Web.ViewComponents
```
AlexJacksonDS, line 2: I don't think @using DigitalLearningSolutions.Web.ViewComponents is necessary in any of these files

livzorn: Yep, removed them all. Must be something in my settings that keeps importing it.

AlexJacksonDS (approved):

PR #894, opened 2022-01-28
DLSV2-489 Show level labels for min/max radio button questions when d…

daniel-manta-hee:
…isplaying in Competency Resource Signposting

### JIRA link
https://hee-dls.atlassian.net/browse/DLSV2-489

### Description
Retrieving label level descriptors for questions when input type is not 2 and displaying the labels in _SignpostingCard.cshtml.

### Screenshots
image

-----
### Developer checks
- When input type is 2 (slider), numbers are shown
- When input type is not 2 (radio or else), level labels are shown
- Values are consistent with Manage Parameters/Summary


DigitalLearningSolutions.Web/Views/Frameworks/Developer/_SignpostingCard.cshtml
```diff
@@ -44,15 +44,15 @@
             Minimum result match
```
kevwhitt-hee, line 44: Can we change this markup to use the same approach we use in the set signposting parameters summary screen for radio button questions? It looks odd if the user only chooses one response.

kevwhitt-hee (changes requested):

kevwhitt-hee (changes requested):
While testing this, I noticed that the labels selected and shown in the summary were sometimes mismatched. See details in Jira.

There are also some merge conflicts that need fixing.

DigitalLearningSolutions.Web/ViewModels/Frameworks/SignpostingCardViewModel.cs
```diff
@@ -1,15 +1,27 @@
...
         public int MinimumResultMatch { get; set; }
         public int MaximumResultMatch { get; set; }
         public string CompareResultTo { get; set; }
         public bool Essential { get; set; }
         public bool ParameterHasNotBeenSet { get; set; }
+
+        public string GetLevelLabel(int value)
+        {
+            return AssessmentQuestionLevelDescriptors != null && AssessmentQuestionInputTypeId != 2 ?
+                AssessmentQuestionLevelDescriptors[value].LevelLabel
```
kevwhitt-hee, line 23: This is using 'value' to retrieve an item from a zero based index. Value isn't zero based.

kevwhitt-hee (changes requested):
See comment. We retrieve AssessmentQuestionLevelDescriptor by using **value**  to retrieve from a zero based index. This needs to retrieve the AssessmentQuestionLevelDescriptor  that has a LevelValue that matches value instead.

kevwhitt-hee (changes requested):
All good now. There are still merge conflicts to resolve, though.

PR #897, opened 2022-01-28
HEEDLS-762 Pricing page feature flag

livzorn:
### JIRA link
[_HEEDLS-762_](https://softwiretech.atlassian.net/browse/HEEDLS-762)

### Description
Added a feature flag for the pricing page so if it is turned off, the Pricing tab will be hidden and the route will return a Not Found page.

### Screenshots
Navbar with PricingPage turned off:
localhost_5001_ApplicationSelector

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Web/Controllers/PricingController.cs
```diff
@@ -1,17 +1,33 @@
...
+            IConfiguration configuration
+        )
+        {
+            this.configuration = configuration;
+        }
+
         [RedirectDelegateOnlyToLearningPortal]
         public IActionResult Index()
         {
+            if (!configuration.IsPricingPageUsed())
```
AlexJacksonDS, line 26: Can we not use `[FeatureGate(FeatureFlags.PricingPage)]` on the controller like we do with the tracking system pages?

AlexJacksonDS (changes requested):
Just one question about how we're preventing the access in the controller

DigitalLearningSolutions.Web/Controllers/PricingController.cs
livzorn: Oh nice, didn't know about that. I've added it in, and this makes the tests that I added unnecessary, right?

AlexJacksonDS: Yeah it does

AlexJacksonDS (approved):

DigitalLearningSolutions.Data/Helpers/ConfigHelper.cs
```diff
@@ -41,6 +42,11 @@ public static bool IsSignpostingUsed(this IConfiguration config)
             return bool.Parse(config[UseSignposting]);
         }
 
+        public static bool IsPricingPageUsed(this IConfiguration config)
+        {
+            return bool.Parse(config[PricingPage]);
+        }
```
SteveJacksonSoft, line 48: Could just tweak the names here to make it slightly nicer - `IsPricingPageUsed` -> `IsPricingPageEnabled` and `PricingPage` -> `(Is)PricingPageEnabled`. It's not quite appropriate to follow the same pattern as signposting because signposting is a feature / process / concept where as PricingPage is a page.
But other than that looks great!

SteveJacksonSoft (changes requested):

DigitalLearningSolutions.Data/Helpers/ConfigHelper.cs
livzorn: Yep, changed the names :)

SteveJacksonSoft: Ah, sorry, I just meant the c# constant `PricingPage` rather than the config setting key, but I now realise my message was a bit unclear on that part. I think the config setting is better being called `PricingPage`, because that fits in with the convention we're using. Could you change it back and then just go ahead and merge? No need for me to re-review.

livzorn: Yep, changed the key back

PR #898, opened 2022-01-28
Heedls 565 course delegates export current

AlexJacksonDS:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-565

### Description
Configured the Export Current button on the Course Delegates page to export an excel file of the delegates on the current course.

### Screenshots
Excerpt from an example excel file:
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Data.Tests/DigitalLearningSolutions.Data.Tests.csproj
```diff
@@ -9,13 +9,17 @@
...
     <None Remove="TestData\DelegateUploadTest.xlsx" />
   </ItemGroup>
 
   <ItemGroup>
     <Content Include="TestData\ActivityDataDownloadTest.xlsx">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </Content>
+    <Content Include="TestData\CourseDelegateExportCurrentDataDownloadTest.xlsx">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </Content>
```
livzorn, line 22: Does this happen automatically when you add the test data file?

livzorn (commented):
Looks great, just a few questions for my own clarity

DigitalLearningSolutions.Data.Tests/Services/CourseDelegatesDownloadFileServiceTests.cs
```diff
@@ -0,0 +1,172 @@
...
+    using DigitalLearningSolutions.Data.Models.Courses;
+    using DigitalLearningSolutions.Data.Models.CustomPrompts;
+    using DigitalLearningSolutions.Data.Services;
+    using DigitalLearningSolutions.Data.Tests.TestHelpers;
+    using FakeItEasy;
+    using NUnit.Framework;
+
+    public class CourseDelegatesDownloadFileServiceTests
+    {
+        public const string CourseDelegateExportCurrentDataDownloadRelativeFilePath = "\\TestData\\CourseDelegateExportCurrentDataDownloadTest.xlsx";
```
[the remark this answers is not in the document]
AlexJacksonDS: \ is the escape character, so e.g. if you want a string with a " in it you need to have \\", this means that when we want to have a \ in the string we need to use \\\\

DigitalLearningSolutions.Data/Services/CourseDelegatesDownloadFileService.cs
```diff
@@ -0,0 +1,177 @@
...
+            dataTable.Columns.AddRange(
+                new[] { new DataColumn("Last name"), new DataColumn("First name"), new DataColumn("Email") }
+            );
+
+            foreach (var prompt in customRegistrationPrompts.CustomPrompts)
+            {
+                dataTable.Columns.Add(
+                    !dataTable.Columns.Contains(prompt.CustomPromptText)
+                        ? prompt.CustomPromptText
+                        : $"{prompt.CustomPromptText} (Prompt {prompt.CustomPromptNumber})"
```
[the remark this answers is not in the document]
AlexJacksonDS: The datatable columns I'm setting up can't have the same name, but there is nothing to stop a centre setting up registration prompts or admin fields with the same prompt. This just handles that case while allowing the file to generate with a sensible column name. I would assume that they wouldn't set up these prompts that this would happen, but given that I know it can, it's probably better to just handle it now.

DigitalLearningSolutions.Data.Tests/DigitalLearningSolutions.Data.Tests.csproj
AlexJacksonDS: Yeah, I'm not sure what the Remove= one at the top does, but we definitely need the lower changes so that the files are copied to the test build directory so that the can be used when the tests runs.

livzorn (approved):

DigitalLearningSolutions.Data/Services/CourseDelegatesDownloadFileService.cs
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        private static void SetUpDataTableColumns(
+            CentreCustomPrompts customRegistrationPrompts,
+            CourseAdminFields adminFields,
+            DataTable dataTable
+        )
+        {
+            dataTable.Columns.AddRange(
+                new[] { new DataColumn("Last name"), new DataColumn("First name"), new DataColumn("Email") }
```
wilkinsm, line 86: I think it would be good to define these magic strings in constants, more so because they are used more than once. Same for the others too.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/CourseDelegatesController.cs
```diff
@@ -89,5 +93,23 @@ public IActionResult AllCourseDelegates(int customisationId)
...
+            var centreId = User.GetCentreId();
+            var adminCategoryId = User.GetAdminCourseCategoryFilter();
+            var content = courseDelegatesDownloadFileService.GetCourseDelegateDownloadFileForCourse(
+                customisationId,
+                centreId,
+                adminCategoryId
+            );
+            return File(
+                content,
+                FileHelper.ExcelContentType,
```
wilkinsm, line 110: There is a library of media types provided in `.net core` albeit slightly more convoluted to obtain one.
Maybe the file helper could under the hood utilise that library meaning calls to things like `FileHelper.ExcelContentType`
would resolve to a method that did :
```
  if (new FileExtensionContentTypeProvider().TryGetContentType(fileName, out var contentType) {
    return contentType;
  }
```

Not necessary to implement but does avoid us storing our own versions of these `mime-types`.

wilkinsm (changes requested):
Looks good @AlexJacksonDS , I just noted a couple of points to think about.

wilkinsm (approved):
Thanks @AlexJacksonDS looks good to merge.

PR #905, opened 2022-02-02
HEEDLS-746 Change the group courses query to also display all centres…

AlexJacksonDS:
… courses

### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-746

### Description
Updated the SQL query used to get courses in a group to also query All Centres courses that the centre has access to

### Screenshots
No visual changes made

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Data/DataServices/GroupsDataService.cs
```diff
@@ -170,7 +170,9 @@ public IEnumerable<GroupDelegate> GetGroupDelegates(int groupId)
         public IEnumerable<GroupCourse> GetGroupCoursesForCentre(int centreId)
         {
             return connection.Query<GroupCourse>(
-                $"{GroupCourseSql} AND c.CentreId = @centreId",
+                @$"{GroupCourseSql}
+                        AND (c.CentreID = @centreId OR c.AllCentres = 1)
+                        AND EXISTS (SELECT CentreApplicationID FROM CentreApplications WHERE (ApplicationID = c.ApplicationID) AND (CentreID = @centreID) AND (Active = 1))",
```
DanBloxham-sw, line 175: I'd argue the new addition would fit better in the GroupCourseSql, and the other methods that use it should be checked and refactored, but that might be unnecessary. Thoughts?

If keeping this new addition where it is, it would be nice to split it up so it it is not just one big line.

DanBloxham-sw (changes requested):
Looks good. Left one comment about where the new SQL should go.

Also, if I am not mistaken, it looks like the altered method gets used in the Group Removal process, so the comment in the ticket should be updated to mention that as well.

AlexJacksonDS:
Realised I named the branch and PR the wrong thing

DigitalLearningSolutions.Data/DataServices/GroupsDataService.cs
AlexJacksonDS: I've modified this so that it is always included in the SQL, which made a method in the data service redundant so I've removed it, and I think I've caught all the places it is used on the ticket.

DanBloxham-sw (approved):
Looks good.

DigitalLearningSolutions.Data/DataServices/GroupsDataService.cs
```diff
@@ -170,24 +174,16 @@ public IEnumerable<GroupDelegate> GetGroupDelegates(int groupId)
         public IEnumerable<GroupCourse> GetGroupCoursesForCentre(int centreId)
```
SteveJacksonSoft, line 174: Sim. to below: `GetGroupCoursesVisibleToCentre()`

SteveJacksonSoft (changes requested):

DigitalLearningSolutions.Data/Services/GroupsService.cs
```diff
@@ -371,7 +373,7 @@ public void UpdateGroupName(int groupId, int centreId, string groupName)
             );
 
             var groupDelegates = GetGroupDelegates(groupId);
-            var groupCourse = groupsDataService.GetGroupCourseById(groupCustomisationId)!;
+            var groupCourse = groupsDataService.GetGroupCourseForCentre(groupCustomisationId, centreId)!;
```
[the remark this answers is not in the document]
AlexJacksonDS: I'm not really sure how we can handle this better, we're selecting the record we just inserted. If this is null the insert must have failed.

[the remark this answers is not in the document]
SteveJacksonSoft: Well not quite - the new code assumes that the course you've added is visible to the centre. Currently this method adds the course to the group either way, then will only return a non-null value if the course is visible to the centre. If we get a `null` back I reckon we should roll back the transaction, ideally log what's happened, and throw an exception. The transaction rollback and exception probably happen already, but it would be nice to make it a bit more obvious.

AlexJacksonDS:
Merge conflicts in GroupCoursesController and GroupCoursesController tests were line ending conflicts, these are all CRLF now

PR #909, opened 2022-02-03
HEEDLS-744 Fix learning resource warning bugs

livzorn:
### JIRA link
[_HEEDLS-744_](https://softwiretech.atlassian.net/browse/HEEDLS-744)

### Description
Made various fixes from issues from failed testing and from merge conflicts with HEEDLS-747:
- Adjusted spacing on warnings on Current and Completed pages so they look okay on tablet and mobile
- Added a config check so that the learning resource warnings will only be displayed is Signposting is used
- Fixed bugs in controllers so that when a resource is removed, the Mark as complete page and the Remove from action plan page will show a page with a warning, while the Set complete by date and Launch learning resource page will show NotFound.

### Screenshots
Spacing on warning: (identical for Completed activities page)
localhost_5001_LearningPortal_Current (1)
localhost_5001_LearningPortal_Current (2)
localhost_5001_LearningPortal_Current (3)

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/BasicAuthenticatedAccessibilityTests.cs
+++ b/DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/BasicAuthenticatedAccessibilityTests.cs
@@ -13,6 +13,7 @@ namespace DigitalLearningSolutions.Web.AutomatedUiTests.AccessibilityTests
         [Theory]
         [InlineData("/MyAccount", "My account")]
         [InlineData("/MyAccount/EditDetails", "Edit details")]
+        [InlineData("/Signposting/LaunchLearningResource/3", "View resource \"Test image resource\"")]
         [InlineData("/TrackingSystem/Centre/Administrators", "Centre administrators")]
         [InlineData("/TrackingSystem/Centre/Administrators/1/EditAdminRoles", "Edit administrator roles")]
         [InlineData(
--- a/DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/BasicAuthenticatedAccessibilityTests.cs
+++ b/DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/BasicAuthenticatedAccessibilityTests.cs
@@ -13,7 +13,6 @@ namespace DigitalLearningSolutions.Web.AutomatedUiTests.AccessibilityTests
         [Theory]
         [InlineData("/MyAccount", "My account")]
         [InlineData("/MyAccount/EditDetails", "Edit details")]
-        [InlineData("/Signposting/LaunchLearningResource/3", "View resource \"Test image resource\"")]
         [InlineData("/TrackingSystem/Centre/Administrators", "Centre administrators")]
         [InlineData("/TrackingSystem/Centre/Administrators/1/EditAdminRoles", "Edit administrator roles")]
         [InlineData(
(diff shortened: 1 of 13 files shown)


DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/BasicAuthenticatedAccessibilityTests.cs
```diff
@@ -13,7 +13,6 @@ public class BasicAuthenticatedAccessibilityTests : AccessibilityTestsBase,
         [Theory]
         [InlineData("/MyAccount", "My account")]
         [InlineData("/MyAccount/EditDetails", "Edit details")]
-        [InlineData("/Signposting/LaunchLearningResource/3", "View resource \"Test image resource\"")]
```
livzorn, line 16: Removed this test because there are no LearningResourceReferences in the remote test db, don't know if it's worth adding one in there?

AlexJacksonDS (approved):

DanBloxham-sw (changes requested):
Looking good, I like the new view component checking the config.

Left a couple of comments that might require some further investigation.

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/BasicAuthenticatedAccessibilityTests.cs
livzorn: Weirdly it kept failing for me the other day, but I've just added it back now and it works fine. Not sure what's up with that but I suppose all's well

DigitalLearningSolutions.Web/Controllers/SignpostingController.cs
```diff
@@ -51,9 +51,11 @@ public async Task<IActionResult> LaunchLearningResource(int resourceReferenceId)
             }
 
             var (resource, apiIsAccessible) =
-                await learningHubResourceService.GetResourceByReferenceId(resourceReferenceId);
+                await learningHubResourceService.GetResourceByReferenceIdAndPopulateDeletedDetailsFromDatabase(
+                    resourceReferenceId
+                );
 
-            if (resource == null)
+            if (resource == null || resource.AbsentInLearningHub)
```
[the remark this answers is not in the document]
livzorn: I added this check because if the user hadn't dismissed the warning, this page would load for a removed resource, but we want it to return not found right? I've now moved this check earlier to before the redirection to ViewResource, because I think that method would still lead to the account linking url even if the resource is deleted, so this seemed the simplest fix. But let me know if that change is a bad idea for any reason

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/BasicAuthenticatedAccessibilityTests.cs
DanBloxham-sw: Odd. Guess we'll keep an eye out for it if it becomes problematic again.

DigitalLearningSolutions.Web/Controllers/SignpostingController.cs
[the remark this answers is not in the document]
DanBloxham-sw: I think this change is what we need. Given that the warning page is just a warning page about account linking, dismissing it should not really produce different functional results to not having it dismissed. So a Not Found in either case is best. This looks like a good solution to me.

DanBloxham-sw (approved):
Think this is good now.

PR #911, opened 2022-02-04
HEEDLS-767 - apply signposting featuregates to relevant endpoints

DanBloxham-sw:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-767

### Description
Applied the FeatureGate to any signposting endpoints that don't already have it (or some dynamic equivalent). This should be all of them, but if you can think of any others, do let me know.
Also moved the Signposting controller into the Signposting folder.

### Screenshots
N/A

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Web/Controllers/Signposting/SignpostingController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/Signposting/SignpostingController.cs
@@ -1,4 +1,4 @@
-﻿namespace DigitalLearningSolutions.Web.Controllers
+﻿namespace DigitalLearningSolutions.Web.Controllers.Signposting
 {
     using System.Threading.Tasks;
     using DigitalLearningSolutions.Data.Services;
@@ -8,10 +8,12 @@
     using DigitalLearningSolutions.Web.ViewModels.Signposting;
     using Microsoft.AspNetCore.Authorization;
     using Microsoft.AspNetCore.Mvc;
+    using Microsoft.FeatureManagement.Mvc;
 
     [Authorize(Policy = CustomPolicies.UserOnly)]
+    [FeatureGate(FeatureFlags.UseSignposting)]
     [SetDlsSubApplication(nameof(DlsSubApplication.LearningPortal))]
-    [Route("Signposting")]
+    [Route("Signposting/LaunchLearningResource/{resourceReferenceId:int}")]
     public class SignpostingController : Controller
     {
         private readonly IActionPlanService actionPlanService;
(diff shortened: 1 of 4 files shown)


AlexJacksonDS (approved):

DigitalLearningSolutions.Web/Controllers/Signposting/SignpostingController.cs
```diff
@@ -8,10 +8,12 @@
     using DigitalLearningSolutions.Web.ViewModels.Signposting;
     using Microsoft.AspNetCore.Authorization;
     using Microsoft.AspNetCore.Mvc;
+    using Microsoft.FeatureManagement.Mvc;
 
     [Authorize(Policy = CustomPolicies.UserOnly)]
+    [FeatureGate(FeatureFlags.UseSignposting)]
     [SetDlsSubApplication(nameof(DlsSubApplication.LearningPortal))]
-    [Route("Signposting")]
+    [Route("Signposting/LaunchLearningResource/{resourceReferenceId:int}")]
```
SteveJacksonSoft, line 16: Have you tested that this works? I'm not sure if removing the `Route` from an action makes it follow conventional routing.
What was the rationale behind this change? I get that it commonises code, but it also makes the controller specific to cases which involve a specific resource reference, which is something the name doesn't really imply - if someone were to add another signposting action to this controller which doesn't deal with a particular resource reference, they'd have to change this line.

DanBloxham-sw: Not sure what you mean with the implication of removing `Route` from an action means it no longer "follows conventional routing". The page is still accessed via a URL of `https://localhost:44363/Signposting/LaunchLearningResource/100` when run locally.

I've tested this, all works as expected, journey seems fine. 

I made the change as it was recommended by Resharper, and it neatens things up. Elsewhere, we have followed of a convention of "do what you need, not what you think you might need in the future" (though I'm not certain how rigorously these has been followed). Currently, we don't have any actions in the controller that don't followed that convention, and as far as I am aware, we don't plan on adding any more. This also helps differentiate it slightly from the SignpostingSSOController that has "[Route("Signposting")]".

SteveJacksonSoft: Perhaps would make sense to combine this controller with SignpostingSSOController come to think of it, but not a big deal. Okay, let's leave it as is.

SteveJacksonSoft (approved):

PR #912, opened 2022-02-04
Heedls 695 open lh resources new window

wilkinsm:
### JIRA link
[https://softwiretech.atlassian.net/browse/HEEDLS-695](https://softwiretech.atlassian.net/browse/HEEDLS-695)

### Description
Update the LH warning page so that it shows a "open in new tab" notice and functionally when clicking "launch resource" that opens in a new tab for "unlinked" users.

### Screenshots
Updated warning page:
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DanBloxham-sw (approved):
Looks good to me.

DigitalLearningSolutions.Web/Helpers/LearningHubWarningTextHelper.cs
```diff
@@ -1,22 +1,23 @@
...
+        public static string ResourceDetailsMayBeOutOfDate =>
+            "Some resource details may be out of date because we are currently unable to retrieve information from the Learning Hub.";
+
+        public static string ResourceHasBeenRemoved =>
+            "This resource has been removed from the Learning Hub so its details may be out of date.";
+
+        public static string LinkAccount =>
+            "The first time you view a resource, you will be taken to the Learning Hub to link your account, " +
+            "or create and link your account if you don't already have one. You will only have to do this once. " +
+            "Clicking Launch resource will start the linking process in a new tab.";
```
SteveJacksonSoft, line 17: Can we have "" around Launch resource?
Also, is this in CRLF now?

DigitalLearningSolutions.Web/Views/Signposting/LearningHubLoginWarning.cshtml
```diff
@@ -30,6 +30,7 @@
 
     <form class="nhsuk-u-margin-bottom-3"
           method="post" novalidate
+          external-link-attributes="@(!Model.LearningHubAccountLinked)"
```
SteveJacksonSoft, line 33: So the effect of this is, "open in new tab if true" right?
`external-link-attributes` doesn't make that very clear to me. Why not call it `submit-in-new-tab`?

SteveJacksonSoft (changes requested):

DigitalLearningSolutions.Web/Views/Signposting/LearningHubLoginWarning.cshtml
wilkinsm: Not strictly true, it set the attributes required but whether it opens in a new tab or not is down to users preference. But most likely new tab as most people I guess leave it as the default.
Yeah I get what you are saying by the name of `external-link-attributes` maybe `set-external-link-attributes` might of been a slightly better name.
I'm not convinced with `submit-in-new-tab`, partly because:
1. It can also be used on `<a>` tags and so its not always a `submit`
2. `submit...` is suggestive of some sort of internal submit behavior.  

Maybe `open-in-new-tab` ? 
But even then that is hiding the fact that the rel values of `noopener noreferrer` are being set. Which in the world of Javascript, for internal links, could cause an issue/confusion if `opener` was ever required.

`use-external-tab-opener` ?

DigitalLearningSolutions.Web/Helpers/LearningHubWarningTextHelper.cs
wilkinsm: According to my IDE it is CRLF. 
Text change is now done.

DigitalLearningSolutions.Web/Views/Signposting/LearningHubLoginWarning.cshtml
SteveJacksonSoft: `use-external-tab-opener` works well I think. I think it's doubtful we'd ever want `opener`. If we're opening something in a new tab, the user has the history in the current tab, so they shouldn't need it in the new tab. Then there's also the security issue.

wilkinsm: Cool, that's done.

SteveJacksonSoft (approved):
Looks good!

PR #925, opened 2022-02-08
HEEDLS-678 throw and log on too-short secret key

JBsoftwire:
### JIRA link
_[HEEDLS-XXXX](https://softwiretech.atlassian.net/browse/HEEDLS-678)_

### Description
Added handling for throwing and logging if the secret key is invalid.

### Screenshots
N/A

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data/Services/LearningHubSsoSecurityService.cs
+++ b/DigitalLearningSolutions.Data/Services/LearningHubSsoSecurityService.cs
@@ -33,7 +33,7 @@
         public string GenerateHash(string state)
         {
             var secondsSinceEpoch = GetSecondsSinceEpoch();
-            var salt = GetSalt();
+            var salt = GetSecretKeyBytes();
 
             var encoder = new UTF8Encoding();
             var timedState = encoder.GetBytes(state + secondsSinceEpoch);
@@ -43,7 +43,7 @@
         public bool VerifyHash(string state, string hash)
         {
             var secondsSinceEpoch = GetSecondsSinceEpoch();
-            var salt = GetSalt();
+            var salt = GetSecretKeyBytes();
 
             var encoder = new UTF8Encoding();
             var toleranceInSec = config.GetLearningHubSsoHashTolerance();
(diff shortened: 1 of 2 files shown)


DigitalLearningSolutions.Data/Services/LearningHubSsoSecurityService.cs
```diff
@@ -56,6 +60,19 @@ public bool VerifyHash(string state, string hash)
             return false;
         }
 
+        private byte[] GetSalt(UTF8Encoding encoder)
```
SteveJacksonSoft, line 63: Nice idea to have a separate method. I don't think there's a benefit to parameterising the encoder, though. There's no reason to believe we'll use this method with a different kind of encoder (though it would be a good thing to parameterise if it were likely).
Could we change the name to make it more obvious that the salt is coming from the secret key? You don't even need to mention salt really - the method is just converting the string into bytes.

DigitalLearningSolutions.Data/Services/LearningHubSsoSecurityService.cs
```diff
@@ -56,6 +60,19 @@ public bool VerifyHash(string state, string hash)
             return false;
         }
 
+        private byte[] GetSalt(UTF8Encoding encoder)
+        {
+            var salt = encoder.GetBytes(config.GetLearningHubSsoSecretKey());
+
+            if (salt.Length < 8)
+            {
+                logger.LogError("Secret key invalid. The secret key must have a length of at least 8 bytes.");
```
SteveJacksonSoft, line 69: We can be more user-friendly - we know we're using UTF8 to convert the string to bytes, so each character is one byte. We can say the key has to be at least 8 characters.

SteveJacksonSoft (changes requested):

DigitalLearningSolutions.Data/Services/LearningHubSsoSecurityService.cs
```diff
@@ -56,6 +61,20 @@ public bool VerifyHash(string state, string hash)
             return false;
         }
 
+        private byte[] GetSalt()
```
SteveJacksonSoft, line 64: Could we change the name to make it more obvious that the salt is coming from the secret key? You don't even need to mention salt really - the method is just converting the string into bytes.

SteveJacksonSoft (approved):
Yup, lovely stuff.

PR #934, opened 2022-02-11
HEEDLS-761 - screen reader pagination announcement fixes

DanBloxham-sw:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-761

### Description
Tackled the comments in the ticket so that the page number is announced when loading the page, and the matching result count is not read out twice. This does impact how things are announced in other circumstances, which I have detailed on comments at the relevant changes. Do let me know if we think any of these side effects should be avoided.

### Screenshots
N/A

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Web/Scripts/searchSortFilterAndPaginate/searchSortFilterAndPaginate.ts
+++ b/DigitalLearningSolutions.Web/Scripts/searchSortFilterAndPaginate/searchSortFilterAndPaginate.ts
@@ -65,7 +65,7 @@ export class SearchSortFilterAndPaginate {
           );
           this.updateSearchableElementLinks(searchableData);
         }
-        this.searchSortAndPaginate(searchableData);
+        this.searchSortAndPaginate(searchableData, false);
       });
   }
 
@@ -97,7 +97,7 @@ export class SearchSortFilterAndPaginate {
     SearchSortFilterAndPaginate.scrollToTop();
   }
 
-  private searchSortAndPaginate(searchableData: ISearchableData): void {
+  private searchSortAndPaginate(searchableData: ISearchableData, updateResultCount = true): void {
     const searchedElements = this.searchEnabled
       ? search(searchableData.searchableElements)
       : searchableData.searchableElements;
@@ -107,9 +107,11 @@ export class SearchSortFilterAndPaginate {
     const sortedElements = sortSearchableElements(filteredElements);
 
     const sortedUniqueElements = _.uniqBy(sortedElements, 'parentIndex');
-    const resultCount = sortedUniqueElements.length;
-    SearchSortFilterAndPaginate
-      .updateResultCount(resultCount);
+
+    if (updateResultCount) {
+      const resultCount = sortedUniqueElements.length;
+      SearchSortFilterAndPaginate.updateResultCount(resultCount);
+    }
 
     const paginatedElements = this.paginationEnabled
       ? paginateResults(sortedUniqueElements, this.page)
--- a/DigitalLearningSolutions.Web/Views/Shared/PaginatedPage/_PaginationNav.cshtml
+++ b/DigitalLearningSolutions.Web/Views/Shared/PaginatedPage/_PaginationNav.cshtml
@@ -14,7 +14,7 @@
       </div>
     </div>
     <div class="page-indicator-container">
-      <span aria-live="polite" aria-atomic="true">
+      <span role="alert">
         <span class="nhsuk-u-visually-hidden">page </span>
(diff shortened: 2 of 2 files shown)


DigitalLearningSolutions.Web/Scripts/searchSortFilterAndPaginate/searchSortFilterAndPaginate.ts
```diff
@@ -107,9 +107,11 @@ export class SearchSortFilterAndPaginate {
...
 
     const sortedUniqueElements = _.uniqBy(sortedElements, 'parentIndex');
-    const resultCount = sortedUniqueElements.length;
-    SearchSortFilterAndPaginate
-      .updateResultCount(resultCount);
+
+    if (updateResultCount) {
+      const resultCount = sortedUniqueElements.length;
+      SearchSortFilterAndPaginate.updateResultCount(resultCount);
+    }
```
DanBloxham-sw, line 114: This has been performed so that the result count is not updated when updateResultCount is false. It is only set to false in the constructor when the page is initially loading. This prevents the result count being announced twice on any page load when JS is enabled.

The reason for this is twofold:

1. When we hit this in the constructor, the page should already have its result count set from the basic GET request (the same that occurs without JS). This shouldn't need to change when JS loads its data, so this is unnecessary to run.
2. The result count was being read out twice on page load because the result count has an alert role. It was read out when the page was initially displayed, and then read out again when the JS loading updated it to the same value. Normally, alerts do not get announced if the text inside is updated to the exact same value. However, we add a non-breaking space to any text value that is identical when updating, see https://github.com/TechnologyEnhancedLearning/DLSV2/pull/648 for further details.

DigitalLearningSolutions.Web/Views/Shared/PaginatedPage/_PaginationNav.cshtml
```diff
@@ -14,7 +14,7 @@
       </div>
     </div>
     <div class="page-indicator-container">
-      <span aria-live="polite" aria-atomic="true">
+      <span role="alert">
```
DanBloxham-sw, line 17: This was previously on `aria-live="polite"` which means it would only be read out when updated. Changing it to an alert role means it will always be announced on page load as well (both when redirected here via the new returnPage functionality, and also when first loading the page).

livzorn (approved):
Looks great to me :)

DigitalLearningSolutions.Web/Views/Shared/PaginatedPage/_PaginationNav.cshtml
SteveJacksonSoft: If we have a filter cookie set, will it now get read out twice on page load? (Once when the HTML loads, and once when the pagination / filter code runs)

DanBloxham-sw: We no longer read them out twice, even if we have a filter cookie set. The initial HTML load is read out, but the SearchSortPaginate function in JS that runs does not update the result count, so the value is not read out twice like it was before.

DigitalLearningSolutions.Web/Scripts/searchSortFilterAndPaginate/searchSortFilterAndPaginate.ts
```diff
@@ -97,7 +97,7 @@ export class SearchSortFilterAndPaginate {
     SearchSortFilterAndPaginate.scrollToTop();
   }
 
-  private searchSortAndPaginate(searchableData: ISearchableData): void {
+  private searchSortAndPaginate(searchableData: ISearchableData, updateResultCount = true): void {
```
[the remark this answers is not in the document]
DanBloxham-sw: That is correct. The Default of true here was implemented so it was easy to see where the change will affect. If the value is true, then nothing will have changed from how the code was previously set out. 

So, we do read out the number of results whenever a user hits next/previous page buttons. This is what how it happened before, and acts consistently with the non-JS behaviour. 

I can see how it might be a bit of a pain though, so with the introduction of this parameter, we could easily prevent JS from updating the result count whenever we know a change in the number will not occur (paginating, updating number of items per page, sorting). Though there won't be any sound notification that things have changed elsewhere on the page.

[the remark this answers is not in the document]
SteveJacksonSoft: Okay, seems fine for now.

PR #948, opened 2022-02-16
HEEDLS-653 - Add precursor page

GeorgeBlackwell:
### JIRA link
[_HEEDLS-653_](https://softwiretech.atlassian.net/browse/HEEDLS-653)

### Description
A new precursor page has been added to _Course delegates_. When a user attempts to access this page, they'll be taken to _Delegate courses_ before moving onto the existing page.

### Screenshots
image
image
image
image
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Web/ViewModels/TrackingSystem/Delegates/DelegateCourses/DelegateCoursesViewModel.cs
```diff
@@ -0,0 +1,59 @@
...
+                courses.AsQueryable(),
+                sortBy,
+                sortDirection
+            );
+            var filteredItems = FilteringHelper.FilterItems(sortedItems.AsQueryable(), filterBy).ToList();
+            var searchedItems = GenericSearchHelper.SearchItems(filteredItems, SearchString).ToList();
+            var paginatedItems = GetItemsOnCurrentPage(searchedItems);
+
+            MatchingSearchResults = searchedItems.Count;
+            SetTotalPages();
```
SteveJacksonSoft, line 43: Can you use methods on the base class(es) instead? That simplifies it a bit. `PaginateItems()` covers the last three lines here.
(For searching you may as well use the searchHelper - you can't simplify it much more)

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/DelegateCoursesController.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        [Route("{page=1:int}")]
+        public IActionResult Index(
+            string? searchString = null,
+            string? sortBy = null,
+            string sortDirection = GenericSortingHelper.Ascending,
+            string? filterBy = null,
+            string? filterValue = null,
+            int page = 1,
+            int? itemsPerPage = null
```
SteveJacksonSoft, line 49: We haven't had a request to set items per page here. Let's check if they want it.
Kevin's away today and tomorrow, but could you leave a question for David and Carolyn (probably include Kevin for when he gets back) on #hee-dls-refactor asking if they want to be able to set the number of courses per page on Delegate courses?

SteveJacksonSoft: If they don't want it, there's a _SearchResultsCount.cshtml partial which doesn't include an `ItemsPerPage` dropdown.

DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/Delegates/DelegateCoursesControllerTests.cs
```diff
@@ -0,0 +1,207 @@
...
+
+            const string cookieName = "DelegateCoursesFilter";
+            const string cookieValue = "Status|Active|false";
+
+            controllerWithCookies = new DelegateCoursesController(
+                courseService,
+                courseCategoryDataService,
+                courseTopicsDataService
+            )
+            .WithMockHttpContext(httpRequest, cookieName, cookieValue, httpResponse)
```
SteveJacksonSoft, line 104: This may not work depending on how the FakeItEasy implements mocks, but can we pull the cookie-setting part of this method out into a separate extension method so we can set cookies in test code rather than setup code?
It's always a bit harder to read a test when some of the important values are set up in a different bit of code. Particularly the first test is a bit confusing.

DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/Delegates/DelegateCoursesControllerTests.cs
```diff
@@ -0,0 +1,207 @@
...
+    using FakeItEasy;
+    using FluentAssertions;
+    using FluentAssertions.Execution;
+    using Microsoft.AspNetCore.Http;
+    using Microsoft.AspNetCore.Mvc;
+    using NUnit.Framework;
+
+    public class DelegateCoursesControllerTests
+    {
+        private readonly List<ApplicationDetails> applicationOptions = new List<ApplicationDetails>
```
SteveJacksonSoft, line 20: Our policy is to use NBuilder to set up test data unless using the constructor is significantly more concise.
It'll populate class members of basic types automatically.
E.g. Builder<ApplicationDetails>.CreateListOfSize(1).Build();
(Gives you an IList instead of a List, but shouldn't be a problem in general)

DigitalLearningSolutions.Web.Tests/ViewModels/TrackingSystem/Delegates/DelegateCourses/DelegateCoursesViewModelTests.cs
```diff
@@ -0,0 +1,147 @@
...
+    using DigitalLearningSolutions.Web.ViewModels.Common.SearchablePage;
+    using DigitalLearningSolutions.Web.ViewModels.TrackingSystem.CourseSetup;
+    using DigitalLearningSolutions.Web.ViewModels.TrackingSystem.Delegates.DelegateCourses;
+    using FluentAssertions;
+    using FluentAssertions.Execution;
+    using NUnit.Framework;
+
+    public class DelegateCoursesViewModelTests
+    {
+        private readonly IEnumerable<CourseStatisticsWithAdminFieldResponseCounts> courses =
```
SteveJacksonSoft, line 16: Can we use NBuilder here instead?

SteveJacksonSoft: Then you can put it in test code without it cluttering them.

DigitalLearningSolutions.Web.Tests/ViewModels/TrackingSystem/Delegates/DelegateCourses/DelegateCoursesViewModelTests.cs
```diff
@@ -0,0 +1,147 @@
...
+                "Category 1",
+                "Category 2",
+            };
+            var topics = new[]
+            {
+                "Topic 1",
+                "Topic 2",
+            };
+
+            var expectedFilters = CourseStatisticsViewModelFilterOptions.GetFilterOptions(categories, topics);
```
SteveJacksonSoft, line 100: Nitpick, but I reckon this belongs in the `Then` section - that's generally where I'd expect to find expected values.

DigitalLearningSolutions.Web.Tests/ViewModels/TrackingSystem/Delegates/DelegateCourses/DelegateCoursesViewModelTests.cs
```diff
@@ -0,0 +1,147 @@
...
+    using DigitalLearningSolutions.Data.Helpers;
+    using DigitalLearningSolutions.Data.Models.Courses;
+    using DigitalLearningSolutions.Web.ViewModels.Common.SearchablePage;
+    using DigitalLearningSolutions.Web.ViewModels.TrackingSystem.CourseSetup;
+    using DigitalLearningSolutions.Web.ViewModels.TrackingSystem.Delegates.DelegateCourses;
+    using FluentAssertions;
+    using FluentAssertions.Execution;
+    using NUnit.Framework;
+
+    public class DelegateCoursesViewModelTests
```
SteveJacksonSoft, line 14: Some of these are really tests of logic in BasePaginatedViewModel. Can we move them into a separate test file for that class?
Obviously because it's an abstract class, we have to pick some subclass to test the logic, but if the tests are in a place dedicated to BasePaginatedViewModel, it means others are less likely to write duplicate tests with other subclasses.

SteveJacksonSoft (changes requested):
Tidy work! Some changes I think would help.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/DelegateCoursesController.cs
```diff
@@ -0,0 +1,101 @@
...
+                CourseStatusFilterOptions.IsActive.FilterValue
+            );
+
+            var centreId = User.GetCentreId();
+            var categoryId = User.GetAdminCourseCategoryFilter();
+            var centreCourses =
+                courseService.GetCentreSpecificCourseStatisticsWithAdminFieldResponseCounts(centreId, categoryId);
+            var categories = courseCategoriesDataService.GetCategoriesForCentreAndCentrallyManagedCourses(centreId)
+                .Select(c => c.CategoryName);
+            var topics = courseTopicsDataService.GetCourseTopicsAvailableAtCentre(centreId).Select(c => c.CourseTopic);
```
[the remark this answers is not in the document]
DanBloxham-sw: Seeing as this is identical to the CourseSetup controller's logic, we'd probably want to reuse it there. So I'd imagine it would instead need to be a new public method in the courseService.

DigitalLearningSolutions.Web.Tests/Controllers/TrackingSystem/Delegates/DelegateCoursesControllerTests.cs
DanBloxham-sw: I think I initially set this up ages ago, and remember it being a bit of a pain. I get the feeling we do need to set it up this way, but can't remember exactly why. If we can get away with it, then extension methods might be nice. Otherwise, we can move the creation of the `controllerWithCookies` into the tests that need it. Though if we do end up needing to create it in multiple locations, then creating it here is probably for the best.

DigitalLearningSolutions.Web.Tests/ViewModels/TrackingSystem/Delegates/DelegateCourses/DelegateCoursesViewModelTests.cs
DanBloxham-sw: I think this sounds like a good idea. I believe CourseSetup was just the first proper implementation of the full BaseSearchableViewModel, so that's where the tests ended up going. I think renaming CourseSetupViewModelTests and moving it to an appropriate folder would suffice. Then this file can be deleted. 

I think some of the CourseSetupControllerTests also follow a similar pattern of testing how we take in different parameters for pagination, but those will have some tests specific to those pages (e.g., default filter). Might be worth having a look through some of those and seeing if they are generic enough to belong in their own file.

DigitalLearningSolutions.Web.Tests/ViewModels/TrackingSystem/Delegates/DelegateCourses/DelegateCoursesViewModelTests.cs
DanBloxham-sw: I asked this in a different PR, seemed like NBuilder wasn't quite sufficient. Link: https://github.com/TechnologyEnhancedLearning/DLSV2/pull/920#discussion_r803573116

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateCourses/Index.cshtml
```diff
@@ -0,0 +1,53 @@
+﻿@using DigitalLearningSolutions.Web.Models.Enums
+@using DigitalLearningSolutions.Web.ViewModels.TrackingSystem.Delegates.DelegateCourses
+@using Microsoft.AspNetCore.Mvc.TagHelpers
```
DanBloxham-sw, line 3: Looks like this TagHelpers is unused and can be safely removed.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateCourses/Index.cshtml
```diff
@@ -0,0 +1,53 @@
...
+        <h1 class="nhsuk-heading-xl">Delegate courses</h1>
+      </div>
+      <div class="nhsuk-grid-column-one-third heading-button-group">
+        <a class="nhsuk-button nhsuk-button--secondary heading-button" role="button">
+          Export all
+        </a>
+      </div>
+    </div>
+
+    <partial name="SearchablePage/Configurations/_FullWidthSearchSortAndFilter" model="Model" />
```
DanBloxham-sw, line 31: This should be the ThreeQuarterWidth one rather than FullWidth as we have a side nav menu taking up some of the space.

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/Delegates/DelegateCourses/DelegateCoursesViewModel.cs
```diff
@@ -0,0 +1,59 @@
...
+            Courses = paginatedItems.Select(c => new SearchableCourseStatisticsViewModel(c));
+            Filters = CourseStatisticsViewModelFilterOptions.GetFilterOptions(categories, topics);
```
DanBloxham-sw, line 45: @SteveJacksonSoft We are going to be displaying all the filterable tags because of this, but the ticket mentions only showing the Active tag. Why do we only want to show the Active tag? I think we show all filterable tags everywhere else, including the Course Setup cards that these should be pretty much identical to.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateCourses/Index.cshtml
```diff
@@ -0,0 +1,53 @@
+﻿@using DigitalLearningSolutions.Web.Models.Enums
+@using DigitalLearningSolutions.Web.ViewModels.TrackingSystem.Delegates.DelegateCourses
+@using Microsoft.AspNetCore.Mvc.TagHelpers
+@model DelegateCoursesViewModel
+
+<link rel="stylesheet" href="@Url.Content("~/css/trackingSystem/courseSetup.css")" asp-append-version="true">
+
+@{
+  ViewData["Title"] = "Delegate courses";
```
DanBloxham-sw, line 9: @SteveJacksonSoft Bit of a blunt question, but what is the purpose of this page? Is it just so the users have an easier time selecting courses before moving to the Course Delegates page? If that is the case, why don't we want to get rid of the course drop down menu on the Course Delegates page?

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/DelegateCoursesController.cs
```diff
@@ -0,0 +1,101 @@
...
+    using Microsoft.AspNetCore.Authorization;
+    using Microsoft.AspNetCore.Mvc;
+    using Microsoft.FeatureManagement.Mvc;
+
+    [FeatureGate(FeatureFlags.RefactoredTrackingSystem)]
+    [Authorize(Policy = CustomPolicies.UserCentreAdmin)]
+    [SetDlsSubApplication(nameof(DlsSubApplication.TrackingSystem))]
+    [SetSelectedTab(nameof(NavMenuTab.Delegates))]
+    [Route("TrackingSystem/Delegates/Courses")]
+    public class DelegateCoursesController : Controller
```
DanBloxham-sw, line 23: Something I forgot to add to the main review - all the new files you've made here have been forced into having LF line endings rather than CRLF. This has probably happened automatically based on your git autocrlf settings. We should have some documentation on that in the Readme. Once that has been set, running the formatter on each of the new files should hopefully correct them to CRLF.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateCourses/Index.cshtml
```diff
@@ -0,0 +1,53 @@
...
+}
+
+<div class="nhsuk-grid-row">
+  <div class="nhsuk-grid-column-one-quarter">
+    <nav class="side-nav-menu" aria-label="Side navigation bar">
+      <partial name="~/Views/TrackingSystem/Delegates/Shared/_DelegatesSideNavMenu.cshtml" model="DelegatePage.DelegateCourses" />
+    </nav>
+  </div>
+
+  <div class="nhsuk-grid-column-three-quarters">
```
DanBloxham-sw, line 19: I've also just merged 597 which adds a loading spinner to a bunch of pages where we use JS search/sort/paginate. This will page will also need the loading spinner added to it once you get the latest changes from master into the branch. The loading spinner is mentioned in the docs: https://softwiretech.atlassian.net/wiki/spaces/HEE/pages/19577569424/Javascript+search+sort+paginate+functionality+with+js-search . I'm happy to help give pointers on its implementation as well.

DigitalLearningSolutions.Web/ViewModels/TrackingSystem/Delegates/DelegateCourses/DelegateCoursesViewModel.cs
SteveJacksonSoft: Oh, good catch. I checked with Kevin whether he wanted the `Visible in Learning Portal` tag and he said no because it's not relevant here: https://softwire.slack.com/archives/C016Q1GR2G5/p1644393276194899?thread_ts=1644339810.552289&cid=C016Q1GR2G5
Though that could have been under the assumption that it would save time not to show it.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateCourses/Index.cshtml
SteveJacksonSoft: Fair question. In the [same Slack thread](https://softwire.slack.com/archives/C016Q1GR2G5/p1644393276194899?thread_ts=1644339810.552289&cid=C016Q1GR2G5) as mentioned in the previous question, Kevin said:

> Hmm. Gut feeling says we keep the Course dropdown on the existing course delegates page, yes, for easy switching.

I think this page was particularly motivated by a certain centre which uses admin fields extensively, and not being able to see these admin fields at the "choose course" level made things much harder for them.
There'll probably be some use cases where admins want to look at delegates for a particular set of courses, and being able to jump from one to another directly without having to change page and search or filter will save them time.

DanBloxham-sw: @SteveJacksonSoft I might be misunderstanding here then, but would that not be satisfied by, instead of making this whole new precursor page, just adding a button over on the Course Setup page cards that says "View delegates on course" which takes you to the relevant course delegates page. I'm just getting suspicious at the amount of similarities between the two pages.

SteveJacksonSoft: It could be done like that I think, but it would require some broader changes to keep a good UX, and doing that would change the layout of the site from what users are used to, so I'm not sure it would be a popular approach.

If we're going to have a Delegates area, users will want to view users by course within that area. If we're going to have a course setup area, it makes sense for that to be a top level area of its own, rather than with the Delegates area. So having two pages has UX value even if they are very similar. And the fact that we're using components mitigates the duplicated-code cost.

DanBloxham-sw: Okay, those points are fair. Just double checking we weren't going to be satisfied with a simple-seeming workaround. Thanks for the explanation.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateCourses/_CentreCourseCard.cshtml
```diff
@@ -0,0 +1,36 @@
...
+    <summary class="nhsuk-details__summary">
+      <span class="nhsuk-details__summary-text searchable-element-title" id="@Model.CustomisationId-name" name="course-name">
+        @Model.CourseName
+      </span>
+    </summary>
+
+    <div class="nhsuk-details__text">
+      <partial name="SearchablePage/_FilterableTags" model="@Model.Tags" />
+
+      <partial name="../../CourseSetup/_CentreCourseCardDetails" model="@Model" />
```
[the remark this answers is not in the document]
SteveJacksonSoft: Nah, I don't think it's worth the bother.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateCourses/Index.cshtml
SteveJacksonSoft: It's a good idea like, thanks for raising it!

PR #952, opened 2022-02-18
DLSV2-491 Rework Learning Hub resource search to reuse existing pagination

daniel-manta-hee:
### JIRA link
https://hee-dls.atlassian.net/browse/DLSV2-491

### Description
Created override methods in CompetencyResourceSignpostingViewModel for using Shared\SearchablePage\_Pagination.cshtml and inherited from BasePaginatedViewModel. The urls and behaviuor should be the same as the implemented in Delegates/All Delegates.

### Screenshots
image

Furthermore this change fixes problem with controller method preserving page number when user runs new search, as it could result in a different number of pages. So it makes more sense to reset current page to 1 when search button is clicked or user presses enter key on search field. 

```asp
        <button class="nhsuk-search__submit sort-submit" type="submit" asp-route-page="1">
          Search
        </button>
```

-----
### Developer checks
Pagination works either when
- Arriving by clicking on "Add resource" button (no parameters supplied for page and search text)
- After entering new search text and navigating to other pages in search results.

And also

- Result count is shown
- Next button is hidden on the last results page
- Previous button is hidden on first results page

### DIFF
--- a/DigitalLearningSolutions.Web/Views/Frameworks/Developer/AddCompetencyLearningResources.cshtml
+++ b/DigitalLearningSolutions.Web/Views/Frameworks/Developer/AddCompetencyLearningResources.cshtml
@@ -37,7 +37,7 @@
       <div class="search-box-container" id="search">
         @Html.LabelFor(m => m.SearchText, "Search", new { @class = "nhsuk-u-visually-hidden", @for = "search-field" })
         @Html.TextBoxFor(m => m.SearchText, new { @class = "search-box", id = "search-field", name = "searchString", type = "search", placeholder = "Search", autocomplete = "off" })
-        <button class="nhsuk-search__submit sort-submit" type="submit">
+        <button class="nhsuk-search__submit sort-submit" type="submit" asp-route-page="1">
           Search
         </button>
       </div>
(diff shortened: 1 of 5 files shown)


kevwhitt-hee (commented):
This works but when I initially search for a resource e.g. with the phrase 'test', it adds an additional `Page=1` parameter to the URL:
`https://localhost:44363/Frameworks/2/Competency/16/CompetencyGroup/16/Signposting/AddResource/1?SearchText=test&Page=1`
Which is unnecessary because page is already a parameter in the route template (see 1 just before the ? in the URL).

kevwhitt-hee (changes requested):
Can we remove the unnecessary extra Page parameter in the URL after performing a search?

DigitalLearningSolutions.Web/Views/Frameworks/Developer/AddCompetencyLearningResources.cshtml
```diff
@@ -37,7 +37,7 @@
       <div class="search-box-container" id="search">
         @Html.LabelFor(m => m.SearchText, "Search", new { @class = "nhsuk-u-visually-hidden", @for = "search-field" })
         @Html.TextBoxFor(m => m.SearchText, new { @class = "search-box", id = "search-field", name = "searchString", type = "search", placeholder = "Search", autocomplete = "off" })
-        <button class="nhsuk-search__submit sort-submit" type="submit">
+        <button class="nhsuk-search__submit sort-submit" type="submit" asp-route-page="1">
```
kevwhitt-hee, line 40: asp-route-page="1" is working fine, here. The problem is the hidden input on line 46 below which needs removing.

PR #959, opened 2022-02-21
Heedls 801 dashboard ticket count

AlexJacksonDS:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-801

### Description
Changes the logic of the ticket count on the Centre Dashboard so that non centre managers only have a count of tickets associated with them.

### Screenshots
No visual changes

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


livzorn (approved):
Looks good!

DigitalLearningSolutions.Data/DataServices/TicketDataService.cs
```diff
@@ -5,7 +5,9 @@
 
     public interface ISupportTicketDataService
     {
-        public int GetNumberOfUnarchivedTicketsForCentreId(int centreId);
+        int GetNumberOfUnarchivedTicketsForCentreId(int centreId);
+
+        int GetNumberOfUnarchivedTicketsForAdminId(int adminId);
     }
 
     public class SupportTicketDataService : ISupportTicketDataService
```
DanBloxham-sw, line 13: Not a proper review, but I spotted this while looking at something else:
The file name is "TicketDataService" but the class name is "SupportTicketDataService" and it would be good to have them consistent (probably both "SupportTicketDataService"

DigitalLearningSolutions.Data/Services/SupportTicketService.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        public SupportTicketService(ISupportTicketDataService ticketDataService, IUserDataService userDataService)
+        {
+            this.ticketDataService = ticketDataService;
+            this.userDataService = userDataService;
+        }
+
+        public int GetNumberOfTicketsForCentreAdmin(int centreId, int adminId)
+        {
+            var admin = userDataService.GetAdminUserById(adminId);
```
DanBloxham-sw, line 24: We should be able to determine if an admin is a centre manager or not from their custom claims. Do you think it is worth grabbing that from the custom claim instead? Would saving making a call to the DB. Think the existing extension method for it also checks if they are a SuperAdmin, so might need to double check if we want that.

Though I've also noticed we make a call to `GetAdminUserById` in the controller action anyway.

DanBloxham-sw (changes requested):
Looks good, one minor question about making things slightly more efficient.

DanBloxham-sw (approved):
Nicely done.

PR #962, opened 2022-02-22
HEEDLS-794 Add PRN edit to Manage Delegate Edit screen

AlexJacksonDS:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-794

### Description
Added the Edit PRN form fields to the Manage Delegate Edit screen

### Screenshots
image
image
image
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data/Services/UserService.cs
+++ b/DigitalLearningSolutions.Data/Services/UserService.cs
@@ -384,6 +384,12 @@ namespace DigitalLearningSolutions.Data.Services
                 editDelegateDetailsData.Email
             );
 
+            userDataService.UpdateDelegateProfessionalRegistrationNumber(
+                delegateUser.Id,
+                editDelegateDetailsData.ProfessionalRegistrationNumber,
+                editDelegateDetailsData.HasBeenPromptedForPrn
+            );
+
             groupsService.SynchroniseUserChangesWithGroups(
                 delegateUser,
                 editDelegateDetailsData,
--- a/DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/EditDelegateController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/EditDelegateController.cs
@@ -72,6 +72,13 @@
                 );
             }
 
+            ProfessionalRegistrationNumberHelper.ValidateProfessionalRegistrationNumber(
+                ModelState,
+                formData.HasProfessionalRegistrationNumber,
+                formData.ProfessionalRegistrationNumber,
+                true
+            );
+
             if (!ModelState.IsValid)
             {
                 return ReturnToEditDetailsViewWithErrors(formData, delegateId, centreId);
--- a/DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/EditDelegateController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/EditDelegateController.cs
@@ -75,8 +75,7 @@
             ProfessionalRegistrationNumberHelper.ValidateProfessionalRegistrationNumber(
                 ModelState,
                 formData.HasProfessionalRegistrationNumber,
-                formData.ProfessionalRegistrationNumber,
-                true
+                formData.ProfessionalRegistrationNumber
             );
 
             if (!ModelState.IsValid)
--- a/DigitalLearningSolutions.Web/Helpers/AccountDetailsDataHelper.cs
+++ b/DigitalLearningSolutions.Web/Helpers/AccountDetailsDataHelper.cs
@@ -53,7 +53,11 @@
                 formData.FirstName!,
                 formData.LastName!,
                 formData.Email!,
-                formData.AliasId
+                formData.AliasId,
+                formData.HasProfessionalRegistrationNumber == true
+                    ? formData.ProfessionalRegistrationNumber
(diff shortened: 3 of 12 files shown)


DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/EditDelegateController.cs
```diff
@@ -72,6 +72,13 @@ public IActionResult Index(EditDelegateFormData formData, int delegateId)
                 );
             }
 
+            ProfessionalRegistrationNumberHelper.ValidateProfessionalRegistrationNumber(
+                ModelState,
+                formData.HasProfessionalRegistrationNumber,
+                formData.ProfessionalRegistrationNumber,
+                true
```
DanBloxham-sw, line 79: `true` matches the same as the optional parameter. Do we want to keep it explicit here for readability?

DigitalLearningSolutions.Web/Helpers/AccountDetailsDataHelper.cs
```diff
@@ -53,7 +53,11 @@ int centreId
                 formData.FirstName!,
                 formData.LastName!,
                 formData.Email!,
-                formData.AliasId
+                formData.AliasId,
+                formData.HasProfessionalRegistrationNumber == true
+                    ? formData.ProfessionalRegistrationNumber
+                    : null,
+                formData.HasProfessionalRegistrationNumber.HasValue
```
DanBloxham-sw, line 60: Does this accurately represent the HasBeenPromptedForPrn? As this is the admin editing the delegate, if the delegate had not been prompted for PRN, but the admin put one in, I'd argue that the user still has not been prompted for the PRN. But then I am not too familiar with this work and what that is used for. Are we good with setting this as true in all cases where the admin supplies it, or false when the admin overrides it with empty?

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -384,6 +384,12 @@ CentreAnswersData centreAnswersData
                 editDelegateDetailsData.Email
             );
 
+            userDataService.UpdateDelegateProfessionalRegistrationNumber(
```
DanBloxham-sw, line 387: Any unit test to check this is called?

Also, should this update the number for any other linked delegate accounts?

DanBloxham-sw (changes requested):
Few minor questions that might just be from my unfamiliarity with the PRN work.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/EditDelegate/Index.cshtml
```diff
@@ -57,6 +59,8 @@
                        autocomplete=""
                        hint-text=""
                        css-class="nhsuk-u-width-one-half" />
+
+        <partial name="_EditRegistrationNumber" model="@Model" />
```
livzorn, line 63: Just noting that it's slightly weird that the label and hint for the prn input address the user as "you", when the user is actually editing someone else's details. Not sure if it's worth modifying the partial to change the wording?

DigitalLearningSolutions.Web/Helpers/AccountDetailsDataHelper.cs
AlexJacksonDS: That's a good question, this is the same code that we're using on the my account edit. @SteveJacksonSoft Do we want this as true when the admin has gone through this page, or only on the my account/registration?

DigitalLearningSolutions.Data/Services/UserService.cs
AlexJacksonDS: Unsure on the other delegates question @SteveJacksonSoft?

SteveJacksonSoft: Let's not update other records right now. After the user account refactor PRN will be specific to a user rather than a delegate account. Let's stick with only updating the name and email for other accounts.

DigitalLearningSolutions.Web/Helpers/AccountDetailsDataHelper.cs
SteveJacksonSoft: Hmm, maybe bad naming. `HasBeenPromptedForPrn` is meant to mean "someone's answered the PRN question".

SteveJacksonSoft: So it should be set to true.
It's there so we can distinguish between the two cases which result in an empty PRN field:
1. The user has told us they have no PRN
2. The user already existed when we added the PRN field and we haven't been told whether they have one yet

AlexJacksonDS: I can't really think of anything much better, so have just left a comment on the DelegateUser property.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/EditDelegate/Index.cshtml
AlexJacksonDS: image

livzorn (approved):

DanBloxham-sw (approved):
I think this satisfies everything.

PR #976, opened 2022-02-25
HEEDLS-806 return no users for no email and test this

JBsoftwire:
### JIRA link
_[HEEDLS-806](https://softwiretech.atlassian.net/browse/HEEDLS-806)_

### Description
Attempting to get all users with a blank email returns zero users rather than all users with no email set.

### Screenshots
N/A

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Data.Tests/Services/UserServiceTests.cs
```diff
@@ -1094,6 +1242,22 @@ public void DeactivateOrDeleteAdmin_calls_delete_if_admin_does_not_have_admin_se
             }
         }
 
+        [Test]
+        public void GetUsersByEmailAddress_returns_no_results_for_empty_address()
```
AlexJacksonDS, line 1246: Is it worth adding the 3 test cases null, empty, and whitespace to this test?

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -348,6 +353,10 @@ CentreAnswersData centreAnswersData
         {
             var delegateUser = userDataService.GetDelegateUserById(editDelegateDetailsData.DelegateId);
             var (adminUser, delegateUsers) = GetUsersByEmailAddress(delegateUser!.EmailAddress!);
+            if (delegateUsers.Count == 0)
+            {
+                delegateUsers.Add(delegateUser);
+            }
```
AlexJacksonDS, line 359: Is there a particular reason to do this this way, rather than bypassing the UpdateDelegateAccountDetails (line 373) call with no other delegate accounts? UpdateDelegate (line 380) updates the names and email on the single delegate anyway.

AlexJacksonDS (changes requested):

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -108,6 +108,11 @@ ISessionDataService sessionDataService
 
         public (AdminUser? adminUser, List<DelegateUser> delegateUsers) GetUsersByEmailAddress(string emailAddress)
         {
+            if (string.IsNullOrWhiteSpace(emailAddress))
+            {
+                return (null, new List<DelegateUser>());
+            }
```
DanBloxham-sw, line 114: Does this impact anywhere else that this method is used?

Additionally, we should probably make the `emailAddress` param nullable as we do handle that case here.

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -348,6 +353,10 @@ CentreAnswersData centreAnswersData
         {
             var delegateUser = userDataService.GetDelegateUserById(editDelegateDetailsData.DelegateId);
             var (adminUser, delegateUsers) = GetUsersByEmailAddress(delegateUser!.EmailAddress!);
```
DanBloxham-sw, line 355: The email address here can definitely be null, so worth removing the final `!`

DanBloxham-sw (changes requested):
Some minor things about pre-existing method that has been modified, and I agree with the suggestions in Alex's comments.

DigitalLearningSolutions.Data/Services/UserService.cs
JBsoftwire: The only potential issue I could find is that `NewEmailAddressIsValid` will no longer report a clash if you try to save an empty email, but empty emails should already have been filtered out by form validation so it's meaningless in practice.

DigitalLearningSolutions.Data.Tests/Services/UserServiceTests.cs
```diff
@@ -1093,61 +1093,6 @@ public void UpdateUserAccountDetailsViaDelegateAccount_calls_UpdateDelegateProfe
             ).MustHaveHappened();
         }
 
-        [Test]
-        public void UpdateUserAccountDetailsViaDelegateAccount_updates_single_account_if_no_email_set()
```
DanBloxham-sw: I think we still want this test in, but modify the assertions to check that the call to update multiple delegates only has an empty array of ids passed in.

DanBloxham-sw (changes requested):
Looks good, but it seems a unit test has been removed that I think we should keep and modify slightly.

DanBloxham-sw (approved):
Yep, think this should cover everything.

PR #979, opened 2022-03-01
HEEDLS-817 Add PRN to delegate approvals reject confirmation and add …

AlexJacksonDS:
…delegate to group

### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-817

### Description
Added PRN to the Reject delegate page and Add delegate to group cards.

### Screenshots
image
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateApprovals/DelegateRejectionPage.cshtml
+++ b/DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateApprovals/DelegateRejectionPage.cshtml
@@ -40,6 +40,12 @@
           @Model.DateRegistered?.ToShortDateString()
         </dd>
       </div>
+      <div class="nhsuk-summary-list__row basic-summary-list__row">
+        <dt class="nhsuk-summary-list__key">
+          Professional Registration Number
+        </dt>
+        <partial name="_SummaryFieldValue" model="@Model.ProfessionalRegistrationNumber" />
+      </div>
     </dl>
 
     <form class="nhsuk-u-margin-bottom-5" asp-controller="DelegateApprovals" asp-action="RejectDelegate" method="post">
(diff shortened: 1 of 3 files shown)


DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateApprovals/DelegateRejectionPage.cshtml
```diff
@@ -40,6 +40,12 @@
           @Model.DateRegistered?.ToShortDateString()
         </dd>
       </div>
+      <div class="nhsuk-summary-list__row basic-summary-list__row">
+        <dt class="nhsuk-summary-list__key">
+          Professional Registration Number
+        </dt>
+        <partial name="_SummaryFieldValue" model="@Model.ProfessionalRegistrationNumber" />
```
DanBloxham-sw, line 47: Just noticed that we get a "Possible null assigned to non-nullable entity" warning on these partial models, when we know full well that the model could be null because that's what the partial is about. Is there a way to set this up so we don't get the warning here? Looks like the partial's view model is a nullable string, so I'm not sure there is much we can do here.

DanBloxham-sw (commented):
Looks good. Left one minor comment about the summary field partial that would be nice to look at, but not necessary.

livzorn (approved):
Looks good, to me, just Dan's question to look at

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateApprovals/DelegateRejectionPage.cshtml
AlexJacksonDS: I can't see any way to avoid this warning other than manually disabling them with resharper each time/each file. We don't want to disable it solution wide though. It doesn't do this in Rider, so I tried updating visual studio and resharper, but it made no difference.

DanBloxham-sw: Fair enough, that's the answer I was expecting. I don't think there are many of us still using visual studio, so probably not worth bothering spending any extra time on this.

DanBloxham-sw (approved):

PR #988, opened 2022-03-03
DLSV2-512 Make supervisor role visible on MyStaffList and AllStaffList

daniel-manta-hee:
### JIRA link
https://hee-dls.atlassian.net/browse/DLSV2-512

### Description
Updated Supervisor\Shared\_StaffDetails.cshtml that is being called from Supervisor\Shared\_StaffMemberCard.cshtml for displaying supervisor details. Only supervisors with `CandidateID` not null would be displaying DLS role.

### Screenshots
image

-----
### Developer checks
- Checked DLS role is displayed for Supervisor, Nominated supervisor and Lerner.
- Checked that MyStaffList and AllStaffList display DLS role in supervisor details.


DigitalLearningSolutions.Web/Controllers/SupervisorController/Supervisor.cs
```diff
@@ -60,6 +63,19 @@ public IActionResult Index()
...
 
+        private DlsRole GetSupervisorDlsRole(SupervisorDelegateDetail supervisor)
+        {
+            var loggedInUserId = User.GetAdminId();
+            var loggedInAdminUser = userDataService.GetAdminUserById(loggedInUserId!.GetValueOrDefault());
+            bool activeAdminRecordAtSameCenter = loggedInAdminUser != null
+                && loggedInAdminUser.CentreId == supervisor.CentreId
+                && loggedInAdminUser.EmailAddress == supervisor.SupervisorEmail
+                && loggedInAdminUser.Active;
+            return activeAdminRecordAtSameCenter && loggedInAdminUser.IsSupervisor ? DlsRole.Supervisor
```
kevwhitt-hee, line 74: This is getting the role (supervisor, nominated supervisor, learner) of the logged in user (the supervisor accessing the staff list interface) rather than each of the members of staff on the staff list. To get this, we will need to change the data service and model that retrieves the supervisorDelegateDetails to find matching AdminUser records for the delegate by outer joining Candidates to AdminUsers on email address, centre and active = 1 and then checking for the supervisor roles in the matched admin records.

kevwhitt-hee (changes requested):
See comment, we need to get supervisor role for the staff members on the list rather than the supervisor viewing the list

kevwhitt-hee (approved):

kevwhitt-hee (changes requested):
All good once conflicts are resolved.

PR #1015, opened 2022-03-14
HEEDLS-798 - switch to using dart-sass for sass compilation

DanBloxham-sw:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-798

### Description
Swap from using node-sass to using dart-sass for compilation. Dart-sass needs to be installed manually on each computer, as the version on NPM has significantly worse performance. The ReadMe modified in this PR explains how each person should install it.

The Task Runner commands have been modified in order to run the new compilation, and the nhsuk-frontend module has been updated to the latest version. 

All files that used `@import` have been modified to use `@use` instead. Anywhere that there is a different change, I will comment on, as some things needed to be modified to get working with the new `@use`. Some files were importing others that then were not used for anything, so I have removed those where appropriate.

Overall, this has slightly improved sass compilation time, but has significantly reduced the file size of search-sort-filter-paginate files. Though we still will get some duplication of style applications, as we will have 4 separate files loaded on the page - one specific to the page, plus index.scss, layout.scss, and nhsuk.scss.

### Screenshots
N/A

-----
### Developer checks

I have:


DigitalLearningSolutions.Web/Styles/frameworks/frameworksShared.scss
```diff
@@ -1,274 +1,274 @@
-﻿@import "~nhsuk-frontend/packages/core/all";
-@import "../shared/headingButtons";
-@import "../shared/searchableElements/searchableElements";
-$nhs-dark-grey: #425563;
+﻿@use "nhsuk-frontend/packages/core/all" as *;
+@use "../shared/headingButtons" as *;
+@use "../shared/searchableElements/searchableElements" as *;
```
DanBloxham-sw, line 3: These are the only real changes on this file, the other "changes" are just line endings being corrected to CRLF.

DigitalLearningSolutions.Web/Styles/learningMenu/section.scss
```diff
@@ -1,11 +1,11 @@
...
-.float-right-additional-information {
-  @include mq($until: desktop) {
-    float: none;
-  }
-
-  @include mq($from: desktop) {
-    float: right;
-  }
-}
+﻿@use "nhsuk-frontend/packages/core/all" as *;
```
DanBloxham-sw, line 1: Again, this line and correcting the line endings on this file are the only changes.

DigitalLearningSolutions.Web/Styles/shared/searchableElements/filter.scss
```diff
@@ -113,7 +113,3 @@
     margin-right: 4px;
   }
 }
-
-.card-filter-tag {
-  @extend .card-tag;
-}
```
DanBloxham-sw, line 119: I had some problems with this file when both tags and common were being used. Specifically, it could not extend the .input-with-submit-button, but when the tags file was removed, it worked fine. Seeing as how this card-filter-tag was the only thing using it, I moved this to the tags file instead.

DigitalLearningSolutions.Web/Styles/trackingSystem/centreAdministrators.scss
```diff
@@ -1,5 +1,6 @@
-﻿@import "~nhsuk-frontend/packages/core/all";
-@import "../shared/searchableElements/searchableElements";
+﻿@use "nhsuk-frontend/packages/core/all" as *;
+@use "../shared/searchableElements/searchableElements" as *;
+@use "../shared/breakpoints" as *;
```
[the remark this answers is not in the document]
DanBloxham-sw: Breakpoints just defines a variable for the xl-desktop width in pixels so that it can be used in certain screen-size conditional styles. Previously, it would have been imported from searchableElements imports, but now we need to explicitly declare it. Think there are a couple more places like this as well.

SteveJacksonSoft (approved):

DigitalLearningSolutions.Web/Styles/shared/searchableElements/common.scss
```diff
@@ -1,5 +1,5 @@
-﻿@import "~nhsuk-frontend/packages/core/all";
-@import "../breakpoints";
+﻿@use "nhsuk-frontend/packages/core/all" as *;
+@use "../breakpoints";
```
livzorn, line 2: What does `as *` do? I noticed some files with "breakpoints" use  `as *`  and some don't.

livzorn (commented):
Looks good, just one question for my own clarification, and the build to fix

DigitalLearningSolutions.Web/Styles/shared/searchableElements/common.scss
DanBloxham-sw: The `as` is a way to set the alias/namespace of the imported files, which is then used to specify the source of variables.

By default (sans `as`) the files have an alias of the file name. So, if we are using the `xl-desktop` variable from the breakpoints file in some styling, we would have to reference it as `breakpoints.$xl-desktop`. Setting the alias to `as *` means that we don't need to use the namespace to specify the variable, and we can just use `$xl-desktop`.

That change just reduces the amount of lines changed, and keeps the scss format the same as we had it before. If you have a file introduced with `@use` that does not have an alias, its contents are not directly references in the scss file it is being introduced to, but the page will still use its styles on the page's elements.

Doc page here: https://sass-lang.com/documentation/at-rules/use#choosing-a-namespace

PR #1029, opened 2022-03-22
DLSV2-474 Add warning if sign-off took place prior to changed self-assessment results

daniel-manta-hee:
Add a warning to the sign-off area if sign-off took place prior to changed self-assessment results

### JIRA link
https://hee-dls.atlassian.net/browse/DLSV2-474

### Description
Checked if there are new results in `SelfAssessmentResults` after self assessment was signed off in `CandidateAssessmentSupervisorVerifications`.

### Screenshots
image

-----
### Developer checks

- Checked that no error is thrown when no results were submitted: prevent `max()` method to throw exception when attempting to find maximum value from an empty collection


DigitalLearningSolutions.Data/DataServices/SelfAssessmentDataService/CompetencyDataService.cs
```diff
@@ -52,6 +53,7 @@ FROM SelfAssessmentResults
                     s.AssessmentQuestionID,
                     s.ID AS ResultID,
                     s.Result,
+                    s.ResultDateTime,
```
kevwhitt-hee, line 56: Should be:
`s.DateTime AS ResultDateTime,`

kevwhitt-hee (changes requested):

PR #1031, opened 2022-03-22
Heedls 799 turn off js pagination

AlexJacksonDS:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-799

### Description
Disabled the javascript search sort filter paginate when the total item count is over 100 items. This causes the page to revert to the no-js styles. This has been done by either removing the script tag when that is the only javascript on the page, or by skipping the setup in the javascript when there is other javascript.

### Screenshots
The styling when there are more than 100 items is now just as per no javascript.
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Web/ViewModels/Common/SearchablePage/BasePaginatedViewModel.cs
```diff
@@ -37,6 +38,10 @@ protected BasePaginatedViewModel(PaginationResult<T> paginationResult)
 
         public int ItemsPerPage { get; set; }
 
+        public int TotalItems { get; set; }
```
GeorgeBlackwell, line 41: `TotalItems` is only used to set `JavascriptSearchSortFilterPaginateEnabled`. You could make it a property and assign it directly in the constructor.

DigitalLearningSolutions.Web/Views/Shared/SearchablePage/_Search.cshtml
```diff
@@ -16,7 +16,7 @@
   <div class="search-box-container" id="search">
     <label class="nhsuk-u-visually-hidden" for="search-field">@Model.SearchLabel</label>
     <input class="search-box" id="search-field" name="searchString" type="search" placeholder="Search" autocomplete="off" value="@Model.SearchString">
-    <button class="nhsuk-search__submit non-js-only" type="submit" asp-action="@ViewContext.RouteData.Values["action"].ToString()" asp-route-page="1">
+    <button class="nhsuk-search__submit search-submit" type="submit" asp-action="@ViewContext.RouteData.Values["action"].ToString()" asp-route-page="1">
       <svg class="nhsuk-icon nhsuk-icon__search" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" aria-hidden="true" focusable="false">
         <path d="M19.71 18.29l-4.11-4.1a7 7 0 1 0-1.41 1.41l4.1 4.11a1 1 0 0 0 1.42 0 1 1 0 0 0 0-1.42zM5 10a5 5 0 1 1 5 5 5 5 0 0 1-5-5z"></path>
       </svg>
```
GeorgeBlackwell, line 22: Difficult to follow this chunk on first glance, might benefit from some formatting

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/EmailDelegates/Index.cshtml
```diff
@@ -67,6 +69,7 @@
   </div>
 </div>
 
+<input type="hidden" id="javascript-search-enabled" asp-for="@Model.JavascriptSearchSortFilterPaginateEnabled" />
 @section scripts {
   <script src="@Url.Content("~/js/trackingSystem/emailDelegates.js")" asp-append-version="true"></script>
 }
```
GeorgeBlackwell, line 75: Should this have a check for `JavascriptSearchSortFilterPaginateEnabled`?

GeorgeBlackwell (changes requested):
Good work! Just a couple of suggestions

DanBloxham-sw (changes requested):
Looks like this should do the trick. Left a few comments, mostly questions.

Also has minor merge conflicts with some of the CourseDelegates tidying I did for 621.

DigitalLearningSolutions.Web/Styles/shared/searchableElements/itemsPerPage.scss
```diff
@@ -47,7 +47,13 @@
 .items-per-page-submit {
   @extend .right-submit-button;
   background-color: $button-dark-blue;
-  
+
+  #area-to-hide-while-loading & {
+    .js-enabled & {
+      display: none;
+    }
+  }
```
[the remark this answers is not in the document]
AlexJacksonDS: We can't use the no-js-only class because we need the styling while js is enabled, but only when we have javascript filters etc, so yeah.
I've tried to come up with a name for both in js-styling-hidden-area-while-loading

DigitalLearningSolutions.Web/Styles/shared/searchableElements/sort.scss
```diff
@@ -39,7 +39,7 @@
     flex: 1;
   }
 
-  .js-enabled & {
+  #area-to-hide-while-loading & {
     width: 100%;
   }
```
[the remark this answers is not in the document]
AlexJacksonDS: I don't think this was actually doing anything useful, so I've just removed it.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/EmailDelegates/Index.cshtml
AlexJacksonDS: Just tested and it doesn't preserve the selected delegates when you interact with the filter controls, but I'm not really sure what we can do about it.
The filtering still works, and the filtered items are the only items selected if you don't mess with the filters before submitting, so it's probably good enough. I'm not sure it's likely to matter for this screen though, but we could just not disable the javascript filtering ever on this page @SteveJacksonSoft ?

DanBloxham-sw: The page should function without JS, so it might be an option just to disable it all. Might need to check the original implementation ticket to see if there were any problems with the No JS version, but I reckon that's the best course of action. But keeping the JS for this page should be fine anyway.

DigitalLearningSolutions.Web/ViewModels/Common/SearchablePage/BasePaginatedViewModel.cs
```diff
@@ -27,6 +27,7 @@ protected BasePaginatedViewModel(PaginationResult<T> paginationResult)
             ItemsPerPage = paginationResult.ItemsPerPage;
             MatchingSearchResults = paginationResult.MatchingSearchResults;
             TotalPages = paginationResult.TotalPages;
+            JavascriptSearchSortFilterPaginateEnabled = paginationResult.TotalItems <= 100;
```
SteveJacksonSoft, line 30: Just had an update from Kevin that this should be 250 instead of 100.
I feel like it might also make sense to put this in the app config so it's a bit easier to change if HEE want to change it themselves for any reason. It also makes it nice and clear why JS pagination only works at certain times rather than having it hidden away in a view model.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/EmailDelegates/Index.cshtml
SteveJacksonSoft: Is this page set up right? When I open this page with no filters at a centre with 135 active delegates it says there are only 50 matching results.

I think it's best to disable the rest of the JS if we have too many delegates on this page.

AlexJacksonDS: It doesn't display all delegates, only those that are approved, non-self-registered, have an email and do not have a password.

DigitalLearningSolutions.Data/Services/SearchSortFilterPaginateService.cs
```diff
@@ -16,17 +18,27 @@ SearchSortFilterAndPaginateOptions searchSortFilterAndPaginateOptions
...
         public SearchSortFilterPaginationResult<T> SearchFilterSortAndPaginate<T>(
             IEnumerable<T> items,
             SearchSortFilterAndPaginateOptions searchSortFilterAndPaginateOptions
         ) where T : BaseSearchableItem
         {
-            var itemsToReturn = items;
+            var allItems = items.ToList();
+            var itemsToReturn = allItems;
             string? appliedFilterString = null;
+            var javascriptShouldBeEnabled =
```
DanBloxham-sw, line 36: Minor, but as there are some pages that will still use JS for non-search/sort/filter/paginate functionality, this could maybe be renamed to be more specific.

DigitalLearningSolutions.Data.Tests/TestHelpers/SearchSortFilterAndPaginateTestHelper.cs
```diff
@@ -30,7 +30,8 @@ ISearchSortFilterPaginateService searchSortFilterPaginateService
                             options!.PaginationOptions?.PageNumber ?? 1,
                             1,
                             options.PaginationOptions?.ItemsPerPage ?? int.MaxValue,
-                            items.Count()
+                            items.Count,
+                            items.Count <= 250
```
DanBloxham-sw, line 34: Do we want to make the 250 a parameter with a default value of 250 on the off chance this gets used in the future for something?

DanBloxham-sw (changes requested):
Couple of very minor comments, but everything else looks great.

DigitalLearningSolutions.Data/Services/SearchSortFilterPaginateService.cs
AlexJacksonDS: I don't think it matters a huge amount since it's a variable within the SearchSortFilterAndPaginate method, but updated it anyway.

DanBloxham-sw (approved):
Think this should be good to merge once the builds pass.

PR #1032, opened 2022-03-23
HEEDLS-797 : Fix return page on remove delegate

GeorgeBlackwell:
### JIRA link
_[HEEDLS-797](https://softwiretech.atlassian.net/browse/HEEDLS-797)_

### Description
_An issue cropped up during testing where the user wasn't being returned to the correct page on Course Delegates when they used the cancel link on the remove confirmation page._

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/ViewDelegateController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/ViewDelegateController.cs
@@ -96,7 +96,8 @@
         public IActionResult ConfirmRemoveFromCourse(
             int delegateId,
             int customisationId,
-            DelegateAccessRoute accessedVia
+            DelegateAccessRoute accessedVia,
+            int? returnPage
         )
         {
             if (!courseService.DelegateHasCurrentProgress(delegateId, customisationId))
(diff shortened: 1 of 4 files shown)


DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/ViewDelegateController.cs
```diff
@@ -116,7 +117,8 @@ DelegateAccessRoute accessedVia
                 customisationId,
                 course!.CourseName,
                 false,
-                accessedVia
+                accessedVia,
+                returnPage
```
DanBloxham-sw, line 121: We are also going to want to maintain the return page value when a validation error occurs during the post. Currently, I don't think it will be retained, so the cancel link will return the user back to the first page if they hit the submit button without checking the checkbox. 

Can't remember if we just add the return page as a query parameter to the post, or have a hidden input somewhere to keep it in the model. Have a look at another example elsewhere and do what that does. E.g. All Delegates --> Set password.

DanBloxham-sw (changes requested):
Looks good, one minor thing about keeping the return page during validation errors.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/ViewDelegateController.cs
GeorgeBlackwell: Good spot! 👍

DanBloxham-sw (approved):
Yeah I think that should do it.

PR #1052, opened 2022-03-31
HEEDLS-810 - Implement All Delegates page export button

DanBloxham-sw:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-810

### Description
Implemented the functionality for exporting spreadsheets from the All Delegates page. Logic is based on the Course Delegates export button.
I've also: 

- Fixed the empty spreadsheet crash in:
  - Delegate Bulk Upload Download journey.
  - Course Delegates Export
  - Delegate Courses Export
- Fiddled with the Delegate Courses Export:
  - Made it use AllCentre courses so it matches the courses actually shown
  - Made the width of the Last Name column also resize to match the contents (so now it will expand to match the course names.

### Screenshots
N/A

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Data/Services/DelegateDownloadFileService.cs
```diff
@@ -59,10 +110,31 @@ private void PopulateDelegatesSheet(IXLWorkbook workbook, int centreId)
...
                     x.Answer6,
                     x.Active,
-                    x.EmailAddress
+                    x.EmailAddress,
                 }
             );
 
+            if (!delegateRecords.Any())
+            {
+                var emptyDelegate = new
```
DanBloxham-sw, line 119: As this file is used in the upload journey, I was unsure if I should leave things blank, or include suitable details for a fake delegate to serve as an example for the user. No idea how likely this will be as a scenario, but for now, all fields will be blank except the Job Group ID and Active columns.

livzorn: I suppose that makes sense though it feels a little weird to me to pre-populate only two values, and to default Job Group ID to 1. I think making a fake delegate with all details filled in would risk the user forgetting to delete the fake delegate and registering them in the upload process.

DigitalLearningSolutions.Data/Services/DelegateDownloadFileService.cs
```diff
@@ -74,7 +146,154 @@ private void PopulateJobGroupsSheet(IXLWorkbook workbook)
...
+            string sortDirection,
+            string? filterString
+        )
+        {
+            var registrationPrompts = centreRegistrationPromptsService.GetCentreRegistrationPromptsByCentreId(centreId);
+            var delegatesToExport = GetDelegatesToExport(centreId, searchString, sortBy, sortDirection, filterString)
+                .ToList();
+
+            var dataTable = new DataTable();
+            // TODO: HEEDLS-810 - display extra PRN details if necessary.
```
livzorn, line 166: Have you found out if the PRN will be necessary?

DigitalLearningSolutions.Data/Services/DelegateDownloadFileService.cs
AlexJacksonDS: I think we need to default to ID 10 - Other for the job group if we're going to default to anything

AlexJacksonDS (changes requested):
I think it mostly looks fine other than the stuff already mentioned

DigitalLearningSolutions.Data/Models/User/CentreAnswersData.cs
```diff
@@ -30,5 +33,19 @@ public class CentreAnswersData
         public string? Answer4 { get; set; }
         public string? Answer5 { get; set; }
         public string? Answer6 { get; set; }
+
+        public string? GetAnswerForRegistrationPromptNumber(int promptNumber)
```
SteveJacksonSoft, line 37: Using the RegistrationField enumeration would be a bit safer than adding knowlege here about which field has which id here (sad that you can't use a switch statement though).

DigitalLearningSolutions.Data/Services/DelegateDownloadFileService.cs
SteveJacksonSoft: A good idea, but there should always be a delegate at any centre with an admin (because they should have a delegate account), so I don't think this will ever be used.

I don't think we'll ever have the case where the centre only has admins and those admins don't have candidate accounts - only admins that registered before November will be in this state, and they'll have delegates by now if they're using the system at all.

DigitalLearningSolutions.Data/Services/DelegateDownloadFileService.cs
```diff
@@ -1,30 +1,57 @@
...
     using ClosedXML.Excel;
     using DigitalLearningSolutions.Data.DataServices;
     using DigitalLearningSolutions.Data.DataServices.UserDataService;
     using DigitalLearningSolutions.Data.Helpers;
+    using DigitalLearningSolutions.Data.Models.CustomPrompts;
+    using DigitalLearningSolutions.Data.Models.User;
 
     public interface IDelegateDownloadFileService
     {
         public byte[] GetDelegateDownloadFileForCentre(int centreId);
```
SteveJacksonSoft, line 16: Can we rename this to "GetDelegatesAndJobGroupsDownloadFileForCentre"? Currently it's a bit unclear what the difference is between the two methods.

DigitalLearningSolutions.Data/Services/DelegateDownloadFileService.cs
```diff
@@ -41,6 +68,30 @@ public byte[] GetDelegateDownloadFileForCentre(int centreId)
...
+                sortDirection,
+                filterString
+            );
+
+            using var stream = new MemoryStream();
+            workbook.SaveAs(stream);
+            return stream.ToArray();
+        }
+
         private void PopulateDelegatesSheet(IXLWorkbook workbook, int centreId)
```
SteveJacksonSoft, line 95: Do we need this method anymore? Does it not do the same thing as PopulateAllDelegatesSheet with no filters/search/sort?

DigitalLearningSolutions.Data/Services/DelegateDownloadFileService.cs
```diff
@@ -74,7 +146,154 @@ private void PopulateJobGroupsSheet(IXLWorkbook workbook)
...
+            }
+
+            ClosedXmlHelper.AddSheetToWorkbook(
+                workbook,
+                AllDelegatesSheetName,
+                dataTable.AsEnumerable(),
+                XLTableTheme.None
+            );
+
+            FormatAllDelegateWorksheetColumns(workbook, dataTable);
```
SteveJacksonSoft, line 187: You could commonise this across all download files I think.
The format method can be commonised by adding a config param which contains the columns that need to be changed and their respective data types.

SteveJacksonSoft (changes requested):

DigitalLearningSolutions.Data/Helpers/DisplayStringHelper.cs
```diff
@@ -1,4 +1,4 @@
-﻿namespace DigitalLearningSolutions.Web.Helpers
+﻿namespace DigitalLearningSolutions.Data.Helpers
```
DanBloxham-sw, line 1: I've moved the DisplayStringHelper over to the Data project. Originally, it was in the Web project to discourage the generation of display strings until there were necessary in view models and views. But this is indeed a case where we want to use the PRN display string in the data project, so I've just moved the whole thing over. Hence there are a lot of file changes to modify used namespaces..

DigitalLearningSolutions.Data/Services/DelegateDownloadFileService.cs
DanBloxham-sw: Seeing as it is incredibly unlikely that this file will be empty, I've removed the fall-back code. Additionally, admins can add delegates the old fashioned way from the "Register" button, which will allow the download to work fine. Though this isn't really obvious to a user.

@SteveJacksonSoft I presume your comment above about how this shouldn't be a problem takes into account your slack message from earlier saying some admins might not have delegate accounts yet?

DigitalLearningSolutions.Data/Services/DelegateDownloadFileService.cs
DanBloxham-sw: Yep, added it. Uses the same wording as we do when we display it as requested by Kevin.

DigitalLearningSolutions.Data/Models/User/CentreAnswersData.cs
DanBloxham-sw: Not quite sure what you mean about "adding knowledge", the fields are intrinsically linked to answers through the Id. I've changed the parameter around so it should be safer. If this isn't what you had in mind, let me know.

DigitalLearningSolutions.Data/Services/DelegateDownloadFileService.cs
DanBloxham-sw: We do need it. It is not quite the same as PopulateAllDelegatesSheets. They use different columns. Additionally, PopulateDelegatesSheet populates the data different, as it uses the anonymous object approach so the column headers are just straight rips of the property names. The upload service does validation on those, and might depending on it in other ways if I recall, so it would be a bigger refactor, probably best suited to another ticket.

DigitalLearningSolutions.Data/Services/DelegateDownloadFileService.cs
DanBloxham-sw: Not easily. Only the Course Delegates download, and the new All Delegates download (which was based off the Course Delegates download) do things in exactly the same way. Every other place has its own subtlety by the looks of it, so again would probably be best suited for its own ticket I think.

DigitalLearningSolutions.Data/Services/DelegateDownloadFileService.cs
SteveJacksonSoft: Ah yeah, you can imagine the slack comment in between the two paragraphs of my comment 😂 
Basically, yes some won't have delegate accounts, but those centres will almost definitely have delegates at them.

DigitalLearningSolutions.Data/Helpers/DisplayStringHelper.cs
AlexJacksonDS: I'm kind of tempted to say you should split off the PRN functionality into its own helper, and leave the rest of the file in the web project.

This is what I did with the display names, splitting off NameQueryHelper which does the "Lastname, Firstname".

AlexJacksonDS (commented):
I think everything except one thing that I'm not sure what you mean either @SteveJacksonSoft has been resolved

DigitalLearningSolutions.Data/Models/User/CentreAnswersData.cs
SteveJacksonSoft: This is fine, sorry.

If your definition of RegistrationField.Id is "the answer number in the database" then yeah, it is intrinsically linked. That's not explicit in the codebase (might have been nicer to make it so by having it as an extra field "int? answerNumber" on RegistrationField, but I don't think it's worth it).

I think this is a big improvement over having an `int` param, because the caller no longer has to worry about what a "registration prompt number" is - it could have been the straight id, or the linkedToFieldId.

DanBloxham-sw: Nice, I'll merge this then.

But for reference, I thought the whole point of the RegistrationField enumeration was so that we had an explicit link between the codebase and answer number. The ID being the answer field number, with 7 reserved for the special case of Job Group. At least when the refactor to include this RegistrationField was implemented, the ID was used in most places where prompt numbers came about, so I thought that was the point.

SteveJacksonSoft: Yeah, and that is how it's used. It's just not explicit from the name. It's not a big logical jump to assume that that's the connection between the RegistrationField enumeration and the Answer numbers so I don't think we need to do anything.

PR #1065, opened 2022-04-11
Uar 859 populate user tables

AlexJacksonDS:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-859

### Description
Migration to transfer data from the DelegateAccounts and AdminAccounts tables into the new Users tables

### Screenshots
N/A

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data.Migrations/202204071000_AddDetailsLastCheckedToUsersTableAndRenameOnDelegateAccounts.cs
+++ b/DigitalLearningSolutions.Data.Migrations/202204071000_AddDetailsLastCheckedToUsersTableAndRenameOnDelegateAccounts.cs
@@ -1,0 +1,23 @@
+﻿namespace DigitalLearningSolutions.Data.Migrations
+{
+    using FluentMigrator;
+
+    [Migration(202204071000)]
+    public class AddDetailsLastCheckedToUsersTableAndRenameOnDelegateAccounts : Migration
+    {
+        public override void Up()
+        {
+            Alter.Table("Users").AddColumn("DetailsLastChecked").AsDateTime().Nullable();
+            Rename.Column("DetailsLastChecked").OnTable("DelegateAccounts").To("CentreSpecificDetailsLastChecked");
+            Delete.Index("IX_AdminAccounts_Email").OnTable("AdminAccounts");
+        }
+
+        public override void Down()
+        {
+            Rename.Column("CentreSpecificDetailsLastChecked").OnTable("DelegateAccounts").To("DetailsLastChecked");
+            Delete.Column("DetailsLastChecked").FromTable("Users");
+            Create.Index("IX_AdminAccounts_Email").OnTable("AdminAccounts").OnColumn("Email").Ascending()
+                .WithOptions().Unique().WithOptions().NonClustered();
+        }
+    }
+}
--- a/DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
+++ b/DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
@@ -8,6 +8,7 @@ namespace DigitalLearningSolutions.Data.Migrations
     using System.Linq;
     using System.Transactions;
     using Dapper;
+    using DigitalLearningSolutions.Data.Migrations.Properties;
     using FluentMigrator;
     using Microsoft.Data.SqlClient;
 
@@ -44,7 +45,8 @@ namespace DigitalLearningSolutions.Data.Migrations
                     HasBeenPromptedForPrn,
                     LearningHubAuthId,
                     HasDismissedLhLoginWarning,
-                    EmailVerified
+                    EmailVerified,
+                    DetailsLastChecked
                 )
                 SELECT Email,
                     Password_deprecated,
@@ -60,6 +62,7 @@ namespace DigitalLearningSolutions.Data.Migrations
                     0,
                     NULL,
                     0,
+                    GETUTCDATE(),
                     GETUTCDATE()
                     FROM AdminAccounts"
             );
@@ -67,15 +70,69 @@ namespace DigitalLearningSolutions.Data.Migrations
             // 3. Update AdminAccounts to reference Users.ID
             connection.Execute(
                 @"UPDATE AdminAccounts
-                SET UserID = (SELECT ID FROM Users WHERE Email = PrimaryEmail)"
+                    SET
+                        UserID = (SELECT ID FROM Users WHERE Email = PrimaryEmail),
+                        Email = NULL"
             );
 
             // 4. DelegateAccount
 
-            var delegateAccounts =
-                connection.Query<DelegateAccount>("SELECT * FROM DelegateAccounts");
+            // Transfer all delegates with unique emails not already in the Users table
+            connection.Execute(
+                @"INSERT INTO dbo.Users (
+                    PrimaryEmail,
+                    PasswordHash,
+                    FirstName,
+                    LastName,
+                    JobGroupID,
+                    ProfessionalRegistrationNumber,
+                    ProfileImage,
+                    Active,
+                    ResetPasswordID,
+                    TermsAgreed,
+                    FailedLoginCount,
+                    HasBeenPromptedForPrn,
+                    LearningHubAuthId,
+                    HasDismissedLhLoginWarning,
+                    EmailVerified,
+                    DetailsLastChecked
+                )
+                SELECT Email,
+                    COALESCE(OldPassword, ''),
+                    COALESCE(FirstName_deprecated, ''),
+                    COALESCE(LastName_deprecated, ''),
+                    JobGroupID_deprecated,
+                    ProfessionalRegistrationNumber_deprecated,
+                    ProfileImage_deprecated,
+                    Active,
+                    ResetPasswordID_deprecated,
+                    NULL,
+                    0,
+                    HasBeenPromptedForPrn_deprecated,
+                    LearningHubAuthID_deprecated,
+                    HasDismissedLhLoginWarning_deprecated,
+                    GETUTCDATE(),
+                    GETUTCDATE()
+                    FROM DelegateAccounts
+                    WHERE Email IN (
+                        SELECT Email FROM DelegateAccounts
+                        WHERE Email IS NOT NULL AND TRIM(Email) IS NOT NULL AND TRIM(Email) <> ''
+                        GROUP BY Email
+                        HAVING COUNT(*) = 1
+                        EXCEPT
+                        SELECT PrimaryEmail FROM Users)"
+            );
+            connection.Execute(
+                @"UPDATE DelegateAccounts
+                    SET
+                        UserID = (SELECT ID FROM Users WHERE Email = PrimaryEmail),
+                        Email = NULL,
+                        CentreSpecificDetailsLastChecked = GETUTCDATE()"
+            );
 
-            var duplicateDelegateAccountsAtCentre = new List<DelegateAccount>();
+            // Get the rest of the delegate accounts we've not resolved yet
+            var delegateAccounts =
+                connection.Query<DelegateAccount>("SELECT * FROM DelegateAccounts WHERE UserId IS NULL");
 
             var delegateAccountsGroupedByEmail = delegateAccounts.GroupBy(da => da.Email);
 
@@ -83,18 +140,33 @@ namespace DigitalLearningSolutions.Data.Migrations
             {
                 if (!string.IsNullOrWhiteSpace(emailGroup.Key))
                 {
-                    // If we have an email, we can check for existing users.
+                    var delegatesToAttemptToLink = new List<DelegateAccount>();
                     var delegateAccountsGroupedByCentre = emailGroup.GroupBy(da => da.CentreId);
 
                     foreach (var centreGroup in delegateAccountsGroupedByCentre)
                     {
-                        // If there are more than one delegate accounts with the same email at a centre
-                        // we only want to keep the email on the first. The rest get reset to having a User with no email
-                        var firstAtCentre = centreGroup.First();
+                        delegatesToAttemptToLink.Add(centreGroup.First());
 
+                        // All duplicate emails at a centre need new User accounts.
+                        // If there are more than one delegate accounts with the same email at a centre
+                        // we only want to keep the email on the first. The rest get reset to having a User with a guid email
+                        var othersAtCentre = centreGroup.Skip(1);
+
+                        foreach (var delegateAccount in othersAtCentre)
+                        {
+                            // Insert a new User with Guid email
+                            InsertNewUserForDelegateAccount(connection, delegateAccount, true);
+                        }
+                    }
+
+                    var allJobGroupsMatch =
+                        delegatesToAttemptToLink.Select(da => da.JobGroupId_deprecated).Distinct().Count() < 2;
+                    foreach (var delegateAccount in delegatesToAttemptToLink)
+                    {
+                        // Check for existing user with email
                         var existingUserWithEmail = connection.QuerySingleOrDefault<User>(
                             "SELECT * FROM Users WHERE PrimaryEmail = @email",
-                            new { email = firstAtCentre.Email }
+                            new { email = emailGroup.Key }
                         );
 
                         if (existingUserWithEmail != null)
@@ -103,74 +175,40 @@ namespace DigitalLearningSolutions.Data.Migrations
                             UpdateExistingUserWithDelegateDetails(
                                 connection,
                                 existingUserWithEmail,
-                                firstAtCentre,
-                                emailGroup
+                                delegateAccount,
+                                allJobGroupsMatch
                             );
                         }
                         else
                         {
-                            // Otherwise we create a new user
-                            InsertNewUserForDelegateAccount(connection, firstAtCentre);
+                            // Otherwise we create a new user with the email address
+                            InsertNewUserForDelegateAccount(connection, delegateAccount, false);
                         }
-
-                        var othersAtCentre = centreGroup.Skip(1);
-
-                        duplicateDelegateAccountsAtCentre.AddRange(othersAtCentre);
                     }
                 }
                 else
                 {
-                    // If we don't have a valid email we create a new user
+                    // If we don't have a valid email we create new users for each with a guid email
                     foreach (var delegateAccount in emailGroup)
                     {
-                        InsertNewUserForDelegateAccount(connection, delegateAccount);
+                        InsertNewUserForDelegateAccount(connection, delegateAccount, true);
                     }
                 }
-            }
-
-            // All duplicate emails at a centre need new User accounts
-            foreach (var delegateAccount in duplicateDelegateAccountsAtCentre)
-            {
-                InsertNewUserForDelegateAccount(connection, delegateAccount, true);
             }
 
             transactionScope.Complete();
         }
 
-        private static void UpdateDelegateAccountUserIdAndDetailsLastChecked(
-            IDbConnection connection,
-            int userId,
-            int delegateAccountId,
-            DateTime? detailsLastChecked
-        )
-        {
-            connection.Execute(
-                @"UPDATE DelegateAccounts
-                            SET
-                                UserID = @userId,
-                                DetailsLastChecked = @detailsLastChecked
-                            WHERE ID = @delegateAccountId",
-                new
-                {
-                    userId,
-                    delegateAccountId,
-                    detailsLastChecked,
-                }
-            );
-        }
-
         public override void Down()
         {
-            Execute.Sql("UPDATE DelegateAccounts SET UserId = NULL, DetailsLastChecked = NULL");
-            Execute.Sql("UPDATE AdminAccounts SET UserId = NULL");
-            Execute.Sql("DELETE Users");
+            Execute.Sql(Resources.UAR_859_PopulateUsersTableFromAccountsTables_DOWN);
         }
 
         private static void UpdateExistingUserWithDelegateDetails(
             IDbConnection connection,
             User existingUserWithEmail,
             DelegateAccount delegateAccount,
-            IEnumerable<DelegateAccount> delegateAccountsWithMatchingEmail
+            bool allJobGroupsMatch
         )
         {
             connection.Execute(
@@ -187,7 +225,8 @@ namespace DigitalLearningSolutions.Data.Migrations
                                 HasBeenPromptedForPrn = @hasBeenPromptedForPrn,
                                 LearningHubAuthId = @learningHubAuthId,
                                 HasDismissedLhLoginWarning = @hasDismissedLhLoginWarning,
-                                EmailVerified = @emailVerified
+                                EmailVerified = GETUTCDATE(),
+                                DetailsLastChecked = CASE WHEN @detailsMatched = 0 OR DetailsLastChecked IS NULL THEN NULL ELSE GETUTCDATE()
                             WHERE ID = @userId",
                 new
                 {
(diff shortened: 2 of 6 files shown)


DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
```diff
@@ -0,0 +1,360 @@
...
+                    FROM AdminAccounts"
+            );
+
+            // 3. Update AdminAccounts to reference Users.ID
+            connection.Execute(
+                @"UPDATE AdminAccounts
+                SET UserID = (SELECT ID FROM Users WHERE Email = PrimaryEmail)"
+            );
+
+            // 4. DelegateAccount
```
SteveJacksonSoft, line 73: This probably doesn't match the algorithm I laid out in the ticket (sorry!), but to shortcut pulling the vast majority of our delegates into memory and making individual queries for them, can we start the DelegateAccount step with a query which copies delegates with new (and unique) emails directly into Users? Then we can forget about these for the rest of the migration.

SteveJacksonSoft (changes requested):

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
```diff
@@ -0,0 +1,415 @@
...
+                    0,
+                    HasBeenPromptedForPrn_deprecated,
+                    LearningHubAuthID_deprecated,
+                    HasDismissedLhLoginWarning_deprecated,
+                    GETUTCDATE(),
+                    GETUTCDATE()
+                    FROM DelegateAccounts
+                    WHERE Email IN (
+                        SELECT Email FROM DelegateAccounts
+                        WHERE Email IS NOT NULL AND TRIM(Email) IS NOT NULL AND TRIM(Email) <> ''
```
SteveJacksonSoft, line 119: Is "Trim(Email) is not null" not enough?

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
```diff
@@ -0,0 +1,415 @@
...
+                                JobGroupID = @jobGroupId,
+                                ProfessionalRegistrationNumber = @professionalRegistrationNumber,
+                                ProfileImage = @profileImage,
+                                Active = @active,
+                                ResetPasswordID = @resetPasswordId,
+                                HasBeenPromptedForPrn = @hasBeenPromptedForPrn,
+                                LearningHubAuthId = @learningHubAuthId,
+                                HasDismissedLhLoginWarning = @hasDismissedLhLoginWarning,
+                                EmailVerified = GETUTCDATE(),
+                                DetailsLastChecked = CASE WHEN @detailsMatched = 0 OR DetailsLastChecked IS NULL THEN NULL ELSE GETUTCDATE()
```
SteveJacksonSoft, line 229: Looks like there should technically be an END on here. Or you can use IIF if you prefer the look of it, but I'm happy with either.

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
```diff
@@ -0,0 +1,415 @@
...
+
+            var prnMatch = delegateAccount.ProfessionalRegistrationNumber_deprecated == existingUser.ProfessionalRegistrationNumber ||
+                           string.IsNullOrWhiteSpace(existingUser.ProfessionalRegistrationNumber) ||
+                           string.IsNullOrWhiteSpace(delegateAccount.ProfessionalRegistrationNumber_deprecated);
+
+            var profileImageMatch = delegateAccount.ProfileImage_deprecated == existingUser.ProfileImage ||
+                                    existingUser.ProfileImage == null ||
+                                    delegateAccount.ProfileImage_deprecated == null;
+
+            return allJobGroupsMatch && firstNamesMatch && lastNamesMatch && profileImageMatch && prnMatch;
```
SteveJacksonSoft, line 289: This is beautiful.

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
```diff
@@ -0,0 +1,415 @@
...
+                    hasBeenPromptedForPrn = delegateAccount.HasBeenPromptedForPrn_deprecated,
+                    learningHubAuthId = delegateAccount.LearningHubAuthId_deprecated,
+                    hasDismissedLhLoginWarning = delegateAccount.HasDismissedLhLoginWarning_deprecated,
+                }
+            );
+
+            UpdateDelegateAccountUserIdEmailAndDetailsLastChecked(connection, userId, delegateAccount.Id);
+        }
+
+        private static void UpdateDelegateAccountUserIdEmailAndDetailsLastChecked(
```
SteveJacksonSoft, line 360: I think we can call this less, but need to finish work now. Remind me to come back to this.

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
```diff
@@ -0,0 +1,415 @@
...
+                                                 delegateAccount.HasDismissedLhLoginWarning_deprecated,
+                    detailsMatch = DoesDelegateAccountMatchExistingUser(delegateAccount, existingUserWithEmail, allJobGroupsMatch)
+                }
+            );
+
+            UpdateDelegateAccountUserIdEmailAndDetailsLastChecked(
+                connection,
+                existingUserWithEmail.Id,
+                delegateAccount.Id
+            );
```
SteveJacksonSoft, line 268: I think we can do this right at the end of the migration rather than after every individual update.
Because we're running this method on all inserts, the only remaining delegates with null userIds at the end of the migration will have an email address matching their user record.
So then you can do a single query for all of them as you've done for admins. Should save us a lot of time.

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
AlexJacksonDS: TRIM('   ') returns '', I figured this was the safest way to be absolutely sure I caught them all, but if anything I need to remove the TRIM(Email) IS NOT NULL

SteveJacksonSoft: Yeah fair enough, feel free to leave as is.

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
```diff
@@ -0,0 +1,415 @@
...
+
+                    var allJobGroupsMatch =
+                        delegatesToAttemptToLink.Select(da => da.JobGroupId_deprecated).Distinct().Count() < 2;
+                    foreach (var delegateAccount in delegatesToAttemptToLink)
+                    {
+                        // Check for existing user with email
+                        var existingUserWithEmail = connection.QuerySingleOrDefault<User>(
+                            "SELECT * FROM Users WHERE PrimaryEmail = @email",
+                            new { email = emailGroup.Key }
+                        );
```
DanBloxham-sw, line 170: I think this `existingUserWithEmail` definition can be moved outside of this forEach. Saves repeated calling.

UPDATE: realised we are using the repeated calling to check the updated details and compare them in an iterative methodology. I feel like this can all be done in a single call, where we do a lot of checks on the delegatesToAttemptToLink prior to it. I'd prefer that to repeatedly updating the same record multiple times, but not sure how feasible it is.

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
```diff
@@ -0,0 +1,415 @@
...
+
+                        if (existingUserWithEmail != null)
+                        {
+                            // If we find a user, we update any default data on that user
+                            UpdateExistingUserWithDelegateDetails(
+                                connection,
+                                existingUserWithEmail,
+                                delegateAccount,
+                                allJobGroupsMatch
```
DanBloxham-sw, line 179: OLD COMMENT: Shouldn't we be checking not just the jobGroups match, but that all first name, last name, profile image, PRN details, are the same across all DelegateAccounts. Otherwise we are just selecting a seemingly random set of details from one of the accounts to update with? Looks like we are going to updating the single User record with details from each of the DelegateAccounts.

NEW COMMENT AFTER READING THROUGH MORE OF THE CODE: I think this method name could be improved. As you can see from my old comment, I was getting very confused as it looked like we kept overwriting details with other ones. Mostly because the method name does not highlight that differing details won't be overwritten, and exacerbated by the fact we check the jobGroups match and none of the others. 

Why are we checking the jobGroups before calling the method? I feel like they can be done inside it along with the others.

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
```diff
@@ -0,0 +1,415 @@
...
+                                LastName = @lastName,
+                                JobGroupID = @jobGroupId,
+                                ProfessionalRegistrationNumber = @professionalRegistrationNumber,
+                                ProfileImage = @profileImage,
+                                Active = @active,
+                                ResetPasswordID = @resetPasswordId,
+                                HasBeenPromptedForPrn = @hasBeenPromptedForPrn,
+                                LearningHubAuthId = @learningHubAuthId,
+                                HasDismissedLhLoginWarning = @hasDismissedLhLoginWarning,
+                                EmailVerified = GETUTCDATE(),
```
DanBloxham-sw, line 228: Do we need to be setting EmailVerified here?

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
```diff
@@ -0,0 +1,415 @@
...
+                                  string.IsNullOrWhiteSpace(existingUser.FirstName) ||
+                                  string.IsNullOrWhiteSpace(delegateAccount.FirstName_deprecated);
+
+            var lastNamesMatch = delegateAccount.LastName_deprecated == existingUser.LastName ||
+                                 string.IsNullOrWhiteSpace(existingUser.LastName) ||
+                                 string.IsNullOrWhiteSpace(delegateAccount.LastName_deprecated);
+
+            var prnMatch = delegateAccount.ProfessionalRegistrationNumber_deprecated == existingUser.ProfessionalRegistrationNumber ||
+                           string.IsNullOrWhiteSpace(existingUser.ProfessionalRegistrationNumber) ||
+                           string.IsNullOrWhiteSpace(delegateAccount.ProfessionalRegistrationNumber_deprecated);
```
DanBloxham-sw, line 283: Do we also need to take HasBeenPromptedForPRN into account?

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
```diff
@@ -0,0 +1,415 @@
...
+                            WHERE ID = @delegateAccountId",
+                new
+                {
+                    userId,
+                    delegateAccountId
+                }
+            );
+        }
+
+        private class DelegateAccount
```
DanBloxham-sw, line 381: It's very unlikely, but I wonder if in the future, when we have defined other classes for these tables in the Data project's models, if someone is going to accidentally put a reference to these classes in their code, rather than the other classes. Do we think we should rename these to make it obvious these are only used in this migration? e.g. `DelegateAccountForUarMigration`?

DanBloxham-sw (changes requested):
I think I've covered everything. A good number of questions.

Comments might be a little out of order (some of the comment later in the files were made earlier in the reviews), so you might need to look between them a bit.

I've also not tried to answer my own questions as much as usual in the interest of time.

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
```diff
@@ -0,0 +1,415 @@
...
+    using System.Collections.Generic;
+    using System.Data;
+    using System.Linq;
+    using System.Transactions;
+    using Dapper;
+    using DigitalLearningSolutions.Data.Migrations.Properties;
+    using FluentMigrator;
+    using Microsoft.Data.SqlClient;
+
+    [Migration(202204071530, TransactionBehavior.None)]
```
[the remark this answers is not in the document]
AlexJacksonDS: The connection needed for all the stuff I'm doing is different from the one that the Migration would use if it was just doing FluentMigrations stuff like the other migrations. Can't have 2 transactions, but need one on the correct connection.

DigitalLearningSolutions.Data.Migrations/202204071000_AddDetailsLastCheckedToUsersTableAndRenameOnDelegateAccounts.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    using FluentMigrator;
+
+    [Migration(202204071000)]
+    public class AddDetailsLastCheckedToUsersTableAndRenameOnDelegateAccounts : Migration
+    {
+        public override void Up()
+        {
+            Alter.Table("Users").AddColumn("DetailsLastChecked").AsDateTime().Nullable();
+            Rename.Column("DetailsLastChecked").OnTable("DelegateAccounts").To("CentreSpecificDetailsLastChecked");
```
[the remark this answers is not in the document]
AlexJacksonDS: This has been intentionally renamed to avoid confusion with the new one on Users. There was a Slack conversation about this, it just hasn't been documented on the ticket (I'll sort this).

DigitalLearningSolutions.Data.Migrations/202204071000_AddDetailsLastCheckedToUsersTableAndRenameOnDelegateAccounts.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace DigitalLearningSolutions.Data.Migrations
+{
+    using FluentMigrator;
+
+    [Migration(202204071000)]
```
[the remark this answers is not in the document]
AlexJacksonDS: Yeah, I've sorted out the test ones so that it does them all in the ideal order when we merge this, but technically there isn't anything to stop them being in the other order.

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
```diff
@@ -0,0 +1,415 @@
...
+                        GROUP BY Email
+                        HAVING COUNT(*) = 1
+                        EXCEPT
+                        SELECT PrimaryEmail FROM Users)"
+            );
+            connection.Execute(
+                @"UPDATE DelegateAccounts
+                    SET
+                        UserID = (SELECT ID FROM Users WHERE Email = PrimaryEmail),
+                        Email = NULL,
```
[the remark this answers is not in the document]
AlexJacksonDS: I've moved this to the end since Steve also suggested this https://github.com/TechnologyEnhancedLearning/DLSV2/pull/1065#discussion_r850187659 and we can do it all at once at the end

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
```diff
@@ -0,0 +1,415 @@
...
+                    ProfileImage_deprecated,
+                    Active,
+                    ResetPasswordID_deprecated,
+                    TCAgreed_deprecated,
+                    FailedLoginCount_deprecated,
+                    0,
+                    NULL,
+                    0,
+                    GETUTCDATE(),
+                    GETUTCDATE()
```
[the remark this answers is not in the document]
AlexJacksonDS: The default value when running this migration is to try to set the DetailsLastChecked to now. When we transfer admin accounts, we can be sure they are correct, but then when we try to add a delegate to the user, it may then become null or stay populated if everything is fine

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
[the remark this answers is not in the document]
AlexJacksonDS: I don't want any emails that are null or whitespace, I think this suggestion just splits the where clause up but doesn't eliminate it?

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
AlexJacksonDS: Probably not, it'll have been set when creating the user.

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
AlexJacksonDS: I mean they are private to a Migration, I would hope that no-one would do that, and if for some stupid reason they did, then it would be caught by the review process.

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
AlexJacksonDS: No idea, @SteveJacksonSoft?

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
AlexJacksonDS: I don't really see how we can do this when the account will not exist for the first delegate with the email, but not for later ones. The migration doesn't really take very long to run, and given it only needs to run once (assuming all goes well) I don't think it's worth spending forever optimising absolutely everything.

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
```diff
@@ -0,0 +1,415 @@
...
+                            InsertNewUserForDelegateAccount(connection, delegateAccount, false);
+                        }
+                    }
+                }
+                else
+                {
+                    // If we don't have a valid email we create new users for each with a guid email
+                    foreach (var delegateAccount in emailGroup)
+                    {
+                        InsertNewUserForDelegateAccount(connection, delegateAccount, true);
```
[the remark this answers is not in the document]
AlexJacksonDS: See https://github.com/TechnologyEnhancedLearning/DLSV2/pull/1065#discussion_r850381154, same deal. If there is a single user we set this to now, if there are more users it can become unset.

DigitalLearningSolutions.Data.Migrations/Scripts/UAR-859-PopulateUsersTableFromAccountsTables-DOWN.sql
```diff
@@ -0,0 +1,88 @@
...
+    ResetPasswordID_deprecated = snapAA.ResetPasswordID_deprecated,
+    IsFrameworkContributor = snapAA.IsFrameworkContributor,
+    IsWorkforceManager = snapAA.IsWorkforceManager,
+    IsWorkforceContributor = snapAA.IsWorkforceContributor,
+    IsLocalWorkforceManager = snapAA.IsLocalWorkforceManager,
+    IsNominatedSupervisor = snapAA.IsNominatedSupervisor,
+    UserID = snapAA.UserID,
+    EmailVerified = snapAA.EmailVerified
+FROM AdminAccounts AS AA
+JOIN ' + @snapshotName + '.dbo.AdminAccounts AS snapAA ON AA.ID = snapAA.ID'
```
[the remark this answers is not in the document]
AlexJacksonDS: It doesn't have to, but I can change it so it does for clarity?

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
```diff
@@ -0,0 +1,415 @@
...
+                    GETUTCDATE(),
+                    GETUTCDATE()
+                    FROM AdminAccounts"
+            );
+
+            // 3. Update AdminAccounts to reference Users.ID
+            connection.Execute(
+                @"UPDATE AdminAccounts
+                    SET
+                        UserID = (SELECT ID FROM Users WHERE Email = PrimaryEmail),
```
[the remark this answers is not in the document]
AlexJacksonDS: I don't think the perfomance impact is enough to bother changing these for a 1 time thing.

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
AlexJacksonDS: Because we are supposed to check all the jobgroups according to the ticket. @SteveJacksonSoft Should we be changing something here?

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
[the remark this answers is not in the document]
DanBloxham-sw: I mean that, in this particular case, we are inserting a new User for this particular DelegateAccount with a GUID email. We aren't going to be linking any other accounts to this, so we won't get a chance to update the DetailsLastChecked (and the DelegateAccounts equivalent), which will be set to the current date time.

So with that in mind, a populated DetailsLastChecked would imply that a user does not need to check their details, because we assume they are correct after this migration. But in this case, the primary email will be a GUID, which I think we want them to check and correct. Whenever the user logs in to an account with a GUID as the primary email, we plan to redirect them to the Edit Detail page so they can change it.

I think this also applies to the other scenario where we set the primary email to a GUID, but this was easier to explain in this circumstance.

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
DanBloxham-sw: Fair enough. For what it's worth, I was envisioning a process where we check all the accounts first, take whatever details we would need to enter to build the "User" class, and then insert it (or update it if it exists). But then the logic to determine what to place in there would be more pernickety. What we have already implemented here is a lot simpler, I just didn't get what was going on until I'd read through it all.

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
DanBloxham-sw: Good point. I completely forgot that that was thing. Nevermind.

DanBloxham-sw (commented):
Good explanations. I've tried re-explaining a few points just to make them clearer.

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
DanBloxham-sw: I only mention the jobGroups bit because we have this whopping great method to check all the various details match, but the jobGroups are checked outside of it. From my naïve look, it seemed like that check could have been moved inside the method, as it didn't seem to be doing anything different from the others.

AlexJacksonDS: https://github.com/TechnologyEnhancedLearning/DLSV2/pull/1065#discussion_r847457945 another relevant comment chain

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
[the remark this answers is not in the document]
AlexJacksonDS: Ah OK, yeah I think you're right in the GUID case

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
SteveJacksonSoft: The logic here is right. We should set the user detailsLastChecked to null if there are any conflicts in the fields that we're merging. Most of those fields are on both delegateAccounts and adminAccounts, but jobGroup is only on delegateAccounts. Because of the way we're implementing the data migration, it's easiest to check the former iteratively and the latter all at once.
An improvement would be to pull `allJobGroupsMatch` out of `DoesDelegateAccountMatchExistingUser()` and combine the two with a && to calculate `@detailsMatch` in the User update query. But I don't _think_ it's worth doing anything more than that.

SteveJacksonSoft: Could maybe explain why `allJobGroupsMatch` is different in a comment I guess.

DigitalLearningSolutions.Data.Migrations/202204071530_PopulateUsersTableFromAccountsTables.cs
SteveJacksonSoft: Nah, if the user's been prompted for one of their delegate accounts then they've been prompted. So we can safely set that to true if it's true for any delegateAccount. We only need users to check details if we've made arbitrary decisions about which value to give a field when merging.

SteveJacksonSoft (approved):

PR #1082, opened 2022-04-22
HEEDLS-807 : Update tracking system links fixes

GeorgeBlackwell:
### JIRA link
_[HEEDLS-807](https://softwiretech.atlassian.net/browse/HEEDLS-807)_

### Description
_There are a number of links that originally pointed to a placeholder location while awaiting other work to be completed, but have not been updated._

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateProgress/EditSupervisor.cshtml
+++ b/DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateProgress/EditSupervisor.cshtml
@@ -47,7 +47,6 @@
     <form class="nhsuk-u-margin-bottom-3" method="post" asp-action="EditSupervisor">
       <input type="hidden" asp-for="DelegateId" />
       <input type="hidden" asp-for="ProgressId" />
-
       <vc:select-list asp-for="@nameof(Model.SupervisorId)"
                       label="Supervisor"
                       value="@Model.SupervisorId.ToString()"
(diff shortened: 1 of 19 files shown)


AlexJacksonDS (approved):

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateProgress/EditSupervisor.cshtml
```diff
@@ -61,9 +63,18 @@
     </form>
 
     @if (Model.AccessedVia.Equals(DelegateAccessRoute.CourseDelegates)) {
-      <vc:cancel-link asp-controller="DelegateProgress" asp-action="Index" asp-all-route-data="@routeParamsForBackLink" />
+      <vc:cancel-link-with-return-page-query
+        asp-controller="DelegateProgress"
+        asp-action="Index"
+        route-data="@routeParamsForBackLink"
+        return-page-query="@Model.ReturnPageQuery" />
```
DanBloxham-sw, line 70: Couple of things for the modifications to the links of these Edit pages. 

Firstly, the DelegateProgress page doesn't have anywhere that it can use the return page query stuff for, as it does not have any cards. I also don't think that it can pass back the return page query to the view delegates page as the only link backwards would be the breadcrumbs, which we aren't preserving the return page query for. So I'm not sure on the usefulness of this.

Secondly, this has been applied to the CourseDelegates version of the cancel link, but those pages do not pass in the return page query. As such, this page will fail to load when accessing from CourseDelegates/DelegateProgress. Admittedly, we are going to be changing those pages around in the big merge ticket, but it might be nice to not break the functionality until that is complete. 

So, overall, we can probably keep this as it was.

DanBloxham-sw (changes requested):
I think there have been a few unnecessary changes in trying to preserve the page position that have broken some pages in some circumstances.

DanBloxham-sw (approved):
Nice and simple. Would be good to leave a comment about the scrolling being excluded for now in a comment on the ticket if it hasn't already. And its probably an easily missed point in the big merge ticket, so could make an explicit mention of there just so we don't forget.

PR #1091, opened 2022-04-29
HEEDLS-820 - trimmed custom prompts so JS filtering functions correctly

DanBloxham-sw:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-820

### Description
Implemented trimming of custom prompts answers to get rid of any unnecessary whitespace, which would cause the JS filtering to fail. 

Such whitespace should now be removed when prompts answers are added, so future answers should not cause the issue. The retrieval of answers used in producing the filters, and for comparing to delegate values, also are trimmed, so existing prompt answers won't be an issue.

It's worth noting that the Registration Prompt saving does not allow duplicate prompts, including with different variations of whitespace that get trimmed away. However, duplicates are allowed on the Admin Field prompts, I don't know if this was intentional in the past, but I have left it alone for now.

### Screenshots
N/A

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


AlexJacksonDS (commented):
I think this looks mostly fine, but there are other places where we query Answers (found by searching for Answer1 in the solution). Do we need this trimming on any of the other ones?

JBsoftwire:
This looks fine to me, but I'll second Alex's question.

DanBloxham-sw:
I don’t believe we need to trim any other places where we extract answers in the data services. I’ve tried to go for a minimalist approach here to fix the functionality without touching anything else that wouldn’t need the fix, to help reduce redundant testing. 

When devving this, I set up prompts/answers that would cause the bug and retained these to test throughout the devwork. As the cause of the issue is when JS interacts with explicit custom prompt answers for filtering, I made sure that existing prompts would not cause any issues, and newly created or edited prompts would not introduce the bad whitespace. I’m pretty sure the current fixes cover all places where we filter by and create/edit prompt answers.

We could still start chucking TRIMs everywhere, but as everywhere else pertains to displaying answers (or choosing answers for the Candidate record etc.), these won’t functionally make a difference. I can’t really think of any positives to doing this, or any major issues this would solve.

Admittedly, there might be a specific set of circumstances that this would affect. Say a centre has set up custom prompts with bad whitespace prior to this change, and some delegates have selected the bad answers. If an admin then does some minor modification to the options, the options will have their whitespace removed, but the Candidate record will still have the old bad whitespace answer in the DB. There are places on the site where we display the number of answers to prompts. These will suddenly change once the admin removes the whitespaces, as they no longer perfectly match. 

Throwing TRIMs everywhere would fix this I think. But the regression testing for everything that gets affected would be quite large, and mostly unimportant. Do we think it is worth the hassle?

AlexJacksonDS:
> Admittedly, there might be a specific set of circumstances that this would affect. Say a centre has set up custom prompts with bad whitespace prior to this change, and some delegates have selected the bad answers. If an admin then does some minor modification to the options, the options will have their whitespace removed, but the Candidate record will still have the old bad whitespace answer in the DB. There are places on the site where we display the number of answers to prompts. These will suddenly change once the admin removes the whitespaces, as they no longer perfectly match.
>
> Throwing TRIMs everywhere would fix this I think. But the regression testing for everything that gets affected would be quite large, and mostly unimportant. Do we think it is worth the hassle?

Yeah I think that is a fairly unlikely scenario, and one that I don't know it can have occurred yet given the new tracking system is unreleased. It's probably not worth the effort

AlexJacksonDS (approved):

SteveJacksonSoft:
I think it's fine as is. In general, if admins change the options for course admin fields or registration prompts, then we get problems with the count anyway. So I can't imagine they'd be much more put out if they made a whitespace change to an option and it messes things up.

PR #1094, opened 2022-05-03
HEEDLS-861 Remove role alert from results count

AlexJacksonDS:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-861

### Description
Removed the role alert from the results count so that Firefox doesn't read out all the alerts twice.

### Screenshots
N/A

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Web/Views/Shared/SearchablePage/_SearchResultsCount.cshtml
```diff
@@ -5,5 +5,5 @@
   @{
     var count = Model.MatchingSearchResults;
   }
-  <span role="alert" id="results-count">@count matching @(count == 1 ? "result" : "results")</span>
+  <span id="results-count">@count matching @(count == 1 ? "result" : "results")</span>
```
DanBloxham-sw, line 8: Not sure this is the best idea. This will presumably get rid of all announcements of the number of results. If we are fine with that, then cool. Otherwise, there is an alternative method. 

The SearchSortFilterAndPaginate js file updates the alert count, but if the count is the same as it was before, we add some whitespace to it to force screen readers to announce it again. I think I added this to prevent confusion when typing quickly, I can look into that reasoning more if it doesn't make sense. 

Previously, we had a parameter somewhere in the SearchSortFilterPaginate file to not update the result count in JS when searching/sorting/filtering/paginating. This could then be true in the constructor, to avoid "updating" the result count on page load. I'm not sure if that was removed, or the reasoning why, might be able to look through the changes to see. But reinstating that might work, while keeping the regular announcements.

DanBloxham-sw (changes requested):
Not sure this is the best idea, but not exactly sure what we want.

DigitalLearningSolutions.Web/Views/Shared/SearchablePage/_SearchResultsCount.cshtml
SteveJacksonSoft: Alex and I were talking about it and we couldn't think of much benefit in having this read out as an alert. The page number is also an alert field, so if the number of pages change (or the number of the page the user is on changes) then that will get read out. I now realise that if filtering reduces the number of pages to 1 the page element disappears and doesn't get read out.

Revolutionary idea though, what if we were to combine the searchResultsCount alert and the page number alert in a hidden span? Then we have only one alert and get both when either number changes.

AlexJacksonDS: That might work, will give it a try

DanBloxham-sw: I think the main benefit of the result count alert was so users would get updated that something has "happened" when they are typing in a search box for instance. A user that doesn't need a screen reader will easily notice that the page number and result count have changed, and an alert would help a screen reader user be similarly aware of it. 

Hidden span might work, would have to be careful with how we hide it to make sure the screen reader still announces it. Though it does make more sense to me to use the already visually displayed page elements as alerts over introducing new hidden ones.

SteveJacksonSoft: The motivation for removing one of them was that they seemed to be interrupting each other on Firefox so that only one was getting read out. So 1 should be better than 2.

DigitalLearningSolutions.Web/Views/Shared/PaginatedPage/_BottomPaginationNavWithoutAlert.cshtml
```diff
@@ -0,0 +1,6 @@
+﻿@using DigitalLearningSolutions.Web.ViewModels.Common.SearchablePage
+@model IBasePaginatedViewModel
+
+<nav class="nhsuk-pagination @(Model.TotalPages == 1 ? "pagination-hidden" : "")" role="navigation" aria-label="Bottom Pagination">
```
SteveJacksonSoft, line 4: Is the aria-label just there to determine what the screen-reader reads out?

SteveJacksonSoft: The motivation behind my question is, "It would be nice if we can simplify the codebase by making the top and bottom ones the same.

AlexJacksonDS: Also so that the landmarks are unique. We get accessibility errors when both navigation have the same label as they are not unique.
"Landmarks must have a unique role or role/label/title (i.e. accessible name) combination"

PR #1098, opened 2022-05-06
HEEDLS-871 Add category topic tags and fix incorrect behaviour

AlexJacksonDS:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-871

### Description
Fixed some bugs with sort and filter and some screen reader issues. Added category and topic tags.

### Screenshots
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Web/Helpers/DisplayStringHelper.cs
```diff
@@ -83,5 +83,10 @@ public static string GetTimeStringFromMinutes(int minutes)
         {
             return minutes < 60 ? $"{minutes}m" : $"{minutes / 60}h {minutes % 60}m";
         }
+
+        public static string GetTimeStringForScreenReaderFromMinutes(int minutes)
```
DanBloxham-sw, line 87: Do we fancy any unit tests for this?

DigitalLearningSolutions.Web/Views/LearningContent/_BrandCourseCard.cshtml
```diff
@@ -22,29 +31,24 @@
...
       </span>
 
-      <div>
-        <div class="header-row nhsuk-u-font-weight-bold">
-          <div class="value header-row-content">Section</div>
-          <div class="value header-row-content">Tutorials</div>
-        </div>
+      <table>
+        <tr class="header-row nhsuk-u-font-weight-bold">
+          <th class="value header-row-content">Section</th>
```
DanBloxham-sw, line 42: The NHS tables have specific roles for different table elements. Do we want them here?
https://service-manual.nhs.uk/design-system/components/table

DigitalLearningSolutions.Web/Helpers/DisplayStringHelper.cs
```diff
@@ -83,5 +83,10 @@ public static string GetTimeStringFromMinutes(int minutes)
         {
             return minutes < 60 ? $"{minutes}m" : $"{minutes / 60}h {minutes % 60}m";
         }
+
+        public static string GetTimeStringForScreenReaderFromMinutes(int minutes)
+        {
+            return minutes < 60 ? $"{minutes} minutes" : $"{minutes / 60} hours {minutes % 60} minutes";
```
DanBloxham-sw, line 89: Just double checking, I'm guessing we are fine with "1 hour 0 minutes" rather than going for just "1 hour" whenever we have a whole number of hours? Same sort of question would apply to the tag content of "1h 0m".

DanBloxham-sw (changes requested):
Looks good, couple of minor points.

Also I'm not sure Kevin's response gives a clear answer as to which filters we want to display, might be worth following up on that.

DigitalLearningSolutions.Web/Views/LearningContent/_BrandCourseCard.cshtml
AlexJacksonDS: I don't believe we need them since we are using the correct html elements, looking at the various bits of documentation on them e.g. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/row_role.
e.g. I probably should have had them for the table structure with divs that I had before.

GeorgeBlackwell (commented):
Nice work, looks like Dan covered it

DigitalLearningSolutions.Web/Helpers/DisplayStringHelper.cs
AlexJacksonDS: I think it's fine where we have data down to the minute. Also the old site displays these as 1h 0m, and it's been to show and tell.

PR #1103, opened 2022-05-10
HEEDLS-895 : UAR Migration 7: Pull centre emails out into separate table

GeorgeBlackwell:
### JIRA link
_[HEEDLS-895](https://softwiretech.atlassian.net/browse/HEEDLS-895)_

### Description
_In early UAR we added email fields on `AdminAccounts` and `DelegateAccounts` for centre-specific notifications. Rather than allowing an admin to have a primary email, an email for their delegate account and an email for their admin account, we should only allow them to have a single secondary email. We’ll pull Email out of `AdminAccounts` and `DelegateAccounts` and put it in a new table called `UserCentreDetails`_

-----
### Developer checks
(Leave tasks unticked if they haven't been appropriate for your ticket.)

I have:
- [X] Run the formatter and made sure there are no IDE errors.
- [ ] Written tests for the changes (accessibility tests, unit tests for controller, data services, services, view models, etc)
- [ ] Manually tested my work with and without JavaScript. Full manual testing guidelines can be found here: https://softwiretech.atlassian.net/wiki/spaces/HEE/pages/6703648740/Testing
- [ ] Updated/added documentation in Swiki and/or Readme. Links (if any) are below:
  - [doc_1_here](link_1_here)
- [X] Updated my Jira ticket with information about other parts of the system that were touched as part of the MR and have to be sanity tested to ensure nothing’s broken.
- [X] Scanned over my own MR to ensure everything is as expected.

PR #1110, opened 2022-05-12
Heedls 856 update registration data service uar

JBsoftwire:
### JIRA link
_[HEEDLS-856](https://softwiretech.atlassian.net/browse/HEEDLS-856)_

### Description
The `RegisterAdmin` method in the `RegistrationDataService` now takes a userId as parameter (not hooked up to the frontend yet), and makes an `AdminAccount` entry linked to that ID in the `Users` table.

### Screenshots
N/A

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -211,7 +211,7 @@ public void RegisterCentreManager(AdminRegistrationModel registrationModel, int
 
             CreateDelegateAccountForAdmin(registrationModel, jobGroupId);
 
-            registrationDataService.RegisterAdmin(registrationModel);
+            registrationDataService.RegisterAdmin(registrationModel, 0);
```
AlexJacksonDS, line 214: Probably need to stick a TODO against both of these 0s in this file just to make absolutely sure they get done.

SteveJacksonSoft, line 214: Can we stick todos on these two lines? It's a bit scary if you come across a bit of code with weird hard-coded values and no explanation for it.

AlexJacksonDS (commented):
Just a quick comment about marking TODOs, other than that this looks fine.

DigitalLearningSolutions.Data.Tests/DataServices/RegistrationDataServiceTests.cs
```diff
@@ -167,10 +177,14 @@ public void Sets_notification_preferences_correctly_on_centre_manager_admin_regi
...
+                    email: "bn.uhrawol@sk",
+                    categoryId: 1
+                );
 
             // When
-            service.RegisterAdmin(registrationModel);
+            service.RegisterAdmin(registrationModel, 4046);
 
             // Then
             var user = userDataService.GetAdminUserByEmailAddress(registrationModel.PrimaryEmail)!;
```
SteveJacksonSoft, line 190: Does this definitely get the user you've just added? You're not longer setting the email address, so I'm a bit suspicious.

SteveJacksonSoft: Similar in other tests

DigitalLearningSolutions.Data/DataServices/RegistrationDataService.cs
```diff
@@ -174,71 +175,69 @@ OUTPUT Inserted.ID
...
+                isContentCreator = registrationModel.IsContentCreator,
+                isContentManager = registrationModel.IsContentManager,
                 importOnly = registrationModel.ImportOnly,
-                trainer = registrationModel.IsTrainer,
-                supervisor = registrationModel.IsSupervisor,
-                nominatedSupervisor = registrationModel.IsNominatedSupervisor
+                isTrainer = registrationModel.IsTrainer,
+                isSupervisor = registrationModel.IsSupervisor,
+                isNominatedSupervisor = registrationModel.IsNominatedSupervisor,
+                // deprecated columns
```
SteveJacksonSoft, line 196: *"Currently non-null deprecated columns" would be clearer

In a few weeks they'll be nullable just like all the other deprecated columns and people will wonder why we're arbitrarily choosing some deprecated columns to update

SteveJacksonSoft (changes requested):
Nice, just a few tweaks that would be good.

SteveJacksonSoft:
@JBsoftwire Could you also wait for George to merge [his PR](https://github.com/TechnologyEnhancedLearning/DLSV2/pull/1103) and include the UserCentreDetails insert in this once he's done that? What goes around comes around, I hear :P Also he's nearly ready to merge

DigitalLearningSolutions.Data/DataServices/UserDataService/UserCentreDetailsDataService.cs
```diff
@@ -0,0 +1,96 @@
...
+{
+    using System.Collections.Generic;
+    using System.Data;
+    using Dapper;
+    using DigitalLearningSolutions.Data.Enums;
+    using DigitalLearningSolutions.Data.Exceptions;
+
+    public partial class UserDataService
+    {
+        public void CreateOrUpdateUserCentreDetails(
```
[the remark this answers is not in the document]
SteveJacksonSoft: Also, can we leave an XmlDoc warning people that this method takes out an exclusive lock on the Users and UserCentreDetails tables, so they should be very careful about attaching this to their transaction?

DigitalLearningSolutions.Data/DataServices/UserDataService/UserCentreDetailsDataService.cs
```diff
@@ -0,0 +1,96 @@
...
+            int userId,
+            int centreId,
+            string email,
+            IDbTransaction? transaction = null
```
SteveJacksonSoft, line 15: If this is null, we should open our own transaction so that the locks we take out when checking if an email is already present are held until we finish adding the new email.

DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -259,20 +260,28 @@ public void PromoteDelegateToAdmin(AdminRoles adminRoles, int categoryId, int de
...
         {
-            ValidateRegistrationEmail(delegateRegistrationModel);
-
             string candidateNumber;
 
             try
             {
+                ValidateRegistrationEmail(delegateRegistrationModel);
                 candidateNumber =
                     registrationDataService.RegisterNewUserAndDelegateAccount(delegateRegistrationModel);
```
SteveJacksonSoft, line 275: We can remove the `candidateNumber` var and put a `return` here.

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests/UserCentreDetailsDataServiceTests.cs
```diff
@@ -0,0 +1,63 @@
...
+            result.Should().Be(expectedResult);
+        }
+
+        [Test]
+        [TestCase(null, false)]
+        [TestCase("vrmei.@akwnthsbcu", false)]
+        [TestCase("test@gmail.com", true)]
+        [TestCase("sample@admin.email", false)]
+        [TestCase("sample@delegate.email", true)]
+        public void EmailIsInUseByOtherUser_throws_if_email_is_in_use_by_other_user(string email, bool expectedResult)
```
SteveJacksonSoft, line 46: `throws`?

DigitalLearningSolutions.Data/Models/User/UserCentreDetails.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace DigitalLearningSolutions.Data.Models.User
+{
+    public class UserCentreDetails
```
SteveJacksonSoft, line 3: Is this class being used?

DigitalLearningSolutions.Data/DataServices/UserDataService/UserCentreDetailsDataService.cs
```diff
@@ -0,0 +1,96 @@
...
+                );
+            }
+        }
+
+        public bool EmailIsInUseByOtherUser(int userId, string email)
+        {
+            return connection.QuerySingle<bool>(
+                @"SELECT CASE
+                        WHEN EXISTS (SELECT ID FROM USERS WITH (UPDLOCK, TABLOCKX, HOLDLOCK) WHERE PrimaryEmail = @email AND ID <> @userId)
+                        OR EXISTS (SELECT ID FROM UserCentreDetails d WITH (UPDLOCK, TABLOCKX, HOLDLOCK)
```
SteveJacksonSoft, line 75: I'm hoping that we don't need an exclusive lock here. Hoping Mike McLean can give me some answers... Watch this space.

SteveJacksonSoft: Talked to Mike. I don't think the case of two people updating their email to the same thing in between when the check happens and when the update happens is something we have to worry about. We have around 500,000 users, and suppose they each change their email twice a year, and 10% the time that email is the same as someone else's email (very generous), then we have an average of one conflict every ~300s. For the checks/updates to interleave they'd have to be within 100ms of each other, probably. Realistically, it's not going to happen.

So let's check that no other user is using the email just before updating, but not worry about locking.

DigitalLearningSolutions.Data/DataServices/UserDataService/UserCentreDetailsDataService.cs
```diff
@@ -71,8 +72,8 @@ public bool EmailIsInUseByOtherUser(int userId, string email)
         {
             return connection.QuerySingle<bool>(
                 @"SELECT CASE
-                        WHEN EXISTS (SELECT ID FROM USERS WITH (UPDLOCK, TABLOCKX, HOLDLOCK) WHERE PrimaryEmail = @email AND ID <> @userId)
-                        OR EXISTS (SELECT ID FROM UserCentreDetails d WITH (UPDLOCK, TABLOCKX, HOLDLOCK)
+                        WHEN EXISTS (SELECT ID FROM USERS WHERE PrimaryEmail = @email AND ID <> @userId)
+                        OR EXISTS (SELECT ID FROM UserCentreDetails d
                             WHERE d.Email = @email AND d.Email IS NOT NULL AND d.UserID <> @userId)
```
SteveJacksonSoft, line 77: Just noticed we've got `AND d.Email IS NOT NULL` -  I don't think we want that. If `email` is not null then `d.Email = email` implies `d.Email` is not null. If we do pass `email = null` into this method then the method should return true if there are other users with null emails (or that's what its name claims). Same in method below.

DigitalLearningSolutions.Data/DataServices/UserDataService/UserCentreDetailsDataService.cs
```diff
@@ -15,7 +15,13 @@ public partial class UserDataService
             IDbTransaction? transaction = null
         )
         {
-            if (EmailIsInUseByOtherUser(userId, email))
+            if (transaction == null)
+            {
+                connection.Open();
```
SteveJacksonSoft, line 20: I think it'd be good to add an extension method `EnsureOpen()` which checks before calling Open() - it does sound like other people have had issues with calling Open() when the connection is already open, so I'd rather not rely on the implementation. And it wouldn't be much extra work to be careful.

PR #1137, opened 2022-05-24
Uar 884 extend my account journey

VladNicol:
### JIRA link
[HEEDLS-884](https://softwiretech.atlassian.net/jira/software/c/projects/HEEDLS/boards/753?modal=detail&selectedIssue=HEEDLS-884)

### Description

- I changed the information displayed on MyAccount page and EditDetails page (I am failing a UI test and I am not sure why)
- I update DetailsLastChecked and CentreSpecificDetailsLastChecked in the database
- I changed the unique email validation to use `UserDataService.EmailIsInUseByOtherUser`
- I removed the password check when updating details (and I renamed `UserService.UpdateUserAccountDetailsForAllVerifiedUsers` to `UserService.UpdateUserAccountDetailsForAllUsers`)
- When I update the account details, I call `UserDataService.SetCentreEmail` with the `transaction` parameter null. This opens a new transaction inside the method. I added some code to close the transaction in case the parameter in null (I am not sure this is right). I was getting an exception in the next method that was getting called, which was `groupsService.SynchroniseUserChangesWithGroups` and it looked like the transaction not being committed was the reason.
- The 'Default' tag in the centre specific details looks bad, but I am not sure how to make it better (second screenshot)

### Screenshots
image
image
image
image
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


SteveJacksonSoft:
@VladNicol Couple of visual things

1. There's a line at the bottom of the tables, below the profile picture and below the last registration prompt. Can we get rid of those? 
2. On the MyAccount page (not the edit details page), you're showing `Centre Email (optional)`. But there'll always be a value for that, it'll just default to the primary email if no secondary email is set. So it feels a bit strange to say (optional). Can we get rid of that?
3. It would be nicer to have the centre email `Default` tag in the table rather than beside it. You could try just inlining it with the centre email, or you might want to add a new column to the table just for the tag - have a play around and see what works best.
4. On the EditDetails page, there's no space between `Centre Email` and `(optional)`. Can you add one?

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -56,6 +56,7 @@ public IActionResult Index(DlsSubApplication dlsSubApplication)
                     delegateUser
                 );
 
+            delegateUser.EmailAddress = userService.GetCentreEmail(User.GetUserId()!.Value, User.GetCentreId());
```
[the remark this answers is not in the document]
AlexJacksonDS: Also both the index and edit details will fail for basic users with no centre. I assume we're going to sort that out in a leter ticket @SteveJacksonSoft ?

DigitalLearningSolutions.Web/ViewModels/MyAccount/MyAccountViewModel.cs
```diff
@@ -23,8 +24,9 @@ DlsSubApplication dlsSubApplication
             ProfilePicture = adminUser?.ProfileImage ?? delegateUser?.ProfileImage;
             Centre = adminUser?.CentreName ?? delegateUser?.CentreName;
             DelegateNumber = delegateUser?.CandidateNumber;
-            AliasId = delegateUser?.AliasId;
             JobGroup = delegateUser?.JobGroupName;
+            CentreEmail = delegateUser?.EmailAddress;
+            DateRegistered = delegateUser?.DateRegistered?.ToString("dd-MM-yyyy");
```
AlexJacksonDS, line 29: Is this the correct date format? We ususally use DigitalLearningSolutions.Data.Helpers.DateHelper.StandardDateFormat

DigitalLearningSolutions.Data/Models/User/CentreAnswersData.cs
```diff
@@ -28,6 +30,7 @@ public class CentreAnswersData
 
         public int CentreId { get; set; }
         public int JobGroupId { get; set; }
```
AlexJacksonDS, line 32: Job group is no longer centre specific so needs moving to the other object MyAccountDetailsData

DigitalLearningSolutions.Data/DataServices/UserDataService/UserCentreDetailsDataService.cs
```diff
@@ -102,5 +110,23 @@ ELSE 0
...
+                    FROM UserCentreDetails
+                    WHERE UserID = @userId AND CentreID = @centreId",
+                new { userId, centreId}
+            ).ToList();
+            if (emails.Count > 1)
+            {
+                throw new MultipleUserAccountsFoundException(
+                    $"User with id {userId} has several emails at centre with id {centreId}"
+                );
+            }
```
AlexJacksonDS, line 127: I don't think this exception is possible, there is a unique constraint on CentreID & Email (where not null), and one on CentreId & UserID, so each user can only have a single record per centre.

AlexJacksonDS (commented):
Several suggestions, particularly for the tidying up of the UserService logic we're using here so that we're actually updating the correct things at the correct time. Hopefully the comments make sense, but if they become too out of order because of github, I'll edit them to make more sense.

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -182,46 +190,50 @@ List<DelegateUser> delegateUsers
         }
 
         // TODO HEEDLS-887 Make sure this logic is correct with the new account structure
-        public void UpdateUserAccountDetailsForAllVerifiedUsers(
+        public void UpdateUserAccountDetailsForAllUsers(
```
livzorn, line 193: This method name sounds to me like it's going to update all Users in the database. Maybe something like `UpdateUserDetailsForAllLinkedAccounts` would be clearer?

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -145,11 +139,30 @@ DlsSubApplication dlsSubApplication
...
+                    formData.Email!,
+                    userAdminId,
+                    userDelegateId,
+                    userId
+                ))
             {
                 ModelState.AddModelError(
                     nameof(MyAccountEditDetailsFormData.Email),
-                    "A user with this email address is already registered at this centre"
+                    "This email address is already in use"
```
livzorn, line 151: Maybe we could add this error message to the CommonValidationErrorMessages helper where the other email-related error messages are?

DigitalLearningSolutions.Data/Services/UserService.cs
AlexJacksonDS: I'm not sure about that name either, as we have no concept of linked accounts like we used to. I'd probably prefer something like `UpdateUserDetailsAndCentreSpecificDetails`

VladNicol:
I think I addressed your comments, I hope I didn't misunderstand (or completely missed) any. I have also fixed that minor problem I had at the demo today that didn't allow me to set the centre email back to null.

I assigned the ticket to Olivia as second reviewer, but @AlexJacksonDS if you could also have another look that would be great!

DigitalLearningSolutions.Web/Views/MyAccount/EditDetails.cshtml
```diff
@@ -14,7 +14,7 @@
 <div class="nhsuk-grid-row">
   <div class="nhsuk-grid-column-full">
     @if (errorHasOccurred) {
-      <vc:error-summary order-of-property-names="@(new[] { nameof(Model.FirstName), nameof(Model.LastName), nameof(Model.Email), nameof(Model.HasProfessionalRegistrationNumber), nameof(Model.ProfessionalRegistrationNumber), nameof(Model.Password), nameof(Model.JobGroupId), nameof(Model.Answer1), nameof(Model.Answer2), nameof(Model.Answer3), nameof(Model.Answer4), nameof(Model.Answer5), nameof(Model.Answer6) })" />
+      <vc:error-summary order-of-property-names="@(new[] { nameof(Model.FirstName), nameof(Model.LastName), nameof(Model.Email), nameof(Model.JobGroupId), nameof(Model.HasProfessionalRegistrationNumber), nameof(Model.ProfessionalRegistrationNumber), nameof(Model.CentreEmail), nameof(Model.Answer1), nameof(Model.Answer2), nameof(Model.Answer3), nameof(Model.Answer4), nameof(Model.Answer5), nameof(Model.Answer6) })" />
```
livzorn, line 17: Is it possible to break this over multiple lines?

DigitalLearningSolutions.Data/Models/User/CentreAnswersData.cs
livzorn: can we remove job group from this model now that we've added it to MyAccountDetailsData? @VladNicol

DigitalLearningSolutions.Web/Views/MyAccount/EditDetails.cshtml
```diff
@@ -53,35 +53,74 @@
...
-          asp-for="@nameof(Model.JobGroupId)"
-          label="Job group"
-          value="@Model.JobGroupId.ToString()"
-          hint-text=""
-          deselectable="false"
-          css-class="nhsuk-u-width-one-half"
-          default-option="Select a job group"
-          select-list-options="@Model.JobGroups" />
+
+        @{ var centreEmailHintText = "If you supply a centre email, all notifications relating to your activity at this centre will be sent to this email address. Your primary email address will still be used for logging in and notifications for managing your account."; }
```
livzorn, line 114: Can we break this hint text over multiple lines?

livzorn (commented):
Nice work, a few minor things left

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -182,46 +190,50 @@ List<DelegateUser> delegateUsers
...
+                    centreAnswersData!.JobGroupId,
+                    userId
                 );
 
                 var oldDelegateDetails =
-                    verifiedDelegateUsers.SingleOrDefault(u => u.Id == myAccountDetailsData.DelegateId);
+                    delegateUsers.SingleOrDefault(u => u.Id == myAccountDetailsData.DelegateId);
 
-                if (oldDelegateDetails != null && centreAnswersData != null)
+                if (oldDelegateDetails != null)
```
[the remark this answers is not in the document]
AlexJacksonDS: We're still doing some unnecessary stuff here. Why do we need to fetch the oldDelegateUserDetails from the database when we've already fetched them from the database by user ID earlier? I think we should try to get the delegate user out of the previous list by id, and then only update the delegate account if it is found.

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -193,77 +201,74 @@ List<DelegateUser> delegateUsers
...
         {
-            var (verifiedAdminUser, verifiedDelegateUsers) =
-                GetVerifiedLinkedUsersAccounts(
-                    myAccountDetailsData.AdminId,
-                    myAccountDetailsData.DelegateId,
-                    myAccountDetailsData.Password
-                );
+            var delegateUsers = userDataService.GetDelegateAccountsByUserId(userId).ToList();
+
+            var detailsLastChecked = DateTime.Now;
```
AlexJacksonDS, line 212: Didn't notice this before, but we should be using ClockService.UtcNow for this

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/EditDelegateController.cs
```diff
@@ -75,11 +75,11 @@ public IActionResult Index(EditDelegateFormData formData, int delegateId)
                 return ReturnToEditDetailsViewWithErrors(formData, delegateId, centreId);
             }
 
-            if (!userService.NewEmailAddressIsValid(formData.Email!, null, delegateId, centreId))
+            if (!userService.NewEmailAddressIsValid(formData.Email!, null, delegateId, User.GetUserId()!.Value))
             {
                 ModelState.AddModelError(
                     nameof(EditDetailsFormData.Email),
-                    "A user with this email address is already registered at this centre"
+                    "This email address is already in use"
```
AlexJacksonDS, line 82: This also needs to be the common error message

DigitalLearningSolutions.Data/Services/UserService.cs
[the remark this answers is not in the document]
AlexJacksonDS: @VladNicol In case you miss this comment that I've unresolved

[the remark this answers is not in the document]
VladNicol: I kept that `oldDelegateUserDetails` because `userDataService.GetDelegateAccountsByUserId` returns a `IEnumerable<DelegateAccount>`, while `groupsService.SynchroniseUserChangesWithGroups` takes as argument a `DelegateUser`. Do we have a nice way to map a `DelegateAccount` to a `DelegateUser`? They have almost the same fields, in cases like this I am used to AutoMapper but we don't seem to use that in the project.

DigitalLearningSolutions.Data/Models/User/CentreAnswersData.cs
VladNicol: That's still used in some places, it didn't feel like part of the scope of this ticket to remove those. Should I do that?

DigitalLearningSolutions.Web/Views/MyAccount/EditDetails.cshtml
VladNicol: Yeah, the auto formatter put that on one line

DigitalLearningSolutions.Data/Models/User/CentreAnswersData.cs
livzorn: Hmm maybe not, it does seem heavily used in that model. @AlexJacksonDS do you have an opinion?

AlexJacksonDS: I guess it's probably fine to leave for now until remaining UAR work gets to it. I would expect when UAR is complete, this should be removed, so lets leave a TODO pointing to the cleanup ticket 887 that this is expected to be removed at some point.

DigitalLearningSolutions.Data/Services/UserService.cs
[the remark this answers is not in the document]
AlexJacksonDS: We don't have a way to map them, but I do wonder whether changing the group synchronization so that it works with the Delgate accounts would be something in scope here. Otherwise I do think it would be good to at some point at least, since I expect we will probably end up with DelegateAccounts in the other place this is called. @SteveJacksonSoft Do we have a ticket to resolve this bit of functionality specifically?

DigitalLearningSolutions.Web/Views/MyAccount/EditDetails.cshtml
SteveJacksonSoft: A nice way around would be to pull the `new[] {...}` out into a variable. Then the autoformatter would break it into separate lines I think.

DigitalLearningSolutions.Data/Services/UserService.cs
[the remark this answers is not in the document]
VladNicol: I don't think there is a nice and easy way around this, at least until `JobGroupId` is removed from `CentreAnswersData`. We have `GroupsService.SynchroniseUserChangesWithGroups` that calls `LinkedFieldHelper.GetLinkedFieldChanges(CentreAnswersData oldAnswers, CentreAnswersData newAnswers, ...)`.

I think we have to first remove `JobGroupId` from `CentreAnswersData` and then create a `DelegateAccount.GetCentreAnswersData` similar to `DelegateUser.GetCentreAnswersData`

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -193,77 +201,75 @@ List<DelegateUser> delegateUsers
...
+                );
             }
         }
 
-        public bool NewEmailAddressIsValid(string emailAddress, int? adminUserId, int? delegateUserId, int centreId)
+        public bool NewEmailAddressIsValid(
+            string emailAddress,
+            int? adminUserId,
+            int? delegateUserId,
+            int? userId = null
```
SteveJacksonSoft, line 261: I'm a bit scared that this method is allowing people to pass in adminId and delegateId and userId.

What happens if someone decides to call this method with a userId and a delegateId which correspond to different users?

We could avoid this by disallowing any calling code from passing in adminUserId or delegateUserId, demanding that it gives us a userId and centreId instead (which makes the implementation of this method simpler).
Can we do this instead?

VladNicol: I don't think I need the centre id, I can just return `!userDataService.EmailIsInUseByOtherUser(userId, emailAddress)`, or am I missing something?

AlexJacksonDS: For checking UserCentreDetails.

DigitalLearningSolutions.Data/Services/UserService.cs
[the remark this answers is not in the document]
SteveJacksonSoft: The group service method still needs the same data, it's just that that data isn't all in the same place anymore, so some renaming / reshuffling needs to be done in the models. I'll add comments in appropriate places.

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -193,77 +201,75 @@ List<DelegateUser> delegateUsers
         }
 
         // TODO HEEDLS-887 Make sure this logic is correct with the new account structure
-        public void UpdateUserAccountDetailsForAllVerifiedUsers(
+        public void UpdateUserDetailsAndCentreSpecificDetails(
+            int userId,
             MyAccountDetailsData myAccountDetailsData,
-            CentreAnswersData? centreAnswersData = null
+            CentreAnswersData centreAnswersData
```
SteveJacksonSoft, line 207: Split this into two params
* `string? centreEmail`
* `DelegateDetailsData? delegateDetailsData`

Such that if delegateDetailsData is non-null then the user is assumed to have a delegate account. 
Just noticed delegateId is on MyAccountDetailsData - might make sense to port that across to delegateDetailsData.

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -193,77 +201,75 @@ List<DelegateUser> delegateUsers
...
-                    groupsService.SynchroniseUserChangesWithGroups(
-                        oldDelegateDetails,
-                        myAccountDetailsData,
-                        centreAnswersData
-                    );
-                }
+                groupsService.SynchroniseUserChangesWithGroups(
+                    oldDelegateUserDetails!,
+                    myAccountDetailsData,
+                    centreAnswersData
```
SteveJacksonSoft, line 252: Here call constructor on ~CentreAnswersData~ RegistrationFieldAnswers to combine `delegateDetailsData` and `jobGroup`

DigitalLearningSolutions.Data/Services/UserService.cs
SteveJacksonSoft: Use `userDataService.EmailIsInUseByOtherUser(userId, emailAddress)` instead of this method.

DigitalLearningSolutions.Data/Models/User/CentreAnswersData.cs
```diff
@@ -8,6 +8,7 @@ public class CentreAnswersData
         public CentreAnswersData(
```
SteveJacksonSoft, line 8: Rename this to `RegistrationFieldAnswers` and add a class `DelegateDetailsData`, and a constructor to turn `DelegateDetailsData` into `RegistrationFieldAnswers`.

Hopefully you can remove the existing constructor too.
I suggested that one should inherit the other, but that actually feels wrong, conceptually. Adding to the `DelegateDetailsData` model won't necessarily add to the registration fields we can use to generate groups.

(Can't comment on the class declaration 😢)

DigitalLearningSolutions.Data/DataServices/UserDataService/UserCentreDetailsDataService.cs
```diff
@@ -73,6 +77,11 @@ ELSE 0
                     transaction
                 );
             }
+
+            if (transactionShouldBeClosed)
+            {
+                transaction.Commit();
```
SteveJacksonSoft, line 83: Good catch!

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/EditDelegateController.cs
```diff
@@ -75,11 +75,11 @@ public IActionResult Index(EditDelegateFormData formData, int delegateId)
                 return ReturnToEditDetailsViewWithErrors(formData, delegateId, centreId);
             }
 
-            if (!userService.NewEmailAddressIsValid(formData.Email!, null, delegateId, centreId))
+            if (!userService.NewEmailAddressIsValid(formData.Email!, null, delegateId, User.GetUserId()!.Value))
```
VladNicol, line 78: The userId is wrong here, it should be the id of the user whose details are being edited

DigitalLearningSolutions.Data/Services/GroupsService.cs
```diff
@@ -38,8 +38,8 @@ public interface IGroupsService
 
         void SynchroniseUserChangesWithGroups(
             DelegateUser delegateAccountWithOldDetails,
-            AccountDetailsData newDelegateDetails,
-            CentreAnswersData newCentreAnswers
+            AccountDetailsData a,
```
SteveJacksonSoft, line 41: probably a mistake

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -35,17 +31,13 @@ List<DelegateUser> delegateUsers
         );
 
         void UpdateUserDetailsAndCentreSpecificDetails(
-            int userId,
             MyAccountDetailsData myAccountDetailsData,
-            CentreAnswersData centreAnswersData
+            DelegateDetailsData? delegateDetailsData,
+            string? centreEmail,
+            int? centreId
```
SteveJacksonSoft, line 37: I think it makes sense to put centreId in DelegateDetailsData - it is in the DelegateAccounts table after all.

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -218,58 +186,47 @@ CentreAnswersData centreAnswersData
...
-                userId
+                myAccountDetailsData.UserId
             );
 
-            userDataService.SetCentreEmail(
-                userId,
-                centreAnswersData.CentreId,
-                centreAnswersData.CentreEmail
-            );
+            if (centreId != null)
```
SteveJacksonSoft, line 194: Surely this should be if centreEmail not null?

DigitalLearningSolutions.Data/Services/UserService.cs
```diff
@@ -218,58 +186,47 @@ CentreAnswersData centreAnswersData
...
+                    centreId.Value,
+                    centreEmail
+                );
+            }
 
-            if (delegateUsers.Count != 0)
+            if (delegateDetailsData != null)
             {
-                var delegateId = myAccountDetailsData.DelegateId!.Value;
                 // TODO HEEDLS-887 Remove this call once groupsService.SynchroniseUserChangesWithGroups is refactored
```
SteveJacksonSoft, line 205: If we're ever going to refactor that method, now is probably the time to do it.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/EditDelegateController.cs
```diff
@@ -75,7 +75,8 @@ public IActionResult Index(EditDelegateFormData formData, int delegateId)
                 return ReturnToEditDetailsViewWithErrors(formData, delegateId, centreId);
             }
 
-            if (!userService.NewEmailAddressIsValid(formData.Email!, null, delegateId, User.GetUserId()!.Value))
+            // TODO HEEDLS-887 Fix this so that it is bassing the correct User ID, that of the delegate not the logged in admin
```
SteveJacksonSoft, line 78: @GeorgeBlackwell Is this being done in your ticket? We can talk about it tomorrow.

SteveJacksonSoft (changes requested):

DigitalLearningSolutions.Data/Services/UserService.cs
SteveJacksonSoft: Just working out if we actually have sufficient motivation to.

SteveJacksonSoft: I don't think we do actually. Even if we do refactor it to use a `Delegate` rather than a `DelegateUser` you'll still have to call the db to get the delegate's old registration field responses. And this line is all it would change (along with the line in the other place it's called I guess).

DigitalLearningSolutions.Data/Services/UserService.cs
AlexJacksonDS: No, we need to be able to clear centre email

DigitalLearningSolutions.Data/Services/UserService.cs
AlexJacksonDS: Users can have a centre email without a delegate account, we need it out

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/EditDelegateController.cs
AlexJacksonDS: This is really https://softwiretech.atlassian.net/browse/HEEDLS-951 I don't see why @GeorgeBlackwell would be changing this on his ticket? I'll mark it with the correct todo number

DigitalLearningSolutions.Data/Services/UserService.cs
SteveJacksonSoft: Aha, gotcha. Guessing you've left the centreId nullable in preparation for the centre-independent EditDetails journey.
I think we want to be using a new method for that anyway, so I think it'd be better to have centreId not-nullable.

SteveJacksonSoft: For now it's fine if we don't save anything when User.GetCentreId() returns null.

PR #1143, opened 2022-05-25
Uar 898 fix login

AlexJacksonDS:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-898

### Description
Fixes some incorrect logic around inactive accounts and DelegateAccounts with no old password and adds an action link to the Account Locked screen.

Documentation has been sorted as well https://softwiretech.atlassian.net/wiki/spaces/HEE/pages/19467173889/Sign+in

### Screenshots
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data/Services/LoginService.cs
+++ b/DigitalLearningSolutions.Data/Services/LoginService.cs
@@ -75,7 +75,8 @@
                 userEntity.DelegateAccounts.SingleOrDefault(da => da.CentreId == singleCentreToLogUserInto.Value);
 
             var centreIsActive = adminAccountToLogInto?.CentreActive ?? delegateAccountToLogInto?.CentreActive ?? false;
-            var accountAtCentreIsActive = adminAccountToLogInto?.Active ?? delegateAccountToLogInto?.Active ?? false;
+            var accountAtCentreIsActive = (adminAccountToLogInto?.Active == null || adminAccountToLogInto.Active) &&
+                                          (delegateAccountToLogInto?.Active == null || delegateAccountToLogInto.Active);
 
             if (!centreIsActive || !accountAtCentreIsActive || delegateAccountToLogInto is { Approved: false })
             {
--- a/DigitalLearningSolutions.Data/Services/LoginService.cs
+++ b/DigitalLearningSolutions.Data/Services/LoginService.cs
@@ -86,6 +86,7 @@
             return new LoginResult(LoginAttemptResult.LogIntoSingleCentre, userEntity, singleCentreToLogUserInto.Value);
         }
 
+        // If there are no accounts this will also return null, as there is no single centre to log into
         private static int? GetCentreIdIfLoggingUserIntoSingleCentre(UserEntity userEntity, string username)
         {
             // Determine if there is only a single account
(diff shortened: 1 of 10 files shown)


DigitalLearningSolutions.Data/Services/LoginService.cs
```diff
@@ -75,7 +75,8 @@ private LoginResult DetermineDestinationForSuccessfulLogin(UserEntity userEntity
                 userEntity.DelegateAccounts.SingleOrDefault(da => da.CentreId == singleCentreToLogUserInto.Value);
 
             var centreIsActive = adminAccountToLogInto?.CentreActive ?? delegateAccountToLogInto?.CentreActive ?? false;
-            var accountAtCentreIsActive = adminAccountToLogInto?.Active ?? delegateAccountToLogInto?.Active ?? false;
+            var accountAtCentreIsActive = (adminAccountToLogInto?.Active == null || adminAccountToLogInto.Active) &&
+                                          (delegateAccountToLogInto?.Active == null || delegateAccountToLogInto.Active);
```
simonbrent, line 79: This diff leads to changes in behaviour in two circumstances:
* When adminAccountToLogInto.Active is true and delegateAccountToLogInto.Active is false, accountAtCentreIsActive is now false, was previously true
* When both are null, accountAtCentreIsActive is now true, was previously false.

What is the intended behaviour here? The second of those circumstances doesn't look right to me.

DigitalLearningSolutions.Web/Startup.cs
```diff
@@ -70,8 +70,8 @@ public void ConfigureServices(IServiceCollection services)
                 options =>
                 {
                     options.AddPolicy(
-                        CustomPolicies.UserOnly,
-                        policy => CustomPolicies.ConfigurePolicyUserOnly(policy)
+                        CustomPolicies.BasicUser,
+                        policy => CustomPolicies.ConfigurePolicyBasicUser(policy)
```
simonbrent, line 74: I have added CustomPolicies.UserOnly usage to the ChangePasswordController as part of 890, so we need to be careful that this change doesn't break that one when merging into uar-test

simonbrent (changes requested):

JBsoftwire (changes requested):
I don't have anything to add to what Simon caught, but I agree with his flags.

DigitalLearningSolutions.Data/Services/LoginService.cs
AlexJacksonDS: - When adminAccountToLogInto.Active is true and delegateAccountToLogInto.Active is false, accountAtCentreIsActive is now false, was previously true. This is the intended change here, to fix the bug raised during testing where the produced true was incorrect (should not be able to log into either at the same centre when either is inactive)
- When both are null, accountAtCentreIsActive is now true, was previously false. This case isn't possible. They cannot both be null at this point, they would fall out at the `singleCentreToLogUserInto == null` return statement in that case.

DigitalLearningSolutions.Web/Startup.cs
AlexJacksonDS: I suspect this might be merged first as it is the smaller PR, but will just have to check when it comes to merging.

JBsoftwire (approved):
I'll pass this. If you want to add a comment following Simon's suggestion no need to re-review.

PR #1145, opened 2022-05-26
Heedls 904 fix sorting

simonbrent:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-904

### Description
* Fixed sorting by "When" column on Delegate learning log page
* Hide sort button on Delegate learning log when javascript is enabled

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateProgress/LearningLog.cshtml
+++ b/DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateProgress/LearningLog.cshtml
@@ -23,9 +23,13 @@
   }
 }
 
+@if (Model.JavascriptSearchSortFilterPaginateEnabled) {
+  <vc:loading-spinner page-has-side-nav-menu="false" />
+}
 <div class="nhsuk-grid-row" id="@(Model.JavascriptSearchSortFilterPaginateEnabled ? "js-styling-hidden-area-while-loading" : "no-js-styling")">
   <div class="nhsuk-grid-column-full">
     <h1 class="nhsuk-heading-xl">@ViewData["Title"]</h1>
+    <span class="display-none" id="progress-id">@Model.ProgressId</span>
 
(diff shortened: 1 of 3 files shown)


DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateProgress/LearningLog.cshtml
```diff
@@ -23,7 +23,7 @@
   }
 }
 
-<div class="nhsuk-grid-row">
+<div class="nhsuk-grid-row" id="@(Model.JavascriptSearchSortFilterPaginateEnabled ? "js-styling-hidden-area-while-loading" : "no-js-styling")">
```
DanBloxham-sw, line 26: I believe the addition of this class will cause its contents to be hidden on page load until the JS reveals it (which might take a few seconds if there are a significant number of records). We should add the loading spinner partial above this (as we do on other search/sort/filter/paginate pages, such as All Delegates) so that it is more obvious to the user that something is happening in the background.

This has also highlighted another problem that wasn't obvious before - the JS will error when there are no learning log entries. When there are no entries, we replace the table/controls with a simple message, but that prevents the JS from finding elements it needs. Now, this means that page content will never be revealed, as the JS breaks before it can reveal it. So that might require a little further refactoring of either the page or the JS.

DanBloxham-sw (changes requested):
Looks great. I've left a comment about a couple of subtle points that occur because of the new class. Happy to elaborate further if it's not obvious.

AlexJacksonDS (commented):
Looks fine other than Dan's comment to resolve

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/DelegateProgress/LearningLog.cshtml
simonbrent: Fixed both issues

AlexJacksonDS (approved):

PR #1151, opened 2022-05-27
HEEDLS-918 Fixed spacing between applied filter tags and clear filter button when js is disabled

simonbrent:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-918

### Description
* Added margin-left of 5px to clear filters button when javascript is enabled
* Increased margin-right between applied filter tags when javascript is enabled from 4 to 5px

The second change brings the spacing inline with when javascript is disabled. 
When javascript is enabled, DOM elements are added by `appendChild`, and margin is applied by CSS 
When javascript is disabled, DOM elements are added by `DigitalLearningSolutions.Web/Views/Shared/Components/CurrentFilters/Default.cshtml` and `_AppliedFilterTag.cshtml`, and have a space character between them (the MVC framework adds this space automatically) which is 5px wide.

### Screenshots
long filter
multi filters

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


livzorn (commented):
Looks good! Do you know if it's also possible to remove the empty space on the right side of tags that take up multiple lines, like the first two screenshots on this MR?

simonbrent:
flex filters
Like this? It's a bit of a change to how we style the applied filters in general, but it is (obviously) possible

simonbrent:
However, it looks pretty bad in some circumstances, e.g. when you don't have a left nav so the page is wider
too wide
:

livzorn:
@simonbrent Sorry should have been clearer, I actually meant the opposite, removing the empty space inside the tag rather than outside of it. So they wont take up the whole width of the container when split over two lines, like this:
appliedfilter

DigitalLearningSolutions.Web/Styles/shared/searchableElements/filter.scss
```diff
@@ -115,7 +120,7 @@
 
   #js-styling-hidden-area-while-loading & {
     .js-enabled & {
-      margin-right: 4px;
+      margin-right: 5px;
```
DanBloxham-sw, line 123: Is this bit actually necessary if we get rid of the above

>   .js-enabled & {
    margin-right: 0;
  }

I'm wondering about potential consequences for when we have js-enabled, but we have disabled js styling when there are large numbers of items.

simonbrent:
Ah right. I think it _might_ be possible, but flex-box is currently defeating me, and I'm not keen to sink a possibly infinite amount of time into trying to get it to work, which is just how CSS issues go sometimes!

DanBloxham-sw (commented):
I think this is suitable. I've left one question about potentially unnecessary css.

Olivia's comment would also be good to tackle, but I don't think it is necessary if you can't get a quick fix. Firstly, the extra space is caused because the filter in question has a long, unbroken word. Most filters shouldn't have this and so will be broken at a more appropriate point, leaving less obvious space. Secondly, the NHS tags, which these are, undergo the same behaviour, so I'd conclude it is fine to leave as is without making modifications to it. https://service-manual.nhs.uk/design-system/components/tag

DigitalLearningSolutions.Web/Styles/shared/searchableElements/filter.scss
simonbrent: Can you give me an example of when "disabled js styling" happens?

DanBloxham-sw: It occurs when we have JS enabled, but the number of items on a page exceeds a certain limit. By default, it is 250 items. You can change that locally in the appsettings.Development.json file from the ` "JavascriptSearchSortFilterPaginateItemLimit": 250,` field.

There are some pages where this does not apply, but in the majority of cases (including Centre Course Setup) it does apply.

simonbrent: Ok, I've looked at this scenario now, and my changes do not fix the issue when the JavascriptSearchSortFilterPaginateItemLimit is exceeded.

I have an alternative solution which appears to work, which is:
* Remove margin-right between active filter tags
* Add 5px margin-right to the `.applied-filter-container` (this change was part of the original commit)
* When adding applied filter tags in `filter.ts`, append a `document.createTextNode(' ')` after appending each filter

This latest bullet point makes the behaviour consistent whether js is enabled or not, i.e. there is a natural space between the tag elements always, rather than sometimes having a space and sometimes having a css margin.

I have checked the behaviour in the following circumstances, and found it to be correct in all cases (i.e. a space between each tag element, and a space between a long tag element and the clear filters button):
* JS enabled
* JS disabled
* JS enabled, but JavascriptSearchSortFilterPaginateItemLimit exceeded
* The above three cases, but with device dimension width = 400px (mobile)

SteveJacksonSoft:
> @simonbrent Sorry should have been clearer, I actually meant the opposite, removing the empty space inside the tag rather than outside of it. So they wont take up the whole width of the container when split over two lines, like this: appliedfilter

Yeah, this'd be nice, but as Dan says, it's not a big deal I don't think. Probably not a very frequent occurrence, and not massively user-unfriendly.

DanBloxham-sw (approved):
I think I see what is going on. It's a shame we can't get a simple css-only solution, but this seems to make things nicely consistent. Looks good to me.

PR #1173, opened 2022-06-08
Uar 884 extend my account journey fixes

VladNicol:
### JIRA link
[HEEDLS-884](https://softwiretech.atlassian.net/browse/HEEDLS-884)

### Description
Fixes after testing:

- Display job group, PRN and centre email on the MyAccount and EditDetails pages for admin only users
- When submitting an empty centre email, delete the record in the UserCentreDetails table instead of having an entry with a null email
- Validate both primary email and centre email at the same time
- Added a hidden text to the Primary email tag on the MyAccount page
- Changed spacing on the MyAccount page to make it look better on mobile

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests/UserCentreDetailsDataServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests/UserCentreDetailsDataServiceTests.cs
@@ -62,11 +62,11 @@
         }
 
         [Test]
-        [TestCase(true, null)]
-        [TestCase(true, "new@admin.email")]
-        [TestCase(false, null)]
-        [TestCase(false, "new@admin.email")]
-        public void SetCentreEmail_sets_email_if_not_empty(bool detailsExist, string? email)
+        [TestCase(true, null, 1)]
+        [TestCase(true, "new@admin.email", 1)]
+        [TestCase(false, null, 0)]
+        [TestCase(false, "new@admin.email", 1)]
+        public void SetCentreEmail_sets_email_if_not_empty(bool detailsExist, string? email, int entriesCount)
         {
             using var transaction = new TransactionScope();
 
@@ -83,9 +83,11 @@
             userDataService.SetCentreEmail(8, 374, email);
             var result = connection.Query<string?>(@"SELECT Email FROM UserCentreDetails WHERE UserID = 8")
                 .SingleOrDefault();
+            var count = connection.Query<int>(@"SELECT COUNT(*) FROM UserCentreDetails WHERE UserID = 8");
 
             // Then
             result.Should().BeEquivalentTo(email);
+            count.Should().Equal(entriesCount);
         }
     }
 }
--- a/DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests/UserCentreDetailsDataServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests/UserCentreDetailsDataServiceTests.cs
@@ -1,7 +1,7 @@
 ﻿namespace DigitalLearningSolutions.Data.Tests.DataServices.UserDataServiceTests
 {
-    using System;
     using System.Collections.Generic;
+    using System.Linq;
     using System.Transactions;
     using Dapper;
     using FluentAssertions;
@@ -27,7 +27,7 @@
             connection.Execute(
                 @"INSERT INTO UserCentreDetails (UserID, CentreID, Email)
                     VALUES (8, 374, 'sample@admin.email'), (225773, 101, 'sample@delegate.email')"
-                );
+            );
 
             // When
             var result = userDataService.AnyEmailsInSetAreAlreadyInUse(emails);
@@ -41,7 +41,10 @@
         [TestCase("test@gmail.com", true)]
         [TestCase("sample@admin.email", false)]
         [TestCase("sample@delegate.email", true)]
-        public void EmailIsInUseByOtherUser_returns_true_if_email_is_in_use_by_other_user(string email, bool expectedResult)
+        public void EmailIsInUseByOtherUser_returns_true_if_email_is_in_use_by_other_user(
+            string email,
+            bool expectedResult
+        )
         {
             using var transaction = new TransactionScope();
 
@@ -56,6 +59,33 @@
 
             // Then
             result.Should().Be(expectedResult);
+        }
+
+        [Test]
+        [TestCase(true, null)]
+        [TestCase(true, "new@admin.email")]
+        [TestCase(false, null)]
+        [TestCase(false, "new@admin.email")]
+        public void SetCentreEmail_sets_email_if_not_empty(bool detailsExist, string? email)
+        {
+            using var transaction = new TransactionScope();
+
+            // Given
+            if (detailsExist)
+            {
+                connection.Execute(
+                    @"INSERT INTO UserCentreDetails (UserID, CentreID, Email)
+                        VALUES (8, 374, 'sample@admin.email')"
+                );
+            }
+
+            // When
+            userDataService.SetCentreEmail(8, 374, email);
+            var result = connection.Query<string?>(@"SELECT Email FROM UserCentreDetails WHERE UserID = 8")
+                .SingleOrDefault();
+
+            // Then
+            result.Should().BeEquivalentTo(email);
         }
     }
 }
--- a/DigitalLearningSolutions.Data/DataServices/UserDataService/UserCentreDetailsDataService.cs
+++ b/DigitalLearningSolutions.Data/DataServices/UserDataService/UserCentreDetailsDataService.cs
@@ -57,25 +57,22 @@
                 transaction
             );
 
-            if (string.IsNullOrWhiteSpace(email))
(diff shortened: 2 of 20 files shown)


DigitalLearningSolutions.Data/DataServices/UserDataService/UserCentreDetailsDataService.cs
```diff
@@ -57,7 +57,15 @@ ELSE 0
...
 
-            if (detailsAlreadyExist)
+            if (string.IsNullOrWhiteSpace(email))
+            {
+                connection.Execute(
+                    @"DELETE FROM UserCentreDetails WHERE userID = @userId AND centreID = @centreId",
+                    new { userId, centreId },
+                    transaction
+                );
+            }
```
DanBloxham-sw, line 67: Could we include this in the "detailsAlreadyExist" if statement? We don't need to delete anything if the details don't exist. We would need to change things around a bit to make sure we aren't making null records in the first place.

Also it looks like this data service method doesn't have any unit tests. It would be good to implement some here.

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -153,6 +158,7 @@ DlsSubApplication dlsSubApplication
                 return ReturnToEditDetailsViewWithErrors(formData, dlsSubApplication);
             }
```
DanBloxham-sw, line 159: Bad place for this comment, but where we call `ProfessionalRegistrationNumberHelper.ValidateProfessionalRegistrationNumber` we are still worrying about whether a user has a delegate ID. So the validation on the PRN is still being skipped in admin-only cases.

DanBloxham-sw (changes requested):
Looks good, but a couple of bits can be improved I think.

DigitalLearningSolutions.Data/DataServices/UserDataService/UserCentreDetailsDataService.cs
AlexJacksonDS: I don't think we should create the UserCentreDetails record when the email is null though.

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -80,17 +84,18 @@ public IActionResult Index(DlsSubApplication dlsSubApplication)
             bool isCheckDetailsRedirect = false
         )
         {
-            var userAdminId = User.GetAdminId();
-            var userDelegateId = User.GetCandidateId();
-            var (adminUser, delegateUser) = userService.GetUsersById(userAdminId, userDelegateId);
+            var centreId = User.GetCentreId();
+            var userEntity = userService.GetUserById(User.GetUserId()!.Value)!;
```
AlexJacksonDS, line 88: Same here, pull it into a variable

AlexJacksonDS (changes requested):
Some extra things to tidy up. I think we need to check what the expected behaviour is with the updating/inserting etc. of UserCentreDetails records

DigitalLearningSolutions.Data/DataServices/UserDataService/UserCentreDetailsDataService.cs
DanBloxham-sw: In which case, it would be good to keep the record with a null email. We should probably make sure the EmailVerified field is set to null as well.

DigitalLearningSolutions.Data/Services/CentreRegistrationPromptsService.cs
```diff
@@ -260,6 +261,101 @@ public string GetCentreRegistrationPromptNameAndNumber(int centreId, int promptN
...
+            );
+            if (prompt6 != null)
+            {
+                list.Add(prompt6);
+            }
+
+            return list;
+        }
+
+        [Obsolete]
```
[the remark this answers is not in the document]
SteveJacksonSoft: Do we expect people to need to use it in the future, or should they always be using new version? I expect it's the latter, and if so it is appropriate to mark this obsolete.
It would be good to have a message in the attribute explaining what should be used instead.

DigitalLearningSolutions.Data/DataServices/UserDataService/UserCentreDetailsDataService.cs
VladNicol: I added some tests, but it feels like I'm testing the functionality of SQL server, and I'm not sure if I can test something else about that method

DigitalLearningSolutions.Web/ViewModels/MyAccount/MyAccountViewModel.cs
```diff
@@ -11,28 +11,27 @@
...
             CentreEmail = centreEmail;
-            DateRegistered = delegateUser?.DateRegistered?.ToString(DateHelper.StandardDateFormat);
-            ProfessionalRegistrationNumber = delegateUser == null
-                ? null
-                : PrnStringHelper.GetPrnDisplayString(
-                    delegateUser.HasBeenPromptedForPrn,
-                    delegateUser.ProfessionalRegistrationNumber
-                );
+            DateRegistered = delegateAccount?.DateRegistered.ToString(DateHelper.StandardDateFormat);
+            ProfessionalRegistrationNumber = PrnStringHelper.GetPrnDisplayString(
```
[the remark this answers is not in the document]
AlexJacksonDS: Job group is still nullable in this file.

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -50,20 +50,25 @@ ILogger<MyAccountController> logger
...
                 );
 
             var model = new MyAccountViewModel(
-                adminUser,
-                delegateUser,
-                userService.GetCentreEmail(User.GetUserId()!.Value, User.GetCentreId()),
+                userEntity.UserAccount,
+                delegateAccount,
+                (adminAccount?.CentreName ?? delegateAccount?.CentreName)!,
+                userService.GetCentreEmail(userId, User.GetCentreId()),
```
AlexJacksonDS, line 71: Not using the centreId variable here.

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -80,19 +85,21 @@ public IActionResult Index(DlsSubApplication dlsSubApplication)
...
+                promptsService.GetEditDelegateRegistrationPromptViewModelsForCentre(delegateAccount, centreId);
 
             var model = new MyAccountEditDetailsViewModel(
-                adminUser,
-                delegateUser,
+                userEntity.UserAccount,
+                delegateAccount,
                 jobGroups,
-                userService.GetCentreEmail(User.GetUserId()!.Value, User.GetCentreId()),
+                userService.GetCentreEmail(userId, User.GetCentreId()),
```
AlexJacksonDS, line 102: Not using the centreId variable here

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests/UserCentreDetailsDataServiceTests.cs
```diff
@@ -57,5 +60,32 @@ public void EmailIsInUseByOtherUser_returns_true_if_email_is_in_use_by_other_use
...
+                );
+            }
+
+            // When
+            userDataService.SetCentreEmail(8, 374, email);
+            var result = connection.Query<string?>(@"SELECT Email FROM UserCentreDetails WHERE UserID = 8")
+                .SingleOrDefault();
+
+            // Then
+            result.Should().BeEquivalentTo(email);
```
AlexJacksonDS, line 88: It would be good to have some way of distinguishing between null because the record doesn't exist and null because the record contains a null if possible so that we can confirm the "false, null" case doesn't create a record . If not, I guess this is enough

DigitalLearningSolutions.Data/DataServices/UserDataService/UserCentreDetailsDataService.cs
```diff
@@ -60,14 +60,19 @@ ELSE 0
...
             {
                 connection.Execute(
-                    @"UPDATE UserCentreDetails
-                    SET Email = @email
-                    WHERE userID = @userId AND centreID = @centreId",
+                    string.IsNullOrWhiteSpace(email)
+                        ? @"UPDATE UserCentreDetails
+                        SET Email = @email,
+                            EmailVerified = null
+                        WHERE userID = @userId AND centreID = @centreId"
```
AlexJacksonDS, line 67: I'm tempted to say we should set the email to NULL as well when we're in this email is null or whitespace case since a whitespace email is invalid.
Also can you capitalise the NULL for consistency

AlexJacksonDS (commented):
Just a few things left to sort out. Also note the unresolved previous comment above.

DigitalLearningSolutions.Web/ViewModels/MyAccount/MyAccountViewModel.cs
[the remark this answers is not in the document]
AlexJacksonDS: @VladNicol Missed this, but once that's resolved this should be good to merge

PR #1175, opened 2022-06-08
HEEDLS-958 - Refactor Promote To Admin to reactivate inactive account…

DanBloxham-sw:
… at centre if it exists

### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-958

### Description
Promote to Admin now reactivates the inactive admin account for that delegate at that centre (if it exists). It will also update the record with personal details (name, etc.) from the delegate record to make sure it is up to date, before setting the admin roles as the user specified on the form page.

### Screenshots
N/A, all backend changes.

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/Services/RegistrationServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/Services/RegistrationServiceTests.cs
@@ -17,6 +17,7 @@ namespace DigitalLearningSolutions.Data.Tests.Services
     using DigitalLearningSolutions.Data.Tests.TestHelpers;
     using FakeItEasy;
     using FluentAssertions;
+    using FluentAssertions.Execution;
     using Microsoft.Extensions.Configuration;
     using Microsoft.Extensions.Logging.Abstractions;
     using NUnit.Framework;
@@ -32,7 +33,6 @@ namespace DigitalLearningSolutions.Data.Tests.Services
         private ICentresDataService centresDataService = null!;
         private IConfiguration config = null!;
         private IEmailService emailService = null!;
-        private IFrameworkNotificationService frameworkNotificationService = null!;
         private IPasswordDataService passwordDataService = null!;
         private IPasswordResetService passwordResetService = null!;
         private IRegistrationDataService registrationDataService = null!;
@@ -50,7 +50,6 @@ namespace DigitalLearningSolutions.Data.Tests.Services
             centresDataService = A.Fake<ICentresDataService>();
             config = A.Fake<IConfiguration>();
             supervisorDelegateService = A.Fake<ISupervisorDelegateService>();
-            frameworkNotificationService = A.Fake<IFrameworkNotificationService>();
             userDataService = A.Fake<IUserDataService>();
 
             A.CallTo(() => config["CurrentSystemBaseUrl"]).Returns(OldSystemBaseUrl);
@@ -72,7 +71,6 @@ namespace DigitalLearningSolutions.Data.Tests.Services
                 centresDataService,
                 config,
                 supervisorDelegateService,
-                frameworkNotificationService,
                 userDataService,
                 new NullLogger<RegistrationService>()
             );
@@ -543,7 +541,7 @@ namespace DigitalLearningSolutions.Data.Tests.Services
                     model.Email,
                     NewCandidateNumber,
                     baseUrl,
-                    model.NotifyDate.Value,
+                    model.NotifyDate!.Value,
                     "RegisterDelegateByCentre_Refactor"
                 )
             ).MustHaveHappened(1, Times.Exactly);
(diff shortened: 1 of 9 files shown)


DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -162,33 +159,58 @@ public void PromoteDelegateToAdmin(AdminRoles adminRoles, int categoryId, int de
...
+                userDataService.ActivateAdmin(adminUser.Id);
+                userDataService.UpdateAdminUser(
+                    delegateUser.FirstName,
+                    delegateUser.LastName,
+                    delegateUser.EmailAddress,
+                    delegateUser.ProfileImage,
+                    adminUser.Id
+                );
+                passwordDataService.SetPasswordByAdminId(adminUser.Id, delegateUser.Password);
+                userDataService.UpdateAdminUserPermissions(
```
DanBloxham-sw, line 173: Worth noting that this does not force the IsCentreManager permission to false like we do when registering a new admin.  So if a user was a centre manager, and was then deactivated, then reactivated, they would still be a centre manager.

I don't _think_ this will cause any issues, but it is a scenario where we have the potential for an admin to "create" an admin with higher permissions than they themselves have. @SteveJacksonSoft shall I leave this as it is, or ensure the IsCentreManager permission is set to false here?

DanBloxham-sw: I've made the ReactivateAdmin data service method revoke permissions higher than CentreAdmin (CentreManager and UserAdmin but UserAdmin is unlikely to be true in any case).

Though I now realise that the promote to admin button and controller require the user to be a centre manager anyway. So centre managers are probably safe reactivating other centre managers. Given that centre manager is a substantial permission, we might still want to revoke it (and UserAdmin) by default. Not sure if there are limits on the number of centre managers, but I don't think centre managers are something admins can easily promote others to. @SteveJacksonSoft  thoughts?

DigitalLearningSolutions.Data/DataServices/UserDataService/AdminUserDataService.cs
```diff
@@ -170,6 +170,23 @@ public void DeactivateAdmin(int adminId)
...
+        /// "create" admins with more permissions than themselves.
+        /// </summary>
+        public void ReactivateAdmin(int adminId)
+        {
+            connection.Execute(
+                @"UPDATE AdminUsers SET
+                        Active = 1,
+						IsCentreManager = 0,
+						UserAdmin = 0
+                    WHERE AdminID = @adminId",
```
simonbrent, line 185: Nasty mix of tabs and spaces

DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -162,33 +159,58 @@ public void PromoteDelegateToAdmin(AdminRoles adminRoles, int categoryId, int de
 
             var adminUser = userDataService.GetAdminUserByEmailAddress(delegateUser.EmailAddress);
 
-            if (adminUser != null)
+            if (adminUser?.Active is false && adminUser.CentreId == delegateUser.CentreId)
```
simonbrent, line 162: What is the difference between `is false` and `== false`?

DigitalLearningSolutions.Data.Tests/Services/RegistrationServiceTests.cs
```diff
@@ -674,7 +673,80 @@ public void PromoteDelegateToAdmin_throws_email_in_use_AdminCreationFailedExcept
...
+            using (new AssertionScope())
+            {
+                UpdateToExistingAdminAccountMustNotHaveHappened();
+                result.Error.Should().Be(AdminCreationError.EmailAlreadyInUse);
+            }
+        }
+
+        [Test]
+        public void
+            PromoteDelegateToAdmin_throws_email_in_use_AdminCreationFailedException_if_inactive_admin_at_different_centre_already_exists()
```
simonbrent, line 685: I'm a bit confused by this. Is there a business requirement which states that if you are an inactive admin at any centre, you can't be made an active admin at a different centre?

simonbrent (changes requested):

DigitalLearningSolutions.Data.Tests/Services/RegistrationServiceTests.cs
AlexJacksonDS: Pre-UAR, the AdminUsers table had a unique email constraint, so if you try to promote a delegate at your centre with an email address matching an admin at a different centre, it is not possible. With the changes here to prevent errors when attempting to do this, we want to allow reactivating the admin at the current centre, but not allow reactivating an admin at a different centre.
To get around this behaviour, they would probably have to change the email of the delegate + then promote. This would make them have the pair of accounts at the same correct centre. (The old account system is pretty confusing/bad)

AlexJacksonDS (commented):
Nothing else to add, other than the stuff already noted

DigitalLearningSolutions.Data/Services/RegistrationService.cs
DanBloxham-sw: In this case, basically nothing, so I've changed it to `==` for consistency as that seems to what be what we use elsewhere.

AlexJacksonDS (approved):

PR #1181, opened 2022-06-13
Uar 877 internal delegate registration

AlexJacksonDS:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-877

### Description
Added a new internal registration journey for existing users. Since this shares quite a lot of stuff with the external delegate registration I've done a fair bit of refactoring common code to reduce duplication.

### Screenshots
image
image
image
image
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


simonbrent (changes requested):

livzorn (commented):
Just some minor points

DigitalLearningSolutions.Web.AutomatedUiTests/AccessibilityTests/RegistrationJourneyAccessibilityTests.cs
```diff
@@ -130,6 +130,31 @@ public void Registration_by_centre_journey_with_set_password_has_no_accessibilit
...
+
+            // when
+            Driver.Navigate().GoToUrl(BaseUrl + registerUrl);
+            var registerResult = new AxeBuilder(Driver).Analyze();
+            Driver.SelectDropdownItemValue("Centre", "2");
+            Driver.SubmitForm();
+
+            var learnerInformationResult = new AxeBuilder(Driver).Analyze();
+            Driver.SubmitForm();
+            Driver.LogOutUser(BaseUrl);
```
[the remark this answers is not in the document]
AlexJacksonDS: If you don't log out, the user will stay logged in for other tests in the file, and fail because the other register journeys kick you out if you are already logged in

DigitalLearningSolutions.Web/Controllers/Register/RegisterAtNewCentreController.cs
```diff
@@ -0,0 +1,349 @@
...
+                )
+            );
+
+            return RedirectToAction("PersonalInformation");
+        }
+
+        [HttpGet]
+        [Authorize(Policy = CustomPolicies.BasicUser)]
+        [ServiceFilter(typeof(RedirectEmptySessionData<InternalDelegateRegistrationData>))]
+        public IActionResult PersonalInformation()
```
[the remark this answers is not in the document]
AlexJacksonDS: I'm not sure what value there is to adding a unit test for these Get methods since there isn't really any logic in them. They are technically tested that the load except the Confirmation one in the UI tests though.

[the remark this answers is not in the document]
simonbrent: Fair enough. I wasn't sure how committed we are to getting as close as possible to 100% coverage

DigitalLearningSolutions.Data.Tests/DataServices/RegistrationDataServiceTests.cs
```diff
@@ -209,6 +209,89 @@ void Action()
...
+                    currentTime
+                );
+            var userCentreDetailsCount = connection.QuerySingle<int>(
+                "SELECT COUNT(*) FROM UserCentreDetails WHERE CentreID = 3 AND UserID = 2"
+            );
+            var user = await connection.GetDelegateUserByCandidateNumberAsync(candidateNumber);
+
+            // Then
+            user.Id.Should().Be(delegateId);
+            userCentreDetailsCount.Should().Be(0);
```
SteveJacksonSoft, line 290: Feels like we should be checking that there's no UserCentreDetails entity with a non-null email because that's the minimum requirement. The app will work perfectly well if the registration does create a UserCentreDetails entity with a null email.

DigitalLearningSolutions.Data.Tests/Services/RegistrationServiceTests.cs
```diff
@@ -932,6 +938,181 @@ public void PromoteDelegateToAdmin_calls_data_service_with_expected_value()
...
+            // Then
+            A.CallTo(
+                () =>
+                    emailService.SendEmail(
+                        A<Email>.That.Matches(
+                            e =>
+                                e.To[0] == ApproverEmail &&
+                                e.Cc.IsNullOrEmpty() &&
+                                e.Bcc.IsNullOrEmpty() &&
+                                e.Subject == "Digital Learning Solutions Registration Requires Approval" &&
```
SteveJacksonSoft, line 1092: I'd be inclined to move these conditions into another test, and just check the approval link in these tests. It makes it easier to see what's actually being tested, and easier to debug if the test breaks as a result of one of these things is suddenly not true.

DigitalLearningSolutions.Web.Tests/Controllers/Register/RegisterAtNewCentreControllerTests.cs
```diff
@@ -0,0 +1,319 @@
...
+        }
+
+        [Test]
+        public void PersonalInformationPost_with_invalid_emails_fails_validation()
+        {
+            // Given
+            const int centreId = 3;
+            controller.TempData.Set(new InternalDelegateRegistrationData());
+            var userAccount = UserTestHelper.GetDefaultUserAccount();
+            var delegateAccount = UserTestHelper.GetDefaultDelegateAccount();
```
SteveJacksonSoft, line 74: Do you need to use these methods? Can you just do `new UserAccount()` etc? If so, that would make it more obvious to readers that you're not relying on the user accounts having particular properties.

DigitalLearningSolutions.Web.Tests/Controllers/Register/RegisterAtNewCentreControllerTests.cs
```diff
@@ -0,0 +1,319 @@
...
+                .MustHaveHappened();
+            result.Should().BeViewResult().WithDefaultViewName();
+        }
+
+        [Test]
+        public void PersonalInformationPost_with_valid_emails_is_allowed()
+        {
+            // Given
+            controller.TempData.Set(new InternalDelegateRegistrationData());
+            var duplicateUser = UserTestHelper.GetDefaultDelegateUser();
```
SteveJacksonSoft, line 100: Why do we need `duplicateUser`? Can't we just specify a random centreId?

SteveJacksonSoft (changes requested):

DigitalLearningSolutions.Data/DataServices/RegistrationDataService.cs
```diff
@@ -87,19 +237,35 @@ OUTPUT Inserted.ID
...
+            int userIdToLinkDelegateAccountTo,
+            DateTime currentTime,
+            IDbTransaction transaction
+        )
+        {
+            var initials = (delegateRegistrationModel.FirstName.Substring(0, 1) +
+                            delegateRegistrationModel.LastName.Substring(0, 1)).ToUpper();
 
             // this SQL is reproduced mostly verbatim from the uspSaveNewCandidate_V10 procedure in the legacy codebase.
             var candidateNumber = connection.QueryFirst<string>(
```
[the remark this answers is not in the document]
AlexJacksonDS: CandidateNumber has a unique constrain on it, so the registrations will fail if we do not insert a unique value. I think I've added what we would want to catch/log this.

DigitalLearningSolutions.Web.Tests/Controllers/Register/RegisterAtNewCentreControllerTests.cs
AlexJacksonDS: We just need to specify that it doesn't match the ControllerContextHelper.CentreId, so I've changed the following line to check that.

PR #1183, opened 2022-06-14
HEEDLS-900 admin self-registration fixes

livzorn:
### JIRA link
[_HEEDLS-900_](https://softwiretech.atlassian.net/browse/HEEDLS-900)

### Description
Removed the word "address" from what the screen reader reads out for changing an email.
Added Terms Agreed to the columns updated when registering a new user.

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Data/DataServices/RegistrationDataService.cs
```diff
@@ -47,6 +47,7 @@ public string RegisterNewUserAndDelegateAccount(DelegateRegistrationModel delega
                 delegateRegistrationModel.Active,
                 PasswordHash = "temp",
                 ProfessionalRegistrationNumber = (string?)null,
+                TermsAgreed = currentTime,
```
DanBloxham-sw, line 50: Couple of things on this.
Firstly, I'm not sure if we do want to set TermsAgreed. Looking at the old code, and the old ticket, we never set TC_Agreed which I think was the old version of this field (used to be on AdminUsers). @SteveJacksonSoft do you know if the requirement to set this has slipped off the radar somehow?

Secondly, this method is currently used in other delegate registration process. So I think we want to be careful about setting this here. I presume a non-null date basically means the user has checked the "I agree to the Terms and Conditions..." check box when registering on said date. But it looks like this method is used in the DelegateBulkUpload journey, where the user certainly wouldn't have checked that check box themselves.

DanBloxham-sw (commented):
Looks like this will solve the issues, but I am unsure on the TermsAgreed portion.

DigitalLearningSolutions.Data/DataServices/RegistrationDataService.cs
simonbrent: I agree...I think we'd need to pass the state of "has the user being registered checked a Ts&Cs checkbox during this journey" through the code until we reach this point in order to set `TermsAgreed` in the database

SteveJacksonSoft: Yeah, we should be populating TC_Agreed in admin registration (external and internal). Don't know how that got missed. And we should be passing the checkbox value through in order to set it, as Simon says.

simonbrent (changes requested):
It would be good to:
* Update the `RegisterNewUserAndDelegateAccount_sets_all_fields_correctly_on_registration` test so that `registerJourneyContainsTermsAndConditions` is true, and the test checks that `TermsAgreed` is set to a reasonable timestamp
* A new test is added to check that `TermsAgreed` is null if `registerJourneyContainsTermsAndConditions` is false

DigitalLearningSolutions.Data/DataServices/RegistrationDataService.cs
SteveJacksonSoft: I've raised [HEEDLS-964](https://softwiretech.atlassian.net/browse/HEEDLS-964) to fix this for the admin registration journey on master (we need to include it in the upcoming release).

simonbrent (approved):

PR #1187, opened 2022-06-16
HEEDLS-412 - Implement new StoreASPAssessNoSession tracker endpoint

DanBloxham-sw:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-412

### Description
New tracker endpoint StoreAspAssessNoSession. Some minor renaming/refactoring of related methods has also been performed.

As part of this, the StoreAspProgressService was renamed to StoreAspService as it contains methods used across StoreAspProgress and StoreAspAssess tracker endpoints. A similar renaming was performed on the test file for this service. It looks like GitHub hasn't compared the details effectively, and considers the old file deleted. So there are a few changes in each which aren't actually changed.

### Screenshots
N/A

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


AlexJacksonDS (approved):
I think this all looks fine

DigitalLearningSolutions.Data/Enums/TrackerEndpointResponse.cs
```diff
@@ -11,8 +11,11 @@ public class TrackerEndpointResponse : Enumeration
         public static readonly TrackerEndpointResponse InvalidAction =
             new TrackerEndpointResponse(-2, nameof(InvalidAction));
 
-        public static readonly TrackerEndpointResponse NullTutorialStatusOrTime =
-            new TrackerEndpointResponse(-14, nameof(NullTutorialStatusOrTime));
+        public static readonly TrackerEndpointResponse StoreAspAssessException =
+            new TrackerEndpointResponse(-6, nameof(StoreAspAssessException));
```
simonbrent, line 15: What's the deal with these negative ids?

simonbrent:
> It looks like GitHub hasn't compared the details effectively, and considers the old file deleted

For reference, this is a git issue, not a github issue, and can be resolved on the command line as follows:

```
git diff master -M01 -- DigitalLearningSolutions.Data/Services/StoreAspProgressService.cs DigitalLearningSolutions.Data/Services/StoreAspService.cs
```
(http://git-scm.com/docs/git-diff#Documentation/git-diff.txt--Mltngt)

DigitalLearningSolutions.Data.Tests/Services/StoreAspServiceTests.cs
```diff
@@ -0,0 +1,703 @@
...
+            using (new AssertionScope())
+            {
+                var expectedProgressRecord = progressRecords[3];
+                result.validationResponse.Should().BeNull();
+                result.progress.Should().BeEquivalentTo(expectedProgressRecord);
+                A.CallTo(() => courseDataService.GetDelegateCoursesInfo(DefaultDelegateId))
+                    .MustHaveHappenedOnceExactly();
+            }
+        }
+
```
simonbrent, line 422: It'd have been easier to review this file if the `GetProgressAndValidateInputsForStoreAspAssess` and `GetProgressAndValidateCommonInputsForStoreAspProgressEndpoints` tests were grouped separately, rather than interspersed.

DigitalLearningSolutions.Data.Tests/Services/StoreAspServiceTests.cs
```diff
@@ -0,0 +1,703 @@
...
+                result.validationResponse.Should().Be(TrackerEndpointResponse.StoreAspAssessException);
+                result.progress.Should().BeNull();
+                A.CallTo(() => courseDataService.GetDelegateCoursesInfo(DefaultDelegateId))
+                    .MustHaveHappenedOnceExactly();
+            }
+        }
+
+        [Test]
+        public void
+            GetProgressAndValidateCommonInputsForStoreAspProgressEndpoints_returns_progress_record_and_null_exception_when_all_is_valid()
```
simonbrent, line 362: Instead of `null_exception`, I'd prefer either `no_exception` or `null_validationResponse`, since I immediately thought: "Is a null exception like Java's null pointer exception?"

Note: there are numerous other tests with this naming convention both in this file and in TrackerActionServiceTests

DigitalLearningSolutions.Data.Tests/Services/TrackerActionServiceTests.cs
```diff
@@ -418,10 +476,495 @@ public void
             }
         }
 
-        private void StoreAspProgressServiceReturnsDefaultDetailedCourseProgressOnValidation()
+        [Test]
+        public void
+            StoreAspAssessNoSession_returns_non_null_exception_from_query_param_and_progress_validation()
```
simonbrent, line 481: "non null exception" = "an exception"?

DigitalLearningSolutions.Data/Enums/TrackerEndpointResponse.cs
DanBloxham-sw: When the process is complete, we expect the tracker endpoint to return some text in the format of 
`"[#Result:1]"`
If the process errors, we expect to return a similar output along the lines of
`"[#Result:{errorCode}]"` 
where the `errorCode` is a specific number. These error codes are all negative to indicate the a problem occurred. Hence we save the IDs of each of these as negative so we can easily return the right code from the enum.
Why the numbers are what they are, I am unsure. But I believe these tracker endpoints are all used in Learning Hub, which will expect specific results in specific circumstances - I am unsure how they will behave if any of this changed. But what we have here more or less emulates the old code.

DigitalLearningSolutions.Data.Tests/Services/StoreAspServiceTests.cs
DanBloxham-sw: Yeah I realise that mistake now. I was trying to go with "near identical tests for different methods are together" that would make the whole file easier to understand, but from a comparison perspective it ends up worse. I've left them how they were for now so comparing between commits should be simple.

PR #1191, opened 2022-06-17
HEEDLS-926 Verify your email page

VladNicol:
### JIRA link
[HEEDLS-926](https://softwiretech.atlassian.net/browse/HEEDLS-926)

### Description
I created the verify email page. I made some changes compared to the designs, see screenshots below (@SteveJacksonSoft it would be nice if you could have a look at these):

- I slightly changed the wording to make it clearer (in my opinion) and to make the cases a bit more consistent with each other (e.g. instead of "until it is verified" I said "until you verify the address" to make it clear that the address is the one being verified)
- In the navigation menu at the top, I changed the order of the buttons. I found this easier to implement because in _Layout.cshtml the buttons are in the order NavMenuItems section (in my case, this is the Log out button), Help, Switch application. Having the Help and Switch application buttons be last is also more consistent with other pages (because they use the same _Layout.cshtml).

Things I'm not sure about this MR and where I expect possible markups:

- The Index.cshtml file is a bit messy. This is because all the cases have different but similar texts. I tried to not duplicate the texts and instead pull out the differences into variables that I then use in string interpolations. Maybe the code would be easier to follow otherwise?
- I am not sure what to test about the Controller and the ViewModel because they don't do much.
- In case the user sets the same email for multiple centres, the same email address will be listed in multiple paragraphs. I think this is fine because it tells the user "hey, you need to verify this email address before you can access any of the centres listed here"

### Screenshots
The cases can be found in the ticket description. I didn't set up my database to test case G, but it should be fine.

Case A:
image

Case B:
image

Case C:
image

Case D:
image

Case E:
image

Case F:
image

No unverified emails:
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Web/Views/VerifyEmail/Index.cshtml
```diff
@@ -0,0 +1,130 @@
...
+  <div class="nhsuk-grid-column-full">
+    <h1 class="nhsuk-heading-xl">Verify your email address</h1>
+
+    @if (Model.UnverifiedEmailsCount == 0) {
+      <p class="nhsuk-body-m"> All your email addresses are verified.</p>
+    } else {
+      var centreEmailsExcludingFirstParagraph = (Model.PrimaryEmail == null
+        ? Model.CentreSpecificEmails.Skip(1)
+        : Model.CentreSpecificEmails).ToList();
+      const string centreEmailExplanation = "You will not be able to access that account until you verify the address.";
```
AlexJacksonDS, line 23: I agree that this is pretty messy, but I'm not 100% sure how best to tidy it up.
I think lets start by making these model properties, so that we can pull apart the later logic into partials more easily.

DigitalLearningSolutions.Web/Views/VerifyEmail/Index.cshtml
```diff
@@ -0,0 +1,130 @@
...
+
+      @if (Model.EmailVerificationReason.Equals(EmailVerificationReason.EmailNotVerified)) {
```
AlexJacksonDS, line 25: We can then move all of the stuff in this, and the EmailChanged if statement into a pair of corresponding partials. This makes the Index view nice and neat, and pulls all the messy stuff into the ViewModel/Partial views

DigitalLearningSolutions.Web/Views/VerifyEmail/Index.cshtml
```diff
@@ -0,0 +1,130 @@
...
+          pronoun = "it";
+        } else {
+          primaryEmail = $", {Model.PrimaryEmail},";
+          emailAddresses = "each of the addresses above";
+          link = "links";
+          address = "addresses";
+          email = "the emails";
+          folder = "folders";
+          pronoun = "them";
+        }
```
AlexJacksonDS, line 50: I think I'd probably inline each of these to a ternary operator (?:) statement. I know it means that we're checking the condition multiple times, but it reduces the places you need to check what value is what. 
It might be worth defining a model property for `Model.UnverifiedEmailsCount == 1` to make it more clear?
(convert code style to var if you do this)

AlexJacksonDS: We can also then move these variables closer to where they are actually used. Maybe even put all the ones that are only used in the `verifyEmailExplanation` directly into the string interpolation? Not really sure that is better though.

DigitalLearningSolutions.Web/Views/VerifyEmail/Index.cshtml
```diff
@@ -0,0 +1,130 @@
...
+        @if (Model.PrimaryEmail != null) {
+          var primaryEmailText = $"Your primary email address{primaryEmail} has not been verified yet." +
+                                 " You can edit your account details, but you cannot access any centre accounts until it is verified.";
+
+          <p class="nhsuk-body-m">@primaryEmailText</p>
```
AlexJacksonDS, line 55: I know the wireframe has the phrasing so that it needs this string interpolation stuff, but I think it would be OK to always just have the ', Model.PrimaryEmail,' chunk of text, even if it is the only email to be verified.
We'd have to confirm that HEE are OK with the final wording, but since we've already changed some wording, we probably need to do that anyway.

DigitalLearningSolutions.Web/Views/VerifyEmail/Index.cshtml
```diff
@@ -0,0 +1,130 @@
...
+
+        var verifyEmailExplanation = $"An email with a verification link should have been sent to {emailAddresses}." +
+                                     $" Please click the {link} to verify your email {address}." +
+                                     $" If you have not received {email} after a few minutes, check your Junk {folder}," +
+                                     $" or return to the My account page to resend {pronoun}.";
+
+        <p class="nhsuk-body-m">@verifyEmailExplanation</p>
+      }
+
+      @if (Model.EmailVerificationReason.Equals(EmailVerificationReason.EmailChanged)) {
```
AlexJacksonDS, line 77: Contents of this if to partial

DigitalLearningSolutions.Web/Views/VerifyEmail/Index.cshtml
```diff
@@ -0,0 +1,130 @@
...
+        var endingText = Model.UnverifiedEmailsCount == 1
+          ? "If you have not received the email after a few minutes, check your Junk folder," +
+            " or return to the My account page to resend it."
+          : "An email with a verification link has been sent to each of the changed addresses." +
+            " Please click the links to verify your email addresses. If you have not received the emails after a few minutes," +
+            " check your Junk folders, or return to the My account page to resend them.";
```
AlexJacksonDS, line 83: Define this just before it is used.

DigitalLearningSolutions.Web/Views/VerifyEmail/Index.cshtml
```diff
@@ -0,0 +1,130 @@
...
+          if (Model.UnverifiedEmailsCount == 1) {
+            primaryEmail = null;
+            primaryEmailOnlyExplanation = $"An email with a verification link has been sent to {Model.PrimaryEmail}." +
+                                          " Please click the link to verify your email address. ";
+          } else {
+            primaryEmail = $": {Model.PrimaryEmail}";
+            primaryEmailOnlyExplanation = null;
+          }
+
+          var primaryEmailChangedText = $"You've updated your primary email address{primaryEmail}." +
```
AlexJacksonDS, line 97: Again, I don't see an issue with having the primary email displayed here in all cases. This would get rid of half the logic above, which should then be reduced to a ternary operator.

DigitalLearningSolutions.Web/Views/VerifyEmail/_NavMenuItems.cshtml
```diff
@@ -0,0 +1,10 @@
+﻿@using DigitalLearningSolutions.Data.Enums
+@using DigitalLearningSolutions.Web.Extensions
+@using Microsoft.AspNetCore.Mvc.TagHelpers
+
+<li class="nhsuk-header__navigation-item @Html.GetSelectedCssClassIfTabSelected(NavMenuTab.LogOut)">
+  <a class="nhsuk-header__navigation-link" asp-controller="Logout" asp-action="Index">
+    Log out
+    <partial name="_NhsChevronRight" />
+  </a>
+</li>
```
AlexJacksonDS, line 10: Does this actually work? The logout endpoint should now be a POST/. I also don't think that this can ever be a selected tab, so I'm not sure we need the css stuff

AlexJacksonDS (changes requested):
Left some changes, and thoughts about how the Index could be tidied up.

JBsoftwire (commented):
I think Alex's proposals for refactoring make sense.

DigitalLearningSolutions.Web/Controllers/VerifyEmailController.cs
```diff
@@ -0,0 +1,50 @@
...
+            var unverifiedPrimaryEmail =
+                userEntity.UserAccount.EmailVerified == null ? userEntity.UserAccount.PrimaryEmail : null;
+            var unverifiedCentreSpecificEmails = userService.GetUnverifiedCentreEmailsForUser(userId).Where(
+                pair => userEntity.AdminAccounts.Any(
+                            account => account.CentreName.Equals(pair.centreName) && account.Active
+                        ) ||
+                        userEntity.DelegateAccounts.Any(
+                            account => account.CentreName.Equals(pair.centreName) && account.Active
+                        )
+            );
```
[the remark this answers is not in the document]
VladNicol: I see the point in moving this logic to the userService, but I don't understand how that allows me to delete the method from the data service. I pushed my changes while I'm still working on the other markups so we can discuss.

DigitalLearningSolutions.Web/Views/VerifyEmail/Index.cshtml
VladNicol: I find having ternary operators inside string interpolations harder to follow, so I'll leave them as variables

DigitalLearningSolutions.Data.Tests/TestHelpers/UserTestHelper.cs
```diff
@@ -31,7 +31,7 @@ public static class UserTestHelper
             DateTime? detailsLastChecked = null
         )
         {
-            emailVerified ??= DateTime.Parse("2022-04-27 16:28:55.247");
+            // emailVerified ??= DateTime.Parse("2022-04-27 16:28:55.247");
```
JBsoftwire, line 34: This has just been commented out, did you mean to remove it?

JBsoftwire (changes requested):
There's an issue with a stray commented line, but besides that it looks alright. No need to re-review once it's cleared up.

PR #1196, opened 2022-06-17
HEEDLS-964 Add setting of TCAgreed to register admin

AlexJacksonDS:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-964

### Description
Set TCAgreed in the Register centre manager journey in the same way it has been done on UAR.

### Screenshots
No visual changes.

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/DataServices/RegistrationDataServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/DataServices/RegistrationDataServiceTests.cs
@@ -1,17 +1,22 @@
 ﻿namespace DigitalLearningSolutions.Data.Tests.DataServices
 {
+    using System;
     using System.Linq;
     using System.Threading.Tasks;
     using System.Transactions;
     using DigitalLearningSolutions.Data.DataServices;
     using DigitalLearningSolutions.Data.DataServices.UserDataService;
+    using DigitalLearningSolutions.Data.Services;
     using DigitalLearningSolutions.Data.Tests.TestHelpers;
+    using FakeItEasy;
     using FluentAssertions;
+    using FluentAssertions.Execution;
     using Microsoft.Data.SqlClient;
     using NUnit.Framework;
 
     public class RegistrationDataServiceTests
     {
+        private IClockService clockService = null!;
         private SqlConnection connection = null!;
         private INotificationPreferencesDataService notificationPreferencesDataService = null!;
         private RegistrationDataService service = null!;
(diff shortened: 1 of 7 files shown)


DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -136,7 +136,7 @@ public void RegisterCentreManager(AdminRegistrationModel registrationModel, int
 
             CreateDelegateAccountForAdmin(registrationModel, jobGroupId);
 
-            registrationDataService.RegisterAdmin(registrationModel);
+            registrationDataService.RegisterAdmin(registrationModel, true);
```
JBsoftwire, line 139: Is it intended that this value not actually be passed through from anywhere but just assumed to be true if this location in the code is reached? That seems a strange way to do it but I can be convinced.

JBsoftwire (changes requested):

DigitalLearningSolutions.Data.Tests/DataServices/RegistrationDataServiceTests.cs
```diff
@@ -51,6 +57,53 @@ public async Task Sets_all_fields_correctly_on_registration()
             user.Answer6.Should().Be(delegateRegistrationModel.Answer6);
         }
 
+        [Test]
+        public async Task RegisterAdmin_Sets_all_fields_correctly_on_centre_manager_admin_registration()
```
DanBloxham-sw, line 61: This looks to be a duplicate of Sets_all_fields_correctly_on_centre_manager_admin_registration but now with the additional TCAgreed stuff. Do we want to get rid of that other test and move its contents to the new "...is_false" test?

DanBloxham-sw (changes requested):
Looks good. Minor question about potential duplicate unit test, and ignorable suggestion about moving where we set the datetime.

DigitalLearningSolutions.Data/Services/RegistrationService.cs
AlexJacksonDS: I've pushed the value up to the service method parameters so that it matches what I did for UAR.
I don't think we should move the clockService call just so that it more closely matches UAR and will cause less confusion when that gets merged.

DigitalLearningSolutions.Data.Tests/DataServices/RegistrationDataServiceTests.cs
AlexJacksonDS: This was accidental duplication that I didn't tidy up. I was thinking of having the original test + a pair of true false tests that only checked TCAgreed, then decided not to do it, but forgot to delete the duplicate test.

DanBloxham-sw (approved):
Looks good.

PR #1208, opened 2022-06-22
HEEDLS-979 Change username and password validation failures to creden…

AlexJacksonDS:
…tials failure

### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-979

### Description
Combined the password and username validation failures in the login journey into a credentials failure so that existence of usernames cannot be discovered from the login form.

Note that this is the same as #1207 but for UAR 

### Screenshots
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data/Services/LoginService.cs
+++ b/DigitalLearningSolutions.Data/Services/LoginService.cs
@@ -31,7 +31,7 @@
 
             if (userEntity == null)
             {
-                return new LoginResult(LoginAttemptResult.InvalidUsername);
+                return new LoginResult(LoginAttemptResult.InvalidCredentials);
             }
 
             if (!userEntity.UserAccount.Active)
@@ -54,7 +54,7 @@
 
                 return userEntity.IsLocked
                     ? new LoginResult(LoginAttemptResult.AccountLocked, userEntity)
-                    : new LoginResult(LoginAttemptResult.InvalidPassword);
+                    : new LoginResult(LoginAttemptResult.InvalidCredentials);
             }
 
             return userEntity.IsLocked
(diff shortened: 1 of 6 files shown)


DigitalLearningSolutions.Data/Services/LoginService.cs
```diff
@@ -31,7 +31,7 @@ public LoginResult AttemptLogin(string username, string password)
 
             if (userEntity == null)
             {
-                return new LoginResult(LoginAttemptResult.InvalidUsername);
+                return new LoginResult(LoginAttemptResult.InvalidCredentials);
             }
 
             if (!userEntity.UserAccount.Active)
```
livzorn, line 37: Do we want to do this check after we know the passwords match? Seems like an attacker could find out a username exists this way as well, though it would be for an inactive account. But also I don't know if we care about that since we're okay with showing them the account locked page.

VladNicol: Just stating some ideas, feel free to ignore them, but I think Olivia makes a good point. Maybe we want to move
`if (!verificationResult.PasswordMatchesAtLeastOneAccountPassword)` right after `if (userEntity == null)`? Going a bit far here, but having the ifs that return the same result one after the other also theoretically diminishes the possibility of a timing attack (of course, the round trip time of the request is much bigger than the processing time of 2-3 instructions).

However, the implications of that would be that we are able to lock an inactive account, and I'm not sure we want that. We could maybe move the ifs around and increase the `FailedLoginCount` only if the account is active?

VladNicol (approved):

PR #1210, opened 2022-06-22
HEEDLS-959 Add recurring SQL jobs

livzorn:
### JIRA link
[_HEEDLS-959_](https://softwiretech.atlassian.net/browse/HEEDLS-959)

### Description
Added two recurring SQL scripts:
- a script to delete MultipageFormData older than 30 days, to be run nightly at 2AM
- a script to delete ResetPassword records older than 4 days, to be run nightly at 3AM

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data.Migrations/RecurringSqlJobs/DeleteMultipageFormDataOlderThan30Days.sql
+++ b/DigitalLearningSolutions.Data.Migrations/RecurringSqlJobs/DeleteMultipageFormDataOlderThan30Days.sql
@@ -1,0 +1,4 @@
+﻿-- Deletes MultiPageFormData records older than 30 days. Should be run every night at 2AM.
+DELETE
+FROM MultiPageFormData
+WHERE CAST(CreatedDate AS DATE) <= GETDATE() - 30
--- a/DigitalLearningSolutions.Data.Migrations/RecurringSqlJobs/DeleteMultipageFormDataOlderThan30Days.sql
+++ b/DigitalLearningSolutions.Data.Migrations/RecurringSqlJobs/DeleteMultipageFormDataOlderThan30Days.sql
@@ -1,4 +1,4 @@
 ﻿-- Deletes MultiPageFormData records older than 30 days. Should be run every night at 2AM.
 DELETE
 FROM MultiPageFormData
-WHERE CAST(CreatedDate AS DATE) <= GETDATE() - 30
+WHERE CreatedDate < DATEADD(day, -30, GETDATE())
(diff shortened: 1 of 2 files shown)


DigitalLearningSolutions.Data.Migrations/RecurringSqlJobs/DeleteMultipageFormDataOlderThan30Days.sql
```diff
@@ -0,0 +1,4 @@
+﻿-- Deletes MultiPageFormData records older than 30 days. Should be run every night at 2AM.
+DELETE
+FROM MultiPageFormData
+WHERE CAST(CreatedDate AS DATE) <= GETDATE() - 30
```
DanBloxham-sw, line 4: I presume we are casting to DATE here so that we aren't concerned about the hours/minutes/seconds. Is that something we are sure we want? Bit confusing as GETDATE also returns a DATETIME, which we have just cast away from on the CreatedDate.

I don't _think_ there are any scenarios where ignoring the hours/minutes/seconds will be important, but thought it is worth double checking.

DigitalLearningSolutions.Data.Migrations/RecurringSqlJobs/DeleteResetPasswordRecordsOlderThan4Days.sql
```diff
@@ -0,0 +1,12 @@
+﻿-- Deletes ResetPassword records older than 4 days. Should be run every night at 3AM.
+BEGIN
+UPDATE c
+SET c.ResetPasswordID = NULL FROM Candidates AS c
+INNER JOIN ResetPassword AS r
+ON r.ID = c.ResetPasswordID
+WHERE CAST (r.PasswordResetDateTime AS DATE) <= GETDATE() -4
```
DanBloxham-sw, line 7: Could do with a space in-between `-4` to keep the formatting consistent.

DanBloxham-sw (changes requested):
Very minor point and question to double check

DigitalLearningSolutions.Data.Migrations/RecurringSqlJobs/DeleteMultipageFormDataOlderThan30Days.sql
AlexJacksonDS: Yeah, I agree that the casting is strange here. I would have done `WHERE CreatedDate < DATEADD(day, -30, GETDATE())`
It feels wrong to me to just have a `-30` -30 what? Days? Minutes?

AlexJacksonDS (commented):
Agree about the WHERE being a bit strange in both these queries.

DigitalLearningSolutions.Data.Migrations/RecurringSqlJobs/DeleteMultipageFormDataOlderThan30Days.sql
livzorn: Makes sense to me, I've gone with your suggestion, Alex.

AlexJacksonDS (approved):

PR #1217, opened 2022-06-24
Heedls 919 send centre notifications to centre email

AlexJacksonDS:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-919

### Description
Change various places where we're sending an email, to use the centre email where possible, falling back to the primary email if a suitable one isn't present.

### Screenshots
no visual changes

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


VladNicol (approved):
I think the code looks really good, just one tiny question.

I will probably base my branch for 897 off this one because it looks like there are several things I could reuse. It is very unlikely that I will merge before you, so I'll deal with any conflicts that might appear.

DigitalLearningSolutions.Data/Services/GroupsService.cs
```diff
@@ -445,7 +454,7 @@ public void GenerateGroupsFromRegistrationField(GroupGenerationDetails groupDeta
             var isJobGroup = groupDetails.RegistrationField.Equals(RegistrationField.JobGroup);
             var linkedToField = groupDetails.RegistrationField.LinkedToFieldId;
 
-            (List<(int id, string name)> newGroupNames, string groupNamePrefix) = isJobGroup
+            (var newGroupNames, var groupNamePrefix) = isJobGroup
```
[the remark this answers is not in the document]
AlexJacksonDS: No, this is just what the formatter automatically did when getting rid of the explicit typing in favour of `var`. In fact I get an IDE recommendation to change this to `var (newGroupNames, groupNamePrefix)`
image

DigitalLearningSolutions.Data/DataServices/GroupsDataService.cs
```diff
@@ -187,18 +187,22 @@ public IEnumerable<GroupDelegate> GetGroupDelegates(int groupId)
...
+                        gd.GroupDelegateID,
+                        gd.GroupID,
+                        gd.DelegateID,
+                        gd.AddedDate,
+                        da.CandidateNumber,
+                        u.FirstName,
+                        u.LastName,
+                        u.HasBeenPromptedForPrn,
+                        u.ProfessionalRegistrationNumber,
+                        u.PrimaryEmail,
```
VladNicol, line 199: After a second look, I realised here I did something similar for 897, but instead of selecting both primary email and centre email I did `COALESCE(ucd.Email, u.PrimaryEmail)` and I kept the `GroupDelegate` class the same. The primary and centre emails are used together as `centreEmail ?? primaryEmail` everywhere, and basically I did this operation directly on the database. Not sure which approach is better.

AlexJacksonDS: It depends on whether we're wanting to email unverified centre emails or not. For 919, this is the case according to the ticket, so I needed the all the details to be pulled through (and didn't want to cause horrible duplicate logic in SQL code). 

It looks like the criteria for 897 is to display always, so where we're not sending emails, COALESCE is probably fine, otherwise do it in c#.

VladNicol: That makes sense, thanks!

DigitalLearningSolutions.Data/DataServices/GroupsDataService.cs
```diff
@@ -187,18 +187,22 @@ public IEnumerable<GroupDelegate> GetGroupDelegates(int groupId)
...
+                        ucd.Email AS CentreEmail,
+                        ucd.EmailVerified AS CentreEmailVerified
                     FROM GroupDelegates AS gd
-                    JOIN Candidates AS c ON c.CandidateID = gd.DelegateID
+                    JOIN DelegateAccounts AS da ON da.ID = gd.DelegateID
+                    JOIN Users AS u ON u.ID = da.UserID
+                    LEFT JOIN UserCentreDetails AS ucd ON ucd.UserID = u.ID AND ucd.CentreId = da.CentreID
```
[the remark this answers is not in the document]
AlexJacksonDS: tbh I find it easier to read like this (I intentionally undid putting it on a separate line). But I don't really mind either way given that I won't be looking at it much longer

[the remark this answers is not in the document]
AlexJacksonDS: ah OK, yeah it should probably match the actual column names where possible. Will get that updated

DigitalLearningSolutions.Data.Tests/Services/GroupServiceTests/GroupsServiceEnrolDelegateTests.cs
```diff
@@ -588,6 +602,7 @@ public void EnrolDelegateOnGroupCourses_sends_correct_email_with_additional_Comp
             groupsService.EnrolDelegateOnGroupCourses(
                 oldDelegateDetails,
                 newAccountDetails,
+                null,
```
DanBloxham-sw, line 605: Should we have an equivalent for this, but with a non-null centre email, to make sure it gets used here?

DigitalLearningSolutions.Data/Helpers/CentreEmailHelper.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace DigitalLearningSolutions.Data.Helpers
+{
+    public static class CentreEmailHelper
+    {
+        public static string GetEmailForCentreNotifications(
```
DanBloxham-sw, line 5: Could do with a very basic unit test for this.

Additionally, as the scope has changed, we are going to want the ticket description changed to match the new requirements.

DigitalLearningSolutions.Data/Services/GroupsService.cs
```diff
@@ -418,7 +427,7 @@ int centreId
             if (groupCourse == null)
             {
                 transaction.Dispose();
-                logger.LogError("Attempted to add a course that a centre does not have access to to a group.");
+                logger.LogError("Attempted to add a course that a centre does not have access to a group.");
```
DanBloxham-sw, line 430: Minor, but I think the "duplicated" `to` makes sense here.

DanBloxham-sw (changes requested):
A few minor things, plus some stuff that might be out of scope.

I'm debating what we should do about the unit tests where we call methods using the EmailForCentreNotifications property. It feels like we need to double most of our unit test so we can ensure we are calling them in situations where the Primary Email should be used, and situations where the Centre Email should be used. But this feels like we are retesting the method. Thoughts?

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests/AdminUserDataServiceTests.cs
```diff
@@ -11,6 +11,19 @@
 
     public partial class UserDataServiceTests
     {
+        [Test]
+        public void GetAdminById_Returns_admin_user()
+        {
+            // Given
+            var expectedAdminEntity = UserTestHelper.GetDefaultAdminEntity();
```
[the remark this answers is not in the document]
AlexJacksonDS: I think we should replicate what the delegate one is doing. 2 tests, one that is this one, and another that populates the UserCentreDetails

DigitalLearningSolutions.Data/Services/DelegateApprovalsService.cs
```diff
@@ -56,55 +58,58 @@ IConfiguration config
         {
             var users = userDataService.GetUnapprovedDelegateUsersByCentreId(centreId);
```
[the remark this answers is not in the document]
AlexJacksonDS: I'm assuming this is covered by https://softwiretech.atlassian.net/browse/HEEDLS-897 @VladNicol?

VladNicol:
> I'm debating what we should do about the unit tests where we call methods using the EmailForCentreNotifications property. It feels like we need to double most of our unit test so we can ensure we are calling them in situations where the Primary Email should be used, and situations where the Centre Email should be used. But this feels like we are retesting the method. Thoughts?

I might be wrong, but I think that if `GetEmailForCentreNotifications` works as expected then we don't need to duplicate the tests

DigitalLearningSolutions.Data/Services/GroupsService.cs
AlexJacksonDS: Stupid english

DigitalLearningSolutions.Data/Services/PasswordResetService.cs
```diff
@@ -256,13 +241,13 @@ string baseUrl
...
-            string emailAddress,
+            DelegateEntity delegateEntity,
             string setPasswordHash,
-            string baseUrl,
-            DelegateUser delegateUser
+            string baseUrl
         )
         {
-            UriBuilder setPasswordUrl = new UriBuilder(baseUrl);
+            var emailAddress = delegateEntity.EmailForCentreNotifications;
```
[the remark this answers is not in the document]
SteveJacksonSoft: I'm starting to wonder if that's a good rule of thumb.
These welcome emails are concerning specific-delegate accounts, and often will be sent off the back of an admin registering a new delegate account for someone who may or may not already have an account.

I think these should be sent to `EmailForCentreNotification`.

AlexJacksonDS:
> > I'm debating what we should do about the unit tests where we call methods using the EmailForCentreNotifications property. It feels like we need to double most of our unit test so we can ensure we are calling them in situations where the Primary Email should be used, and situations where the Centre Email should be used. But this feels like we are retesting the method. Thoughts?
> 
> I might be wrong, but I think that if `GetEmailForCentreNotifications` works as expected then we don't need to duplicate the tests

Agreed, I think as long as the thing we're calling (in this case the helper) is tested and working, then duplicating the tests doesn't really add anything. If we were doing this same logic inline in the method being tested, then I would want the tests to be there. Like this comment https://github.com/TechnologyEnhancedLearning/DLSV2/pull/1217#discussion_r908465410

DigitalLearningSolutions.Data.Tests/Services/UnlockServiceTests.cs
```diff
@@ -1,149 +0,0 @@
...
-            A.CallTo(() => notificationDataService.GetUnlockData(A<int>._)).Returns(
-                new UnlockData
-                {
-                    ContactEmail = "recipient@example.com",
-                    ContactForename = "Forename",
-                    CourseName = "Activity Name",
-                    CustomisationId = 22,
-                    DelegateId = 1,
-                }
-            );
```
DanBloxham-sw, line 38: We seem to be missing this set up, but it looks like everything works fine without it. So I guess it is just useless.

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests/AdminUserDataServiceTests.cs
```diff
@@ -12,16 +12,43 @@
...
+        {
+            using var transaction = new TransactionScope();
+
+            // Given
+            connection.Execute(
+                @"INSERT INTO UserCentreDetails (UserID, CentreID, Email)
+                    VALUES (2, 2, 'centre@email.com')"
+            );
+            var expectedUserCentreDetails = UserTestHelper.GetDefaultAdminEntity(
+                userCentreDetailsId: 1,
```
DanBloxham-sw, line 38: I think it is wise to not assert that the ID must match some specific value. As we are setting the Id here in the expectedUserCentreDetails, would it be helpful to add a comment in the assertions to state why we aren't comparing it, and why we are setting it here? Or do we think it is obvious enough as is?

DanBloxham-sw (changes requested):
One very minor thing to change, and two potentially useless comments.

DigitalLearningSolutions.Data.Tests/Services/UnlockServiceTests.cs
AlexJacksonDS: None of the tests actually tested that any of these values were part of the email, so it didn't actually do anything.

DigitalLearningSolutions.Data.Tests/DataServices/UserDataServiceTests/AdminUserDataServiceTests.cs
AlexJacksonDS: Probably needs a comment at it isn't obvious behaviour.

PR #1220, opened 2022-06-27
UAR-906 - Implemented Reregistering functionality

DanBloxham-sw:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-906

### Description
Modified the Internal Registration journey to allow users with inactive delegate accounts to reregister. Changed the "Reactivate" button to read "Reregister" and hooked it up to that journey.

### Screenshots
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/DataServices/RegistrationDataServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/DataServices/RegistrationDataServiceTests.cs
@@ -51,7 +51,7 @@
                 delegateRegistrationModel,
                 false
             );
-            
+
             // Then
             var delegateEntity = userDataService.GetDelegateByCandidateNumber(candidateNumber);
             using (new AssertionScope())
@@ -89,7 +89,7 @@
                 delegateRegistrationModel,
                 true
             );
-           
+
             // Then
             var delegateEntity = userDataService.GetDelegateByCandidateNumber(candidateNumber);
             using (new AssertionScope())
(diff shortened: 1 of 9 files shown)


DigitalLearningSolutions.Web.Tests/Controllers/Register/RegisterAtNewCentreControllerTests.cs
```diff
@@ -135,11 +138,37 @@ public void PersonalInformationPost_with_invalid_emails_fails_validation()
...
-            A.CallTo(() => userService.NewEmailAddressIsValid(model.CentreSpecificEmail!, ControllerContextHelper.UserId))
+            A.CallTo(
+                    () => userService.NewEmailAddressIsValid(model.CentreSpecificEmail!, ControllerContextHelper.UserId)
+                )
                 .MustHaveHappened();
             result.Should().BeViewResult().WithDefaultViewName();
         }
 
+        [Test]
+        public void PersonalInformationPost_with_null_centre_email_skips_email_validation()
```
DanBloxham-sw, line 149: This isn't testing anything new that I've implemented, just looked like a good test to have that didn't already exist for some reason.

DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -343,32 +389,11 @@ public void PromoteDelegateToAdmin(AdminRoles adminRoles, int? categoryId, int d
...
-
-                logger.LogError(exception, errorMessage);
-
-                throw new DelegateCreationFailedException(errorMessage, exception, error);
-            }
+            ValidateRegistrationEmail(delegateRegistrationModel);
+            return registrationDataService.RegisterNewUserAndDelegateAccount(
+                delegateRegistrationModel,
+                registerJourneyContainsTermsAndConditions
+            );
```
AlexJacksonDS, line 396: How come we're removing the try catch here, What about other places that use this that don't have the error handling/logging? This is also used in the first time registration and the delegate upload journey.

DigitalLearningSolutions.Web/Views/Login/ChooseACentre.cshtml
```diff
@@ -72,10 +72,10 @@
                   </button>
                 </form>
               } else if (centreRow.Status.ActionButton == ChooseACentreButton.Reactivate) {
-                @* TODO HEEDLS-906: Link to Reactivate centre account page *@
-                <button class="nhsuk-button nhsuk-u-margin-bottom-3 nhsuk-u-padding-bottom-0 nhsuk-u-padding-top-0 nhsuk-u-margin-top-0">
-                  Reactivate <span class="nhsuk-u-visually-hidden">@centreRow.CentreName</span>
-                </button>
+                var reactivateRouteData = new Dictionary<string, string> {{"centreId", centreRow.CentreId.ToString()}};
+                <a asp-controller="RegisterAtNewCentre" asp-action="Index" asp-all-route-data="@reactivateRouteData" class="nhsuk-button nhsuk-u-margin-bottom-3 nhsuk-u-padding-bottom-0 nhsuk-u-padding-top-0 nhsuk-u-margin-top-0">
```
AlexJacksonDS, line 76: Why not use `asp-route-centreId` here instead of `asp-all-route-data`?

This also needs the correct aria role adding to it (link styled as button)

AlexJacksonDS (changes requested):
I think removing the try catch I commented on is potentially incorrect, and a couple of small changes for the link button on choose a centre.

DigitalLearningSolutions.Data/Services/RegistrationService.cs
DanBloxham-sw: Whoops. The try catch was not supposed to be removed from this method. A near identical try catch was supposed to be removed from the private method below it, but I must have got them mixed up when doing the replacement.

The actual try catch that was supposed to be moved (and now should be) simplifies the two private methods for registering/reregistering at a centre for an existing user to prevent a duplication of try catches. As the private methods are only used once, and are within the moved try catch, this should now be fine.

Good spot.

DigitalLearningSolutions.Web/Views/Login/ChooseACentre.cshtml
DanBloxham-sw: Good calls, added.

DigitalLearningSolutions.Web/Controllers/Register/RegisterAtNewCentreController.cs
```diff
@@ -102,13 +102,11 @@ public IActionResult PersonalInformation(InternalPersonalInformationViewModel mo
             {
                 var delegateAccount = userService.GetUserById(User.GetUserIdKnownNotNull())!.DelegateAccounts
                     .SingleOrDefault(da => da.CentreId == model.Centre);
```
simonbrent, line 104: This can be replaced with
```
var delegateAccount = userService.GetUserById(User.GetUserIdKnownNotNull())!.GetCentreAccountSet(model.Centre.Value)?.DelegateAccount;
```
if we would like to use `GetCentreAccountSet` in more (all?) appropriate places

DigitalLearningSolutions.Data.Tests/Services/RegistrationServiceTests.cs
```diff
@@ -1102,6 +1103,97 @@ public void CreateDelegateAccountForExistingUser_with_unapproved_IP_registers_de
...
+                             d.Answer2 == model.Answer2 && d.Answer3 == model.Answer3 && d.Answer4 == model.Answer4 &&
+                             d.Answer5 == model.Answer5 && d.Answer6 == model.Answer6
+                    ),
+                    userId,
+                    existingDelegateId,
+                    currentTime
+                )
+            ).MustHaveHappenedOnceExactly();
+        }
+
```
simonbrent, line 1196: At least one of the other successful `CreateDelegateAccountForExistingUser` tests should check that a call to `registrationDataService.ReregisterDelegateAccountAndCentreDetailForExistingUser` `MustNotHaveHappened`

DigitalLearningSolutions.Data.Tests/TestHelpers/UserTestHelper.cs
```diff
@@ -570,5 +573,31 @@ UserAccount user
...
+        }
+
+        public static (string? email, DateTime? emailVerified) GetEmailAndVerifiedDateFromUserCentreDetails(
+            this DbConnection connection,
+            int userId,
+            int centreId
+        )
+        {
+            return connection.QuerySingle<(string? email, DateTime? emailVerified)>(
+                $"SELECT Email, EmailVerified FROM UserCentreDetails WHERE CentreID = {centreId} AND UserID = {userId}"
```
simonbrent, line 599: Should this query be parametised?

DigitalLearningSolutions.Data.Tests/DataServices/RegistrationDataServiceTests.cs
```diff
@@ -195,6 +195,181 @@ public async Task
...
+
+                userAfterUpdate!.Id.Should().Be(userBeforeUpdate.Id);
+                userAfterUpdate.FirstName.Should().Be(userBeforeUpdate.FirstName);
+                userAfterUpdate.LastName.Should().Be(userBeforeUpdate.LastName);
+                userAfterUpdate.PrimaryEmail.Should().Be(userBeforeUpdate.PrimaryEmail);
+                userAfterUpdate.DetailsLastChecked.Should().Be(userBeforeUpdate.DetailsLastChecked);
+                userAfterUpdate.JobGroupId.Should().Be(userBeforeUpdate.JobGroupId);
+                userAfterUpdate.PasswordHash.Should().Be(userBeforeUpdate.PasswordHash);
+                userAfterUpdate.ProfessionalRegistrationNumber.Should()
+                    .Be(userBeforeUpdate.ProfessionalRegistrationNumber);
```
simonbrent, line 254: It looks like we're checking that all(?) the properties of `userAfterUpdate` are the same as those in `userBeforeUpdate`. Is there a simple one-line test we can do to test this instead? 

The above code made me wonder if there are some other properties that we aren't testing that might have changed.

It actually seems to me that we don't really care about the property values of `userBeforeUpdate` or `userAfterUpdate` at all, just that they are the same, and not null.

DigitalLearningSolutions.Data.Tests/DataServices/RegistrationDataServiceTests.cs
```diff
@@ -195,6 +195,181 @@ public async Task
...
+            service.ReregisterDelegateAccountAndCentreDetailForExistingUser(
+                delegateRegistrationModel,
+                userId,
+                existingDelegateId,
+                currentTime
+            );
+
+            // Then
+            var userCentreDetails = connection.GetEmailAndVerifiedDateFromUserCentreDetails(userId, centreId);
+            userCentreDetails.email.Should().Be(newCentreEmail);
```
simonbrent, line 325: Missing `using (new AssertionScope())`

simonbrent (changes requested):

DigitalLearningSolutions.Web/Controllers/Register/RegisterAtNewCentreController.cs
DanBloxham-sw: Might as well. I've changed to use it here.

DigitalLearningSolutions.Data.Tests/Services/RegistrationServiceTests.cs
DanBloxham-sw: I've added them to all the successful new account registrations, because why not.

simonbrent (approved):

PR #1232, opened 2022-06-29
HEEDLS-899 Fix supervisor delegate relationships

livzorn:
### JIRA link
[_HEEDLS-899_](https://softwiretech.atlassian.net/browse/HEEDLS-899)

### Description
Ensured all registration journeys still update the supervisor delegate record as expected, and changed them to match the delegate email to either the primary or centre specific email.
Also updated SubmitSummary in SelfAssessment.cs to save the CentreSpecificEmail ?? PrimaryEmail as the delegate's email for a new SupervisorDelegate record, and updated GetUserEmail in the SupervisorController to get the supervisor's CentreSpecificEmail ?? PrimaryEmail.

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data.Tests/DataServices/SupervisorDelegateDataServiceTests.cs
+++ b/DigitalLearningSolutions.Data.Tests/DataServices/SupervisorDelegateDataServiceTests.cs
@@ -1,0 +1,123 @@
+﻿namespace DigitalLearningSolutions.Data.Tests.DataServices
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Transactions;
+    using Dapper;
+    using DigitalLearningSolutions.Data.DataServices;
+    using DigitalLearningSolutions.Data.Models.Email;
+    using DigitalLearningSolutions.Data.Models.Supervisor;
+    using DigitalLearningSolutions.Data.Tests.TestHelpers;
+    using FakeItEasy;
+    using FluentAssertions;
+    using Microsoft.Data.SqlClient;
+    using Microsoft.Extensions.Logging;
+    using NUnit.Framework;
+
+    internal class SupervisorDelegateDataServiceTests
+    {
+        private readonly Guid inviteHashForFirstSupervisorDelegateRecord =
+            Guid.Parse("72e44c4d-77bd-4bed-a254-7cc27ab32927");
+
+        private SqlConnection connection = null!;
+        private SupervisorDelegateDataService supervisorDelegateDataService = null!;
+
+        [SetUp]
+        public void Setup()
+        {
+            connection = ServiceTestHelper.GetDatabaseConnection();
+            var logger = A.Fake<ILogger<SupervisorDelegateDataService>>();
+            supervisorDelegateDataService = new SupervisorDelegateDataService(connection, logger);
+        }
+
+        [Test]
+        public void GetSupervisorDelegateRecordByInviteHash_returns_correct_record()
+        {
+            // Given
+            var expectedRecord = new SupervisorDelegate
+            {
+                ID = 8,
+                SupervisorAdminID = 1,
+                DelegateEmail = "kevin.whittaker@hee.nhs.uk",
+                CandidateID = 254480,
+                Added = DateTime.Parse("2021-06-28 16:40:35.507"),
+                NotificationSent = DateTime.Parse("2021-06-28 16:40:35.507"),
+                Removed = null,
+                SupervisorEmail = "kevin.whittaker@hee.nhs.uk",
+                AddedByDelegate = false,
+                CentreId = 101,
+            };
+
+            // When
+            var result =
+                supervisorDelegateDataService.GetSupervisorDelegateRecordByInviteHash(
+                    inviteHashForFirstSupervisorDelegateRecord
+                );
+
+            // Then
+            result.Should().BeEquivalentTo(expectedRecord);
+        }
+
+        [Test]
+        public void GetPendingSupervisorDelegateRecordsByEmailsAndCentre_returns_correct_records()
+        {
+            using var transaction = new TransactionScope();
+
+            // Given
+            const int centreId = 101;
+            const string delegateEmailForValidRecord = "primary@email.com";
+            const string delegateEmailForRecordWithNonNullCandidateId = "kevin.whittaker@hee.nhs.uk";
+            const string delegateEmailForRemovedRecord = "louis.theroux@gmail.com";
+            const string? nullDelegateEmail = null;
+            var currentTime = DateTime.UtcNow;
+
+            connection.Execute(
+                @"INSERT INTO SupervisorDelegates (SupervisorAdminID, DelegateEmail, CandidateID, Added,
+                                 NotificationSent, Removed, SupervisorEmail, AddedByDelegate, InviteHash)
+                    OUTPUT INSERTED.ID
+                    VALUES (1, @delegateEmailForValidRecord, null, @currentTime, @currentTime, null,
+                            'supervisor@email.com', 0, '72e44c4d-77bd-4bed-a254-7cc27ab32928')",
+                new { delegateEmailForValidRecord, currentTime }
+            );
+
+            // When
+            var result = supervisorDelegateDataService.GetPendingSupervisorDelegateRecordsByEmailsAndCentre(
+                centreId,
+                new List<string?>
+                {
+                    delegateEmailForValidRecord,
+                    delegateEmailForRecordWithNonNullCandidateId,
+                    delegateEmailForRemovedRecord,
+                    nullDelegateEmail,
+                }
+            ).ToList();
+
+            // Then
+            result.Count.Should().Be(1);
+            result.First().DelegateEmail.Should().Be(delegateEmailForValidRecord);
+        }
+
+        [Test]
+        public void UpdateSupervisorDelegateRecordsCandidateId_updates_record_correctly()
+        {
+            using var transaction = new TransactionScope();
+
+            // Given
+            const int candidateId = 1;
+
+            // When
+            supervisorDelegateDataService.UpdateSupervisorDelegateRecordsCandidateId(
+                new List<int> { 6, 7, 8 },
+                1
+            );
+
+            // Then
+            var updatedRecord =
+                supervisorDelegateDataService.GetSupervisorDelegateRecordByInviteHash(
+                    inviteHashForFirstSupervisorDelegateRecord
+                );
+            updatedRecord!.CandidateID.Should().Be(candidateId);
+        }
+    }
+}
--- a/DigitalLearningSolutions.Data/DataServices/SupervisorDelegateDataService.cs
+++ b/DigitalLearningSolutions.Data/DataServices/SupervisorDelegateDataService.cs
@@ -11,7 +11,10 @@
     {
         SupervisorDelegate? GetSupervisorDelegateRecordByInviteHash(Guid inviteHash);
 
-        IEnumerable<SupervisorDelegate> GetPendingSupervisorDelegateRecordsByEmailAndCentre(int centreId, string email);
+        IEnumerable<SupervisorDelegate> GetPendingSupervisorDelegateRecordsByEmailsAndCentre(
+            int centreId,
+            IEnumerable<string?> emails
+        );
 
         void UpdateSupervisorDelegateRecordsCandidateId(IEnumerable<int> supervisorDelegateIds, int candidateId);
     }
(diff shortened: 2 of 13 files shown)


DigitalLearningSolutions.Data.Tests/DataServices/SupervisorDelegateDataServiceTests.cs
```diff
@@ -0,0 +1,123 @@
+﻿namespace DigitalLearningSolutions.Data.Tests.DataServices
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Transactions;
+    using Dapper;
+    using DigitalLearningSolutions.Data.DataServices;
+    using DigitalLearningSolutions.Data.Models.Email;
```
simonbrent, line 9: `DigitalLearningSolutions.Data.Models.Email` is unused

DigitalLearningSolutions.Data.Tests/DataServices/SupervisorDelegateDataServiceTests.cs
```diff
@@ -0,0 +1,123 @@
...
+        [SetUp]
+        public void Setup()
+        {
+            connection = ServiceTestHelper.GetDatabaseConnection();
+            var logger = A.Fake<ILogger<SupervisorDelegateDataService>>();
+            supervisorDelegateDataService = new SupervisorDelegateDataService(connection, logger);
+        }
+
+        [Test]
+        public void GetSupervisorDelegateRecordByInviteHash_returns_correct_record()
```
simonbrent, line 35: The query in `GetSupervisorDelegateRecordByInviteHash` is wrapped in a try/catch, but there is only a test that it returns the correct record. There should also be a test that it logs an error and returns null if the same `inviteHash` exists for multiple rows

DigitalLearningSolutions.Data.Tests/DataServices/SupervisorDelegateDataServiceTests.cs
```diff
@@ -0,0 +1,123 @@
...
+                {
+                    delegateEmailForValidRecord,
+                    delegateEmailForRecordWithNonNullCandidateId,
+                    delegateEmailForRemovedRecord,
+                    nullDelegateEmail,
+                }
+            ).ToList();
+
+            // Then
+            result.Count.Should().Be(1);
```
simonbrent, line 97: Multiple assertions should be wrapped in `using (new AssertionScope())`

DigitalLearningSolutions.Data.Tests/DataServices/SupervisorDelegateDataServiceTests.cs
```diff
@@ -0,0 +1,123 @@
...
+            const int candidateId = 1;
+
+            // When
+            supervisorDelegateDataService.UpdateSupervisorDelegateRecordsCandidateId(
+                new List<int> { 6, 7, 8 },
+                1
+            );
+
+            // Then
+            var updatedRecord =
```
simonbrent, line 116: This isn't explicitly testing that the candidateId has been changed, just that it is 1 after the update method has been called...it could have been 1 beforehand too. To remedy this:
* `var oldRecord = supervisorDelegateDataService.GetSupervisorDelegateRecordByInviteHash(inviteHashForFirstSupervisorDelegateRecord)` before updating
* Test that `oldRecord.CandidateID` is not 1
* Test that `oldRecord.ID` is in the list [6,7,8] (otherwise you aren't explicitly proving that the record you're updating is the same as the record you're getting from the database after the update)

simonbrent (changes requested):
`DigitalLearningSolutions.Data/Services/SupervisorService.cs` contains lots of functions which do SQL queries. These should be probably be moved to a data service.

DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -503,11 +500,13 @@ private void ValidateCentreEmail(int userId, string centreEmail)
             DelegateRegistrationModel delegateRegistrationModel
         )
         {
+            var delegateEmails = new List<string?>
+                { delegateRegistrationModel.PrimaryEmail, delegateRegistrationModel.CentreSpecificEmail };
+
```
VladNicol, line 505: Unless I missed something, it looks like this is the only place which can trigger a call to `SupervisorDelegateDataService.GetPendingSupervisorDelegateRecordsByEmailsAndCentre` with an empty email in the list. Related to Simon's comment about having nulls in the email list, maybe it makes sense to only add to the list if the emails are not null here?

VladNicol (approved):
No further comments, other than what Simon spotted.

SteveJacksonSoft:
> `DigitalLearningSolutions.Data/Services/SupervisorService.cs` contains lots of functions which do SQL queries. These should be probably be moved to a data service.

I mean ideally yes... but this is code that's being managed by HEE which we rarely touch. I'm not convinced it's worth the effort / risk of trying to split the data layer from the service layer.

SteveJacksonSoft:
Ideally we should wait until https://softwiretech.atlassian.net/browse/HEEDLS-932 is merged as that will change the update logic slightly, so we'll have to make changes afterwards.

DigitalLearningSolutions.Data/DataServices/SupervisorDelegateDataService.cs
```diff
@@ -73,10 +79,10 @@ public IEnumerable<SupervisorDelegate> GetPendingSupervisorDelegateRecordsByEmai
                     FROM SupervisorDelegates sd
                     INNER JOIN AdminUsers au ON sd.SupervisorAdminID = au.AdminID
                     WHERE au.CentreID = @centreId
-                      AND sd.DelegateEmail = @email
+                      AND sd.DelegateEmail IN @emails
```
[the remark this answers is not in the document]
SteveJacksonSoft: I'm not sure how Dapper implements this - could you test it? I'd expect it to work fine if `emails` is empty though...
It's possible for multiple supervisors to have invited the same delegate, so this method should return an IEnumerable.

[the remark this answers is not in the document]
simonbrent: Ok, it looks like an empty list results in `AND sd.DelegateEmail IN (SELECT @emails WHERE 1 = 0)`, i.e. a safe but unnecessary query.

If the first item in `emails` is null, you get `System.NotSupportedException : The first item in a list-expansion cannot be null` from Dapper, so that is a problem. Looking at the usages, it should not be possible to create a list where the first entry is null, because the list is created with primary email first, then centre specific, but this is non-obvious and fluky. I therefore propose:

* Change `GetPendingSupervisorDelegateRecordsByEmailsAndCentre` so that `emails` are non-nullable strings
* Change `GetPendingSupervisorDelegateIdsMatchingDelegate` so that `delegateRegistrationModel.CentreSpecificEmail` is only added to `delegateEmails` if it is not null

DigitalLearningSolutions.Web/Controllers/SupervisorController/SupervisorController.cs
```diff
@@ -12,28 +12,28 @@
...
         private readonly IConfigDataService configDataService;
-        private readonly ICentreRegistrationPromptsService centreRegistrationPromptsService;
-        private readonly IUserDataService userDataService;
+        private readonly IFrameworkNotificationService frameworkNotificationService;
         private readonly ILogger<SupervisorController> logger;
-        private readonly IConfiguration config;
         private readonly ISearchSortFilterPaginateService searchSortFilterPaginateService;
+        private readonly ISelfAssessmentService selfAssessmentService;
+        private readonly ISupervisorService supervisorService;
+        private readonly IUserDataService userDataService;
```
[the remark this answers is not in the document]
livzorn: removed all that. Yeah not sure either but HEE wrote those files so I'll leave them as is

DigitalLearningSolutions.Web.Tests/Controllers/LearningPortal/LearningPortalControllerTests.cs
```diff
@@ -74,9 +75,9 @@ public void SetUp()
...
                 searchSortFilterPaginateService
             );
-            controller.ControllerContext = new ControllerContext { HttpContext = new DefaultHttpContext { User = user } };
+            controller.ControllerContext = new ControllerContext
+                { HttpContext = new DefaultHttpContext { User = user } };
             controller = controller.WithMockTempData();
-
         }
     }
 }
```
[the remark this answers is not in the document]
livzorn: It looks like there's no tests at all for the method I made my small change to. Since HEE wrote the method I'm not sure if we want to spend the time writing tests for it. @SteveJacksonSoft do you have an opinion?

DigitalLearningSolutions.Data/DataServices/SupervisorDelegateDataService.cs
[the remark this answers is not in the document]
livzorn: Ah I see, I figured it would be fine since the primary email will never be null so it would always be called with at least one not null value, but better to write it so we only give it non-null values. 

I've changed the service method to filter out null/empty values so we only have valid emails when we call the data service

DigitalLearningSolutions.Data.Tests/DataServices/SupervisorDelegateDataServiceTests.cs
livzorn: Wrote another test that it returns null. I was having toruble checking that a call to the logger was made though, got this error: `The current proxy generator can not intercept the method Microsoft.Extensions.Logging.LoggerExtensions.LogError(Microsoft.Extensions.Logging.ILogger logger, System.String message, System.Object[] args) for the following reason:
    - Extension methods can not be intercepted since they're static.`

DigitalLearningSolutions.Data.Tests/DataServices/SupervisorDelegateDataServiceTests.cs
livzorn: done :)

DigitalLearningSolutions.Data.Tests/DataServices/SupervisorDelegateDataServiceTests.cs
livzorn: Thanks, made those changes

DigitalLearningSolutions.Data.Tests/DataServices/SupervisorDelegateDataServiceTests.cs
livzorn: removed

DigitalLearningSolutions.Web/Controllers/SupervisorController/SupervisorController.cs
```diff
@@ -60,16 +60,9 @@ private int GetAdminID()
...
-            var userEmail = User.GetUserPrimaryEmail();
-            if (userEmail == null)
-            {
-                return "";
-            }
-            else
-            {
-                return userEmail;
-            }
+            var adminId = GetAdminID();
```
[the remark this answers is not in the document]
livzorn: done

DigitalLearningSolutions.Web/Controllers/LearningPortalController/SelfAssessment.cs
```diff
@@ -3,6 +3,7 @@
     using System;
     using System.Collections.Generic;
     using System.Linq;
+    using DigitalLearningSolutions.Data.DataServices.UserDataService;
```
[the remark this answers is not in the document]
livzorn: removed

DigitalLearningSolutions.Data.Tests/DataServices/SupervisorDelegateDataServiceTests.cs
simonbrent: Ok

DigitalLearningSolutions.Data/Services/SupervisorDelegateService.cs
```diff
@@ -30,17 +34,27 @@ ISupervisorDelegateDataService supervisorDelegateDataService
...
         }
 
-        public void AddDelegateIdToSupervisorDelegateRecords(IEnumerable<int> supervisorDelegateIds, int delegateId)
+        public IEnumerable<SupervisorDelegate> GetPendingSupervisorDelegateRecordsByEmailsAndCentre(
+            int centreId,
+            IEnumerable<string?> emails
+        )
         {
-            supervisorDelegateDataService.UpdateSupervisorDelegateRecordsCandidateId(supervisorDelegateIds, delegateId);
+            emails = emails.Where(e => !string.IsNullOrWhiteSpace(e)).ToList();
```
simonbrent, line 47: It's normally not considered good practice to mutate input parameters to a function, so I think I'd prefer this to be 
```
var nonEmptyEmails = emails.Where(e => !string.IsNullOrWhiteSpace(e)).ToList();
```

simonbrent (changes requested):

DigitalLearningSolutions.Web/Controllers/LearningPortalController/SelfAssessment.cs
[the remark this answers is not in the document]
livzorn: Sorry that's weird. Officially gone now

livzorn:
I'll tackle the further changes once HEEDLS-932 is merged, or I'll leave TODO comments for them if it's not merged before I go on holiday next week

PR #1236, opened 2022-06-29
HEEDLS-990 Add DefaultContentTypeId filter to queries for customisati…

AlexJacksonDS:
…ons/applications

### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-990

### Description
Added DefaultContentTypeId <> 4 to all places where we're querying Customisations/Applications.

Fixed some mixed tabs and spaces and line endings.

### Screenshots
N/A

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data/DataServices/CourseAdminFieldsDataService.cs
+++ b/DigitalLearningSolutions.Data/DataServices/CourseAdminFieldsDataService.cs
@@ -121,10 +121,12 @@
             return connection.Query<string>(
                 @$"SELECT
                         cp.CoursePrompt
-                    FROM Customisations c
+                    FROM Customisations AS c
+                    INNER JOIN Applications AS ap ON ap.ApplicationID = c.ApplicationID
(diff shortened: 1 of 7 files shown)


DanBloxham-sw (changes requested):
Yeah, I think these changes are fine, but we still have a few queries that don't use the new filter that I found from searching for "Customisations" or "Applications" in the data project. E.g. `GetCourseFieldPromptIdsForCustomisation`. 

Don't know if the logo needs changing - we normally only pass in the centre id for the centre logo. I'd assume if we reach a page where we are passing in a customisationId, it's one where it already would have been caught in an earlier filter.

There are a number of other places that don't have this filter, but are accessed from paths other than the customisation ID (e.g., Progress ID to Progress table joined on to others) which presumably should be fine as is (assuming the rest of the codebase is good), but we haven't explicitly applied the filter. Do we want those here as well?

DanBloxham-sw (approved):
I think this looks good now.

PR #1251, opened 2022-07-01
TD-267 add new foreign keys to self assessment tables for passporting

JBsoftwire:
### JIRA link
_[TD-267](https://hee-tis.atlassian.net/browse/TD-267)_

### Description
Up-migration, population script, and down-migration for UAR on the SupervisorDelegates and CandidateAssessments tables.

### Screenshots
N/A

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


JBsoftwire:
Data tests for this are failing, due to failure to find the path to the sql script in the migration. I'm not sure why right now and out of time, but I'll figure it out Tuesday morning.

DigitalLearningSolutions.Data.Migrations/202206281724_AddUniquenessConstraintsToCandidateAssessmentsAndSupervisorDelegates.cs
```diff
@@ -0,0 +1,28 @@
...
+        private const string SupervisorDelegatesIndexName = "IX_SupervisorDelegates_DelegateUserId_SupervisorAdminId";
+
+        public override void Up()
+        {
+            Create.Index(CandidateAssessmentsIndexName).OnTable("CandidateAssessments").WithOptions()
+                .UniqueNullsNotDistinct().OnColumn("DelegateUserID").Ascending()
+                .OnColumn("SelfAssessmentID").Ascending();
+            Create.Index(SupervisorDelegatesIndexName).OnTable("SupervisorDelegates").WithOptions()
+                .UniqueNullsNotDistinct().OnColumn("DelegateUserID").Ascending()
+                .OnColumn("SupervisorAdminID").Ascending();
```
VladNicol, line 19: I'm not sure what the usages of this table will be. Do we also want an index only on `SupervisorAdminID` only? For queries by `SupervisorAdminID` without querying by `DelegateUserID` I think this index won't be useful, will we encounter this case often?

DigitalLearningSolutions.Data.Migrations/202206281412_AddForeignKeysToSupervisorDelegatesAndCandidateAssessments.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        public override void Down()
+        {
+            Rename.Column("CandidateID_deprecated").OnTable("SupervisorDelegates").To("CandidateID");
+            Rename.Column("CandidateID_deprecated").OnTable("CandidateAssessments").To("CandidateID");
+
+            Alter.Table("SupervisorDelegates").AlterColumn("CandidateID").AsInt32().Nullable()
+                .ForeignKey("DelegateAccounts", "ID");
+            Alter.Table("CandidateAssessments").AlterColumn("CandidateID").AsInt32().Nullable()
+                .ForeignKey("DelegateAccounts", "ID");
```
VladNicol, line 48: (Not sure if this comment will make any sense) It looks like these columns are not currently nullable and they are made nullable by the Up migration. In general, I think I would expect that after applying the Down migration, the database has the same structure as it had before applying the Up migration.

In the Up migration we populate the new columns and make the deprecated columns nullable. It is possible that we then add new data with the deprecated columns as null. If we were to revert the database to the original structure, the deprecated columns should be non nullable and we would have trouble with those new records where they are null. In theory, we could take those values from the new columns and populate the old columns (kinda like a reversed populating script). It sounds like quite a lot of effort, @SteveJacksonSoft do you think it's worth trying?

VladNicol (approved):
I'm relatively new to these migrations stuff, so I just a couple of questions, not necessarily requesting changes.

SteveJacksonSoft:
Sorry, I'm going to need to take some more time to think through the ramifications of this migration before I'm confident it should be merged. Hopefully it won't take too long...

DigitalLearningSolutions.Data.Migrations/202206281724_AddUniquenessConstraintsToCandidateAssessmentsAndSupervisorDelegates.cs
simonbrent: If the index is created as `(supervisorAdminId, delegateUserId)`, a query on `supervisorAdminId` only will be able to use the index. 

However, these indexes are both created with `DelegateUserID` first. In both cases, the order of the `OnColumn` statements should be reversed so that the indexes can be used in queries which only consider `selfAssessmentID` or `supervisorAdminId`.

DigitalLearningSolutions.Data.Migrations/202206281412_AddForeignKeysToSupervisorDelegatesAndCandidateAssessments.cs
simonbrent: I think this might not be worth worrying about because we should not be inserting data into deprecated columns via the application, so the solution to "it is possible that we then add new data with the deprecated columns as null" is: Don't Do That

DigitalLearningSolutions.Data.Migrations/202206281412_AddForeignKeysToSupervisorDelegatesAndCandidateAssessments.cs
```diff
@@ -0,0 +1,63 @@
...
+                .ForeignKey("Users", "ID");
+
+            Alter.Table("CandidateAssessments")
+                .AddColumn("DelegateUserID").AsInt32().Nullable()
+                .ForeignKey("Users", "ID")
+                .AddColumn("CentreID").AsInt32().Nullable()
+                .ForeignKey("Centres", "CentreID");
+
+            Execute.Script(
+                "DigitalLearningSolutions.Data.Migrations/Scripts/HEEDLS-932-PopulateNewKeysOnSupervisorDelegatesAndCandidateAssessments.sql"
```
simonbrent, line 21: I could not get the migration to run, because this relative path resolved to the wrong absolute path: trying to start the localhost website resulted in the path being `/path/to/my/repo/DigitalLearningSolutions.Data.Web/DigitalLearningSolutions.Data.Migrations/Scripts/HEEDLS-932-PopulateNewKeysOnSupervisorDelegatesAndCandidateAssessments.sql`

You can fix this by changing `Execute.Script` to `Execute.Sql`, putting the SQL from `HEEDLS-932-PopulateNewKeysOnSupervisorDelegatesAndCandidateAssessments.sql` in as an argument, and removing that file

simonbrent (changes requested):

DigitalLearningSolutions.Data.Migrations/202206281724_AddUniquenessConstraintsToCandidateAssessmentsAndSupervisorDelegates.cs
JBsoftwire: Good question, I should have made it clear: The index will not actually be referenced by anything (at least not that I'm aware of). It's a way to create the complex uniqueness constraint required by the ticket.

DigitalLearningSolutions.Data.Migrations/202206281412_AddForeignKeysToSupervisorDelegatesAndCandidateAssessments.cs
JBsoftwire: I discussed this with Steve while coding, we're willing to assume that the down migration is run immediately after the up migration in case of some problem, and that no new data has been added to the DB. So we don't need to worry about null values. If some data is added and then the down migration needs to be run then it will need to be cleaned up manually.

DigitalLearningSolutions.Data.Migrations/202206281724_AddUniquenessConstraintsToCandidateAssessmentsAndSupervisorDelegates.cs
simonbrent: We have existing queries with joins or where clauses on `supervisorAdminId` and `selfAssessmentID`. SQL server does not automatically add indexes on FK relationships, so we don't currently have them. Reversing the order of the columns in these new indexes therefore seems like a good way to kill two birds with one stone

DigitalLearningSolutions.Data.Migrations/202206281412_AddForeignKeysToSupervisorDelegatesAndCandidateAssessments.cs
JBsoftwire: No, that was just my misreading my own code - they should be `NotNullable()`, good catch.

JBsoftwire: Correction, only the CA table column should be not nullable. The SA table column is nullable to begin with.

PR #1275, opened 2022-07-08
DLSV2-573 Add catalogue filtering to signposting

daniel-manta-hee:
### JIRA link
https://hee-dls.atlassian.net/browse/DLSV2-573

### Description
Making a request to learning hub for retrieving catalogues. When user selects a resource, catalogue id is passed as parameter in order to return to the original search when needed.

### Screenshots
image

-----
### Developer checks

I have:

Checked that it works for searches returning multiple pages and it returns to the original search and page when cancel button is clicked.

### DIFF
--- a/DigitalLearningSolutions.Web/Controllers/FrameworksController/Signposting.cs
+++ b/DigitalLearningSolutions.Web/Controllers/FrameworksController/Signposting.cs
@@ -20,6 +20,8 @@ namespace DigitalLearningSolutions.Web.Controllers.FrameworksController
 {
     public partial class FrameworksController
     {
+        private static List<Catalogue> Catalogues { get; set; }
+
         [Route("/Frameworks/{frameworkId}/Competency/{frameworkCompetencyId}/CompetencyGroup/{frameworkCompetencyGroupId}/Signposting")]
         public IActionResult EditCompetencyLearningResources(int frameworkId, int frameworkCompetencyGroupId, int frameworkCompetencyId)
         {
@@ -29,9 +31,12 @@ namespace DigitalLearningSolutions.Web.Controllers.FrameworksController
         }
 
         [Route("/Frameworks/{frameworkId}/Competency/{frameworkCompetencyId}/CompetencyGroup/{frameworkCompetencyGroupId}/Signposting/AddResource/{page=1:int}")]
-        public async Task<IActionResult> SearchLearningResourcesAsync(int frameworkId, int frameworkCompetencyId, int? frameworkCompetencyGroupId, string searchText, int page)
+        public async Task<IActionResult> SearchLearningResourcesAsync(int frameworkId, int frameworkCompetencyId, int? frameworkCompetencyGroupId, int? catalogueId, string searchText, int page)
         {
             var model = new CompetencyResourceSignpostingViewModel(frameworkId, frameworkCompetencyId, frameworkCompetencyGroupId);
+            Catalogues = Catalogues ?? (await this.learningHubApiClient.GetCatalogues())?.Catalogues;
+            model.Catalogues = Catalogues;
+            model.CatalogueId = catalogueId;
             if (frameworkCompetencyGroupId.HasValue)
             {
                 var competency = frameworkService.GetFrameworkCompetencyById(frameworkCompetencyId);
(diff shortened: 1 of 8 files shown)


DigitalLearningSolutions.Web/Controllers/FrameworksController/Signposting.cs
```diff
@@ -29,22 +31,25 @@ public IActionResult EditCompetencyLearningResources(int frameworkId, int framew
         }
 
         [Route("/Frameworks/{frameworkId}/Competency/{frameworkCompetencyId}/CompetencyGroup/{frameworkCompetencyGroupId}/Signposting/AddResource/{page=1:int}")]
-        public async Task<IActionResult> SearchLearningResourcesAsync(int frameworkId, int frameworkCompetencyId, int? frameworkCompetencyGroupId, string searchText, int page)
+        public async Task<IActionResult> SearchLearningResourcesAsync(int frameworkId, int frameworkCompetencyId, int? frameworkCompetencyGroupId, int? catalogueId, string searchText, int page)
         {
-            var model = new CompetencyResourceSignpostingViewModel(frameworkId, frameworkCompetencyId, frameworkCompetencyGroupId);
+            var model = new CompetencyResourceSignpostingViewModel(frameworkId, frameworkCompetencyId, frameworkCompetencyGroupId);
+            Catalogues = Catalogues ?? (await this.learningHubApiClient.GetCatalogues())?.Catalogues;
```
kevwhitt-hee, line 37: Can we do something (maybe with linq) to put this list of catalogues in alphabetical order?

DigitalLearningSolutions.Web/Views/Frameworks/Developer/AddCompetencyLearningResources.cshtml
```diff
@@ -1,8 +1,15 @@
...
-  ViewData["Application"] = "Framework Service";
+    ViewData["Title"] = "Add Competency Learning Resource";
+    ViewData["Application"] = "Framework Service";
+    var dropdownItems = Model.Catalogues
+          .Select(c => new SelectListItem(
+            text: c.Name,
+            value: c.Id.ToString(),
+            selected: c.Id == Model.CatalogueId)
+          ).ToList();
+    dropdownItems.Insert(0, new SelectListItem("Any", "null", !Model.CatalogueId.HasValue));
```
kevwhitt-hee, line 12: Can we change "Any" to "Any catalogue", otherwise we'll need to label the dropdown.

DigitalLearningSolutions.Web/Views/Frameworks/Developer/AddCompetencyLearningResources.cshtml
```diff
@@ -46,11 +66,25 @@
...
+                asp-route-page="@Model.Page"
+                asp-route-frameworkId="@Model.FrameworkId"
+                asp-route-frameworkCompetencyId="@Model.FrameworkCompetencyId"
+                asp-route-frameworkCompetencyGroupId="@Model.FrameworkCompetencyGroupId"
+                asp-route-catalogueId="@Model.CatalogueId">
           Search
         </button>
       </div>
+      <div class="filter-value-container">
+        @Html.DropDownListFor(c => c.CatalogueId, dropdownItems,
```
kevwhitt-hee, line 80: Accessibility tests are complaining that this dropdown doesn't have a unique ID. There is a hidden input on the page with the same Id - "CatalogueId"

kevwhitt-hee (changes requested):
Working well. Please could we list the catalogues in alphabetical order, in the dropdown? Also could we sort the duplicate "CatalogueId" control id issue, please?

kevwhitt-hee:
Could we also use cookies to store the catalogue a user has selected and select it in the dropdown (if present) when the signposting page loads? Most users will be looking for resources in a single catalogue and this will make things much easier for them.

kevwhitt-hee (changes requested):
Please could we change the use of tempdata to store the catalogue filter to use the approach used elsewhere in the application?
Suggested approach:
Use the Web.Helpers.FilterCookieHelper UpdateFilterCookie method to store the catalogue filter.
Add a second GetFilterCookie method to that helper based on the implementation in Data.Helpers.FilteringHelpers GetFilterString method to retrieve the filter value from the cookie.

kevwhitt-hee:
> Please could we change the use of tempdata to store the catalogue filter to use the approach used elsewhere in the application? Suggested approach: Use the Web.Helpers.FilterCookieHelper UpdateFilterCookie method to store the catalogue filter. Add a second GetFilterCookie method to that helper based on the implementation in Data.Helpers.FilteringHelpers GetFilterString method to retrieve the filter value from the cookie.

Instead of the above which handles more complex filters as you pointed out, please use your existing cookie implementation with the same expiry as that used elsewhere:
`private static readonly DateTimeOffset CookieExpiry = DateTimeOffset.UtcNow.AddDays(30);`

kevwhitt-hee (approved):

PR #1287, opened 2022-07-13
Heedls 902 internal admin self registration journey fixes

VladNicol:
### JIRA link
[HEEDLS-902](https://softwiretech.atlassian.net/browse/HEEDLS-902)

### Description
I addressed the testing comments:

- I ended up making the page look more like what Olivia suggested on the ticket because it was easy and I also think it looks a bit better.
- If the user already has a delegate at the same centre, I don't update the centre specific email for the delegate because it is already updated when creating the admin account
- If the user does not have a delegate account at the centre, I create one and pass a null centre email to avoid `UserDataService.CentreSpecificEmailIsInUseAtCentre` returning false (the previous creation of the admin account sets the centre email anyway)
- I fixed displaying both errors if the centre specific email in the form is invalid and also doesn't match the centre's email
- (I think) I fixed the redirecting if the form is accessed when not logged in

### Screenshots
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Web/Controllers/Register/RegisterInternalAdminController.cs
```diff
@@ -101,10 +101,11 @@ public async Task<IActionResult> Index(InternalAdminInformationViewModel model)
 
             if (delegateAccount == null)
             {
+                // We can have the centreSpecificEmail = null here because we already set it for the admin account
                 registrationService.CreateDelegateAccountForExistingUser(
                     new InternalDelegateRegistrationModel(
                         model.Centre!.Value,
-                        model.CentreSpecificEmail,
+                        null,
```
SteveJacksonSoft, line 108: I think the root of the problem is within `CreateDelegateAccountForExistingUser`. It shouldn't assume that I don't already have an admin account with a centre email set at the centre. Can we fix it there instead?

DigitalLearningSolutions.Web/Views/RegisterInternalAdmin/Index.cshtml
```diff
@@ -1,51 +1,51 @@
...
+      </div>
+
+      <div class="nhsuk-form-group">
+        <label class="nhsuk-label">
+          Primary email
+        </label>
+        <div class="nhsuk-hint">
+          This will be used for logging into the system. All communications will be sent to this email by default. You can change your primary email on the My Account page.
+        </div>
+        <p class="nhsuk-body-m nhsuk-u-margin-top-2"><b>@Model.PrimaryEmail</b></p>
```
SteveJacksonSoft, line 34: Don't know if I like it being bold, makes it look... just.... inconsistent.

SteveJacksonSoft (changes requested):

DigitalLearningSolutions.Web/Controllers/Register/RegisterInternalAdminController.cs
SteveJacksonSoft: Could you change this back so it's just setting the centre-specific email again? Then it'll be fixed when [HEEDLS-1002](https://github.com/TechnologyEnhancedLearning/DLSV2/pull/1286) is merged.

SteveJacksonSoft (approved):
Merge away!

PR #1311, opened 2022-07-27
Heedls 975 claim account set password

simonbrent:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-975

### Description
* Added a new view for claiming an account when the user needs to set a password, making the wireframe in Jira
* Added a new POST route and modified the ClaimAccountService to handle the case where the user needs to set a password
  * The old POST route has been changed to `CompleteRegistrationWithoutPassword`, the new one is `CompleteRegistration`. This is because there is a requirement to show a view with errors when there are password fields, and we want to keep the URL the same (i.e. /ClaimAccount/CompleteRegistration) when this happens.
  * The `CompleteRegistration` POST route has password/confirm password in the HTTP body, but `email` and `code` as part of the query string. This is also so that the URL will be unchanged if the POST request results in errors which need to be displayed on the page.
* Created a partial view for fields required when changing a user's password - Shared/_SetPasswordFields.cshtml - used by
  * ChangePassword/Index.cshtml
  * Register/Password.cshtml
  * RegisterAdmin/Password.cshtml
  * ResetPassword/Index.cshtml
  * SetPassword/Index.cshtml

#### Complete registration page with password fields
complete_registration

#### Error messages due to bad password input
bad_password

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/Controllers/Register/ClaimAccountControllerTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/Controllers/Register/ClaimAccountControllerTests.cs
@@ -81,15 +81,10 @@
         }
 
         [Test]
-        public void IndexGet_with_no_existing_user_redirects_to_AccessDenied()
+        public void IndexGet_with_no_user_to_be_claimed_redirects_to_AccessDenied()
         {
             // Given
-            A.CallTo(
-                () => userDataService.GetUserIdAndCentreForCentreEmailRegistrationConfirmationHashPair(
-                    DefaultEmail,
-                    DefaultCode
-                )
-            ).Returns((null, null, null));
+            GivenEmailAndCodeDoNotMatchAUserToBeClaimed(DefaultEmail, DefaultCode);
 
             // When
             var result = controller.Index(DefaultEmail, DefaultCode);
@@ -159,15 +154,10 @@
         }
 
         [Test]
-        public void CompleteRegistrationGet_with_no_existing_user_redirects_to_AccessDenied()
+        public void CompleteRegistrationGet_with_no_user_to_be_claimed_redirects_to_AccessDenied()
         {
             // Given
-            A.CallTo(
-                () => userDataService.GetUserIdAndCentreForCentreEmailRegistrationConfirmationHashPair(
-                    DefaultEmail,
-                    DefaultCode
-                )
-            ).Returns((null, null, null));
+            GivenEmailAndCodeDoNotMatchAUserToBeClaimed(DefaultEmail, DefaultCode);
 
             // When
             var result = controller.CompleteRegistration(DefaultEmail, DefaultCode);
@@ -223,7 +213,7 @@
         public async Task CompleteRegistrationWithoutPassword_with_email_in_use_returns_NotFound()
         {
             // Given
-            var model = GivenClaimAccountViewModel();
+            var model = GivenClaimAccountViewModel(wasPasswordSetByAdmin: true);
 
             A.CallTo(() => userDataService.PrimaryEmailIsInUse(model.Email)).Returns(true);
 
@@ -270,7 +260,11 @@
         )
         {
             // Given
-            var model = GivenClaimAccountViewModel(email: email, registrationConfirmationHash: code);
+            var model = GivenClaimAccountViewModel(
+                wasPasswordSetByAdmin: true,
+                email: email,
+                registrationConfirmationHash: code
+            );
 
             // When
             var result = await controller.CompleteRegistrationWithoutPassword(
@@ -284,17 +278,12 @@
         }
 
         [Test]
-        public async Task CompleteRegistrationWithoutPassword_with_no_existing_user_redirects_to_AccessDenied()
+        public async Task CompleteRegistrationWithoutPassword_with_no_user_to_be_claimed_redirects_to_AccessDenied()
         {
             // Given
-            var model = GivenClaimAccountViewModel();
+            var model = GivenClaimAccountViewModel(wasPasswordSetByAdmin: true);
 
-            A.CallTo(
-                () => userDataService.GetUserIdAndCentreForCentreEmailRegistrationConfirmationHashPair(
-                    model.Email,
-                    model.RegistrationConfirmationHash
-                )
-            ).Returns((null, null, null));
+            GivenEmailAndCodeDoNotMatchAUserToBeClaimed(model.Email, model.RegistrationConfirmationHash);
 
             // When
             var result = await controller.CompleteRegistrationWithoutPassword(
@@ -313,7 +302,7 @@
         public async Task CompleteRegistrationWithoutPassword_with_logged_in_user_redirects_to_LinkDlsAccount()
         {
             // Given
-            var model = GivenClaimAccountViewModel();
+            var model = GivenClaimAccountViewModel(wasPasswordSetByAdmin: true);
 
             // When
             var result = await controllerWithLoggedInUser.CompleteRegistrationWithoutPassword(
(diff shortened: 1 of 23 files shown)


DigitalLearningSolutions.Web.Tests/Controllers/Register/ClaimAccountControllerTests.cs
```diff
@@ -171,63 +188,216 @@ public void CompleteRegistrationGet_with_logged_in_user_redirects_to_LinkDlsAcco
...
+                model.RegistrationConfirmationHash
+            );
 
             // Then
+            result.Should().BeRedirectToActionResult().WithControllerName("LearningSolutions")
+                .WithActionName("AccessDenied");
+        }
+
+        [Test]
+        public async Task CompleteRegistrationWithoutPassword_with_no_existing_user_redirects_to_AccessDenied()
```
SteveJacksonSoft, line 287: Is this for the case where the email and hash don't match any user-to-be-claimed?

DigitalLearningSolutions.Web.Tests/Controllers/Register/ClaimAccountControllerTests.cs
```diff
@@ -240,48 +410,109 @@ public void CompleteRegistrationPost_with_password_not_set_returns_NotFound()
...
+            );
 
             // Then
+            ACallToConvertTemporaryUserToConfirmedUserMustNotHaveHappened();
+
             result.Should().BeRedirectToActionResult().WithActionName("LinkDlsAccount");
         }
 
+        [Test]
+        public async Task CompleteRegistrationPost_with_invalid_model_display_view_with_errors()
```
SteveJacksonSoft, line 482: Can we say "with_validation_errors" instead? Currently the name reads as if the user will be redirected to an error page.

DigitalLearningSolutions.Web/ViewModels/Common/ViewComponents/SetPasswordViewModel.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace DigitalLearningSolutions.Web.ViewModels.Common.ViewComponents
+{
+    public class SetPasswordViewModel : ConfirmPasswordViewModel
+    {
+        public SetPasswordViewModel(
+            string label = "Password"
```
SteveJacksonSoft, line 6: Feels a bit weird that you're constructing a separate confirm password view model rather than taking in the one that the view model has.
Because we don't send passwords back to the user for security's sake, it doesn't make a difference, but if we did, I think this setup would stop dotnet from populating the input fields with the data we've returned.

SteveJacksonSoft (changes requested):

DigitalLearningSolutions.Web/ViewModels/Common/ViewComponents/SetPasswordViewModel.cs
SteveJacksonSoft: (I think this should be changed)

DigitalLearningSolutions.Web/Controllers/Register/ClaimAccountController.cs
```diff
@@ -85,16 +109,21 @@ public IActionResult CompleteRegistrationPost(string email, string code)
                 return NotFound();
             }
 
-            if (!model.PasswordSet)
+            if (model.WasPasswordSetByAdmin != passwordMustAlreadyBeSet)
```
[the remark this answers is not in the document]
simonbrent: It does mean we have to put the 
```
            if (model == null)
            {
                return RedirectToAction("AccessDenied", "LearningSolutions");
            }
```
validation into both public methods too, but ok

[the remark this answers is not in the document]
simonbrent: It also means we change the order of precedence of the sad paths.

Previously:
* If you're logged in, redirect to /LinkDlsAccount
* THEN redirect to /AccessDenied if the code/email don't match
* THEN redirect to NotFound if the email is in use or the password is already set/not already set depending on the route

Now:
* Redirect to /AccessDenied if the code/email don't match
* THEN redirect to NotFound if password is already set/not already set depending on the route
* THEN redirect to /LinkDlsAccount if you're logged in
* THEN redirect to NotFound if the email is in use

This is a slight deviation from the description in Jira

DigitalLearningSolutions.Web/Controllers/Register/ClaimAccountController.cs
```diff
@@ -239,5 +268,21 @@ public IActionResult AccountAlreadyExists(string email, string centreName)
...
+        )
+        {
+            return new ClaimAccountCompleteRegistrationViewModel
+            {
+                Email = model.Email,
+                Code = model.RegistrationConfirmationHash,
+                CentreName = model.CentreName,
+                WasPasswordSetByAdmin = model.WasPasswordSetByAdmin,
+                Password = formData?.Password,
+                ConfirmPassword = formData?.ConfirmPassword,
```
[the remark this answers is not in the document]
simonbrent: Oh yeah, that was dumb!

DigitalLearningSolutions.Web/Controllers/Register/ClaimAccountController.cs
```diff
@@ -61,12 +63,34 @@ public IActionResult CompleteRegistration(string email, string code)
...
+        public async Task<IActionResult> CompleteRegistrationWithoutPassword(string email, string code)
+        {
+            return await CompleteRegistrationPost(email, code, true);
+        }
+
+        private async Task<IActionResult> CompleteRegistrationPost(
+            string email,
+            string code,
+            bool passwordMustAlreadyBeSet,
+            ConfirmPasswordViewModel? formData = null
```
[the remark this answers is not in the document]
simonbrent: Done

DigitalLearningSolutions.Web/Views/ClaimAccount/AccountAlreadyExists.cshtml
```diff
@@ -1,4 +1,6 @@
-﻿@using DigitalLearningSolutions.Web.ViewModels.Register
+﻿@using DigitalLearningSolutions.Web.ViewComponents
+@using DigitalLearningSolutions.Web.ViewModels.Register.ClaimAccount
+@using Microsoft.AspNetCore.Mvc.TagHelpers
```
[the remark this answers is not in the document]
simonbrent: They got auto-added by Rider. Is this wrong?

DigitalLearningSolutions.Web.Tests/Controllers/Register/ClaimAccountControllerTests.cs
simonbrent: Yes. I've renamed these tests `*_with_no_user_to_be_claimed_redirects_to_AccessDenied` and added another private method: `GivenEmailAndCodeDoNotMatchAUserToBeClaimed`

DigitalLearningSolutions.Web.Tests/Controllers/Register/ClaimAccountControllerTests.cs
simonbrent: Ok

DigitalLearningSolutions.Web/Views/ClaimAccount/AccountAlreadyExists.cshtml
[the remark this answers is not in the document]
simonbrent: Removed

DigitalLearningSolutions.Web/ViewModels/Common/ViewComponents/SetPasswordViewModel.cs
simonbrent: Done

DigitalLearningSolutions.Web.Tests/Controllers/Register/ClaimAccountControllerTests.cs
SteveJacksonSoft: Very nice!

SteveJacksonSoft (approved):
Looking great!

PR #1316, opened 2022-07-28
Heedls 874b fix synchronisation

JBsoftwire:
### JIRA link
_[HEEDLS-874](https://softwiretech.atlassian.net/browse/HEEDLS-874)_

### Description
The groups to synchronise with are now properly selected for new registrations or edits to established users, and a bug in the composition of the `RegistrationFieldAnswers` object in the `UserService` has been fixed.

Some simple new methods have not had tests added. 

### Screenshots
N/A

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Web.Tests/Services/GroupServiceTests/GroupsServiceSynchroniseGroupsTests.cs
+++ b/DigitalLearningSolutions.Web.Tests/Services/GroupServiceTests/GroupsServiceSynchroniseGroupsTests.cs
@@ -32,7 +32,7 @@
                 centreAnswersData,
                 delegateDetails.GetRegistrationFieldAnswers(),
                 null,
-                TODO
+                new List<Group>()
             );
 
             // Then
@@ -57,9 +57,6 @@
                 linkedToField: 2,
                 changesToRegistrationDetailsShouldChangeGroupMembership: true
             );
-            A.CallTo(() => groupsDataService.GetGroupsForCentre(A<int>._)).Returns(
-                new List<Group> { synchronisedGroup }
-            );
 
             // When
             groupsService.SynchroniseUserChangesWithGroups(
@@ -68,7 +65,7 @@
                 centreAnswersData,
                 UserTestHelper.GetDefaultRegistrationFieldAnswers(reusableDelegateDetails.CentreId),
                 null,
-                TODO
+                new List<Group> { synchronisedGroup }
             );
 
             // Then
@@ -94,9 +91,6 @@
                 linkedToField: 1,
                 changesToRegistrationDetailsShouldChangeGroupMembership: true
             );
-            A.CallTo(() => groupsDataService.GetGroupsForCentre(A<int>._)).Returns(
-                new List<Group> { synchronisedGroup }
-            );
 
             // When
             groupsService.SynchroniseUserChangesWithGroups(
@@ -105,7 +99,7 @@
                 centreAnswersData,
                 UserTestHelper.GetDefaultRegistrationFieldAnswers(reusableDelegateDetails.CentreId),
                 null,
-                TODO
+                new List<Group> { synchronisedGroup }
             );
 
             // Then
@@ -131,9 +125,6 @@
                 linkedToField: 1,
                 changesToRegistrationDetailsShouldChangeGroupMembership: true
             );
-            A.CallTo(() => groupsDataService.GetGroupsForCentre(A<int>._)).Returns(
-                new List<Group> { synchronisedGroup }
-            );
 
             // When
             groupsService.SynchroniseUserChangesWithGroups(
@@ -142,7 +133,7 @@
                 centreAnswersData,
                 UserTestHelper.GetDefaultRegistrationFieldAnswers(reusableDelegateDetails.CentreId, answer1: "old answer"),
                 null,
-                TODO
+                new List<Group> { synchronisedGroup }
             );
 
             // Then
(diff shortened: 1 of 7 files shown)


DigitalLearningSolutions.Web/Services/GroupsService.cs
```diff
@@ -601,9 +645,19 @@ private IEnumerable<Group> GetSynchronisedGroupsForCentre(int centreId)
...
+        private IEnumerable<Group> GetAddNewRegistrantGroupsForCentre(int centreId)
+        {
+            return groupsDataService.GetGroupsForCentre(centreId)
+                .Where(g => g.ShouldAddNewRegistrantsToGroup);
+        }
+
         private static bool GroupLabelMatchesAnswer(string groupLabel, string? answer, string linkedFieldName)
         {
-            return !string.IsNullOrEmpty(answer) && string.Equals(groupLabel, answer, StringComparison.CurrentCultureIgnoreCase) || string.Equals(
+            return !string.IsNullOrEmpty(answer) && string.Equals(
```
simonbrent, line 656: I should have spotted this in the earlier review: this code contains an incorrect order of operations.

`a && b || c` === `(a && b) || c`, whereas I think what we want here is `a && (b || c)`. 

As it stands, given the condition that `answer` is an empty string and `groupLabel` is `{linkedFieldName} - `, this function will return `true`, which would _probably_ be surprising. Obviously we know this situation is not going to exist, but I generally consider it code-smell to have a mixture of `&&` and `||` without brackets to explicitly show what is happening. 

As an added bonus, changing this to `a && (b || c)` makes it slightly more optimized: currently if `answer` is null or empty string, the second `string.Equals` gets executed, and as it _should_ be impossible for this to return true (given sensible values for `groupLabel` in the database), this is an unnecessary operation.

DigitalLearningSolutions.Web.Tests/Services/RegistrationServiceTests.cs
```diff
@@ -1292,6 +1239,22 @@ public void CreateDelegateAccountForExistingUser_adds_new_delegate_to_groups()
...
+                            m.Answer3 == model.Answer3 &&
+                            m.Answer4 == model.Answer4 &&
+                            m.Answer5 == model.Answer5 &&
+                            m.Answer6 == model.Answer6 &&
+                            m.JobGroup == userEntity.UserAccount.JobGroupId &&
+                            m.Centre == model.Centre
+                            ))
+            ).MustHaveHappenedOnceExactly();
+            /*
             A.CallTo(
```
simonbrent, line 1258: Remove commented out code

DigitalLearningSolutions.Web.Tests/Services/GroupServiceTests/GroupsServiceSynchroniseGroupsTests.cs
```diff
@@ -359,17 +344,15 @@ public void
...
 
             // When
             groupsService.SynchroniseUserChangesWithGroups(
                 reusableDelegateDetails.Id,
                 reusableEditAccountDetailsData,
                 centreAnswersData,
                 UserTestHelper.GetDefaultRegistrationFieldAnswers(reusableDelegateDetails.CentreId),
-                null
+                null,
+                new List<Group> { synchronisedGroup1, synchronisedGroup2 }
```
simonbrent, line 355: I think there should be a couple of new tests added:
* Call `SynchroniseUserChangesWithSynchronisedGroups` with `groupsDataService.GetGroupsForCentre` returning one group with `ChangesToRegistrationDetailsShouldChangeGroupMembership` = true, and one where it = false, where both groups fulfil the criteria required to appear in `groupsToAddDelegateTo`. Check that a call to `groupsDataService.AddDelegateToGroup` was only made for the correct group.
* Do the same for `AddNewlyRegisteredUserToAddNewRegistrantGroups`, but setting `ShouldAddNewRegistrantsToGroup` instead of `ChangesToRegistrationDetailsShouldChangeGroupMembership`.

(Note: I have suggested new names for these functions, below)

simonbrent (changes requested):

DigitalLearningSolutions.Web.Tests/Services/RegistrationServiceTests.cs
JBsoftwire: What's distressing me here is I checked this file twice.

DigitalLearningSolutions.Web.Tests/Services/GroupServiceTests/GroupsServiceSynchroniseGroupsTests.cs
```diff
@@ -359,6 +370,97 @@ public void
...
+            // Then
+            DelegateMustHaveBeenAddedToGroups(new List<int> { synchronisedGroup.GroupId });
+            A.CallTo(
+                () => groupsDataService.AddDelegateToGroup(
+                    A<int>._,
+                    A<int>._,
+                    A<DateTime>._,
+                    A<int>._
+                )
+            ).MustHaveHappenedOnceExactly();
```
simonbrent, line 414: It's not immediately clear from looking at this test that you are testing both that 
1. the delegate has been added to `synchronisedGroup`, and
2. the delegate has not been added to `unsynchronisedGroup`

Testing that `AddDelegateToGroup` only happened once is a proxy for 2. but it would be better to explicitly test that 
```
 A.CallTo(
    () => groupsDataService.AddDelegateToGroup(
        reusableDelegateDetails.Id,
        unsynchronisedGroup.GroupId,
        A<DateTime>._,
        A<int>._
    )
).MustNotHaveHappened();
````

File: DigitalLearningSolutions.Web.Tests/Services/GroupServiceTests/GroupsServiceSynchroniseGroupsTests.cs [hunk 5]
```diff
@@ -359,6 +370,97 @@ public void
...
```
simonbrent, line 461: Same as above

simonbrent (changes requested):

simonbrent (approved):
Great, thanks

PR #1323, opened 2022-08-03
Heedls 1024 verify email

simonbrent:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-1024

### Description
* Added a VerifyEmailController with a single endpoint: `/VerifyEmail`
* Added service and data service methods to get user/user centre details records matching the given email and code
* Added service and data service methods to update `EmailVerified` and `EmailVerificationHashID` for all user/user centre details records matching the given email if any email/code match is found in either `Users` or `UserCentreDetails` tables

#### Notes:
* I have done one thing that isn't specified by the issue: I have used `[Authorize(Policy = CustomPolicies.BasicUser)]` in the controller. It isn't strictly necessary, but I don't think there's a downside to doing it - anyone who wants to verify an email will already have a login.
* I have added a `TransactionScope` to the route (in the controller) to wrap the service method which updates the database. The scope of https://softwiretech.atlassian.net/browse/HEEDLS-994 states that we will not automatically add `TransactionScope` for GET routes, so this is necessary. I think that we _should_ add them for GET routes, but that is a decision for @SteveJacksonSoft.
* I have not added an accessibility test for the new endpoint, because it modifies the database, so the test will only pass the first time it is run - subsequent runs will result in NotFound instead of the page we expect.

### Screenshots
verified
---
verification_link_expired

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data/Models/EmailVerificationDetails.cs
+++ b/DigitalLearningSolutions.Data/Models/EmailVerificationDetails.cs
@@ -1,0 +1,21 @@
+﻿namespace DigitalLearningSolutions.Data.Models
+{
+    using System;
+    using DigitalLearningSolutions.Data.Utilities;
+
+    public class EmailVerificationDetails
+    {
+        public int UserId { get; set; }
+        public string Email { get; set; } = null!;
+        public string EmailVerificationHash { get; set; } = null!;
+        public DateTime? EmailVerified { get; set; }
+        public DateTime EmailVerificationHashCreatedDate { get; set; }
+
+        public bool IsEmailVerified => EmailVerified != null;
+
+        public bool HasVerificationExpired(IClockUtility clockUtility)
+        {
+            return EmailVerificationHashCreatedDate <= clockUtility.UtcNow - TimeSpan.FromDays(3);
+        }
+    }
+}
--- a/DigitalLearningSolutions.Web/Controllers/VerifyEmailController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/VerifyEmailController.cs
@@ -1,5 +1,6 @@
 ﻿namespace DigitalLearningSolutions.Web.Controllers
 {
+    using System.Transactions;
     using DigitalLearningSolutions.Data.Utilities;
     using DigitalLearningSolutions.Web.Attributes;
     using DigitalLearningSolutions.Web.Helpers;
@@ -45,11 +46,15 @@
                 return View("VerificationLinkExpired");
             }
 
+            using var transaction = new TransactionScope();
+
             userService.SetEmailVerified(
                 emailVerificationDetails.UserId,
                 emailVerificationDetails.Email,
                 clockUtility.UtcNow
             );
+
+            transaction.Complete();
 
             return View();
         }
--- a/DigitalLearningSolutions.Web/Controllers/VerifyEmailController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/VerifyEmailController.cs
@@ -1,0 +1,57 @@
+﻿namespace DigitalLearningSolutions.Web.Controllers
+{
+    using DigitalLearningSolutions.Data.Utilities;
+    using DigitalLearningSolutions.Web.Attributes;
+    using DigitalLearningSolutions.Web.Helpers;
+    using DigitalLearningSolutions.Web.Models.Enums;
+    using DigitalLearningSolutions.Web.Services;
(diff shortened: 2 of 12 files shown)


DigitalLearningSolutions.Web/Controllers/VerifyEmailController.cs
```diff
@@ -0,0 +1,62 @@
...
+    using System.Transactions;
+    using DigitalLearningSolutions.Data.Utilities;
+    using DigitalLearningSolutions.Web.Attributes;
+    using DigitalLearningSolutions.Web.Helpers;
+    using DigitalLearningSolutions.Web.Models.Enums;
+    using DigitalLearningSolutions.Web.Services;
+    using Microsoft.AspNetCore.Authorization;
+    using Microsoft.AspNetCore.Mvc;
+
+    [Authorize(Policy = CustomPolicies.BasicUser)]
```
SteveJacksonSoft, line 12: This endpoint should be unauthenticated.

simonbrent: Done

simonbrent:
Rebased from uar-test + force push to get the chromedriver update so that the Jenkins pipeline doesn't fail

DigitalLearningSolutions.Data/Models/EmailVerificationDetails.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace DigitalLearningSolutions.Data.Models
+{
+    using System;
+    using DigitalLearningSolutions.Data.Utilities;
+
+    public class EmailVerificationDetails
+    {
+        public int UserId { get; set; }
+        public string Email { get; set; } = null!;
+        public string EmailVerificationHash { get; set; } = null!;
```
livzorn, line 10: For my own clarification, why are you setting these to null and not make these nullable strings?

DigitalLearningSolutions.Web/Services/UserService.cs
```diff
@@ -411,6 +415,35 @@ public void SetCentreEmails(int userId, Dictionary<int, string?> centreEmailsByC
...
+            }
+
+            return null;
+        }
+
+        public void SetEmailVerified(int userId, string email, DateTime verifiedDateTime)
+        {
+            userDataService.SetPrimaryEmailVerified(userId, verifiedDateTime);
+            userDataService.SetCentreEmailVerified(userId, email, verifiedDateTime);
+        }
```
livzorn, line 445: Shouldn't we be passing `email `to the SetPrimaryEmailVerified method as well so it only sets email to verified if the email matches the user's primary email?

livzorn (commented):
Just a couple questions

DigitalLearningSolutions.Data/Models/EmailVerificationDetails.cs
simonbrent: `EmailVerificationHashes.EmailVerificationHash` is a not nullable database column. `GetCentreEmailVerificationDetails` (which is the only place this model is created) will always return either a `EmailVerificationDetails` (when in input code matches a value in `EmailVerificationHashes.EmailVerificationHash`), or null (when it doesn't). You can therefore guarantee that none of `UserId`, `Email`, `EmailVerificationHash`, or `EmailVerificationHashCreatedDate` will ever be null, because the model can only exist if they are not null. (`EmailVerified` is nullable because `UserCentreDetails.EmailVerified` is a nullable database column.)

DigitalLearningSolutions.Web/Services/UserService.cs
simonbrent: I suppose we _could_, but it's not strictly speaking necessary (for the current implementation, at least). You won't have got here without first having confirmed that an email/code resolves to a particular userId. It is necessary to pass email into `SetCentreEmailVerified` because we don't have `delegateId` available here, and need to be able to update just one row in the `UserCentreDetails` table, but this doesn't hold for the `Users` table where there is only one row per `userId`.

livzorn: Maybe I'm missing something but it seems like if a user clicks a link to verify their centre email address, it will automatically verify their primary email as well?

simonbrent: You are not missing something. I was missing something. You are right, and I will change it!

livzorn (approved):

PR #1325, opened 2022-08-04
HEEDLS-915 Send verification emails when editing My Account details

VladNicol:
### JIRA link
[HEEDLS-915](https://softwiretech.atlassian.net/browse/HEEDLS-915)

### Description
I created the `EmailVerificationService` that sends verification emails to unverified addresses when a user changes their primary/centre emails.

I added an `EmailVerified` parameter to `UserDataService.UpdateUser`. When calling `UserDataService.UpdateUser` and `UserDataService.SetCentreEmai`l, I check whether the email is getting changed or if it is already verified or set by an admin for a delegate.

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -239,8 +249,29 @@ DlsSubApplication dlsSubApplication
                 delegateAccount?.Id
             );
 
+            var modifiedEmails = new List<(string, int?)>();
```
SteveJacksonSoft, line 252: We should make changes to this method, because it is now _massive_
We should split it into smaller methods, and move as many of them as we can into a new MyAccount service
* Validate data requirements and formatting
* If invalid return
* Validate email uniqueness
* If invalid return
* Get list of changed emails
* Save changes
* Send the user on

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -239,8 +249,29 @@ DlsSubApplication dlsSubApplication
                 delegateAccount?.Id
             );
 
+            var modifiedEmails = new List<(string, int?)>();
+            var shouldLogIntoCentrelessAccount = false;
```
SteveJacksonSoft, line 253: This is really `shouldRedirectToVerifyEmail && centreId != null`. It would be safer not to track it separately.

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
SteveJacksonSoft: I'd also like to separate the centreless scenario from the logged-into-centre scenario a bit more strongly, rather than repeatedly checking which scenario you're in. Probably branch after doing validation.

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -415,5 +468,19 @@ DlsSubApplication dlsSubApplication
 
             return delegateAccount is { Active: true } ? delegateAccount : null;
         }
+
+        private async Task LogOutAndLogIntoCentrelessAccount(UserAccount userAccount)
+        {
+            var claims = LoginClaimsHelper.GetClaimsForCentrelessSignIn(userAccount);
+            var claimsIdentity = new ClaimsIdentity(claims, "Identity.Application");
+            var authProperties = new AuthenticationProperties
```
SteveJacksonSoft, line 476: You should check whether remember-me has been ticked too. I think the property is called `IsPersistent` or something. Preserve that in the AuthenticationProperties.

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -415,5 +468,19 @@ DlsSubApplication dlsSubApplication
 
             return delegateAccount is { Active: true } ? delegateAccount : null;
         }
+
+        private async Task LogOutAndLogIntoCentrelessAccount(UserAccount userAccount)
```
SteveJacksonSoft, line 472: Rather than duplicating the logout and login logic, we should reuse the methods used by LoginController for logging users in and out.

DigitalLearningSolutions.Web/Services/EmailVerificationService.cs
```diff
@@ -0,0 +1,133 @@
...
+    using DigitalLearningSolutions.Data.Models.Email;
+    using DigitalLearningSolutions.Data.Models.User;
+    using DigitalLearningSolutions.Data.Utilities;
+    using Microsoft.Extensions.Configuration;
+    using MimeKit;
+
+    public interface IEmailVerificationService
+    {
+        // This method returns true if any email needs verification
+        bool SendVerificationEmails(UserAccount userAccount, List<(string, int?)> emails);
```
SteveJacksonSoft, line 17: We talked about pulling two methods out of this: 

```
public void EnsureEmailVerificationStatusIsAccurate() { }
public bool AccountEmailRequiresVerification() { }
```

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
VladNicol: I split that huge method into several smaller ones, but I didn't move them in a service, I don't see the point of doing that. The controller will still call all those methods and will have the same level of complexity. We will not call those methods in another controller, so I don't see the difference between having those methods in a service and having them in the controller.

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -248,42 +387,41 @@ DlsSubApplication dlsSubApplication
...
+                if (shouldLogIntoCentrelessAccount)
                 {
-                    ModelState.AddModelError(
-                        $"{nameof(MyAccountEditDetailsFormData.AllCentreSpecificEmailsDictionary)}_{centreId}",
-                        CommonValidationErrorMessages.CentreEmailAlreadyInUse
-                    );
+                    var claim = ((ClaimsIdentity)User.Identity).FindFirst("IsPersistent");
+                    var isPersistent = claim != null && Convert.ToBoolean(claim.Value);
+                    await HttpContext.Logout();
+                    await this.CentrelessLogInAsync(userAccount, isPersistent);
```
VladNicol, line 415: This doesn't work because var claim is null. I wasn't very confident about it, I found the code here: https://stackoverflow.com/questions/40341591/how-to-get-is-session-ispersistent-on-asp-net-mvc

I'm also getting a `DuplicateKeyException` in `NoCachingAttribute.OnActionExecuted` sometimes when I run the code, I assume it's because of this part.

DigitalLearningSolutions.Web/Services/EmailVerificationService.cs
```diff
@@ -0,0 +1,115 @@
...
+
+                foreach (var emailUpdate in emailGroup)
+                {
+                    UpdateEmailVerificationHashId(userAccount.Id, emailUpdate.CentreId, hashId);
+                }
+
+                emailService.SendEmail(
+                    GenerateVerificationEmail(userAccount, hash, emailGroup.Key!, baseUrl)
+                );
+            }
```
simonbrent, line 65: I have a suggestion in another comment which should allow you to simplify this method to
```
            foreach (var emailGroup in unverifiedEmails.GroupBy(emailUpdate => emailUpdate.NewEmail))
            {
                var hash = Guid.NewGuid().ToString();
                var hashId = emailVerificationDataService.CreateEmailVerificationHash(hash, clockUtility.UtcNow);
                var emailAddress = emailGroup.Key!;

                UpdateEmailVerificationHashId(userAccount.Id, emailAddress, hashId);

                emailService.SendEmail(
                    GenerateVerificationEmail(userAccount, hash, emailAddress, baseUrl)
                );
            }
```

DigitalLearningSolutions.Web/Services/EmailVerificationService.cs
```diff
@@ -0,0 +1,115 @@
...
+                emailVerificationDataService.UpdateEmailVerificationHashIdForPrimaryEmail(userId, hashId);
+            }
+            else
+            {
+                emailVerificationDataService.UpdateEmailVerificationHashIdForCentreEmail(
+                    userId,
+                    centreId.Value,
+                    hashId
+                );
+            }
```
simonbrent, line 81: I have a suggestion in another comment which should allow you to simplify this method to:
```
emailVerificationDataService.UpdateEmailVerificationHashIdForPrimaryEmail(userId, emailAddress, hashId);
emailVerificationDataService.UpdateEmailVerificationHashIdForCentreEmails(userId, emailAddress, hashId);
```

Note: you shouldn't need to update centre emails in a loop - it is possible to update the verification hash ids for all user/email combinations at once

DigitalLearningSolutions.Data/DataServices/EmailVerificationDataService.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            return connection.QuerySingle<int>(
+                @"INSERT INTO EmailVerificationHashes (EmailVerificationHash, CreatedDate)
+                        OUTPUT INSERTED.ID
+                        VALUES (@hash, @created)",
+                new { hash, created }
+            );
+        }
+
+        public void UpdateEmailVerificationHashIdForPrimaryEmail(int userId, int hashId)
```
simonbrent, line 38: I recommend you change this to
```
        public void UpdateEmailVerificationHashIdForPrimaryEmail(int userId, string emailAddress, int hashId)
        {
            connection.Execute(
                @"UPDATE Users SET EmailVerificationHashID = @hashId WHERE ID = @userId AND PrimaryEmail = @emailAddress",
                new { hashId, userId, emailAddress }
            );
        }
```
to ensure that the `EmailVerificationHashID` is only set when the email address is what we expect it to be

DigitalLearningSolutions.Data/DataServices/EmailVerificationDataService.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            connection.Execute(
+                @"UPDATE Users SET EmailVerificationHashID = @hashId WHERE ID = @userId",
+                new { hashId, userId }
+            );
+        }
+
+        public void UpdateEmailVerificationHashIdForCentreEmail(int userId, int centreId, int hashId)
```
simonbrent, line 46: I recommend you change this to
```
        public void UpdateEmailVerificationHashIdForCentreEmails(int userId, string emailAddress, int hashId)
        {
            connection.Execute(
                @"UPDATE UserCentreDetails SET EmailVerificationHashID = @hashId WHERE UserID = @userId AND Email = @emailAddress",
                new { hashId, userId, emailAddress }
            );
        }
```
to ensure that the EmailVerificationHashID is only set when the email address is what we expect it to be, and to allow us to update all matching `UserCentreDetails` records at once
```

File: DigitalLearningSolutions.Web/Services/EmailVerificationService.cs (hunk 8 above)
Comment (simonbrent) [reply DigitalLearningSolutions.Web/Services/EmailVerificationService.cs:81]: As a side-effect of these changes, `PossibleEmailUpdate` no longer needs to have a `CentreId` property

File: DigitalLearningSolutions.Web/Services/RegistrationService.cs [hunk 11]
```diff
@@ -200,16 +203,33 @@ IGroupsService groupsService
 
             try
             {
+                var emailUpdate = new PossibleEmailUpdate
+                {
+                    OldEmail = userDataService.GetCentreEmail(userId, internalDelegateRegistrationModel.Centre),
+                    NewEmail = delegateRegistrationModel.CentreSpecificEmail,
+                    NewEmailIsVerified = false,
```
simonbrent, line 210: `NewEmailIsVerified` will be true if:
* `delegateRegistrationModel.CentreSpecificEmail` matches `userEntity.UserAccount.PrimaryEmail` and `userEntity.UserAccount.EmailVerified` is not null, 
* `delegateRegistrationModel.CentreSpecificEmail` matches any verified `UserCentreDetails` for the user

so you'll need to use `EmailVerificationService.AccountEmailIsVerifiedForUser` here

DigitalLearningSolutions.Web/Services/RegistrationService.cs
```diff
@@ -200,16 +203,33 @@ IGroupsService groupsService
 
             try
             {
+                var emailUpdate = new PossibleEmailUpdate
```
simonbrent, line 206: Can you please rename all the `var emailUpdate` to `var possibleEmailUpdate`, since out of context (i.e. where it is passed into another method) it can be a bit unclear what this thing is for

DigitalLearningSolutions.Web/Services/RegistrationService.cs
```diff
@@ -349,7 +378,19 @@ public void CreateCentreManagerForExistingUser(int userId, int centreId, string?
...
+            registrationDataService.RegisterAdmin(
+                new AdminAccountRegistrationModel(registrationModel, userId),
+                new PossibleEmailUpdate
+                {
+                    OldEmail = userDataService.GetCentreEmail(userId, centreId),
+                    NewEmail = centreSpecificEmail,
+                    NewEmailIsVerified = emailVerificationDataService.AccountEmailIsVerifiedForUser(
+                                             userId,
+                                             centreSpecificEmail
+                                         ),
```
simonbrent, line 390: This whitespace is wrong (if I save the file the closing bracket moves to inline with the N of `NewEmailIsVerified`

DigitalLearningSolutions.Web/Services/RegistrationService.cs
```diff
@@ -349,7 +378,19 @@ public void CreateCentreManagerForExistingUser(int userId, int centreId, string?
...
+                    CentreId = null,
```
simonbrent, line 391: I don't think that `CentreId` should be null here, but also given the other changes I am suggesting, we can remove the need for `PossibleEmailUpdate.CentreId`

DigitalLearningSolutions.Web/Services/RegistrationService.cs
```diff
@@ -398,7 +439,7 @@ public void PromoteDelegateToAdmin(AdminRoles adminRoles, int? categoryId, int u
                     true
                 );
 
-                registrationDataService.RegisterAdmin(adminRegistrationModel);
+                registrationDataService.RegisterAdmin(adminRegistrationModel, new PossibleEmailUpdate());
```
simonbrent, line 442: Might be better for `RegisterAdmin` (and subsequent functions) to take a `PossibleEmailUpdate? = null`, so that this isn't necessary

DigitalLearningSolutions.Web/Services/UserService.cs
```diff
@@ -393,25 +416,55 @@ bool changeMadeBySameUser
...
-                centreId,
-                centreEmail,
-                changeMadeBySameUser ? (DateTime?)null : clockUtility.UtcNow
-            );
+            if (isCentreEmailUpdated)
+            {
+                var emailVerified =
+                    changeMadeBySameUser &&
+                    !emailVerificationDataService
+                        .AccountEmailIsVerifiedForUser(
```
simonbrent, line 430: I think this is maybe _slightly_ easier to read if `emailVerificationDataService` and `AccountEmailIsVerifiedForUser` are on the same line

DigitalLearningSolutions.Web/Services/UserService.cs
```diff
@@ -393,25 +416,55 @@ bool changeMadeBySameUser
...
+        )
         {
+            var currentTime = clockUtility.UtcNow;
+
             foreach (var (centreId, email) in centreEmailsByCentreId)
             {
-                userDataService.SetCentreEmail(userId, centreId, email);
+                if (!string.Equals(
+                        email,
+                        userCentreDetails.Where(ucd => ucd.CentreId == centreId).Select(ucd => ucd.Email)
```
simonbrent, line 458: I think this can be `userCentreDetails.SingleOrDefault(ucd => ucd.CentreId == centreId)?.Email`

DigitalLearningSolutions.Data/DataServices/RegistrationDataService.cs
```diff
@@ -123,19 +138,29 @@ bool registerJourneyContainsTermsAndConditions
...
-                userId,
-                delegateRegistrationModel.Centre,
-                delegateRegistrationModel.CentreSpecificEmail,
-                null,
-                transaction
-            );
+            if (emailUpdate.IsEmailUpdating)
+            {
+                var emailVerified =
+                    string.IsNullOrWhiteSpace(delegateRegistrationModel.CentreSpecificEmail) ||
```
simonbrent, line 151: I'm a bit concerned that this is the only place we care about whether the email is null or whitespace. Shouldn't we do this always/never? (Never = `NewEmailIsVerified` should already consider whether the email is null or whitespace)

simonbrent (changes requested):

DigitalLearningSolutions.Web/Controllers/MyAccountController.cs
```diff
@@ -227,38 +303,64 @@ DlsSubApplication dlsSubApplication
...
-                delegateAccount?.Id
+        private IEnumerable<PossibleEmailUpdate> GetPossibleEmailUpdates(
+            UserEntity userEntity,
+            int? centreId,
+            AccountDetailsData accountDetailsData,
+            MyAccountEditDetailsFormData formData,
+            List<UserCentreDetails> userCentreDetails
+        )
+        {
+            var isNewPrimaryEmailVerified = emailVerificationService.AccountEmailIsVerifiedForUser(
```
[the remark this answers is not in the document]
VladNicol: I don't think so, `userEntity.UserAccount.EmailVerified` will say if the current primary email is verified. I want to check that the updated primary email is already verified - perhaps as a centre email.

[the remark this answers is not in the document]
simonbrent: Oh yeah, duh!

DigitalLearningSolutions.Web.Tests/Controllers/MyAccount/MyAccountControllerTests.cs
```diff
@@ -387,16 +410,89 @@ public void
...
                         A<string?>._,
                         A<int>._,
+                        true,
+                        false,
                         true
                     )
                 )
                 .MustHaveHappened();
 
+            A.CallTo(
```
simonbrent, line 420: I don't think you should test this here. You are testing implicit behaviour (i.e. the primary email happens to have changed), when the test name is `EditDetailsPostSave_for_admin_only_user_with_missing_delegate_answers_doesnt_fail_validation_or_update_delegate` - nothing about whether the email should be sent or not.

simonbrent (changes requested):

DigitalLearningSolutions.Data.Tests/DataServices/RegistrationDataServiceTests.cs
```diff
@@ -315,21 +450,28 @@ public void
...
+                }
             );
 
             // Then
             using (new AssertionScope())
             {
                 var userCentreDetails = connection.GetEmailAndVerifiedDateFromUserCentreDetails(userId, centreId);
                 userCentreDetails.email.Should().Be(newCentreEmail);
-                userCentreDetails.emailVerified.Should().BeNull();
+                userCentreDetails.emailVerified.Should().Be(null);
```
simonbrent, line 474: Isn't `BeNull()` is better than `Be(null)`?

DigitalLearningSolutions.Web.Tests/Controllers/MyAccount/MyAccountControllerTests.cs
```diff
@@ -396,19 +415,103 @@ public void
...
+                .DoesNothing();
+
+            // When
+            var result = await myAccountController.EditDetails(model, "save", DlsSubApplication.Default);
+
+            // Then
+            A.CallTo(
+                () => emailVerificationService.CreateEmailVerificationHashesAndSendVerificationEmails(
+                    testUserEntity.UserAccount,
+                    A<IEnumerable<PossibleEmailUpdate>>._,
```
simonbrent, line 478: You should specify what the `PossibleEmailUpdate`s argument is (matches) here

DigitalLearningSolutions.Web.Tests/Services/UserServiceTests.cs
```diff
@@ -157,23 +159,32 @@ public void GetDelegateUserById_returns_null_from_data_service()
...
-            UpdateUserDetailsAndCentreSpecificDetails_with_null_delegate_details_only_updates_user_and_centre_email()
+            UpdateUserDetailsAndCentreSpecificDetails_with_null_delegate_details_only_updates_user_and_centre_email(
+                bool isPrimaryEmailUpdated,
+                bool isCentreEmailUpdated
+            )
         {
             // Given
             const int centreId = 1;
             const string centreEmail = "test@email.com";
             const bool shouldUpdateProfileImage = true;
```
simonbrent, line 175: `shouldUpdateProfileImage` should be `changesMadeBySameUser`

DigitalLearningSolutions.Web.Tests/Services/UserServiceTests.cs
```diff
@@ -187,22 +198,42 @@ public void
...
-                        A<IDbTransaction?>._
+
+            if (isCentreEmailUpdated)
+            {
+                A.CallTo(
+                        () => userDataService.SetCentreEmail(
+                            accountDetailsData.UserId,
+                            centreId,
+                            centreEmail,
+                            null,
```
simonbrent, line 217: this is only null because `shouldUpdateProfileImage` is true. if it were false, it would be `currentTime`

DigitalLearningSolutions.Web.Tests/Services/EmailVerificationServiceTests.cs
```diff
@@ -0,0 +1,168 @@
...
+                    A<int>._
+                )
+            ).MustNotHaveHappened();
+
+            A.CallTo(() => emailService.SendEmail(A<Email>._)).MustNotHaveHappened();
+        }
+
+        [Test]
+        public void
+            CreateEmailVerificationHashesAndSendVerificationEmails_updates_hashId_and_sends_email_when_primary_email_requires_verification()
```
simonbrent, line 89: `CreateEmailVerificationHashesAndSendVerificationEmails` doesn't actually care whether you're updating a primary or centre email, so I don't think you should be referencing them in the test names, since it is not possible to distinguish between what emails are updating in the Given/When sections. What you should be testing is:
* `SendEmail` and both `UpdateEmailVerificationHashId` methods are called if `unverifiedEmails` is not empty
* Each `UpdateEmailVerificationHashId` is only called once for each distinct email address in `unverifiedEmails`
* `SendEmail` is only called once for each distinct email address in `unverifiedEmails`

simonbrent (changes requested):

DigitalLearningSolutions.Web/Services/RegistrationService.cs
simonbrent: `emailUpdate` is still present in:

- DigitalLearningSolutions.Data/DataServices/RegistrationDataService.cs
- DigitalLearningSolutions.Web/Services/EmailVerificationService.cs
- DigitalLearningSolutions.Web/Services/RegistrationService.cs

Can you rename it there too please?

DigitalLearningSolutions.Web.Tests/Services/EmailVerificationServiceTests.cs
```diff
@@ -101,68 +103,57 @@ public void
...
                 () => emailVerificationDataService.UpdateEmailVerificationHashIdForCentreEmails(
                     userAccount.Id,
-                    email,
+                    newEmail2,
                     hashId
                 )
             ).MustHaveHappenedOnceExactly();
 
-            A.CallTo(() => emailService.SendEmail(A<Email>._)).MustHaveHappenedOnceExactly();
+            A.CallTo(() => emailService.SendEmail(A<Email>._)).MustHaveHappenedTwiceExactly();
```
simonbrent, line 156: You should explicitly test that SendEmail is called once for newEmail1 and once for newEmail2

simonbrent (changes requested):

simonbrent (approved):
Great! Merge away

PR #1336, opened 2022-08-10
HEEDLS-975 Throw an exception when trying to claim a user with unexpe…

simonbrent:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-975

### Description
Throw an exception according to the spec:
> The user entity associated with the new delegate record must not have any other centre account records. If it does, return a server error (you can just use e.g. IEnumerable.Single()). This shouldn’t happen anyway, since the user entity is a (possibly) temporary entity created by the register-by-centre or bulk upload journey and can’t be logged into until claimed.

(Note: you can't just use IEnumerable.Single())

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)


VladNicol (approved):
Looks good

PR #1339, opened 2022-08-10
HEEDLS-1036 Only request 150 resources from Learning Hub API

simonbrent:
### JIRA link
https://softwiretech.atlassian.net/browse/HEEDLS-1036

### Description
* Migration to add `MaxSignpostedResources` = 150 to the Config database table.
* Limit the number of resource ids given to `learningHubResourceService.GetBulkResourcesByReferenceIds` to `Config.MaxSignpostedResources`
* Order resources returned from the API by descending `RecommendationScore`

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data.Migrations/202208101435_AddMaxSignpostedResourcesConfigValue.cs
+++ b/DigitalLearningSolutions.Data.Migrations/202208101435_AddMaxSignpostedResourcesConfigValue.cs
@@ -1,0 +1,22 @@
+﻿namespace DigitalLearningSolutions.Data.Migrations
+{
+    using FluentMigrator;
+
+    [Migration(202208101435)]
+    public class AddMaxSignpostedResourcesConfigValue : Migration
+    {
+        private const string MaxSignpostedResourcesValue = "150";
+
+        public override void Up()
+        {
+            Execute.Sql(
+                @$"IF NOT EXISTS (SELECT ConfigID FROM Config WHERE ConfigName = 'MaxSignpostedResources')
+                    BEGIN
+                      INSERT INTO Config VALUES ('MaxSignpostedResources', '{MaxSignpostedResourcesValue}', 0)
+                    END"
+            );
+        }
+
+        public override void Down() { }
+    }
+}
--- a/DigitalLearningSolutions.Data/Services/RecommendedLearningService.cs
+++ b/DigitalLearningSolutions.Data/Services/RecommendedLearningService.cs
@@ -1,5 +1,6 @@
 ﻿namespace DigitalLearningSolutions.Data.Services
 {
+    using System;
     using System.Collections.Generic;
     using System.Linq;
     using System.Threading.Tasks;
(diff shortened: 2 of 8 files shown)


DigitalLearningSolutions.Data/Services/RecommendedLearningService.cs
```diff
@@ -45,27 +49,41 @@ ILearningLogItemsDataService learningLogItemsDataService
...
+                clr => clr.LearningHubResourceReferenceId
+            ).Distinct();
 
             var resources =
-                await learningHubResourceService.GetBulkResourcesByReferenceIds(uniqueLearningHubReferenceIds);
+                await learningHubResourceService.GetBulkResourcesByReferenceIds(
+                    hasMaxSignpostedResources
+                        ? uniqueLearningHubReferenceIds.Take(maxSignpostedResources).ToList()
+                        : uniqueLearningHubReferenceIds.ToList()
+                );
```
VladNicol, line 86: Isn't this going to take 150 resources and then order them by score, instead of having the best 150 resources?

The problem was that the response from the method `GetBulkResourcesByReferenceIds` was too big when more than 150 resources were queried. I think we could make multiple queries for 150 resources each and then sort them all.

VladNicol (changes requested):
I have some suggestions, not sure if they make sense

DigitalLearningSolutions.Data.Migrations/202208101435_AddMaxSignpostedResourcesConfigValue.cs
```diff
@@ -0,0 +1,22 @@
...
+        {
+            Execute.Sql(
+                @$"IF NOT EXISTS (SELECT ConfigID FROM Config WHERE ConfigName = 'MaxSignpostedResources')
+                    BEGIN
+                      INSERT INTO Config VALUES ('MaxSignpostedResources', '{MaxSignpostedResourcesValue}', 0)
+                    END"
+            );
+        }
+
+        public override void Down() { }
```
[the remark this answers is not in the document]
simonbrent: Possibly/probably. In this case I was following the example set in https://github.com/TechnologyEnhancedLearning/DLSV2/blob/uar-test/DigitalLearningSolutions.Data.Migrations/202208031701_AddSupportEmailConfigValue.cs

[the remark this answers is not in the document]
VladNicol: yeah makes sense, keeping a config value when reverting migration can't do much harm

DigitalLearningSolutions.Data/Services/RecommendedLearningService.cs
simonbrent: No, the problem was that the _request to_ the learning hub API (as made in `GetBulkResourcesByReferenceIds`) was too big: the URL exceeded 2048 characters because of the format (`resourceReferenceIds=1&resourceReferenceIds=2&...`)

https://softwiretech.atlassian.net/browse/HEEDLS-1033 will resolve that, by changing the URL we use/having a different query string format, but the requirement of HEEDLS-1036 is still to limit to only 150 ids in the request. 

When imposing this limit, it is not possible request the "best" 150 resources - we only know recommendation score after we make the request. I agree that we _could_ make multiple queries to get all the data and then sort them all, but it seems unlikely that, in the situation where there are more than 150, the user will actually want to wade through even the first 150, let alone anything beyond that. (I seem to recall having a conversation with Steve where he was saying this, but can't find it in slack)

VladNicol: I agree that a user will only go through the first few resources, but that's exactly why this method sorts by score, so those first few resources are actually the best. If it's not too much effort, I think we should have multiple requests, sort all resources and optionally just return the top few instead of all of them.

VladNicol (approved):
Looks good to me

PR #1343, opened 2022-08-11
Heedls 1018 update user centre job groups

JBsoftwire:
### JIRA link
_[HEEDLS-1018](https://softwiretech.atlassian.net/browse/HEEDLS-1018)_

### Description
Added a method to get delegate accounts associated with a user besides the one being updated, and then update their job group group memberships depending on 

### Screenshots
N/A

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Web/Services/GroupsService.cs
+++ b/DigitalLearningSolutions.Web/Services/GroupsService.cs
@@ -295,6 +295,62 @@
             transaction.Complete();
         }
 
+        // TODO HEEDLS-1018 rename this
+        public void SynchroniseJobGroupsOnOtherCentres(
+            int delegateId,
+            int oldJobGroupId,
+            int newJobGroupId,
+            AccountDetailsData accountDetailsData,
+            string? centreEmail
+        )
+        {
+            if (oldJobGroupId == newJobGroupId)
+            {
+                return;
+            }
+
+            var userId = userDataService.GetUserIdFromDelegateId(delegateId);
+            var delegateAccounts = userDataService.GetDelegateAccountsByUserId(userId);
+            // possibly exclude existing account here
+
+            foreach (var account in delegateAccounts)
+            {
+                var groupsLinkedToJobGroup = GetGroupsLinkedToJobGroupForCentre(account.CentreId).ToList();
+
+                var oldJobGroupName = jobGroupsDataService.GetJobGroupName(oldJobGroupId);
+                var newJobGroupName = jobGroupsDataService.GetJobGroupName(newJobGroupId);
+
+                // TODO HEEDLS-1018 magic strings?
+
+                var groupToRemoveDelegateFrom = groupsLinkedToJobGroup.Single(
+                    g =>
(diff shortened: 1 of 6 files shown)


DigitalLearningSolutions.Web/Services/GroupsService.cs
```diff
@@ -275,26 +299,89 @@ List<Group> groupsForSynchronisation
...
+                    account.Id,
+                    accountDetailsData,
+                    centreEmail,
+                    groupToAddDelegateTo,
+                    account.CentreId
+                );
+            }
+        }
+
+        private void AddDelegateToGroupAndEnrolOnGroupCourses(
```
simonbrent, line 361: I find it a bit confusing to have a private and public function with the same name, but with slightly different functionality: 
* calling `AddDelegateToGroup` with different `addedByFieldLink` values
* one having a TransactionScope, the other not

I think ideally, one of these functions would call the other, and they would be one-after-the-other in the file

DigitalLearningSolutions.Web/Services/GroupsService.cs
```diff
@@ -275,26 +299,89 @@ List<Group> groupsForSynchronisation
...
+                .Where(da => da.Id != originalDelegateId);
+
+            foreach (var account in delegateAccounts)
+            {
+                var groupsLinkedToJobGroup = GetGroupsLinkedToJobGroupForCentre(account.CentreId).ToList();
+
+                var oldJobGroupName = jobGroupsDataService.GetJobGroupName(oldJobGroupId);
+                var newJobGroupName = jobGroupsDataService.GetJobGroupName(newJobGroupId);
+
+                var groupToRemoveDelegateFrom = groupsLinkedToJobGroup.Single(
```
simonbrent, line 339: As discussed, this should be at least `SingleOrDefault`, if not `Where`

DigitalLearningSolutions.Web/Services/GroupsService.cs
```diff
@@ -275,26 +299,89 @@ List<Group> groupsForSynchronisation
...
+                    g =>
+                        GroupLabelMatchesAnswer(g.GroupLabel, oldJobGroupName, JobGroupLinkedFieldName)
+                );
+
+                var groupToAddDelegateTo = groupsLinkedToJobGroup.Single(
```
simonbrent, line 344: Same as above

simonbrent (changes requested):

DigitalLearningSolutions.Web/Services/GroupsService.cs
SteveJacksonSoft: I think it's okay to call the public method `AddDelegateToGroup` because that will always involve enrolling them on courses, and it's natural to assume that GroupsService will deal with the details of what happens when you add a delegate to a group - so we probably don't need to specify at the interface level all the details of what happens.

SteveJacksonSoft (commented):
Looks nice! Just that method name change that I'd recommend.

SteveJacksonSoft (approved):
Nice

PR #1347, opened 2022-08-16
DLSV2-569 Remove Supervisor fails and results in an error

daniel-manta-hee:
Manage Supervisor/Remove Supervisor, fails to remove the supervisor and it results in error.

### JIRA link
https://hee-dls.atlassian.net/browse/DLSV2-569

### Description
**Introduction:** Since I cannot connect VS to UAT database I am guessing what the problem is based on database table relations and the code available.  There is a soft-delete on `SupervisorDelegates` and hard delete on `CandidateAssessmentSupervisor`. The later may be causing an error if the table is referenced from another table.

**Suggested solution:** Soft-delete `CandidateAssessmentSupervisors` if there are no records pointing to that table from `CandidateAssessmentSupervisorVerifications` nor `SelfAssessmentResultSupervisorVerifications`. Otherwise, a hard delete can be performed.

Any matching record in `SelfAssessmentResultSupervisorVerifications` can be hard deleted which has the same effect as pressing "Withdraw" for that request from the user interface.

No action was taken for matching `CandidateAssessmentSupervisorVerfications` at this point. There are queries like `GetSelfAssessmentResultSummary` at SupervisorService.cs line 755, that use these records to join other tables, e.g. `SelfAssessments AS sa INNER JOIN CandidateAssessmentSupervisorVerifications AS casv INNER JOIN CandidateAssessmentSupervisors`

### Implementation

Despite `CandidateAssessments` and `SelfAssessmentSupervisorRoles` (green tables) are referenced from records that are going to be deleted in `CandidateAssessmentSupervisors`, it must be noted that the relation goes in the oposite direction. What we want to delete (`CandidateAssessmentSupervisor`) is not pointed by these tables, hence it can be deleted. It seems blue tables are what we need to care about.

#### Relevant tables:
- In **Red**, tables containing records that need to be deleted.
- In **Blue**: tables pointing **to** red tables.
- In **Green**: tables pointed **by** red tables.

image

### Steps to reproduce the bug

**Add capability confirmation requests for a supervisor**

Untitled1

**Attempt to remove the supervisor**

Untitled2

**In Visual Studio this results in the following error:**

> Microsoft.Data.SqlClient.SqlException: 'The DELETE statement conflicted with the REFERENCE constraint "FK_SelfAssessmentResultSupervisorVerifications_CandidateAssessmentSupervisorID_CandidateAssessmentSupervisors_ID". The conflict occurred in database "mbdbx101", table "dbo.SelfAssessmentResultSupervisorVerifications", column 'CandidateAssessmentSupervisorID'.

Stacktrace points to RemoveCandidateAssessmentSupervisor method in SupervisorService.cs

**From user interface this can be seen:**

image

-----
### Developer checks
Checked that:

- Neither hard or soft delete are resulting in an error in the cases I've tested.
- There are no more tables somewhere pointing to `CandidateAssessmentSupervisors`: run `sp_help 'CandidateAssessmentSupervisors'` as suggested [in this post](https://stackoverflow.com/a/23993240/2516718).
- Soft-deleted records are filtered out in queries: tried to find some queries that need to be updated but it seems they are already doing that filter or filtering the result by not removed supervisor delegate or they are export functions. There are tons of queries using these tables so it's quite possible I could be missing something.

### DIFF
--- a/DigitalLearningSolutions.Data/Services/SupervisorService.cs
+++ b/DigitalLearningSolutions.Data/Services/SupervisorService.cs
@@ -45,7 +45,7 @@
         int InsertCandidateAssessmentSupervisor(int delegateId, int supervisorDelegateId, int selfAssessmentId, int? selfAssessmentSupervisorRoleId);
         bool InsertSelfAssessmentResultSupervisorVerification(int candidateAssessmentSupervisorId, int resultId);
         //DELETE DATA
-        bool RemoveCandidateAssessmentSupervisor(int selfAssessmentId, int supervisorDelegateId);
+        bool RemoveCandidateAssessmentSupervisor(int supervisorDelegateId);
     }
     public class SupervisorService : ISupervisorService
     {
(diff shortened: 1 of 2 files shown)


DigitalLearningSolutions.Data/Services/SupervisorService.cs
```diff
@@ -567,25 +567,46 @@ public bool RemoveCandidateAssessment(int candidateAssessmentId)
...
-         @"DELETE FROM cas
-FROM  CandidateAssessmentSupervisors AS cas INNER JOIN
-         CandidateAssessments AS ca ON cas.CandidateAssessmentID = ca.ID
-WHERE (ca.SelfAssessmentID = @selfAssessmentId) AND (cas.SupervisorDelegateId = @supervisorDelegateId)",
-        new { selfAssessmentId, supervisorDelegateId });
-            if (numberOfAffectedRows < 1)
+        public bool RemoveCandidateAssessmentSupervisor(int supervisorDelegateId)
+        {
+            var numberOfRemovedVerifications = connection.Execute(
+                @"DELETE FROM sarsv
```
kevwhitt-hee, line 573: This is hard deleting SelfAssessmentResultSupervisorVerifications - we should never do that. The implementation should be checking if they exist (or if any related CandidateAssessmentSupervisorVerifications exist) and soft deleting the CandidateAssessmentSupervisors record if they do.

We then need to check queries called from the LearningPortal/SelfAssessments controllers and the Supervisor controllers to make sure they exclude soft deleted CandidateAssessmentSupervisors records.

kevwhitt-hee (changes requested):

DigitalLearningSolutions.Data/Services/SupervisorService.cs
```diff
@@ -47,7 +47,7 @@ public interface ISupervisorService
         int InsertCandidateAssessmentSupervisor(int delegateId, int supervisorDelegateId, int selfAssessmentId, int? selfAssessmentSupervisorRoleId);
         bool InsertSelfAssessmentResultSupervisorVerification(int candidateAssessmentSupervisorId, int resultId);
         //DELETE DATA
-        bool RemoveCandidateAssessmentSupervisor(int selfAssessmentId, int supervisorDelegateId);
```
kevwhitt-hee, line 50: Removing the selfAssessmentId will result in the supervisor being removed from all self assessments that they are supervising for this delegate rather than just the one the delegate has removed them from.

An ID PK column has been added to CandidateAssessments table since this work was originally done so we could pass just a candidateAssessmentId parameter safely but we can't use only selfAssessmentId.

kevwhitt-hee (changes requested):
Change selfAssessmentId parameter to candidateAssessmentId parameter - see comment.

kevwhitt-hee (approved):

PR #1368, opened 2022-08-25
DLSV2-535 Use competency name instead of description in confirmation requests view

daniel-manta-hee:
Use competency name instead of competency description in Confirmation Requests view. Description could be blank in the database.

### JIRA link
https://hee-dls.atlassian.net/browse/DLSV2-535

### Description
Edited `ReviewConfirmationRequests.cshtml` for using competency `Name` property instead of `Description`.

### Screenshots
image

-----
### Developer checks
Checked that there are no blank or null competency names in UAT database table Competencies: `select * from Competencies where [Name] is null or trim([Name]) = ''`

### DIFF
--- a/DigitalLearningSolutions.Web/Views/LearningPortal/SelfAssessments/ReviewConfirmationRequests.cshtml
+++ b/DigitalLearningSolutions.Web/Views/LearningPortal/SelfAssessments/ReviewConfirmationRequests.cshtml
@@ -59,7 +59,7 @@
       <tr role="row" class="nhsuk-table__row">
         <td role="cell" class="nhsuk-table__cell nhsuk-u-font-size-16">
           <span class="nhsuk-table-responsive__heading">Comptenecy </span>
-          @DisplayStringHelper.RemoveMarkup(competency.Description)
+          @competency.Name
         </td>
         <td role="cell" class="nhsuk-table__cell nhsuk-u-font-size-16">
           <span class="nhsuk-table-responsive__heading">Question </span>


kevwhitt-hee (approved):

DigitalLearningSolutions.Web/Views/LearningPortal/SelfAssessments/ReviewConfirmationRequests.cshtml
```diff
@@ -59,7 +59,7 @@
       <tr role="row" class="nhsuk-table__row">
         <td role="cell" class="nhsuk-table__cell nhsuk-u-font-size-16">
           <span class="nhsuk-table-responsive__heading">Comptenecy </span>
-          @DisplayStringHelper.RemoveMarkup(competency.Description)
```
kevwhitt-hee, line 62: Is this string helper used elsewhere in the solution? If not, we should remove it.

daniel-manta-hee: Not at the moment and it can be deleted. But there are two places where this helper could be useful in the future. 1) When validating competency `Description` is empty (_OverviewTable.cshtml:53) as Description allows Html entities like `&npsp;` and tags without content. 2) As a replacement for `SignpostingHelper.DisplayText()` which also removes markup and spaces. For (1) this code would handle null and empty `<p class="@(DisplayStringHelper.RemoveMarkup(competency.Description).Trim() != "" ? "nhsuk-u-font-weight-bold" : "") nhsuk-u-margin-bottom-0">` in the view and `@using DigitalLearningSolutions.Web.Helpers` is needed.

kevwhitt-hee: Thanks Daniel. That makes sense, I will add a tech debt ticket to replace and remove SignpostingHelper.DisplayText() with DisplayStringHelper.RemoveMarkup, then.

PR #1403, opened 2022-09-22
Develop/fixes/dlsv2 641 new framework summary page browser back404

martinfoster-hee:
### JIRA link
https://hee-dls.atlassian.net/jira/software/projects/DLSV2/boards/1?assignee=6321797ece3e476e42ac8a00&selectedIssue=DLSV2-641

### Description
There was an issue when saving a new framework that if the user hit the browser back button to return to the previous /Frameworks/Categorise/New page, a 404 error page was displayed.

This is because the user is trying to refresh the page in 'create new' mode, but the framework data has already been saved.

The error page has been updated to display a 410 error, which means the page is not longer valid (see attached screenshot).

The behaviour can also be observed after saving the new framework by browsing back to the summary page and clicking on any of the four links ('Change Name', 'Change Description', 'Change Vocabulary', 'Change Branding').

In the frameworks controller the 410 errors are now automatically raised via a service filter (RedirectToErrorEmptySessionData).

### Screenshots
See attached screenshot of new 410 page (in the context of invalid /Frameworks/Catagorise/New page state).

-----
### Developer checks
I have:
  - [doc_1_here](link_1_here)
- [x] Scanned over my own MR to ensure everything is as expected and it looks right in the browser

bug641_new410errorpage


DigitalLearningSolutions.Web/Controllers/FrameworksController/Frameworks.cs
```diff
@@ -157,7 +157,7 @@ public IActionResult CreateNewFramework(string actionname, int frameworkId = 0)
             {
                 if (TempData[MultiPageFormDataFeature.AddNewFramework.TempDataKey] == null)
```
kevwhitt-hee, line 158: Could we replace this repeated redirect code with a ServiceFilter implementation similar to that used in `DigitalLearningSolutions.Web\Controllers\Register\RegisterController.cs` before all GET methods in the multi page form, we have `[ServiceFilter(typeof(RedirectEmptySessionData<DelegateRegistrationData>))]`?

A variation of this called Redirect**ToError**EmptySessionData would allow us to decorate get methods with this service filter rather than repeat code. 

It occurred to me that we could use the existing service filter but to my mind, we don't want to redirect users to the start of the form in most cases and we don't always use the pattern of an `index.cshtml` view to start all of these flows (though perhaps we should), which that filter method is currently reliant on.

kevwhitt-hee (changes requested):
See comment in ticket. We need to handle this scenario throughout the app. Some implementations of the MultiPageFormService use a ServiceFilter method decorator to call RedirectEmptySessionData which redirects the user to the start of the flow if tempdata doesn't exist (see code comment). This is a cleaner approach and we should adapt it to implement the 410 error redirect where the RedirectEmptySessionData isn't already used.

DigitalLearningSolutions.Web/Controllers/FrameworksController/Frameworks.cs
kevwhitt-hee: I notice there is another ServiceFilter that might be of interest to look at: `DigitalLearningSolutions.Web\ServiceFilter\RedirectMissingMultiPageFormData.cs`

martinfoster-hee: Yes that's a good call re picking it up everywhere automatically in the GET method pre-processing. Will implement.

martinfoster-hee: @kevwhitt-hee I've implemented this as a new service filter throughout the frameworks controller (I couldn't find anywhere else where the code usually raises a 404(/410) if TempData is missing).

DigitalLearningSolutions.Web/ServiceFilter/RedirectToErrorEmptySessionData.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            _feature = feature;
+            _multiPageFormService = multiPageFormService;
+        }
+
+        public void OnActionExecuted(ActionExecutedContext context) { }
+
+        public void OnActionExecuting(ActionExecutingContext context)
+        {
+            if (!(context.Controller is Controller controller) || _feature.Name != "AddNewFramework")
```
kevwhitt-hee, line 32: Will this hardcoded reference to "AddNewFramework" prevent this filter from being used in other multi-page forms?

kevwhitt-hee (commented):
See comment - can this filter be genericised to allow it to be used elsewhere where multi-page forms are implemented?

Places in the application that use multi-page forms where this filter could be useful are:

Web\Controllers\FrameworksController\Frameworks.cs → 

StartNewFrameworkSession (Frameworks / Create new framework)

Web\Controllers\FrameworksController\AssessmentQuestions.cs → 

StartAssessmentQuestionSession (Frameworks / Structure / Manage Assessment Questions / Add new question)

Web\Controllers\FrameworksController\Signposting.cs → 

StartSignpostingParametersSession (Frameworks / Structure / Manage competency signposting / Search / Link resource to competency)

Web\Controllers\LearningPortalController\SelfAssessment.cs →

 SearchInSelfAssessmentOverviewGroups (Learning Portal / Current / Launch Self Assessment / View capabilities / Search)

StartRequestVerification (Learning Portal / Current / Launch Self Assessment / Request confirmation)

Web\Controllers\SupervisorController\Supervisor.cs → 

StartEnrolDelegateOnProfileAssessment (Supervisor / My Staff List / View self assessments / Enrol)

We could pick these up in a separate ticket if you prefer, though?

DigitalLearningSolutions.Web/ServiceFilter/RedirectToErrorEmptySessionData.cs
martinfoster-hee: Hmm, yes you're right - it should be passed in as a parameter. Will fix.

kevwhitt-hee (changes requested):
This is working pretty well, now when I test it in the browser and the changes look good.

I did notice when I hit the back button repeatedly through a flow, I eventually hit the exception on line 73 of DigitalLearningSolutions.Data\Services\MultiPageFormService.cs:

`            if (tempData[feature.TempDataKey] == null)
            {
                throw new MultiPageFormDataException("Attempted to get data with no Guid identifier");
            }`

This makes me think that we should throw a "Gone" error here rather than a "Something went wrong".

I'm guessing that we are making the assumption that if one of the controllers using the service filter works, then the others will and that it isn't worth testing them - fair enough.

kevwhitt-hee (approved):

PR #1447, opened 2022-11-01
TD-546 manage supervisor same supervisor gets added multiple times

OluwatobiAwe:
### JIRA link
[TD-546](https://hee-dls.atlassian.net/browse/TD-546)

### Description
Same Supervisor gets added multiple times when submit button continuously clicked

### Screenshots
Kindly refer to the jira link for a video

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)
- [x] Scanned over my own MR to ensure everything is as expected and it looks right in the browser

### DIFF
--- a/DigitalLearningSolutions.Web/Controllers/LearningPortalController/SelfAssessment.cs
+++ b/DigitalLearningSolutions.Web/Controllers/LearningPortalController/SelfAssessment.cs
@@ -4,6 +4,7 @@
     using System.Collections.Generic;
     using System.Linq;
     using System.Net;
+    using System.Threading.Tasks;
     using DigitalLearningSolutions.Data.Enums;
     using DigitalLearningSolutions.Data.Models.SearchSortFilterPaginate;
     using DigitalLearningSolutions.Data.Models.SelfAssessments;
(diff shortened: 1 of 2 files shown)


DigitalLearningSolutions.Web/Controllers/LearningPortalController/SelfAssessment.cs
```diff
@@ -16,6 +17,9 @@
     using DigitalLearningSolutions.Web.ViewModels.Common.SearchablePage;
     using DigitalLearningSolutions.Web.ViewModels.LearningPortal.Current;
     using DigitalLearningSolutions.Web.ViewModels.LearningPortal.SelfAssessments;
+    using DocumentFormat.OpenXml.EMMA;
```
kevwhitt-hee, line 20: What are these `DocumentFormat.OpenXml` libraries being used for?

kevwhitt-hee (commented):
We seem to be using two DocumentFormat.OpenXml libraries and I can't see why.

Is the problem here likely to exist elsewhere in the application? Should we add a ticket to use this async method with request verification token pattern elsewhere, too?

DigitalLearningSolutions.Web/Controllers/LearningPortalController/SelfAssessment.cs
OluwatobiAwe: DocumentFormat.OpenXml must have been added while updating the controller method to be asynchronous.
Iv removed the using statements since they were not being used.
I've also moved the verification token check to an extension method.

kevwhitt-hee: Thanks @OluwatobiAwe, looks much better. Before I merge this, can I get your opinion on whether we should be using this post pattern elsewhere/throughout the system to prevent repeat posting?

OluwatobiAwe: **This would perfectly work anywhere**, However, I'll prefer we use an action attribute where an actual form is being posted, where a model error would simply be appended. I decided to use an extension in this case, because we didn't have a well-defined model in the Post request.

PR #1460, opened 2022-11-08
Removed adminDelegateUser account dependency (work in progress).

martinfoster-hee:
### JIRA link
_DLSV2-XXXX_

### Description
_Describe what has changed and how that will affect the app. If relevant, add references to the resources you used. Use this as your opportunity to highlight anything odd and give people context around particular decisions._

### Screenshots
_Attach screenshots on mobile, tablet and desktop._

-----
### Developer checks

I have:
- [ ] Run the formatter and made sure there are no IDE errors (see [info on Text Editor settings](https://hee-tis.atlassian.net/wiki/spaces/TP/pages/3546185813/DLS+Dev+Process) to avoid whitespace changes)
  - [doc_1_here](link_1_here)
- [ ] Scanned over my own MR to ensure everything is as expected and it looks right in the browser

### DIFF
--- a/DigitalLearningSolutions.Data/Services/RegistrationService.cs
+++ b/DigitalLearningSolutions.Data/Services/RegistrationService.cs
@@ -36,8 +36,11 @@ namespace DigitalLearningSolutions.Data.Services
             AdminRoles adminRoles,
             int categoryId,
             int delegateId,
-            AdminUser? supervisorAdminUser,
-            DelegateUser supervisorDelegateUser);
+            AdminUser supervisorAdminUser
+            );
+
+        //AdminUser? supervisorAdminUser,
+        //DelegateUser supervisorDelegateUser);
     }
 
     public class RegistrationService : IRegistrationService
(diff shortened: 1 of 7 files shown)


DigitalLearningSolutions.Data/Services/RegistrationService.cs
```diff
@@ -320,7 +321,10 @@ private void SendEmailNotification(AdminRegistrationModel adminRegistrationModel
             builder.TextBody += "You will be able to access the Digital Learning Solutions platform with these new access permissions the next time you login.";
             builder.HtmlBody += "You will be able to access the Digital Learning Solutions platform with these new access permissions the next time you login.</body>";
 
-            supervisorService.UpdateNotificationSent(adminRegistrationModel.SupervisorDelegateId);
+            if (adminRegistrationModel.DelegateIdToPromote != 0)
+            {
+                supervisorService.UpdateNotificationSent(adminRegistrationModel.DelegateIdToPromote);
```
kevwhitt-hee, line 326: This still doesn't look right. We are still passing a delegatId into  a method that is expecting a supervisorDelegateId (an entirely different entity type id)

kevwhitt-hee (changes requested):
I think we need to move the call to the `supervisorService.UpdateNotificationSent` method from the RegistrationService to the Supervisor controller (Controllers\SupervisorController\Supervisor.cs).
In line 953, we currently call `registrationService.PromoteDelegateToAdmin` we should call `supervisorService.UpdateNotificationSent` immediately after this and pass in the supervisorDelegateId which is available in the controller as `supervisorDelegate.Id`.

DigitalLearningSolutions.Data/Services/RegistrationService.cs
kevwhitt-hee: If we do this, we can probably revert all of the attempts we've been making to retrieve a valid delegateId to pass into the method in the registration service.

martinfoster-hee: @kevwhitt-hee I've moved the UpdateNotificationSent call to the Supervisor controller and also consolidated the ids as much as possible.

kevwhitt-hee (approved):
This looks good. I have run the code and confirmed that the issue reported by end users is no longer in evidence.

PR #1609, opened 2023-02-07
TD-926 GUID values are displayed for Delegates in Self-Assessment -> …

manishagarwalnhs:
### JIRA link
https://hee-tis.atlassian.net/browse/TD-926

### Description
GUID values are displayed for Delegates in Self-Assessment -> Add Supervisor page , Delegates -> Export Excel sheet data, Delegate group and Course Delegate View

### Screenshots
No Screenshots. The screens do not display GUID anymore and the excel generated does not have any GUID.

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data/DataServices/CourseDataService.cs
+++ b/DigitalLearningSolutions.Data/DataServices/CourseDataService.cs
@@ -593,8 +593,7 @@ namespace DigitalLearningSolutions.Data.DataServices
                                             cap.Active = 1)))
                         AND da.CentreID = @centreId
                         AND pr.CustomisationID = @customisationId
-                        AND ap.DefaultContentTypeID <> 4
-                        AND TRY_CAST(COALESCE(ucd.Email, u.PrimaryEmail) AS UNIQUEIDENTIFIER) IS NULL",
+                        AND ap.DefaultContentTypeID <> 4",
                 new { customisationId, centreId }
             );
         }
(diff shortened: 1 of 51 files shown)


DigitalLearningSolutions.Data/DataServices/CourseDataService.cs
```diff
@@ -593,7 +593,8 @@ public IEnumerable<DelegateCourseInfo> GetDelegateCourseInfosForCourse(int custo
                                             cap.Active = 1)))
                         AND da.CentreID = @centreId
                         AND pr.CustomisationID = @customisationId
-                        AND ap.DefaultContentTypeID <> 4",
+                        AND ap.DefaultContentTypeID <> 4
+                        AND TRY_CAST(COALESCE(ucd.Email, u.PrimaryEmail) AS UNIQUEIDENTIFIER) IS NULL",
```
kevwhitt-hee, line 597: I think it would be easier to achieve this using LINQ on the result set returned rather than in the SQL query? This would avoid the need to introduce an additional excludeGuid parameter for the service method too.

kevwhitt-hee (commented):
See comment. I think we have examples of where we are excluding GUIDs using LINQ elsewhere in the solution, already. Would this approach not be simpler?

DigitalLearningSolutions.Data/DataServices/CourseDataService.cs
martinfoster-hee: @kevwhitt-hee I wonder if we should have some sort of sql variables file for magic numbers like the 4 here also? Just to make it more obvious exactly what the meaning is here?

kevwhitt-hee (approved):

DigitalLearningSolutions.Data/DataServices/CourseDataService.cs
kevwhitt-hee: This is a good idea. Probably not worth holding this ticket up for, though, since we didn't introduce it here. I'll add a maintenance/tech debt ticket to tidy these up.

PR #1663, opened 2023-02-27
TD-837-UAR: Console '500' error when clicked browser back buttons while enrolling a staff member

Auldrin-Possa:
**JIRA link**
https://hee-tis.atlassian.net/browse/TD-837

**Description**
TempData.Clear() statement removed to persist information in browser back action.

Screenshots
N/A

-----
Developer checks

I have:
- [ ] Run the formatter and made sure there are no IDE errors (see [info on Text Editor settings](https://hee-tis.atlassian.net/wiki/spaces/TP/pages/3546185813/DLS+Dev+Process) to avoid whitespace changes)
- [x] Tested any Views or partials created or changed with [Wave Chrome plugin]
- [ ] Updated/added documentation in [Confluence]


DigitalLearningSolutions.Web/Controllers/SupervisorController/Supervisor.cs
```diff
@@ -828,7 +828,7 @@ public IActionResult EnrolDelegateConfirm(int delegateUserId, int supervisorDele
                     GetCentreId()
                 );
             }
-            TempData.Clear();
+
```
kevwhitt-hee, line 831: Not sure we should be retaining the tempdata at the end of this flow in order to solve this problem. I think we should be checking whether it exists on the in the controller method for the page we are going back to and throw a 410 expired error if so. We use an action filter elsewhere to achieve this.

kevwhitt-hee (changes requested):
See code comments.

kevwhitt-hee:
@Auldrin-Possa , tagging you in case you missed the above.

PR #1863, opened 2023-05-02
TD-1467 Subscribe to notification when delegate is registered or prom…

manishagarwalnhs:
…oted to admin

### JIRA link
https://hee-tis.atlassian.net/browse/TD-1467

### Description
1) Subscribe to notifications when delegate is registered internally or externally.
2) Subscribe to notifications when delegate is promoted to admin

### Screenshots
This is backend change so no screenshot available.

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data/DataServices/NotificationDataService.cs
+++ b/DigitalLearningSolutions.Data/DataServices/NotificationDataService.cs
@@ -16,6 +16,8 @@
         IEnumerable<NotificationRecipient> GetAdminRecipientsForCentreNotification(int centreId, int notificationId);
 
         IEnumerable<int> GetRoleBasedNotifications(int isCentreManager, int isContentManager, int isContentCreator);
+
+        void SubscribeDefaultNotifications(int? candidateId, int? adminUserId);
     }
 
     public class NotificationDataService : INotificationDataService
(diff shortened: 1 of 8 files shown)


DigitalLearningSolutions.Data/DataServices/NotificationDataService.cs
```diff
@@ -105,5 +107,24 @@ public IEnumerable<int> GetRoleBasedNotifications(int isCentreManager, int isCon
...
+                connection.Execute(
+                @"INSERT INTO NotificationUsers (
+                    NotificationID,CandidateID)
+                    (SELECT NotificationID,@candidateId FROM NotificationRoles WHERE RoleID = 5)",new {candidateId});
+            }
+
+            if (adminUserId != null)
+            {
+                connection.Execute(
+                @"INSERT INTO NotificationUsers (
```
kevwhitt-hee, line 123: The logic removed at line 237 of RegistrationDataService.cs is better than this, we should revert this code to continue to use that logic. I had assumed that the subscription to notifications had been missed entirely but I see now that it was only present for admin registrations.

DigitalLearningSolutions.Data/DataServices/NotificationDataService.cs
```diff
@@ -105,5 +107,24 @@ public IEnumerable<int> GetRoleBasedNotifications(int isCentreManager, int isCon
                 ).AsEnumerable();
         }
 
+        public void SubscribeDefaultNotifications(int? candidateId, int? adminUserId)
+        {
+            if (candidateId != null)
+            {
+                connection.Execute(
+                @"INSERT INTO NotificationUsers (
+                    NotificationID,CandidateID)
```
kevwhitt-hee, line 116: We should join to the Notifications table here and honour the AutoOptIn flag (only insert a record if it is True) as per the existing logic for admin notification subscriptions.

kevwhitt-hee (changes requested):
The previous logic for admin notifications is better than the new logic - we should keep it.
We should also honour the AutoOptIn flag for delegate notifications - see code comments.

DigitalLearningSolutions.Data/DataServices/NotificationDataService.cs
manishagarwalnhs: This is addressed.

DigitalLearningSolutions.Data/DataServices/NotificationDataService.cs
manishagarwalnhs: This is addressed.

kevwhitt-hee (approved):

PR #1867, opened 2023-05-03
Develop/feature/td 1455 issues with session invalidation

ColinBeebyHEE:
### JIRA link
[TD-1455](https://hee-tis.atlassian.net/browse/TD-1455)

### Description
Reverted all the changes relating to session invalidation, which were introduced for TD-269. The changes were logging out users after 15 mins, even if they had checked the Remember Me checkbox.

### Screenshots
_Attach screenshots on mobile, tablet and desktop._

-----
### Developer checks

I have:
- [X] Run the formatter and made sure there are no IDE errors (see [info on Text Editor settings](https://hee-tis.atlassian.net/wiki/spaces/TP/pages/3546185813/DLS+Dev+Process) to avoid whitespace changes)
- [X] Manually tested my work with and without JavaScript
  - [doc_1_here](link_1_here)
- [X] Updated my Jira ticket with information about other parts of the system that were touched as part of the MR and have to be sanity tested to ensure nothing’s broken
- [X] Scanned over my pull request in GitHub and addressed any warnings from the GitHub Build and Test checks.

[TD-1455]: https://hee-tis.atlassian.net/browse/TD-1455?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kevwhitt-hee:
@rshrirohit, @ColinBeebyHEE, I'm loath to revert these changes since they will reappear in a future pen test. Should we remove the "Remember me" checkbox from login to avoid confusion caused by insecure previous implementation of "remember me"?

rshrirohit:
> @rshrirohit, @ColinBeebyHEE, I'm loath to revert these changes since they will reappear in a future pen test. Should we remove the "Remember me" checkbox from login to avoid confusion caused by insecure previous implementation of "remember me"?

@kevwhitt-hee it definitely appears to be the way to go. Leaving a session open for a long time is a big time risk. We can add the 'remember me' feature later on when it's more matured and honoured and doesn't fiddle around with the session timeouts in the wrong way.

ColinBeebyHEE:
@rshrirohit, @kevwhitt-hee, OK, so once a user has logged in:

- If the user is actively using the site in their browser then we can set a sliding expiration, so that the session will not get logged out - this seems reasonable. 
- 
- If the user leaves the site idle in their browser, how long should we leave it before we log them out. Currently this is 15 mins on most pages and infinitely long on 'courses' pages where the heartbeat keeps the session alive - i'm guessing that we would probably want to increase this idle timeout - question is, what to? Users will probably what it to be really long, pen testers will probably want this really short. 
- 
- If the user closes the browser (or tab), if we remove the Remember Me functionality, then the session will finish and the user will need to log in again when they return to the site. This is a change from what users are used to and I suspect will trigger support calls.

OK, so, if we keep this idle timeout functionality, once a user is logged in and leaves the site idle, how long do we leave it before we log them out? I would suggest that 24 hours is a reasonable compromise (but that is really a guess on my part)?

kevwhitt-hee:
> @rshrirohit, @kevwhitt-hee, OK, so once a user has logged in:
> 
> * If the user is actively using the site in their browser then we can set a sliding expiration, so that the session will not get logged out - this seems reasonable.
> * If the user leaves the site idle in their browser, how long should we leave it before we log them out. Currently this is 15 mins on most pages and infinitely long on 'courses' pages where the heartbeat keeps the session alive - i'm guessing that we would probably want to increase this idle timeout - question is, what to? Users will probably what it to be really long, pen testers will probably want this really short.
> * If the user closes the browser (or tab), if we remove the Remember Me functionality, then the session will finish and the user will need to log in again when they return to the site. This is a change from what users are used to and I suspect will trigger support calls.
> 
> OK, so, if we keep this idle timeout functionality, once a user is logged in and leaves the site idle, how long do we leave it before we log them out? I would suggest that 24 hours is a reasonable compromise (but that is really a guess on my part)?

@ColinBeebyHEE, since the pen test highlighted problems with the standard timeout period and moving this to a sliding 15 minutes addresses this. 

The pen test didn't (as far as I'm aware) highlight issues with the "remember me" facility specifically, so I think your compromise of 24 hours sounds sensible. I wonder whether it is worth changing "Remember me" to "Remember me on this computer for 24 hours" to be more specific too. If a future pen test highlights this as an issue, we can address it then. Sound okay?

kevwhitt-hee:
@ColinBeebyHEE. please could you update the ticket with the above requirements and close this PR?

ColinBeebyHEE:
The ticket has now been updated, so this PR can be closed.

PR #1932, opened 2023-06-01
TD-1392 Fixed mobile view issue on cookie policy page

manishagarwalnhs:
### JIRA link
https://hee-tis.atlassian.net/browse/TD-1392

### Description
1) Fixed mobile view issue on cookie policy page.

### Screenshots
Digital-Learning-Solutions-Cookie-policy-Digital-Learning-Solutions
WavePlugin

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Web/Views/CookieConsent/CookiePolicy.cshtml
+++ b/DigitalLearningSolutions.Web/Views/CookieConsent/CookiePolicy.cshtml
@@ -4,6 +4,17 @@
     ViewData["Title"] = "Digital Learning Solutions - Cookie policy";
     ViewData["DoNotDisplayNavBar"] = true;
 }
+<style>
+    @@media only screen and (max-width: 767px) {
+        .nhsuk-details {
+            word-break: break-word;
+        }
+
+        .nhsuk-table-responsive .nhsuk-table__body .nhsuk-table__row td {
+            display: block;
+        }
+    }
+</style>
 <form method="post" asp-action="CookiePolicy">
     @*<input type="hidden" asp-for="UserConsent" />*@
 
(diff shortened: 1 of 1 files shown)


DigitalLearningSolutions.Web/Views/CookieConsent/CookiePolicy.cshtml
```diff
@@ -18,7 +29,7 @@
                     <p>We'll only use these cookies if you say it's OK. We'll use a cookie to save your settings.</p>
 
                     <h3>Tell us if we can use analytics cookies</h3>
-                    <fieldset class="nhsuk-fieldset" id="cookie-statistics-form" aria-describedby="cookie-statistics">
+                    <fieldset class="nhsuk-fieldset" id="cookie-statistics" aria-describedby="cookie-statistics">
```
rshrirohit, line 32: To tackle the error thrown by wav plugin, you have used a different ID as the existing one was not valid. What's the source/where did you get this new ID from? Are we sure about this accessibility change?

rshrirohit (commented):
Please see inline comment on code regarding wave error tackling

DigitalLearningSolutions.Web/Views/CookieConsent/CookiePolicy.cshtml
manishagarwalnhs: The wave plugin checks for the validity of attributes and in this case the aria-describedby has no referenced element in the html. Hence it was throwing an error. The id is neither used in the js nor it is used to prepare client side script on server side. I did not get this id from anywhere but just making the already existing html correct. I'm not sure about the source you are asking. If you still have doubts then please let me know what do you mean by source and I'll look into it.

rshrirohit: The id used currently should not be used here. It currently means that it is describing itself (aria-describedby is same as id). 
In order to solve the wave error, we should either try and remove the fieldset altogether OR try to add the id in the h3 heading (tell us.......cookies) and use that id in the fieldset.

rshrirohit (changes requested):

DigitalLearningSolutions.Web/Views/CookieConsent/CookiePolicy.cshtml
manishagarwalnhs: This is addressed.

PR #1940, opened 2023-06-06
TD-1470 Fixed issue in generating verify email link by changing the p…

manishagarwalnhs:
…arameters passed for getting hash id and generate mail content

### JIRA link
https://hee-tis.atlassian.net/browse/TD-1470

### Description
1) Fixed issue in generating verify email link by changing the parameters passed for getting hash id and generate mail content in controller action.

### Screenshots
This is backend change so no screenshot available.

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data/DataServices/UserDataService/DelegateUserDataService.cs
+++ b/DigitalLearningSolutions.Data/DataServices/UserDataService/DelegateUserDataService.cs
@@ -125,6 +125,7 @@
                 da.UserID,
                 da.CentreSpecificDetailsLastChecked,
                 da.RegistrationConfirmationHash,
+                da.RegistrationConfirmationHashCreationDateTime,
                 u.ID,
                 u.PrimaryEmail,
                 u.PasswordHash,
--- a/DigitalLearningSolutions.Data/DataServices/UserDataService/DelegateUserDataService.cs
+++ b/DigitalLearningSolutions.Data/DataServices/UserDataService/DelegateUserDataService.cs
@@ -125,7 +125,6 @@
                 da.UserID,
                 da.CentreSpecificDetailsLastChecked,
                 da.RegistrationConfirmationHash,
-                da.RegistrationConfirmationHashCreationDateTime,
                 u.ID,
                 u.PrimaryEmail,
                 u.PasswordHash,
--- a/DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/ViewDelegateController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/ViewDelegateController.cs
@@ -4,8 +4,10 @@
     using DigitalLearningSolutions.Data.DataServices.UserDataService;
     using DigitalLearningSolutions.Data.Enums;
     using DigitalLearningSolutions.Data.Extensions;
+    using DigitalLearningSolutions.Data.Models;
     using DigitalLearningSolutions.Data.Models.Email;
     using DigitalLearningSolutions.Data.Models.User;
+    using DigitalLearningSolutions.Data.Utilities;
     using DigitalLearningSolutions.Web.Attributes;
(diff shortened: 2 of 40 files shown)


kevwhitt-hee (changes requested):
@manishagarwalnhs, this change doesn't seem to make any difference to the email generated. We still pick up the existing `DelegateAccounts.RegistrationConfirmationHash` and the `DelegateAccounts.RegistrationConfirmationHashCreationDateTime` does not change.

Given that the link is time-limited, if the administrator generates the email after the time limit has expired, the learner won't be able to use it. We want to generate a new hash and reset the creation time in the same way that happens when the "Send welcome email" button is clicked. 

Because we are simply including a link to open an email, we can't achieve this, though, so we should load the RegistrationConfirmationHashCreationDateTime into the view model and only show the link if it is more than 1 day in the future. Ideally we should also calculate the number of days and include that detail in the email, though that might be tricky.

manishagarwalnhs:
"Open welcome email" and "Open verification email" links will only be shown if there is more than 1 day left in its expiry.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/ViewDelegateController.cs
```diff
@@ -79,18 +81,25 @@ public IActionResult Index(int delegateId)
                 model.DelegateInfo.Email = null;
 
             var baseUrl = config.GetAppRootPath();
-            if ((model.DelegateInfo?.IsActive ?? false) && (model.DelegateInfo.RegistrationConfirmationHash != null))
+            IClockUtility clockUtility = new ClockUtility();
+            if ((model.DelegateInfo?.IsActive ?? false) && (model.DelegateInfo.RegistrationConfirmationHash != null
+                && model.DelegateInfo.RegistrationConfirmationHashCreationDateTime?.AddDays(2) > clockUtility.UtcNow)
```
kevwhitt-hee, line 86: The validity period will be moved to appsettings in https://hee-tis.atlassian.net/browse/TD-1863 please connect with @rshrirohit to make sure this is picked up here.

kevwhitt-hee (commented):
Happy to approve this once the comment re https://hee-tis.atlassian.net/browse/TD-1863 has been acknowledged.

We should add a ticket to specify the number of days the link will remain valid in the email text too, @rshrirohit, since we could be sending an email stating it is valid for 14 days when in fact it is only valid for 2 or 3.

DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/ViewDelegateController.cs
manishagarwalnhs: @kevwhitt-hee This is now not required anymore I believe because we have reverted back the changes.

DigitalLearningSolutions.Data/DataServices/UserDataService/DelegateUserDataService.cs
```diff
@@ -125,6 +125,7 @@ public partial class UserDataService
                 da.UserID,
                 da.CentreSpecificDetailsLastChecked,
                 da.RegistrationConfirmationHash,
+                da.RegistrationConfirmationHashCreationDateTime,
```
kevwhitt-hee, line 128: We need to revert the changes that bring the `RegistrationConfirmationHashCreationDateTime` into the view model since we no longer do anything with it.

kevwhitt-hee (changes requested):
See code comments, need to revert changes that bring the `RegistrationConfirmationHashCreationDateTime` into the model since it is not used any more.

DigitalLearningSolutions.Data/DataServices/UserDataService/DelegateUserDataService.cs
manishagarwalnhs: This is addressed.

PR #1953, opened 2023-06-12
TD-1577: The "Change" link will be hidden when there is only one centre and when  supervisors enrolling a delegate on a self assessment can only do so in the “Educator manager”

sarathlal-sarangadharan:
### JIRA link
https://hee-tis.atlassian.net/browse/TD-1577

### Description
1. “The "Change" link should be hidden when there is only one centre“
2. Roles selection has been be limited in the Supervise path

### Screenshots
Before code change:
image
image
image

After code change:
image
image

**Change functionality when number of centre > 1** 

Before code change:
image

After code change:
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)
- [ ] Scanned over my pull request in GitHub and addressed any warnings from the GitHub Build and Test checks.

### DIFF
--- a/DigitalLearningSolutions.Web/Controllers/LearningPortalController/SelfAssessment.cs
+++ b/DigitalLearningSolutions.Web/Controllers/LearningPortalController/SelfAssessment.cs
@@ -1003,6 +1003,11 @@
             );
             var roles = supervisorService.GetDelegateNominatableSupervisorRolesForSelfAssessment(selfAssessmentId);
             var supervisor = selfAssessmentService.GetSupervisorByAdminId(sessionAddSupervisor.SupervisorAdminId);
+            var distinctCentres = selfAssessmentService.GetValidSupervisorsForActivity(
+                User.GetCentreIdKnownNotNull(),
+                selfAssessmentId,
+                User.GetUserIdKnownNotNull()
+            ).Select(c => new { c.CentreID, c.CentreName }).Distinct().OrderBy(o => o.CentreName).ToList();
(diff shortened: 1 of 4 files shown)


kevwhitt-hee (changes requested):
It looks like we have removed the logic to check if more than one role is available and display the choose role step if there is. None of the existing self assessments have multiple roles available but that may change in future.

Also, I can see the same logic in the `QuickAddSupervisor` controller action - does this need the same updates too?

sarathlal-sarangadharan:
> It looks like we have removed the logic to check if more than one role is available and display the choose role step if there is. None of the existing self assessments have multiple roles available but that may change in future.
> 
> Also, I can see the same logic in the `QuickAddSupervisor` controller action - does this need the same updates too?

Hi @kevwhitt-hee , It looks we have multiple roles available for a self assessment in local DB as well as in UAT. Please find below screenshot.

Local DB:
image
UAT DB:
image

If we include the logic to check more than one role, again we will get the role selection page. Please correct me if I am wrong.

sarathlal-sarangadharan:
> It looks like we have removed the logic to check if more than one role is available and display the choose role step if there is. None of the existing self assessments have multiple roles available but that may change in future.
> 
> Also, I can see the same logic in the `QuickAddSupervisor` controller action - does this need the same updates too?

Hi @kevwhitt-hee , This action `QuickAddSupervisor`  is being called from role selection page which is not using at the moment. When there is a role selection in future, the functionality will continue working as it is.

kevwhitt-hee (approved):

PR #1997, opened 2023-07-04
TD-2032 Updates User table to trim whitespaces from first name and last name

rshrirohit:
### JIRA link
https://hee-tis.atlassian.net/browse/TD-2032

### Description
This migration will trim off the leading and trailing whitespaces from the FirstName and Lastname columns of the User table.

### Screenshots
Database change so no screenshot is required.

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data.Migrations/202307040730_UpdateFirstnameLastNameUsersTable.cs
+++ b/DigitalLearningSolutions.Data.Migrations/202307040730_UpdateFirstnameLastNameUsersTable.cs
@@ -1,0 +1,16 @@
+﻿namespace DigitalLearningSolutions.Data.Migrations
+{
+    using FluentMigrator;
+
+    [Migration(202307040730)]
+    public class UpdateFirstnameLastNameUsersTable : Migration
+    {
+        public override void Up()
+        {
+            Execute.Sql(@"Update Users set FirstName = TRIM(FirstName), LastName = TRIM(LastName)");
+        }
+        public override void Down()
+        {
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


DigitalLearningSolutions.Data.Migrations/202307040730_UpdateFirstnameLastNameUsersTable.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace DigitalLearningSolutions.Data.Migrations
+{
+    using FluentMigrator;
+
+    [Migration(202307040730)]
+    public class UpdateFirstnameLastNameUsersTable : Migration
+    {
+        public override void Up()
+        {
+            Execute.Sql(@"Update Users set FirstName = TRIM(FirstName), LastName = TRIM(LastName)");
```
kevwhitt-hee, line 10: TRIM was introduced in SQL 2017. The production SQL server is 2016. We will need to use `RTRIM(LTRIM(FirstName))` or similar to achieve this.

kevwhitt-hee (changes requested):
See issue in comments re SQL version compatibility. I think it might be worth seeing how long this query takes on production-sized table, too.

DigitalLearningSolutions.Data.Migrations/202307040730_UpdateFirstnameLastNameUsersTable.cs
rshrirohit: Used the aforementioned as per compatibility with SSMS 2016

rshrirohit:
> See issue in comments re SQL version compatibility. I think it might be worth seeing how long this query takes on production-sized table, too.

Version compatibility issue resolved. Checked with a copy of production database - it has around 26k records that have spaces in either firstname/lastname/both. It takes 15-20 seconds to run the query.

manishagarwalnhs:
It looks good.

PR #2515, opened 2024-02-07
TD-3652 changing not registered to inactive

sherif-olaboye:
### JIRA link
https://hee-tis.atlassian.net/browse/TD-3652

### Description
Changing Not register to Inactive

### Screenshots
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Web/Views/Supervisor/Shared/_StaffMemberCard.cshtml
+++ b/DigitalLearningSolutions.Web/Views/Supervisor/Shared/_StaffMemberCard.cshtml
@@ -18,7 +18,7 @@
                     <div class="nhsuk-grid-column-one-quarter">
                         <span class="float-right-additional-information">
                             <strong class="nhsuk-tag nhsuk-tag--red">
-                                Not registered
+                                Inactive
                             </strong>
                         </span>
                     </div>


rshrirohit:
I don't think changing the label is the right solution here. Please check why an inactive member is showing as not registered. A not registered user is the one that you can't find on Superadmin (or anywhere). In other words, you won't find any record in users table and/or UserCentreDetails table. Please check why these inactive are appearing as not registered and share you findings so that we can look for solutions.

rshrirohit (changes requested):
Please see above comment and correct the code.

sherif-olaboye:
> I don't think changing the label is the right solution here. Please check why an inactive member is showing as not registered. A not registered user is the one that you can't find on Superadmin (or anywhere). In other words, you won't find any record in users table and/or UserCentreDetails table. Please check why these inactive are appearing as not registered and share you findings so that we can look for solutions.

Having look at the issue stated, I believe changing the not register to inactive, is proper and right cause it represent the status of those users on Users table, for the users that shows not register are actually inactive, that is why there delegateid is null from SupervisorDelegates and make it shows Not register on the mystaff page, I am of opinion of changing the Not register to Inactive is appropriate cause the you are actually ‘inactive’ and not ‘Not register’

PR #2611, opened 2024-04-18
TD-3173 Removing role selection from enrol on activity

rshrirohit:
### JIRA link
[TD-3173](https://hee-tis.atlassian.net/browse/TD-3173)

### Description
Removing role selection from enrol on activity

### Screenshots
image

image
-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

[TD-3173]: https://hee-tis.atlassian.net/browse/TD-3173?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/EnrolController.cs
+++ b/DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/EnrolController.cs
@@ -215,7 +215,6 @@ namespace DigitalLearningSolutions.Web.Controllers.TrackingSystem.Delegates
             }
             else
             {
-                var roles = supervisorService.GetSupervisorRolesForSelfAssessment(sessionEnrol.AssessmentID.GetValueOrDefault()).ToArray();
                 var model = new EnrolSupervisorViewModel(
                     delegateId,
                     (int)sessionEnrol.DelegateUserID,
@@ -223,7 +222,6 @@ namespace DigitalLearningSolutions.Web.Controllers.TrackingSystem.Delegates
                     sessionEnrol.IsSelfAssessment,
                    supervisorList,
                    sessionEnrol.SupervisorID.GetValueOrDefault(),
-                   roles,
                    sessionEnrol.SelfAssessmentSupervisorRoleId.GetValueOrDefault());
                 return View(model);
             }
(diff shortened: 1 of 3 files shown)


DigitalLearningSolutions.Web/Controllers/TrackingSystem/Delegates/EnrolController.cs
```diff
@@ -257,11 +254,7 @@ public IActionResult EnrolDelegateSupervisor(int delegateId, EnrolSupervisorView
                 sessionEnrol.SupervisorName = supervisorList.FirstOrDefault(x => x.AdminId == model.SelectedSupervisor).Name;
                 sessionEnrol.SupervisorID = model.SelectedSupervisor;
             }
-            if (model.SelectedSupervisorRoleId.HasValue && model.SelectedSupervisorRoleId.Value > 0)
-            {
-                sessionEnrol.SelfAssessmentSupervisorRoleName = roles.FirstOrDefault(x => x.ID == model.SelectedSupervisorRoleId).RoleName;
-            }
-            sessionEnrol.SelfAssessmentSupervisorRoleId = model.SelectedSupervisorRoleId;
+            sessionEnrol.SelfAssessmentSupervisorRoleId = roles.FirstOrDefault().ID;
```
kevwhitt-hee, line 257: This makes the assumption that there will only be one role returned by the supervisorService (or that the first one should be chosen). While this will work for the existing self assessments, it won't work if a self assessment has more than one role that can be assigned to a supervisor. I think we still need to handle the scenario where there might be more than 1.

DigitalLearningSolutions.Web/Views/TrackingSystem/Delegates/Enrol/EnrolDelegateSupervisor.cshtml
```diff
@@ -27,29 +27,6 @@
                             css-class="nhsuk-u-width-one-half"
                             default-option="--Select--"
                             select-list-options="@Model.SupervisorList" />
-            @if (Model.IsSelfAssessment && Model.SupervisorRoleList.Count() > 1)
```
kevwhitt-hee, line 30: see other comment, I think we may still need this if more than 1 role is returned by the service

kevwhitt-hee (changes requested):
I think we still need to handle a scenario where there is more than 1 supervisor assignable role for the self assessment in case there is in future.

PR #2760, opened 2024-07-30
TD-4322 Introduce "Optional" tag in the learner competencies menu (Overview)

sherif-olaboye:
### JIRA link
https://hee-tis.atlassian.net/browse/TD-4322

### Description
I added a partial view tag to the supervisor shared folder
I added the optional tag to the ReviewSelfAssessment razor page
### Screenshots
image

image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Web/Views/Supervisor/Shared/_TagRow.cshtml
+++ b/DigitalLearningSolutions.Web/Views/Supervisor/Shared/_TagRow.cshtml
@@ -1,14 +1,8 @@
-﻿@using DigitalLearningSolutions.Data.Models.SelfAssessments;
-
-@model bool;
-
+﻿
 
 <div class="tags">
-        @if (Model)
-        {
-            <div class="card-filter-tag">
-                <strong class="nhsuk-tag nhsuk-tag--blue">Optional</strong>
-            </div>
-        }
-    
+
+    <div class="card-filter-tag">
+        <strong class="nhsuk-tag nhsuk-tag--blue">Optional</strong>
+    </div>
 </div>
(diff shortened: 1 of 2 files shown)


DigitalLearningSolutions.Web/Views/Supervisor/Shared/_TagRow.cshtml
```diff
@@ -0,0 +1,14 @@
+﻿@using DigitalLearningSolutions.Data.Models.SelfAssessments;
```
rshrirohit, line 1: _TagRow.cshtml already exists in the solution

DigitalLearningSolutions.Web/Views/Supervisor/Shared/_TagRow.cshtml
```diff
@@ -0,0 +1,14 @@
...
+
+@model bool;
+
+
+<div class="tags">
+        @if (Model)
```
rshrirohit, line 7: don't you think that this condition is better suited in the main view rather than the partial? I mean if the competency is not optional, then we won't have to come to this partial.

rshrirohit (commented):
Please discuss with me before making the suggested changes.

DigitalLearningSolutions.Web/Views/Supervisor/Shared/_TagRow.cshtml
sherif-olaboye: I think there should be separation of concern regarding the _TagRow.cshtml, the one that exist in the solution is in learning portal not in the supervisor

DigitalLearningSolutions.Web/Views/Supervisor/Shared/_TagRow.cshtml
sherif-olaboye: The reason why I did not put condition on the main page is that I was thinking if we want to add additional tag to this _TagRow.cshtml in the future it make it simple and easy doing that

PR #2776, opened 2024-08-13
TD-4502 Issue not listing the delegates on 'Send welcome message' screen for whose Password is set and unclaimed

sherif-olaboye:
### JIRA link
https://hee-tis.atlassian.net/browse/TD-4502

### Description
I added  password set  and unclaimed delegate to the list delegate on the send welcome message

### Screenshots
image
image
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Web/Services/UserService.cs
+++ b/DigitalLearningSolutions.Web/Services/UserService.cs
@@ -275,7 +275,7 @@ namespace DigitalLearningSolutions.Web.Services
         public IEnumerable<DelegateUserCard> GetDelegateUserCardsForWelcomeEmail(int centreId)
         {
             return userDataService.GetDelegateUserCardsByCentreId(centreId).Where(
-                user => user.Approved && !user.SelfReg && (string.IsNullOrEmpty(user.Password) || !string.IsNullOrEmpty(user.Password)) &&
+                user => user.Approved && !user.SelfReg && 
                         !string.IsNullOrEmpty(user.EmailAddress)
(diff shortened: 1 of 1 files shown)


DigitalLearningSolutions.Web/Services/UserService.cs
```diff
@@ -275,7 +275,7 @@ public void UpdateFailedLoginCount(UserAccount userAccount)
         public IEnumerable<DelegateUserCard> GetDelegateUserCardsForWelcomeEmail(int centreId)
         {
             return userDataService.GetDelegateUserCardsByCentreId(centreId).Where(
-                user => user.Approved && !user.SelfReg && string.IsNullOrEmpty(user.Password) &&
+                user => user.Approved && !user.SelfReg && (string.IsNullOrEmpty(user.Password) || !string.IsNullOrEmpty(user.Password)) &&
```
rshrirohit, line 278: I think this condition is not at all required. The condition (string.IsNullOrEmpty(user.Password) || !string.IsNullOrEmpty(user.Password)) will always return true because a Password can only be either empty or not empty. This means the condition will pass the filter regardless of the Password, effectively ignoring the Password condition entirely.

rshrirohit (changes requested):
Please see the in-line comment. Remove the password condition entirely so that the function returns a list which is independent of the password status.

DigitalLearningSolutions.Web/Services/UserService.cs
sherif-olaboye: Done

PR #2974, opened 2024-11-26
TD-4893 Delegate activities - view course delegates resulted in 500 error when course admin field applied filters got deleted

sherif-olaboye:
### JIRA link
https://hee-tis.atlassian.net/browse/TD-4893

### Description
I added Q1Options property to Customisation class
I add Q1Options as a parameter in the GetFilterString method
I added a check to the GetFilterString method to clear the cookievalue where is neccessary
### Screenshots
image

-----
### Developer checks

I have:
  - [doc_1_here](link_1_here)

### DIFF
--- a/DigitalLearningSolutions.Data/DataServices/CourseDataService.cs
+++ b/DigitalLearningSolutions.Data/DataServices/CourseDataService.cs
@@ -2040,6 +2040,7 @@ namespace DigitalLearningSolutions.Data.DataServices
                         ,DiagObjSelect
                         ,HideInLearnerPortal
                         ,NotificationEmails
+                        ,Q1Options
                     FROM Customisations 
 						WHERE CustomisationID = @customisationID ",
                 new { customisationId }).FirstOrDefault();
--- a/DigitalLearningSolutions.Data/Helpers/FilteringHelper.cs
+++ b/DigitalLearningSolutions.Data/Helpers/FilteringHelper.cs
@@ -66,7 +66,34 @@
                 ? defaultFilterValue
                 : AddNewFilterToFilterString(existingFilterString, newFilterToAdd);
         }
+        public static string? GetFilterString(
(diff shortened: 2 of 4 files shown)


Auldrin-Possa (commented):
The proposed solution does not seem appropriate. This may give error if we use number of admin fields with different option names.

kevwhitt-hee (changes requested):
As per Auldrin's comment. This looks like it will only partially fix the problem. There are 6 custom prompt fields / answers. I suggest we try to a generic approach of catching the error when applying filters, clearing the filters if there is an error and reloading page with cleared filters, if possible.

sherif-olaboye:
> The proposed solution does not seem appropriate. This may give error if we use number of admin fields with different option names.

Please kindly take a look at the latest push

sherif-olaboye:
> As per Auldrin's comment. This looks like it will only partially fix the problem. There are 6 custom prompt fields / answers. I suggest we try to a generic approach of catching the error when applying filters, clearing the filters if there is an error and reloading page with cleared filters, if possible.

Please kindly take a look at the latest push

Auldrin-Possa (commented):
Looks good, however some changes need to apply
1. To CheckIfFilterisValid(+) 
     - remove response param if not being used.
     - clearFilters param check can be handled by existingFilterString param. Can be removed. (Need to test)
2. Duplicate filters should be handled within or outside of CheckIfFilterisValid(+)

sherif-olaboye:
> Looks good, however some changes need to apply
> 
> 1. To CheckIfFilterisValid(+)
>    
>    * remove response param if not being used.
>    * clearFilters param check can be handled by existingFilterString param. Can be removed. (Need to test)
> 2. Duplicate filters should be handled within or outside of CheckIfFilterisValid(+)

Done

PR #3074, opened 2025-01-22
Improved Dependency Update Workflow and Auto-Merge Validation

binon:
Description:
This PR addresses issues where Dependabot fails to create a PR due to connectivity problems with our Tel Azure NuGet source.

Changes include:

Added checks to handle scenarios where Dependabot cannot connect to the NuGet source.
Enhanced validation for auto-merge functionality:
Auto-merge will now wait for pipelines that are not only In_Progress but also in the Queued status, preventing merges when pipelines are stuck in the queue.
This ensures smoother dependency updates and avoids potential CI/CD disruptions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,23 +4,35 @@
 # https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
 
 version: 2
+registries:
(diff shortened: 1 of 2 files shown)


kevwhitt-hee (approved):
Looks good, Binon.

